PR #4, opened 2017-08-14
C/Linux: Mark the load of non-value-return atomics as "noreturn"

fbq:
On some architectures, there exists atomic memory operations which will
be run directly at memory controller after issued by a CPU. Such an
operation doesn't return a value to a register, therefore the load part
is not treated as a load by barriers. In order to differ this kind of
operations, we mark the load part of a non-value-return atomic as
"noreturn" annotation. The memory model's cat file can add special rules
to it afterwards.

### DIFF
--- a/herd/CSem.ml
+++ b/herd/CSem.ml
@@ -37,6 +37,7 @@ module Make (Conf:Sem.Config)(V:Value.S)
 
     module MOorAN = MemOrderOrAnnot
     let a_once = ["once"]
+    let a_noreturn = ["noreturn"]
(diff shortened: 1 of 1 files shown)


maranget:
I am affraid that this patch intoduces incompatibilities with the current (and submitted) for publication
linux kernel model. Thus I'd rather wait the end of the refereeing process before merging the changes.

maranget:
I have re-target the pull request to the "dev" branch and merged it. I'll now add the rest of
modification (macros, bell ans cat files)

PR #5, opened 2017-08-29
x86 mixed-size: new instructions, truncated sentinel

lindenr:
Added movb, movw etc instructions to litmus' x86 parser for writing mixed-size litmus tests. Also truncated the sentinel value in some cases.


maranget:
Ok for merging those small changes.

PR #6, opened 2017-09-15
install.sh: make installation directories before trying to copy files to them

chutz:


maranget:
Thank you for this useful addition to th einstallation script.

PR #11, opened 2018-09-24
[RFC] Litmus translation

fbq:
Hi Luc,

This PR consists of a new tool called "gen_theme" which generates a theme for jingle to do litmus translation.

Also I "fixed" some problems like labels and register allocations in jingle, so as a result I could translate the currently litmus tests from Linux Kernel (C) to AArch64. Although I didn't check whether we can get the same results for both side (i.e. the correctness of the translation), but I think it's a good time for you to review the code, and see if I did something wrong.

### DIFF
--- a/defs.sh
+++ b/defs.sh
@@ -4,7 +4,7 @@ HERD="herd.native"
 LITMUS="litmus.native klitmus.native"
 TOOLS="mfind.native moutcomes.native splitcond.native mshowhashes.native mlog2cond.native mflags.native mdiag.native recond.native mcycles.native mmixer.native knames.native mdiff.native mcmp.native madd.native mtopos.native mfilter.native mapply.native mcompare.native mhash.native mrcu.native mprog.native mnames.native ksort.native mobserved.native msort.native msum.native mselect.native mcond.native mproj.native rehash.native splitdot.native mlock.native mtrue.native mlisa2c.native"
 GEN="readRelax.native atoms.native dont.native diycross.native mexpand.native atomize.native diyone.native nexts.native classify.native diy.native norm.native"
-JINGLE="jingle.native"
+JINGLE="jingle.native gen_theme.native"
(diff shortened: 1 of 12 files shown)


maranget:
Dear Boqun,

I found a regression bug in jingle. consider the following X.theme
```
C to AArch64

"%x = constvar:c;" -> "MOV %x,&c"
"*%x = constvar:c;" -> "MOV %tmp,&c;
                       STR %tmp,[%x]"
"*%x = %y;" -> "STR %y,[%x]"
"%x = *%y;" -> "LDR %x,[%y]"
```
And the following test :
```
C X
{}
P0(int *x) { *x = 1; }
forall (x=1)
```

Then, `jingle7 -theme X.theme X.litmus` yields
```% jingle7 -theme X.theme X.litmus
AArch64 X
Mapping=0:X0=x
Hash=0c2859f07532855764fb37b8d40298a5

{0:X0=x;}

 P0          ;
 MOV X2,#1   ;
 STR X1,[X0] ;

forall (x=1)
Generated 1 tests
```

The code above is wrong, registers X2 and X1 should be the same.

fbq:
Hi Luc,

I've add a new commit dd28da3 to fix the issue you found. Although I may take a different way to fix this than last time we talk. But I hope you can try it
to see if that fixes the problem.

Regards,
Boqun

maranget:
Dear Boqun,

I have corrected some definitions in Linux.call/Linux2AArch64.map. Please feel free to comment. In particular I could not avoid a direct coding of spin_lock. See my_lock in Linux2AArch64.map. If you have a better idea...

fbq:
Thanks for looking into this, Luc ;-) especial on the fix of cmpxchg_release.

About the "my_lock", how about introducing another function "out_to_loop" as I did in e397cc7 ?

maranget:
Merged...

PR #14, opened 2019-04-15
jingle: libdir: Add more rules to support translation from C

fbq:
Add a few more rules:

* if-expression
* Addition in WRITE_ONCE()
* atomic_{cmp}xchg_*()
* atomic_inc() // using stadd


maranget:
Just merged,

PR #18, opened 2019-07-15
Lib: Misc: Fix split_on_char

fbq:
split_on_char should skip the 'sep', so we need to change "k" with
"(k+1)" in hit branch


maranget:
Thanks Boqun, as you see I am trying to preserve herdtools7 backward compatibility down to ocaml 4.02.0 included. This may exagerated :) Thanks again for you reactivity.

PR #19, opened 2019-07-22
jingle: libdir: Add Linux2PPC.map

fbq:
This will generate a theme file based on Linux.call file and we can
convert a Linux litmus to PPC assembly one with the help of the theme
file.

Currently, we are only able to translate litmus tests in Linux kernel.

Signed-off-by: Paul E. McKenney <paulmck@linux.ibm.com>


maranget:
Merged-in, thanks all.

PR #23, opened 2019-08-27
Hopefully-improved macOS detector

### LINKED ISSUE #21: Litmus operating system check doesn't recognise macOS correctly
The check for macOS in [litmus/runUtils.ml](https://github.com/herd/herdtools7/blob/43cfe8db2959559c1c34dfa13cdff965e952897d/litmus/runUtils.ml#L73) does so by testing the presence of the file `/mach_kernel`.  This file [hasn't existed since OSX Yosemite](https://discussions.apple.com/thread/6735137) and as such the check erroneously reports modern macOS as Linux.  This then leads to:

```
Fatal error: open_in failed: /proc/cpuinfo: No such file or directory
```

I have a quick-ish fix that replaces the file test with a scrape on the output of `uname -s`, which I'll post as a PRQ, but there might be a more principled way of doing it that I hadn't thought of!

MattWindsor91:
Attempt to fix #21, though there are probably more principled versions, and I haven't been able to test this as exhaustively as I'd like.  (It gives the right response on macOS, though may need a bit more checking on other *nix.)

Commit message reads thusly:

`/mach_kernel` hasn't existed for several macOS revisions, so the
existing detection heuristic (checking that `/mach_kernel` exists) fails on
modern macOS.

This change replaces that detection logic with a test to make sure that
1) the system is a Unix system; and 2) `uname -s` is `Darwin`.

Ideally, I'd have used `Unix` to execute `uname -s` and read its stdout,
but the `Unix` library is unavailable (herdtools7 targeting things other
than *nix!)

### DIFF
--- a/litmus/runUtils.ml
+++ b/litmus/runUtils.ml
@@ -65,12 +65,15 @@ let open_quote chan =
 and close_quote chan =
   if O.is_out then output_line chan "EOF"
 
+let is_darwin () : bool =
+  Sys.unix && Sys.command "sh -c 'test \"$(uname -s)\" = Darwin'" = 0
+
 module W = Warn.Make(O)
 let target_os =
   if O.is_out then
     O.targetos
   else begin
-    if Sys.file_exists "/mach_kernel"
+    if is_darwin ()
     then begin
       W.warn "OS: Darwin" ;
       TargetOS.Mac


maranget:
Thanks for the pull-request.

I am not aware of any non-unix deployment of herdtools. Notice that litmus naturally features cross-compilation, when given the command line option ``-o <name|name.tar>``. In that case, the target OS is specified on the the command-line (option ``-os <os>`` or in a configuration file (line ``os <os>``. In cross compilation mode the target ps is linux.

As a consequence your patch applies to the specific use case of a direct execution of a test on MacOS.

PR #24, opened 2019-10-26
Support dune 1.2

fshaked:
This works with dune 1.2 and later


maranget:
Cherry-picked into master, thanks Shaked.

PR #28, opened 2020-02-14
[herd,C] support "(void) expr;" statement

fbq:
In C language, a "(void) expr;" statement is used for executing "expr"
while discarding the return value of "expr". And currently herd doesn't
support this kind of statement.

In order to support the following litmus test, this functionality is
added.

	C Atomic-set-observable-to-RMW

	{
	  atomic_t v = ATOMIC_INIT(1);
	}

	P0(atomic_t *v)
	{
	  (void)atomic_add_unless(v,1,0);
	}

	P1(atomic_t *v)
	{
	  atomic_set(v, 0);
	}

	locations [v]
	exists
	(v=2)

The implementation is quite simple: we convert the "(void) expr;"
statement into a StoreReg whose destination is register named "", since
no one can declare a identifier named "", so we don't need to worry
about the side effect of storing to this register.

### DIFF
--- a/lib/CParser.mly
+++ b/lib/CParser.mly
@@ -239,6 +239,8 @@ instruction:
   { $1 }
 | IDENTIFIER EQ expr SEMI
   { StoreReg(None,$1,$3) }
+| LPAR VOID RPAR expr SEMI
+  { StoreReg(None,"",$4) }
 | STAR location EQ expr SEMI
   { StoreMem($2,$4,AN []) }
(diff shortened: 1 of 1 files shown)


maranget:
Hi Boqun, thank you for the PR.

I merge it, I may provide a specific AST constructor later. The "" trick is neat but may lead to difficulties later.

--Luc

fbq:
Sure, thanks!

Issue #2, opened 2017-02-28
Another litmus test that runs on old herd but fails on herd7


AndriyLin:
Hi,

While trying multiple C11 tests, I found another test from <http://multicore.doc.ic.ac.uk/overhauling/C11/cppmem_unsequencedrace.litmus>:

```
C cppmem_unsequenced_race

{
[x] = 2;
[y] = 0;
}

// This litmus test is ported from an example distributed
// with the CppMem tool. Original is available online at:
// http://svr-pes20-cppmem.cl.cam.ac.uk/cppmem/index.html

P0 (int* x, int* y) {
  *y = (*x == (*x = 3));
}

exists
()
```

which will complain:

```
line 13, characters 18-19: unexpected '=' (in prog)
```

in herd7 but runs with old version herd.

maranget:
Hi,

Hum, for herd7, stores cannot occur in expression position.
I admit this is a significant shortcoming.

However, fixing this would mean important changes, and would be of marginal
interest. I will not fix this now.

--Luc

Issue #1, opened 2017-02-28
Handling "atomic_compare_exchange_strong_explicit" in C11


AndriyLin:
Hi,

I was using herd7 (7.43, installed via OPAM) to run against the C11 litmus from <https://github.com/nchong/c11popl15/blob/master/tests/a2.litmus>:

```
C a2
{ [x] = 0; [y] = 0; [z] = 0; }

P0 (atomic_int* x, atomic_int* y, atomic_int *z) {
  int r0 = atomic_load_explicit(y, memory_order_relaxed);
  int r1 = atomic_compare_exchange_strong_explicit(x, z, 1, memory_order_release, memory_order_relaxed);
}

P1 (atomic_int* x, volatile int* y) {
  int r2 = atomic_load_explicit(x, memory_order_acquire);
  if (r2) {
    *y = 1;
  }
}
```

But herd7 complains:

```
line 6, characters 11-50: unexpected 'atomic_compare_exchange_strong_explicit' (in prog)
```

It seems "atomic_load_explicit" (and also "atomic_store_explicit" according to other tests) are recognized, but not "atomic_compare_exchange_strong_explicit".

I also tried with old version herd in <https://github.com/herd/herdtools>, and it can successfully finish the run there.

So is herd7 indeed missing the part for "atomic_compare_exchange_strong_explicit" at the moment?
Thanks.

maranget:
Hi,

Thank you for your report.

This is a isfeature, a particurily silly one, as herd7 implements
atomic_compare_exchange_strong and atomic_compare_exchange_weak.

Correction is a matter of minutes. That said semantics is untested.

Thanks again.

--Luc

maranget:
The issue should be solved in 7.44, now available on github
as source.

<https://github.com/herd/herdtools7/tree/7.44>

That said, I was too eager to bumb version and publish on opam, as I see
you have sent another PR..


--Luc

PR #29, opened 2020-04-21
[all]: LDP/STP Support + concrete numeric addresses

lukeg101:
This patch upstreams support for the AArch64 LDP/STP instructions.
This also upstreams support for providing numeric addresses in the
initial state. The patch contributions are outlined as follows
with motivations:

- Upstream LDP/STP support
- Added new address operator to the initial state
    you can write &x = 0xf00, then when a numeric address
    is passed to ldp/stp, it can dereference this address to
    get x. Later patches will upstream support for many more
    data access instructions like this. This kind of code is
    frequently emitted by compilers.
- Added address_of and deref unops to eventsMonad, this allows
  the solver to move between concrete '0xf00' and symbolic
  representations 'x' of addresses so that we may increment addresses,
  such as x+64 in ldp/stp, or acquire a symbolic address in the above
  example.
- This patch is supported by multiple unittests in herd/unittests to
  be used as regressions, see herd/unittests/README for instructions
  on what this means. As a policy any new instruction/variant should
  be supported with unittests and should be rejected without.

Some notes:
- Passing in symbolic variables still works as expected, this patch
  simply increases the kinds of input to the instructions we support
- Later patches will upstream support for many more lx**p, st**p, ld**r,
  st**r instructions, as well as pre-index, and barrel-shift
instructions.


maranget:
Hi Luke, 

Before I review the code, what would be your evaluation on merging into master directly instead of merging  into the telechat branch.

The advantage would be faster availability of the new instructions.  However I am concerned that merging into master may complicate the later merge of the telechat branch into master.

What is your opinion ?

lukeg101:
Hi Luc,
Thanks for taking the time to look into this. As far as changes go this could be rather a controversial one due to the introduction of numeric values as addresses. This has an effect on multiple things that are worth looking into to ensure you are happy with them before merging into master. My reasons for doing this on a non-master branch are outlined below:

First, I will be changing many data access instructions in subsequent patches to use this feature and as such changing the semantics of AArch64 instructions in the large. Existing code should work fine, but it is worth evaluating how this feature works in practice (and with unittests) before completely overwriting master.

Second, this feature is useful and works for instructions however I am aware down the line there is a bug pertaining to how we handle pointer types in the context of numeric addresses (in mixed mode). For instructions, this feature works (as demonstrated by unittests), but later I would like to add pointer support which will inevitably need fixing -  I will need your keen eyes here as I suspect it is in the code above.

As for merging, I am happy to mirror master on this branch and rebase changes as they come, this will make merging later trivial as we will have a linear history (which is an approach I use e.g in llvm dev work)

As for future patches, they will be smaller than this one but they use the functionality outlined here

herd/AArch64Sem.ml
```diff
@@ -245,6 +245,52 @@ module Make
 
       and str sz rs rd kr ii = do_str sz AArch64.N rs (get_ea rd kr ii) ii
 
+      (* Store Pair *)
```
maranget, line 248: I have a double, should not the address increment be 4 and 8 for V32 and V64,  respectively?

maranget:
Hi Luke,

First I totally agree with you that merging into master is prematurate.

I have not looked in detail at the code, I have preliminary remarks. Those are thought I want to share and does not mean that I do not approve the PR.

Consider that STP and LDP are already available in litmus. It would be a real bonus if herd and litmus could handle the same tests. For instance, first consider this tests:
```
Arch64 A00-INT
"stp vs. ldp, int cell"

 {
   int z[2];
   0:X2=z; 1:X2=z;
}
 P0               | P1             ;
 MOV W0,#1        | LDP W1,W0,[X2] ;
 MOV W1,#2        |                ;
 STP W1, W0, [X2] |                ;
locations [z;]
~exists (1:X0=1 /\ 1:X1=0)
```
The test is runnable by litmus, as the declaration int z[2] suffices for the runtime skeleton to allocate the correct amount of memory with correct alignement. However z, z[0], z[1], cannot appear in the final conditions, which precludes writing tests as your A02 etc. unit tests. As a sometimes handy workaround, one can print the z array by the ``locations`` construct.

One can also test the quadword form of the ldp and stp instructions:
```
Arch64 A00
"stp vs. ldp, quad"

 {
   int64_t z[2];
   uint64_t 1:X1; uint64_t 1:X0;
   0:X2=z; 1:X2=z;
}
 P0               | P1             ;
 MOV X0,#1        | LDP X1,X0,[X2] ;
 MOV X1,#2        |                ;
 STP X1, X0, [X2] |                ;
locations [z;]
exists (1:X0=1 /\ 1:X1=0)
```
With respect to test ``A00-INT``, we replace int with uint64_t and ``W`` registers by ``X`` registers. The type declaration of registers ``uint64_t 1:X1`` and ``uint64_t 1:X0`` are not really useful, apart from sparing us a warning.

I attach a run of the two tests on some ODROID-C2 machine.
[LOG.txt](https://github.com/herd/herdtools7/files/4517641/LOG.txt)

Now how could we  proceed ? May I suggest that you use litmus test meta-datas, _i.e._ additional information in the litmus file that tools are free to ignore.
Meta-data are introduced by a keyword followed by "=". As an exemple, diy generators produce litmus sources with meta-data. Try ```diyone7 -arch AArch64 PodWW Rfe PodRR Fre`` to see some, 
``Generator=...`` etc.

Now my first idea would by to express the numerical value of symbolic locations as meta-data: ``Load=x:0x200`` for instance. Another solution would be to keep you idea of two symbolic addresses x and z (ideally in some meta-data! for herd to interpret them as your ``&x=..``) and to inform litmus (with appropriate meta-data!) that x and z have to be allocated together, with a distance that would be the difference of their numerical value. This would implies changing litmus a bit, nothing that cannot be managed.

Admittedly, my suggestion significantly departs from your initial proposal. But I do think we should strive at having tests that are food for both litmus and herd. One important benefit of this approach is debugging instruction semantics by comparison with actual hardware.

Finally, I'd like to observe that LDP/STP can be implemented on top of symbolic locations, as indexed locations (such as x+4, or x+8) are handled naturally by herd. Hence as far as LDP/STP are concerned the address_of and deref operations are not necessary, strictly speaking.

lukeg101:
> Hi Luke,
> 
> First I totally agree with you that merging into master is prematurate.
> 
> I have not looked in detail at the code, I have preliminary remarks. Those are thought I want to share and does not mean that I do not approve the PR.
> 
> Consider that STP and LDP are already available in litmus. It would be a real bonus if herd and litmus could handle the same tests. For instance, first consider this tests:
> 
> ```
> Arch64 A00-INT
> "stp vs. ldp, int cell"
> 
>  {
>    int z[2];
>    0:X2=z; 1:X2=z;
> }
>  P0               | P1             ;
>  MOV W0,#1        | LDP W1,W0,[X2] ;
>  MOV W1,#2        |                ;
>  STP W1, W0, [X2] |                ;
> locations [z;]
> ~exists (1:X0=1 /\ 1:X1=0)
> ```
> 
> The test is runnable by litmus, as the declaration int z[2] suffices for the runtime skeleton to allocate the correct amount of memory with correct alignement. However z, z[0], z[1], cannot appear in the final conditions, which precludes writing tests as your A02 etc. unit tests. As a sometimes handy workaround, one can print the z array by the `locations` construct.
>

I agree that herd and litmus should handle the same tests, however I believe any notion of vectors is out of scope for this commit - I intend to look into handling vectors or collections properly in the future. 

This commit can show individual variables in the final state without the need for the `locations` workaround. This can be achieved like in the unittest examples that name x,y,z etc... in the final state. Later commits in this series will add regular loads etc... with numeric support which better demonstrates the notion that numeric addresses are independent of vectors.

This commit just introduces the notion of numeric address, and if the solver understands an address `0xf00` corresponds to a symbolic one `x`. This only occurs _only_ when the `address_of` operator is needed by data access instructions such as `ldr xd, [xm, #imm]` (here `#imm` must be added to some `0xf00` to gain `y=0xf00+imm` which we can encode in the initial state).

With this commit, we can also handle code stored relatively close to each other (0xf04, 0xf08, etc...) just by knowing each address, without enforcing a notion of spacial restraint - such as vectors. ie we don't need a notion of `allocate z=x+4` or vectors, just that `&x = 0xf00` and `&y=0xf04`.

Furthermore this would not reflect the architectural intent of programs, unless explicitly in the case where ordered collections are needed. For instance a compiler may compile `x=1; y=2` into two stores at `0xf00` and `0xf04` respectively. There is no relationship between these variables other than that the compiler infered it would be cheap to store them together in memory.
 
> One can also test the quadword form of the ldp and stp instructions:
> 
> ```
> Arch64 A00
> "stp vs. ldp, quad"
> 
>  {
>    int64_t z[2];
>    uint64_t 1:X1; uint64_t 1:X0;
>    0:X2=z; 1:X2=z;
> }
>  P0               | P1             ;
>  MOV X0,#1        | LDP X1,X0,[X2] ;
>  MOV X1,#2        |                ;
>  STP X1, X0, [X2] |                ;
> locations [z;]
> exists (1:X0=1 /\ 1:X1=0)
> ```
> 
> With respect to test `A00-INT`, we replace int with uint64_t and `W` registers by `X` registers. The type declaration of registers `uint64_t 1:X1` and `uint64_t 1:X0` are not really useful, apart from sparing us a warning.
> 
> I attach a run of the two tests on some ODROID-C2 machine.
> [LOG.txt](https://github.com/herd/herdtools7/files/4517641/LOG.txt)
> 
> Now how could we proceed ? May I suggest that you use litmus test meta-datas, _i.e._ additional information in the litmus file that tools are free to ignore.
> Meta-data are introduced by a keyword followed by "=". As an exemple, diy generators produce litmus sources with meta-data. Try ```diyone7 -arch AArch64 PodWW Rfe PodRR Fre`to see some, `Generator=...`` etc.
> 
> Now my first idea would by to express the numerical value of symbolic locations as meta-data: `Load=x:0x200` for instance. Another solution would be to keep you idea of two symbolic addresses x and z (ideally in some meta-data! for herd to interpret them as your `&x=..`) and to inform litmus (with appropriate meta-data!) that x and z have to be allocated together, with a distance that would be the difference of their numerical value. This would implies changing litmus a bit, nothing that cannot be managed.
> 

I don't think that numeric addresses should be put into metadata as numeric addresses are part of the semantics of the program itself and metadata is optional. see points below.

I agree that we should inform litmus that code should be stored together, but I think this is out of scope for now - see my above point about vectors.

ldp/stp instructions implicitly store data next to each other by design (x, x+64 for instance) we do not need any notion of collected data for now - out of scope until vectors are handled properly.

> Admittedly, my suggestion significantly departs from your initial proposal. But I do think we should strive at having tests that are food for both litmus and herd. One important benefit of this approach is debugging instruction semantics by comparison with actual hardware.
> 

I agree that tools should agree on the semantics, however this feature was motivated by the code emitted by llvm, gcc, and armclang, to be run on hardware and so I believe my approach is the correct one.

Notably program symbol tables emit numeric addresses (e.g `0xf00)` which are used during program execution on hardware, this is also what litmus does under the hood when it executes programs it compiles using gcc. This feature is explicitly modelling such an execution, with the intent to test real flows emitted by llvm such as :

```
{y=42; 0:X1=0}
ADR X0, label_pointing_to_0xf00
LDR X0, [X0, #imm] // 0xf00 + imm = y, load value at y
STR X0, [X1]
exists (0:X1=42)
```

This commit makes explicit in the initial state the addresses that are used in such an execution, and as such they are part of the semantics of the compiled program itself - I don't think it's right to put numeric addresses in metadata where they can be ignored.

Additionally, I should mention at the point where data accesses are executed in herd we are using the symbolic location `y` that corresponds to `0xf00+imm` above. the address_of/deref operators are simply there to facilitate code that uses numeric addresses.

> Finally, I'd like to observe that LDP/STP can be implemented on top of symbolic locations, as indexed locations (such as x+4, or x+8) are handled naturally by herd. Hence as far as LDP/STP are concerned the address_of and deref operations are not necessary, strictly speaking.

Symbolic addresses still work if provided, however I believe that address_of/deref is needed. As noted in the initial commit message, symbolic addresses (such as x) are used if a numeric address isn't present so we still retain the behaviour of x+4 etc... as a fallback. This commit adds numeric addressing so we can handle code that is emitted by compilers as well, but we retain the existing behaviour too. This means all the historical litmus tests should still _just work_.

For example, if we know 
`&x + 4` and we know `&x=0xf00` then the solver will return `0xf00+4`
however
if we don't know `&x= 0xf00`the standard behaviour of `x+4`  

We obtain both the existing symbolic behaviour and the concrete behaviour from compilers. If we exclude `address_of` and `deref` then we would need two different semantics for numeric addresses and symbolic addresses for each instruction. These unops seem to be the cleanest way to retain the semantics in the same place and get both behaviours.

Finally, in future commits I shall be adding infrastructure to automatically generate litmus tests from compiled code, this infrastructure automatically constructs litmus tests with the initial state populated with the necessary addresses. This might give you an idea of what kinds of things I am trying to test here

herd/AArch64Sem.ml
lukeg101: I think we are talking about the same thing. but as bytes, versus bits, ie `sizeof(byte) * tr_variant(V64) = 8 * MachSize.Quad = 8 * 8 = 64` and `sizeof(byte) * tr_variant(V32) = 8 * MachSize.Word = 8 * 4 = 32`

I think we could simplify this like you say, perhaps as one single function `incr_a` which takes the variant as an argument and can be inlined as a monadic op such as `M.unitT a >>| (M.address_of a >>= incr_a var >>= M.deref)` in stp

maranget: Hum, I was not clear I guess, this how i see things:

I thought the address "granule" is always the byte. Hence +1 means addressing next byte. +4 then next int32_t etc.

Hence, with concrete value of location x being say, 0x100, with X1=x initially
LDP W0,W1,[X2]
will load the contents of location 0x100 into X0 and the contents of location 0x104 into X1.

Hence, the increment to effective address performed by ``ldp`` would be 4.

maranget:
Ok I think I now understand why litmus compatibility is not urgent, the (hopefully temporary) incompatibility with litmus being more than compensated by the compatibility with compiler generated assembly. I will look at the code itself tomorrow.

I am also glad that symbolic behaviour will remain the default, when no concrete values of locations are given. As, a consequence I guess it is possible to consider (all symbolic) unit test that will rune both on herd and litmus, similar to the modified A02 test above. I'll also try this soon.

herd/mem.ml
```diff
@@ -44,7 +44,7 @@ module type S = sig
       loop_present : bool ;
      }
 
-  val glommed_event_structures : S.test -> result
+  val glommed_event_structures : S.test -> ((string * string) list) -> result
```
maranget, line 47: External parenthesis of second tye argument can be removed
```suggestion
  val glommed_event_structures : S.test -> (string * string) list -> result
```

herd/valconstraint.ml
```diff
@@ -32,6 +32,8 @@ module type S = sig
 
   type expr =
     | Atom of atom
+    | Address of atom
+    | Deref of atom
```
maranget, line 36: I am wondering: would'nt be possible to have  the address and deref operations implemented as unary operation. That is extend type op1 in lib/op.{mi,mli} with constructors Address and Deref, then encode your `Address at` by `Unop (Address,at)`.
This would simplify some functions a bit.

herd/valconstraint.ml
```diff
@@ -219,18 +240,81 @@ and type state = A.state =
...
+    (* This step simplifies variables to their numeric addresses *)
+    (* and dereferences addresses to their symbolic variables *)
+    (* We recursively resolve addresses as much as we can until no changes*)
+    (* are made to the constraints *)
+    let rec simplify_addr_in_cnstrnts cs =
+      let bcs = List.map (simplify_addr_in_cnstrnt cs) cs in
+      if List.exists fst bcs then
+        simplify_addr_in_cnstrnts (List.map snd bcs)
+      else
+        List.map snd bcs
```
maranget, line 311: As I understand the Inv constraint are rigid (_i.e._ do not change). Hence it can be slightly more efficient to extract them from the constraint list before simplifying addresses.
```suggestion
    let simplify_addr_in_cnstrnts cs =
      let invs,cs = List.partition (fun c -> match c with Inv _ -> true | _ -> false) cs in
      
      let rec do_simpiify cs = 
        let bcs = List.map (simplify_addr_in_cnstrnt inv) cs in
        if List.exists fst bcs then
          do_simplify (List.map snd bcs)
      else
        List.map snd bcs in
        
      inv@do_simplify cs
```
Another solution, which would imply more changes, would be to replace the constraint list by a pair or record that would keep invariants and standard constraints separated.

lib/testHash.ml
```diff
@@ -59,7 +60,8 @@ let digest_init debug init =
       init in
   let init =
     Misc.rem_dups
-      (fun (loc1,_) (loc2,_) -> location_compare loc1 loc2 = 0)
+      (fun (loc1,(t1,_)) (loc2,(t2,_)) -> location_compare loc1 loc2 = 0
+                                        && not (is_address t1 <> is_address t2))
```
maranget, line 64: Again, double negation can be shunted.

lib/testHash.ml
```diff
@@ -48,7 +48,8 @@ let digest_init debug init =
         | 0 ->
             if
               ParsedConstant.compare v1 v2 <> 0 &&
-              compare t1 t2 <> 0
+              compare t1 t2 <> 0 &&
+              (not (is_address t1 <> is_address t2));
```
maranget, line 52:
```suggestion
              is_address t1 = is_address t2
```
Incidentally, the final semicolon is accepted  and immediately forgotten by the `ocaml` parser. That is expression `e;` simply is `e`, except in sequence `e1; e2`. Something I have always found very bizarre.

litmus/compile.ml
```diff
@@ -57,6 +57,7 @@ module Generic (A : Arch_litmus.Base)
         | MiscParser.TyDefPointer  -> pointer
         | MiscParser.Ty t -> Base t
         | MiscParser.Atomic t -> Atomic (Base t)
+        | MiscParser.Address -> base
```
maranget, line 60: This can be problematic in the future, as base is `int`, while addresses can take concrete values up to 64bits. However, as  addresses numerical values should probably be ignored by litmus, it probably does not harm. Perhaps using `intptr_t` ?

litmus/compile.ml
```diff
@@ -104,6 +105,7 @@ module Generic (A : Arch_litmus.Base)
                     | MiscParser.TyDefPointer -> Some pointer
                     | MiscParser.Ty s -> Some (Base s)
                     | MiscParser.Atomic s -> Some (Atomic (Base s))
+                    | MiscParser.Address -> None
```
maranget, line 108: May be `assert false` or `Warn.fatal ...`. Not important. This "typing" code is not as armored as it should anyway.

herd/AArch64Sem.ml
lukeg101: I see what you mean now, yeah I will fix that

herd/valconstraint.ml
lukeg101: it's difficult to say, on the one hand these are special unary operations in eventsMonad, and we add invariants corresponding to the address of each location`&x = 0xf00`. However these operations should be used only in data access instructions like ldr etc... and require the inclusion of a symbol table (`symbs` in mem.ml) so they are not normal.

They act like normal unary operations, but they need special treatment with the symbol table. If we were to treat them as unary operations,  then we would need to pass the symbol table into op/symbValue etc... I think it's simpler, and more elegant, to restrict the scope of the symbol table to the initial state in litmus tests, and the frontend of herd where such constraints are parsed and sent to the solver only. 

This would prevent numeric addresses from leaking into the symbolic part of herd as well which would prevent accidental conflation of symbols/addresses in the rest of herd.

herd/valconstraint.ml
```diff
@@ -42,7 +44,9 @@ module type S = sig
 
   type cnstrnt =
     | Assign of atom * rvalue
+    | AssignAddr of atom * rvalue (* Handle locality and relativity, Einstein approved*)
     | Unroll of string (* unrolling stopped *)
+    | Inv of cnstrnt (* These are Invariants*)
```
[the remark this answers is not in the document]
lukeg101: I agree with you here, will fix

herd/mem.ml
lukeg101: good spot, will fix

herd/valconstraint.ml
```diff
@@ -219,18 +240,81 @@ and type state = A.state =
...
     | Assign (v,e) ->
         let v = subst_atom m v
         and e = subst_expr m e in
         Assign (v,e)::k
-    | Unroll _ -> cn::k
+    | AssignAddr (v,e) ->
+        let v = subst_atom m v
+        and e = subst_expr m e in
+        AssignAddr (v,e)::k
+    | Unroll _ | Inv _ -> cn::k
```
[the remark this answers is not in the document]
lukeg101: agreed - will remove `Inv` altogether and replace `Inv _` with `AssignAddr _`

herd/valconstraint.ml
```diff
@@ -219,18 +240,81 @@ and type state = A.state =
...
 
     let subst_cns soln cns = List.fold_right (subst_cn soln) cns []
 
+    let match_addr v1 = function
+      | Inv (AssignAddr (v2, Atom v)) when v1 = v2-> Some v
```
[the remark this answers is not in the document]
lukeg101: good spot, will fix

herd/valconstraint.ml
```diff
@@ -219,18 +240,81 @@ and type state = A.state =
...
+      | _ -> None
+
+    let match_loc v1 = function
+      | Inv (AssignAddr(v, Atom v2)) when v1 = v2-> Some v
```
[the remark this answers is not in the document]
lukeg101: ditto

herd/valconstraint.ml
```diff
@@ -219,18 +240,81 @@ and type state = A.state =
...
+        let vs = List.filter_map (match_addr a) cs in
+        if List.length vs == 0 then begin
+          (* If we can't find the address for a, and it is concrete,
+           * or symbolic (but not known) *)
+          if V.is_concrete a || V.is_symbol a then
+            (false, Atom a)
+          else
+            (false, Address a)
+        end else
+          (true, Atom (List.hd vs)) in
```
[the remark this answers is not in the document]
lukeg101: when we collect addresses in the initial state we use the same `Set` library to collect unique assignments to addresses so we should never have more than 1 numeric address for a location - the `_::_::_` should never occur.

More generally anything other than `[]` or `[v]` should fail and so I can add a failure case for just `_ -> Warn.fatal "Multiple ... "`

Perhaps a better data structure would work here, using `List.find_map` to return the first element that succeeds, or returning `None`.

What do you think?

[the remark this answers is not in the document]
lukeg101: more specifically something that returns either nothing or one_element is the safest like you say, hmm. I'll do what your suggestion says

herd/valconstraint.ml
```diff
@@ -219,18 +240,81 @@ and type state = A.state =
...
+
+      let find_loc a =
+        let vs = List.filter_map (match_loc a) cs in
+        if List.length vs == 0 then
+          if V.is_symbol a then (false, Atom a) else (false, Deref a)
+        else
+          (true, Atom (List.hd vs)) in
```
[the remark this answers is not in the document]
lukeg101: also fixed

herd/valconstraint.ml
lukeg101: good point, I will fix this, but I wonder if the if statement could be expressed in a more idiomatic fashion as well?
I've added your fix and will think about the if

lib/testHash.ml
lukeg101: yeah this annoys me more than it should - it makes parser construction much harder than it needs to be. Will fix

lib/testHash.ml
lukeg101: fixed

litmus/compile.ml
lukeg101: changed to `pointer`

litmus/compile.ml
lukeg101: ok added

lukeg101:
Hi Luc, sorry about the force-push but I have managed to push to this fork from my branch. I have addressed everything above but the main points are:
- fixed ldp/stp semantics to use byte level granularity, corrected unittests
- fixed valconstraints issues, refactored code
- cleaned up other ticket issues

herd/valconstraint.ml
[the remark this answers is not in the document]
maranget: Will see that once I can look at the new code.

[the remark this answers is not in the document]
maranget: Well I am not that sure since your solution, although a bit less efficient performs a sanity check.

Anyhow, there is such a function: [`List.find_opt`](https://caml.inria.fr/pub/docs/manual-ocaml/libref/List.html#VALfind_opt) :
```
let rec find_opt p = function
| [] -> None
| x::xs ->
    if p x then Some x else find_opt p xs
```
There is also [`List.find`](https://caml.inria.fr/pub/docs/manual-ocaml/libref/List.html#VALfind) that raises exception `Not_found` when search fail.

maranget:
Hi Luke, never mind the forced push, It did not introduce confusion. I am ready to merge. I have answered to what I think was a question of yours in the last unresolved comment.

maranget:
Merged, waiting for the next 101 PR's

PR #30, opened 2020-04-27
Telechat

lukeg101:
This Pull Request upstreams assembly support and additional features. Unittest tests are provided where appropriate, and otherwise each commit is self-contained with a description. In summary the commits are: 
- LDP/STP: Don't know why this is here again
- merge: probably to fix the above (and the previous force push)
- as HINTs as NOPs in AArch64
- add swp file to gitignore, useful when you have multiple windows open during dev
- added addret flag to generate parsable C++ code
- Added SP as a special register in AArch64
- made check_name public so it can be reused in other AArch64 parsers
- added useful helper function uniq_lists to myset
- added \_ support to mcompare to support more names
- added LDR literal instruction to AArch64
- Added MOVZ instruction to AArch64 
- Added/updated ADRP/ADR instruction to AArch64 - does not check 4kb page boundary


lib/AArch64Base.ml
```diff
@@ -607,7 +611,7 @@ let do_pp_instruction m =
   | I_OP3 (v,op,r1,r2,kr) ->
       pp_rrkr (pp_op op) v r1 r2 kr
   | I_ADDR (r,lbl) ->
-      sprintf "ADDR %s,%s" (pp_xreg r) (pp_label lbl)
+      sprintf "ADR %s,%s" (pp_xreg r) (pp_label lbl)
```
maranget, line 644: Well done.

herd/dune
```diff
@@ -5,5 +5,5 @@
    herd
    )
    (public_names herd7)
-   (libraries herdtools)
+   (libraries herdtools str)
```
maranget, line 8: I'd rather avoid this build dependency on str , if possible.

lib/AArch64Base.ml
```diff
@@ -33,7 +36,7 @@ type gpr =
   | R16 | R17 | R18 | R19
   | R20 | R21 | R22 | R23
   | R24 | R25 | R26 | R27
-  | R28 | R29 | R30
+  | R28 | R29 | R30 | RSpecial
```
maranget, line 39: Why not calling the register SP ?

herd/AArch64Sem.ml
```diff
@@ -299,6 +318,30 @@ module Make
...
+          let v = V.op1 (Op.LeftShift s) (V.intToV k) in
+          M.op Op.Or v V.zero
+        | _, Some(LSL(n)) -> Warn.fatal
+                   "illegal shift immediate %d in %s instruction movz"
+                   n
+                   (pp_variant sz)
+        | _, Some(s) -> Warn.fatal
+                   "illegal shift operand %s in %s instruction movz"
+                   (pp_barrel_shift s pp_imm)
+                   (pp_variant sz)
```
maranget, line 340: It may be clearer to users to print the wrong instruction (with AArch64.dump_instruction).

lib/AArch64Base.ml
```diff
@@ -296,6 +300,10 @@ let inverse_cond = function
 type op = ADD | ADDS | SUB | SUBS | AND | ANDS | ORR | EOR
 type variant = V32 | V64
 
+let pp_variant = function
+  | V32 -> "V32"
+  | V64 -> "V64"
+
```
maranget, line 306: Is this function useful ?

lib/AArch64Lexer.mli
```diff
@@ -18,4 +18,5 @@
 
 module Make : functor(O:LexUtils.Config) -> sig
   val token : Lexing.lexbuf -> AArch64Parser.token
+  val check_name : string -> AArch64Parser.token
```
maranget, line 21: I am wondering, why export check_name ?

litmus/AArch64Compile_litmus.ml
```diff
@@ -118,6 +118,16 @@ module Make(V:Constant.S)(C:Config) =
...
     let str_memo t = Misc.lowercase (str_memo t)
+    let load_literal tr_lab memo v rD lbl = match v with
+      | V32 ->
+          { empty_ins with
+            memo = sprintf "%s ^wo0, %s" memo (A.Out.dump_label (tr_lab lbl));
+            outputs=[rD]; reg_env=[(rD, word)]; }
+      | V64 ->
+          { empty_ins with
+            memo = sprintf "%s ^o0, %s" memo (A.Out.dump_label (tr_lab lbl));
+            outputs=[rD]; reg_env=[(rD, word)]; }
```
maranget, line 129: Should not "word" be replaced by "quad"?

litmus/AArch64Compile_litmus.ml
```diff
@@ -431,7 +441,32 @@ module Make(V:Constant.S)(C:Config) =
           memo=sprintf "%s %s,%s" memo f1 f2;
           inputs = r2; outputs=r1; reg_env=add_q (r1@r2);}
 
+    let do_mov memo v rd k os = match v, k, os with
```
maranget, line 444: Nipicking:  "do_movz" would be a better name.

litmus/AArch64Compile_litmus.ml
```diff
@@ -431,7 +441,32 @@ module Make(V:Constant.S)(C:Config) =
...
+        { empty_ins with
+          memo=sprintf "%s %s, #%d, LSL #%d" memo f1 k s;
+          outputs=r1; reg_env=add_q r1;}
+    | V64, K k, None ->
+        let r1,f1 = arg1 "xzr" (fun s -> "^o"^s) rd in
+        { empty_ins with
+          memo=sprintf "%s %s, #%d" memo f1 k;
+          outputs=r1; reg_env=add_q r1;}
+    | _ -> Warn.fatal "Illegal form of %s instruction" memo
+
```
maranget, line 466: Barrel shifter argument handling could probably be factored out. This will prove convenient if one add the other variants (LSR etc)  later.

herd/AArch64Sem.ml
```diff
@@ -283,6 +294,14 @@ module Make
 
       and stlr sz rs rd ii = do_str sz AArch64.L rs (read_reg_ord rd ii) ii
 
+      (* Load literal *)
+      and ldr_lit rd lbl ii =
+        (* We do not use the reg size as we load a word *)
+        let open AArch64Base in
+        M.deref (parse_lbl lbl)
+        >>= fun a -> read_mem MachSize.Word rd a ii
+        >>! B.Next
```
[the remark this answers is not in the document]
lukeg101: Yes you're right, I read the definition of load literal from [1] literally - the pseudocode is better here. Upon reading the pseudocode it's clear we should use the variant to pick the size as you suggest.

What bug do you mean? is it related to any work I've done? is there anything I can do other than pulling from master?

[1] https://developer.arm.com/docs/ddi0596/e/base-instructions-alphabetic-order/ldr-literal-load-register-literal

herd/AArch64Sem.ml
```diff
@@ -490,9 +535,15 @@ module Make
...
-
+        | I_MOVZ(var,rd,K k,os) ->
+            movz var rd k os ii
+        | I_MOVZ(_,_,_,_) ->
+            Warn.fatal "Illegal argument in movz, expecting imm16 only"
+        
         | I_ADDR (r,lbl) ->
-            write_reg r (V.nameToV lbl) ii >>! B.Next
+            write_reg r (parse_lbl lbl) ii >>! B.Next
+        | I_ADRP (r,lbl) ->
```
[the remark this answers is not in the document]
lukeg101: good point, will combine

lib/AArch64Base.ml
```diff
@@ -418,6 +439,7 @@ type 'k kinstruction =
   | I_STOPBH of  atomic_op * bh * w_type  * reg * reg
 (* Operations *)
   | I_MOV of variant * reg * 'k kr
+  | I_MOVZ of variant * reg * 'k kr * 'k s option
```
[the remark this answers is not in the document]
lukeg101: good point, will modify

lib/AArch64Base.ml
lukeg101: I can't remember why I did this, this commit is old, will fix

herd/AArch64Sem.ml
lukeg101: good idea, will fix

lib/AArch64Base.ml
lukeg101: it is used in the error message in a later patch for `I_OP3`

lib/AArch64Lexer.mli
lukeg101: Later I shall use it to parse assembly using a different grammar parsed to Menhir. 
Details of how Menhir does this comes under _partial specifications_ in the Menhir manual http://gallium.inria.fr/~fpottier/menhir/manual.pdf

lukeg101: ie this allows us to reuse assembly mnemonics elsewhere

litmus/AArch64Compile_litmus.ml
lukeg101: good spot, will fix

litmus/AArch64Compile_litmus.ml
lukeg101: will fix

litmus/AArch64Compile_litmus.ml
lukeg101: good idea, will fix

herd/AArch64Sem.ml
```diff
@@ -171,6 +171,17 @@ module Make
       let read_mem_acquire_pc sz = do_read_mem sz AArch64.Q
       let read_mem_noreturn sz = do_read_mem sz AArch64.NoRet
 
+      (* We calculate addresses from labels L0xf00 -> 0xf00 *)
+      (* This is needed to for literal instructions *)
+      let parse_lbl lbl =
+        (* We should use more robust types for this in the future *)
+        let addr = Str.regexp "L.*" in
```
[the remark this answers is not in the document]
lukeg101: if the search fails on `Lout` it should fallback to parse the label as a non-numeric address (such as `x`) in the else statement, this gives us both a means to parse numeric addresses and non-numeric.

I will add `user_error` to change `int_of_string` 

Do you know an alternative to the use of `Str` as I use it in one or two other places in a later patch?

herd/AArch64Sem.ml
```diff
@@ -299,6 +318,30 @@ module Make
                   read_mem_acquire_pc sz rd a ii)
           (read_reg_ord rs ii) ii
 
+      let movz sz rd k os ii =
+        let open AArch64Base in
+        let is_imm16 n = n >= 0 && n < 65535 in
```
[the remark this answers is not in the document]
lukeg101: good idea, will try to add to machsize

herd/AArch64Sem.ml
```diff
@@ -171,6 +171,17 @@ module Make
...
       let read_mem_noreturn sz = do_read_mem sz AArch64.NoRet
 
+      (* We calculate addresses from labels L0xf00 -> 0xf00 *)
+      (* This is needed to for literal instructions *)
+      let parse_lbl lbl =
+        (* We should use more robust types for this in the future *)
+        let addr = Str.regexp "L.*" in
+        if Str.string_match addr lbl 0 then
+          V.intToV (int_of_string ("0x"^String.sub lbl 1 (String.length lbl -1)))
+        else
```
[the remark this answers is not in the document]
lukeg101: will modify to parse both hexadecimal and decimal numbers

herd/dune
lukeg101: do you know an alternative for simple string checking?

herd/AArch64Sem.ml
```diff
@@ -299,6 +318,30 @@ module Make
...
           (read_reg_ord rs ii) ii
 
+      let movz sz rd k os ii =
+        let open AArch64Base in
+        let is_imm16 n = n >= 0 && n < 65535 in
+        assert (is_imm16 k);
+        begin match sz, os with
+        | V32, None | V64, None ->
+          (* Or'ing zero with value should zero out what's left *)
+          M.op Op.Or (V.intToV k) V.zero
```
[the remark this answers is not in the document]
lukeg101: good point, will remove

lukeg101:
Addressed all comments - details are in the new commit message. Some important details about how to parse labels are in there.

Let me know if there's anything else, or indeed if you need me to rebase from master to fix that bug

herd/AArch64Sem.ml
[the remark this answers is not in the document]
maranget: Well, in fact herd has two modes, mixed-size, where memory accesses are divided into smaller memory accesses of a fixed "byte" size, and non-mixed where accesses w.r.t. a given location should be of the same size. Hence, giving the right size is important in both mode.

Besides all computations are performed on int64's and the OCaml state also holds 64 bit integers, hence the importance of masking  (everywhere for robustness), There may be forgotten maskings, I have added some in master after the review.

herd/AArch64Sem.ml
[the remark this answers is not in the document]
maranget: Fine,

herd/AArch64Sem.ml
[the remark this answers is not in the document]
maranget: Ack

herd/AArch64Sem.ml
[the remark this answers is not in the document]
maranget: Ack

lukeg101:
That's good to know, thanks. I have removed Str altogether in the previous commit - it's easier. Is there anything else I need to do now?

maranget:
Hi Luke, thanks the changes, I have made a few comments on the new commits. Nothing serious. I am waiting for your reaction before merging.

herd/AArch64Sem.ml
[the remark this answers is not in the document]
lukeg101: ok

herd/AArch64Sem.ml
```diff
@@ -171,15 +171,13 @@ module Make
...
       (* This is needed to for literal instructions *)
       let parse_lbl lbl =
         (* We should use more robust types for this in the future *)
-        let addr = Str.regexp "L.*" in
-        if Str.string_match addr lbl 0 then
-          V.intToV (int_of_string ("0x"^String.sub lbl 1 (String.length lbl -1)))
-        else
-          V.nameToV lbl
+          try V.intToV (int_of_string lbl)
+          with _ -> V.nameToV lbl
```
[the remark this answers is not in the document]
lukeg101: Good idea, I shall fix this. Sorry I cut a corner here, I normally try to avoide using exceptions and use total functions after programming in Haskell for so long

herd/unittests/AArch64/A17.litmus
```diff
@@ -4,6 +4,6 @@ AArch64 A17
 { &x=2293760; x=42}
 
 P0;
-  LDR X0, L230000;
+  LDR X0, 0x230000;
 
 exists (0:X0=42)
```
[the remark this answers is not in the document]
lukeg101: Apologies I got these swapped in the commit message - the unit tests have the right behaviour. 

However I still worry that this specific example isn't part of the arm arm as there is no variant with a number as the second operand. What do you think? we need this behaviour as it is the kind of thing compilers emit like:
```
LDR X0, 0x230000 <__DYNAMIC_FOO+0xdeadbeef> //loads numeric address 0x230000 into X0
```
Similarly, this patch allows us to do
```
LDR X0, x // loads address x into X0
```
both examples are at the boundary of the arm arm and the symbolic nature of herd, but the former isn't strictly in the arm arm but useful

herd/AArch64Sem.ml
```diff
@@ -320,24 +318,19 @@ module Make
...
+        | V32, NOEXT | V64, NOEXT ->
           (* Or'ing zero with value should zero out what's left *)
-          M.op Op.Or (V.intToV k) V.zero
-        | V32, Some(LSL(0|16 as s))
-        | V64, Some(LSL((0|16|32|48 as s))) ->
+          M.unitT (V.intToV k)
+        | V32, LSL(0|16 as s)
+        | V64, LSL((0|16|32|48 as s)) ->
           let v = V.op1 (Op.LeftShift s) (V.intToV k) in
           M.op Op.Or v V.zero
```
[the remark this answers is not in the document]
lukeg101: will fix

herd/AArch64Sem.ml
```diff
@@ -320,24 +318,19 @@ module Make
...
-        | _, Some(s) -> Warn.fatal
-                   "illegal shift operand %s in %s instruction movz"
-                   (pp_barrel_shift s pp_imm)
-                   (pp_variant sz)
+        | _, LSL(_) | _, _ ->
+            Warn.fatal
+              "illegal instruction %s"
+              (dump_instruction (I_MOVZ (sz, rd, K k, os)))
         end
           >>= (fun v -> write_reg rd v ii)
```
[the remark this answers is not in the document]
lukeg101: I'm not sure what you mean here, we can still shift a 16-bit value by 16 bits, which is stored in a 32-bit reg ie:
`0000000000000001` ->_{16} `00000000000000010000000000000000` when stored in a wreg. ie I think the shift is applied as it is loaded into the register

lukeg101:
Thanks for your help Luc, I've addressed what I can but there are 2 open questions still

herd/AArch64Sem.ml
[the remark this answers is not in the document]
maranget: We agree.

maranget:
Thanks Luke. See you soon for the next PR....

PR #33, opened 2020-05-18
Telechat: C type support

lukeg101:
This patch upstreams a 2 small features, before I spend some time to implement vectors in Herd. In detail this patch provides support for: 
- a `NoVolatile` variant to diy, in the case we want to test litmus tests without volatile constraints.
- adding C Parser support (no sem) for the `const` qualifier in C.


lib/CType.ml
```diff
@@ -21,6 +21,7 @@ type base = string
 type t =
   | Base of base
   | Volatile of t
+  | Const of t
```
maranget, line 24: Adding the `Const` constructor impacts many functions of this module, which should strip  the new attribute, as they strip `Volatile` and `Atomic`. Those are `is_ptr`, `is_array`, `is_atomic`, strip_atomic`, `strip_volatile`, `strip_attributes` and `is_ptr_to_atomic`. Here is for instance haw `is_ptr` can be rewritten:
```
let rec is_ptr = function
| Atomic t|Volatile t|Const t -> is_ptr t
| Pointer _ -> true
| _ -> false
```

Obviously the initial code is not 1/ very robust 2/ poorly written initially, (adding a constructor should fire so much changes).
 1.  Underscores can be replaced by completing or-patterns. For instance the last clause of `is_ptr` can be:
```
| Array _|Base _ -> false
```
2. More involved, one can think about having a `Attribute of attribute * t` constructor (with `type attribute = Volatile|Atomic|Const`), and/or restructuring code (_e.g._ the "is_*" functions could call strip_attributes.

litmus/skelUtil.ml
```diff
@@ -47,7 +47,7 @@ let dump_global_type loc t = match t with
 
 let rec nitems t = match t with
 | Array (_,sz) -> sz
-| Volatile t|Atomic t -> nitems t
+| Volatile t|Atomic t|Const t -> nitems t
 | Base _|Pointer _ -> 1
```
maranget, line 51: That one was properly written and the warning draw you attention onto the problematic situation to correct...

lukeg101:
Hi Luc, I did a quick clean of the CType code, and related code that uses e.g `strip_volatile`. I strongly suspect this file will be rewritten at some point when we have a more powerful C frontend

maranget:
> Hi Luc, I did a quick clean of the CType code, and related code that uses e.g `strip_volatile`. I strongly suspect this file will be rewritten at some point when we have a more powerful C frontend so I didn't do too much

I agree, I merge.

maranget:
Perhaps you can find a way not to re-introduce old previous commits in PR ?

PR #40, opened 2020-07-14
Unaligned to master integration

jalglave:


maranget:
Merging, we should had a few examples.

PR #42, opened 2020-07-28
Add string field to Fault

ambroise-arm:
This field is used to differentiate different kinds of faults that can
happen on the same memory location.


maranget:
Hi, we are currently working for various tools (e.g. msum, mcompare...) to handle faults, so I'd rather not modify the type of falults at the moment. Would it be possible to delay your pull request?

jalglave:
hi Luc, please can we consider this one now?
Thanks,
Jade

maranget:
> hi Luc, please can we consider this one now?
> Thanks,
> Jade

It would be nice if the changes had no impact on tools, one simple solution would be to ignore the extra messages while parsing the logs for tools.  That said, the tools in master are not handling faults at the moment. Branch KVM does.

PR #43, opened 2020-07-28
[tools] Handle faults in tools

maranget:


maranget:
Too early to merge...

PR #44, opened 2020-08-06
[AArch64 MTE] Fix tag init on DpAddr and DpData

ambroise-arm:
Currently, using diy with address or data dependencies generates litmus tests where some addresses have no tag. From what I have observed it doesn't change anything functionnally, as the missing tags are the default ones, but there may be other cases.
In any case, having all addresses be tagged seems better for consistency.


jalglave:
hi Luc, please can we consider this one now?
Thanks,
Jade

PR #45, opened 2020-08-24
Lockref

jalglave:


lib/AArch64Parser.mly
```diff
@@ -180,6 +183,12 @@ cond:
 | LE { A.LE }
 | LT { A.LT }
 
+label_addr:
+| HEX LABEL { Printf.sprintf "0x%x" $1 }
+| HEX       { Printf.sprintf "0x%x" $1 }
+| NUM LABEL { Printf.sprintf "%d" $1 }
+| NUM       { Printf.sprintf "%d" $1 }
+| NAME      { $1 }
```
maranget, line 205: Why introducing numerical labels? Cannot TBNZ and TBZ have a NAME last argument as other branch instructions do?

jalglave:
hi Luc, 
thanks I just pushed a new version - please can you have a look?

Thanks,

Jade

maranget:
Ok fine, I merge. Notice that having hexadecimal constants is a good idea. It could have been done like in stateLexer.mll.

PR #50, opened 2020-09-11
Add initial Pre-Commit config

eth-arm:
This PR:

- Adds a [Pre-Commit](https://pre-commit.com) config to enforce properties that the repository already meets.
- Adds a short howto for contributors.
- Applies some fixes to do with executable text files.
- Adds a Pre-Commit check that executable text files have shebangs (e.g. `#!/bin/sh`).


dune-build.sh
```diff
@@ -1,3 +1,5 @@
+#!/bin/sh
+
```
maranget, line 2: No set -eu here?

maranget (commented):
The ALL.sh and NR.sh shell scripts can be deleted.

I am wondering how so many files had spurious executable permission...

maranget:
Hi Eth,

I have made two quick comments, one of which is a real question.

Thanks,

dune-build.sh
eth-arm: `$EXE` is unbound, because it is commented-out in [dune-defs.sh](https://github.com/herd/herdtools7/blob/0607078b1bddcaa9e74dcae5168fb816652e102e/dune-defs.sh#L21). The same unbound `$EXE` exists in [dune-uninstall.sh](https://github.com/herd/herdtools7/blob/0607078b1bddcaa9e74dcae5168fb816652e102e/dune-uninstall.sh#L28), so I was going to fix those together in a separate PR.

maranget: Fine, I merge.

PR #52, opened 2020-09-11
Fix whitespace and trailing newlines for all Litmus, Cat, and CFG files

eth-arm:
This PR removes whitespace at the ends of lines and normalizes newlines at the ends of files for all *.litmus, *.cat, and *.cfg files.


maranget:
I agree that cleaning trailing whitespace is a good idea. I also agree with cleaning the catalogue (merged)

I wish we could do the same for source files, however I am afraid of adding useless conflicts with active branches (eg KVM). May I have your opinion on this?

PR #57, opened 2020-09-22
Run `make all` at pre-commit when changing OCaml & Dune files

eth-arm:
This should help ensure that changes being committed definitely don't break anything, but allowing e.g. documentation to be updated without triggering a rebuild.

From here, I want to add some tests, then add a similar pre-commit hook to run those via `make test` or similar.


maranget:
Let us try this.

PR #58, opened 2020-09-22
Merging telechat base branch with fork to prepare for more commits

lukeg101:
This doesn't add new code, just removes the annoying merge commit so that the next pull request doesn't have it


maranget:
Hi Luke, I do not quite understand. Does this PR regroup the previous ones, saved for a merged commit? In that case, the code has already been reviewed.

Moreover rebasing this PR directly is not possible due to conflicts. However it can be squashed and merged. Is this the intention?

lukeg101:
> PR regroup the previous ones
yeah, everytime I make a PR from my fork of herdtools I fetch + merge the telechat branch, this means I have to make this merge and I thought I would separate this commit for existing work, and then commit other patches ready to go on top.

Do you know a better way for me to do this to prevent the need for me to do this after every PR? I agree this will be painful when we eventually rebase onto master...

maranget:
> Do you know a better way for me to do this to prevent the need for me to do this after every PR? I agree this will be painful when we eventually rebase onto master...

A better way than squashing and merging?

lukeg101:
I'm worried about squashing because there are 45 commits on this branch so far, and when we rebase onto master that will be pretty difficult/long to ensure we don't lose anything

maranget:
I do not know enough to be sure. I have followed the 'command line instructions' (basically merge -ff of the new branch into telechat). The final result does not look so bad : The above list + one additional 'merge commit. Notice that the history looks non-linear.

jalglave:
Hi Luke, I believe we've now settled on another approach by email. If yes please may I close this PR?
Thanks,
Jade

PR #59, opened 2020-09-24
[build] Refactor ocb-* and dune-* build scripts

eth-arm:
A summary of changes in this PR:

- Adding `set -eu` to the top of all scripts, so that they fail if any commands fail (`-e`) or if any variables are unbound (`-u`).
- Making various variables `readonly` to ensure they can't be overridden later.
- Wrapping many variables in `""` to make the script safe for paths with whitespace, e.g. `/tmp/foo bar`.
- Merging `dune-defs.sh` into `defs.sh`.
- Removing dead code & duplication.

All existing behavior should be maintained. For example, I did not make `ocb-build.sh` call `version-gen.sh` because it has different behavior for generating `Version.ml`.

I have standardized on tabs instead of spaces, but could change that to standardize on spaces if you would prefer.

This is leading toward automated testing of building with OCamlbuild, and automated regression testing.


maranget:
Eth, I'd rather not to have tabs in files. Would you please revert to spaces.
--Luc

eth-arm:
I have changed tabs into 2 spaces.

maranget:
Fine merged, thanks.

PR #62, opened 2020-09-28
[herd]: Added partial support for AArch64 LDP/STP

lukeg101:
Following the review of https://github.com/herd/herdtools7/pull/29

This patch upstreams support for Load/Store pair instructions for
AArch64 in Herd. The patch linked above contains more code for numeric
addressing which has been removed here. In detail this patch does:

- partial support for Wreg load/store pairs, we cannot handle the xreg
  case just yet as it requires more work.

This patch does not:

- Add any kind of numeric addressing capabilities as alluded to in #29.
- Add any kind of support for vectors - or indeed a large 128-bit type.

This patch raises questions about what we do in the fully symbolic case
for:

STP Xn, Xm, [Xn] where Xn and Xm contain 64-bit values and Xn points to
a value

In herd the largest possible sized type is int64_t and so storing Xn
causes out-of-bounds errors as we are writing 128-bits worth of data. We
have a solution to this in the works, but for now we can safely upstream
this patch so that it works for wregs. It may be preferable to add a
pattern for 'unsupported xregs' until we upstream the solution for xregs.

The same reasoning applies to LDP instructions

### DIFF
--- a/herd/AArch64Sem.ml
+++ b/herd/AArch64Sem.ml
@@ -248,6 +248,17 @@ module Make
         else
           mop ma >>! B.Next
 
+      (* Helper function to increment address for ldp/stp*)
+      let incr_a var =
+        let n = V.intToV (MachSize.nbytes (AArch64.tr_variant var)) in
+        M.add n
+
+      (* add immediate to initial base address *)                                    let add_imm =
+        function
+        | AArch64.K 0 -> M.unitT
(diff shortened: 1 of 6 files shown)


jalglave:
hi Luke, I believe we've now settled on another approach. If yes, please may I close this PR?
Thanks,
Jade

PR #63, opened 2020-09-29
Unify Dune & OCamlbuild built binaries

eth-arm:
Add dont7 to Dune, and mcat2includes7 to OCamlbuild.


maranget:
Ethel,  having discussed with Jade, we agree to deprecate `dont`. It would be nice if you could change the PR accordingly.
Thanks.

eth-arm:
I have removed the addition of `dont` to Dune from the first commit, and removed the `dont` binary itself in a second commit. It's still referenced in the documentation; should I also remove that?

maranget:
> I have removed the addition of `dont` to Dune from the first commit, and removed the `dont` binary itself in a second commit. It's still referenced in the documentation; should I also remove that?

Thanks, yes please remove any reference to `dont` from the documentation.

eth-arm:
I think I have removed `dont`/`auto` from the codebase.

PR #64, opened 2020-09-29
[herd] Add a flag -set-libdir to set the libdir

eth-arm:
Also remove Version_herd.ml, because libdir is now defined in opts.ml.

This means that the binary can be run as, e.g.

    % _build/herd/herd.native -set-libdir=./herd/libdir foo.litmus

Comparing with & without the flag:

    % _build/install/default/bin/herd7 -libdir
    /Users/ethmor02/share/herdtools7/herd
    % _build/install/default/bin/herd7 -set-libdir=./herd/libdir -libdir
    ./herd/libdir

### DIFF
--- a/herd/Version_herd.ml
+++ b/herd/Version_herd.ml
@@ -1,3 +1,0 @@
-include Version
-
-let libdir = Filename.concat libdir "herd"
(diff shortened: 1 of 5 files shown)


maranget:
Easy one for a saturday morning. Thanks.

PR #66, opened 2020-10-01
Added MOVZ operation with barrel shift operands

lukeg101:
following https://github.com/herd/herdtools7/pull/30

This patch upstreams support for the MOVZ operation in AArch64
with logical shift left (LSL)  operand:

movz Wn, #imm{, LSL #shift1}
movz Xn, #imm{, LSL #shift2}

where shift1 is 0 or 16 for 32-bit register
where shift2 is 0, 16, 32, 48 for 64-bit register

### DIFF
--- a/lib/AArch64Base.ml
+++ b/lib/AArch64Base.ml
@@ -441,6 +441,7 @@ type 'k kinstruction =
   | I_STOPBH of  atomic_op * bh * w_type  * reg * reg
 (* Operations *)
   | I_MOV of variant * reg * 'k kr
+  | I_MOVZ of variant * reg * 'k kr * 'k s
   | I_SXTW of reg * reg
   | I_OP3 of variant * op * reg * reg * 'k kr * 'k s
   | I_ADDR of reg * lbl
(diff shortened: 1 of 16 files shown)


maranget:
Luke, haven't those additions been reviewed already?

jalglave:
hi Luc,

yes - the point of those PR is to add those to master now.

Thanks,
Jade

lib/AArch64Base.ml
```diff
@@ -441,6 +441,7 @@ type 'k kinstruction =
   | I_STOPBH of  atomic_op * bh * w_type  * reg * reg
 (* Operations *)
   | I_MOV of variant * reg * 'k kr
+  | I_MOVZ of variant * reg * 'k kr * 'k s
```
maranget, line 444: As I understand it the instruction MOVZ always takes an immediate constant as second argument. As a consequence, the following definition would be more logical.
```suggestion
  | I_MOVZ of variant * reg * 'k * 'k s
```
Of course, such a change will entail many other changes.Those should quite easy, with typechecker guidance.

PR #70, opened 2020-10-03
Added all to gitignore

lukeg101:
This would only be included in the repo as an example, most of the time
we don't want to include this

Following https://github.com/herd/herdtools7/pull/30

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ jerd.byte
 catIncludes.ml
 *~
 *.swp
+@all


maranget:
I am not very found of this PR. If present `@all`  (the conventional name for listing all .litmus files) should probably be managed. Perhaps choose an alternative name for transient index files.

lukeg101:
I am happy to leave it if you are not fond of it, it's purely a convenience so I'm happy to close this if you are.

PR #74, opened 2020-10-06
[herd]: Updated label support for BNE,BEQ,CBZ,CBNZ,B,and BL

lukeg101:
Added support for labels, unittests also provided.

The label_addr production rule will be extended later to support
numeric addresses.

Following https://github.com/herd/herdtools7/pull/32


maranget:
Fine supplying   the unit tests, thanks.

PR #82, opened 2020-10-13
[herd/unittests]: Fixed pre-commit failures in unittests

lukeg101:
This commit does not add new content, this patch fixes pre-commit checks

After running pre-commit run --all-files we found some issues
with the file format of some of the unittests, it seems they were
executable format. This is now fixed so they do not fail pre-commit.


lukeg101:
Updated commit message with reported-by

PR #83, opened 2020-10-14
[litmus] Clean cache before changing PTE attributes

relokin:
In AArch64 accessing a single location with different cacheability
attributes can lead to loss of coherence. This patch ensures that
initial writes and cache maintenance prior to the code of the test are
performed after setting the relevant PTE. In addition, and to make
sure that there are no stale copies in the cache, we also add a cache
invalidation.

### DIFF
--- a/litmus/preSi.ml
+++ b/litmus/preSi.ml
@@ -1010,10 +1010,25 @@ module Make
           O.o "static void vars_init(vars_t *_vars,intmax_t *_mem) {" ;
           O.oi "const size_t _sz = LINE/sizeof(intmax_t);";
           O.oi "pteval_t *_p;" ;
-          O.o "" ;
+          let mem_map =
+            let open BellInfo in
+            match test.T.bellinfo with
+            | None|Some { regions=None; } -> []
+            | Some { regions=Some map } -> map in
           List.iter
             (fun (a,_) ->
+              O.o "" ;
               O.fi "_vars->%s = _mem;" a ;
+              begin try
+                let rs = Misc.Simple.assoc a mem_map in
+                O.fi "cache_flush((void *)%s);" (OutUtils.fmt_pte_kvm a);
+                List.iter
+                  (fun r ->
+                    let r = sprintf "attr_%s" (MyName.name_as_symbol r) in
+                    O.fi "litmus_set_pte_attribute(%s,%s);" (OutUtils.fmt_pte_kvm a) r)
+                  rs;
+                O.fi "litmus_flush_tlb((void *)%s);" (OutUtils.fmt_pte_kvm a);
+              with Not_found -> () end;
               O.fi "_vars->%s = _p = litmus_tr_pte((void *)_mem);" (OutUtils.fmt_pte_tag a) ;
               O.fi "_vars->%s = *_p;" (OutUtils.fmt_phy_tag a) ;
               O.oi "_mem += _sz ;")
@@ -1134,18 +1149,13 @@ module Make
                 (pctag (proc,addr)) addr)
             addrs
         end ;
-        let mem_map =
-          let open BellInfo in
-          match test.T.bellinfo with
-          | None|Some { regions=None; } -> []
-          | Some { regions=Some map } -> map in
-        begin match pte_init,mem_map with
-        | [],[] -> ()
-        | bds,_ ->
+        begin match pte_init with
+        | [] -> ()
+        | bds ->
             O.oii "barrier_wait(_b);" ;
             List.iter
               (fun x ->
-                let ok1 = try
+                try
                   begin match Misc.Simple.assoc x bds with
                   | P phy ->
                       O.fii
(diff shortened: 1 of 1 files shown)


relokin:
On a second look, I'm still missing cases where the pte is not valid and we'll emit code to initialize, prime/clean the cache, I will need to fix this before this gets merged.

relokin:
I think this solves the problem. On a high level the logic is: if the PTE is not valid then we don't perform init writes or cache maintenance. In addition, we also avoid init writes/cache mainenance if we create an alias. For example, if pte_y = pte_x, we skip the init write to Y and rely on the init write to X. That way, we don't have to worry here if pte_x is valid or not.

maranget:
Hi Nikos,

> I think this solves the problem. On a high level the logic is: if the PTE is not valid then we don't perform init writes or cache maintenance. In addition, we also avoid init writes/cache mainenance if we create an alias. For example, if pte_y = pte_x, we skip the init write to Y and rely on the init write to X. That way, we don't have to worry here if pte_x is valid or not.
>

I think it is not possible to proceed that way. The (implicit) semantics of litmus tests is PTE initialisation take place last before the test and PTE restoring to default values takes place first after the code. As a consequence 'virtual' global locations such as x,y etc. are stable in initialisations and final condition they always correspond to the same memory locations, regardless of what happens to PTE's in initialisations or in code.

 As an example of a potential change of behaviour, consider this test:
```
AArch64 T
{
int x=1;
pte_x=(valid:0);
0:X2=x;
1:X4=pte_x;
1:X6=(oa:phy_x,valid:1);
}
  P0                 |. P1.                ;
LDR W0,[X2] |  STR X6,[X4] ;
exists (0:X0=1)
```
At the moment we sometime observe the output 0:X0=1, with your modification it will not longer be the case.

Similar differences in semantics occur with hardware management of the AF and DB flags, initial writes would change those to 1, voiding the initialisations of PTE.
 
I cannot say that I fully understand the reason why you put PTE initialisation first. My current opinion is that we just can't do that...

relokin:
Hi Luc,

Thanks for reviewing the change!

The primary motivation for this change is that I am trying to avoid mixing cacheability/shareability attributes unintentionally. Before my patch, if I were to run the following limtus test:
```
AArch64 MP+dmb.sys-Ish-WT
{
0:X1=x; 0:X3=y;
1:X1=y; 1:X3=x;
}
 P0          | P1          ;
 MOV W0,#1   | LDR W0,[X1] ;
 STR W0,[X1] | DMB SY      ;
 DMB SY      | LDR W2,[X3] ;
 MOV W2,#1   |             ;
 STR W2,[X3] |             ;
scopes: (Ish P0 P1)
regions: x: normal inner-write-through outer-write-through, y: normal inner-write-through outer-write-through
exists
(1:X0=1 /\ 1:X2=0)
```
Litmus would allocate memory for `x` and `y` and map them as inner write-back and outer write-back (default) and initialize them to 0. Then we would change the cacheability attributes to the desired ones, and execute the instructions of the litmus test. This means that the init writes to `x` and `y` and the other memory events to 'x' and 'y' have different memory attributes (described as Mismatched Memory Attributes in the Arm ARM) leading to loss of uniprocessor semantics, coherence, etc. 

There are two ways to solve this, either make sure that the init writes use the same memory attributes as the test or clean and invalidate the cache between the init writes and the beginning of the test. Does that make sense?

If we need to preserver the order of initializing the pte and the corresponding variable as-is, we could also clean the cache if we change set the memory attributes in the init phase.

Thanks for the detailed explanation, I missed the point you raised on the init of ptes and I can have a look whether there is an easy way to fix it.

maranget:
Perhaps the cacheability attributes can be set elsewhere, here the x,y, fields are initialised. (in `vars_init`).  Or, maybe, where each HW thread set its role, in `set_role`. Namely the cacheability attributes do not change from run to run.

relokin:
Again thanks for the review!

Indeed there is no need for setting them there. I've now moved it to ```vars_init```. Hope I haven't missed something.

relokin:
Just realized this doesn't allow me to compile the following test the way I want:
```
AArch64 MP+dmb.sys-Ish-0xWByWB-1xWByWT
{
pte_z=(oa:phy_y);
0:X1=x; 0:X3=y;
1:X1=z; 1:X3=x;
}
 P0          | P1          ;
 MOV W0,#1   | LDR W0,[X1] ;
 STR W0,[X1] | DMB SY      ;
 DMB SY      | LDR W2,[X3] ;
 MOV W2,#1   |             ;
 STR W2,[X3] |             ;
regions: x: normal inner-write-back outer-writeback, y: normal inner-write-back outer-write-back, z: normal inner-write-through outer-write-through
exists (1:X0=1 /\ 1:X2=0)
```
My intention here is for `z` to be an alias for `y` with different cacheability attributes.

What I get is that in `vars_init` we allocate memory for `z` and give it the right memory attributes. Then in `do_run` we set:
```
*_vars->pte_z = _vars->saved_pte_y;
```

PR #89, opened 2020-10-20
[fix] fix regression caused by other regression fix

lukeg101:
The recent commit 60eb4359aa2da588db2b caused a regression, this fixes it


maranget:
Well done

PR #90, opened 2020-10-20
[diy]: Add NoVolatile value to Variant.t in diy

lukeg101:
Following https://github.com/herd/herdtools7/pull/33

Added the option to generate C tests without the Volatile option.
Volatile is not currently used for any other reason than to prevent
compiler optimisations to be applied in litmus. If we are to use such
optimisations we must trust they are correct.

This patch removes the qualifier on types as it is not needed in this
case.

No unittests are provided but you can run test the implementation by
running

diy7 -arch C -variant NoVolatile

and check the output C tests do not have volatile qualified types,
you can check these are then accepted by Herd using

herd7 -model rc11.cat *.litmus

We can follow this up with tests once the infrastructure to do this exists.

### DIFF
--- a/gen/CCompile_gen.ml
+++ b/gen/CCompile_gen.ml
@@ -811,7 +811,10 @@ module Make(O:Config) : Builder.S
           List.map
             (fun (t,loc) -> match t with
             | A.Plain _ ->
-                sprintf "volatile %s* %s" (A.dump_typ t) loc
+                let novolatile = O.variant Variant_gen.NoVolatile in
+                let volatile = if novolatile then
+                    "" else "volatile " in
+                sprintf "%s%s* %s" volatile (A.dump_typ t) loc
             | A.Atomic _ ->
                 sprintf "%s* %s" (A.dump_typ t) loc)
             args in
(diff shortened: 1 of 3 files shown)


maranget:
Hi Luke, I have a simplifying suggestion

It may in fact be more simple not to add an novolatile boolean field to `CCompile_gen.Make` configuration argument, as the `variant` field is already present (from `Top_gen.Config`). Then one may simple define `novolatile` inside 
`CCompile_gen.Make`  as follows:
```
let novolatile = !Config.variant Variant_gen.NoVolatile
```

That way you need not modify the various generator drivers (`diyone.ml`, `diy.ml`, `diycross.ml`)

lukeg101:
That is a good idea thanks! I will add that tomorrow. 
I think we should hold off merging this until we have a good way to test this, just so we don't miss anything

lukeg101:
- Added Luc's simplifying suggestion
- reworded the Commit message and description, + code comments following feedback from Eth

I think the code portion of this commit is ready now, but we should wait to commit until we can test this automatically

maranget:
> I think the code portion of this commit is ready now, but we should wait to commit until we can test this automatically
Let's merge this PR does not look dangerous as it is short and does not alter anything when option is not selected.

PR #93, opened 2020-10-28
[herd] Extract parts of test_herd to a library internal_lib

eth-arm:
This PR is the first in a series that will add `test_herd_diycross` and `test_herd_catalogue`, for testing `herd` with `diycross` and against the Catalogue respectively.

It is being sent as a separate PR because it already introduces multiple new concepts (a new library, tests written in OCaml for testing that library directly), so this will be enough to review on its own.

If approved, future PRs will further extract "running `herd` against an expected output" to a library.

Where possible, functions are named similarly to standard library functions (e.g. `Channel.iter_lines` like `List.iter`).

I have been using exceptions instead of error values, because to my current understanding, that is more "the OCaml way".

`Command.run_with_stdout` is called that because later PRs will include `Command.run` and `Command.run_with_stdin_and_stdout`.

### DIFF
--- a/internal/lib/channel.ml
+++ b/internal/lib/channel.ml
@@ -1,0 +1,38 @@
+(****************************************************************************)
+(*                           the diy toolsuite                              *)
+(*                                                                          *)
+(* Jade Alglave, University College London, UK.                             *)
+(* Luc Maranget, INRIA Paris-Rocquencourt, France.                          *)
+(*                                                                          *)
+(* Copyright 2010-present Institut National de Recherche en Informatique et *)
+(* en Automatique and the authors. All rights reserved.                     *)
+(*                                                                          *)
+(* This software is governed by the CeCILL-B license under French law and   *)
+(* abiding by the rules of distribution of free software. You can use,      *)
+(* modify and/ or redistribute the software under the terms of the CeCILL-B *)
+(* license as circulated by CEA, CNRS and INRIA at the following URL        *)
+(* "http://www.cecill.info". We also give a copy in LICENSE.txt.            *)
+(****************************************************************************)
+
+(** Utilities for handling in_channel and out_channel. *)
+
+(* in_channel utilities *)
+
+let iter_lines f chan =
+  try
+    let rec iter () = f (input_line chan) ; iter () in iter ()
+  with End_of_file -> ()
+
+let map_lines f chan =
+  let ret = ref [] in
+  iter_lines (fun l -> ret := f l :: !ret) chan ;
+  List.rev !ret
+
+let read_lines chan =
+  map_lines (fun l -> l) chan
+
+
+(* out_channel utilities *)
+
+let write_lines chan lines =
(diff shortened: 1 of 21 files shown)


eth-arm:
I split this PR into multiple commits to make it easier to review.

internal/lib/test.ml
```diff
@@ -0,0 +1,67 @@
...
+    |> List.map (fun (x, y) -> c x y)
+    |> find_comparison
+  end else
+    compared_length
+
+let string_list_compare xs ys =
+  list_compare String.compare xs ys
+
+let int_list_compare xs ys =
+  list_compare Int.compare xs ys
```
maranget, line 67: Module `Int` is available from OCaml 4.08.

Yu may define your own int_compare as follows:
```
let in_compare (i:int) (j:int) = compare i j
```
`compare` above is  the polymorphic compare. 

You can also write
```
external int_compare : int -> int -> -> int = "caml_int_compare"
```
Instructing the compiler to call a specific primitive (written in C).

internal/lib/channel.ml
```diff
@@ -0,0 +1,39 @@
...
+
+(** Utilities for handling in_channel and out_channel. *)
+
+(* in_channel utilities *)
+
+let rec iter_lines f chan =
+  try begin
+    (f (input_line chan)) ; iter_lines f chan
+  end
+  with End_of_file -> ()
```
maranget, line 25: Why not, your implementation is not tail-recursive, Consider
```
try while true do
  f (input_line chan)
done
with End_of_file -> ()
```
The `while true` loop can be replaced by a recursive function. The point is the recursion not to stack exception handlers.
```
let rec iter_lines f chan =
  try
    let rec iter () = f (input_line chan) ; iter () in iter ()
  with End_of_file -> ()
```

```

File: internal/lib/channel.ml [hunk 3]
```diff
@@ -0,0 +1,39 @@
...
+
+let map_lines f chan =
+  let ret = ref [] in
+  iter_lines (fun l -> ret := !ret @ [ f l ]) chan ;
+  !ret
```
maranget, line 30: This code is quadratic (construction a list by repetitive append at the end).
Better to construct the list by `ret := f l :: !ret` and return `List.rev !ret`.

internal/lib/test.ml
eth-arm: Done.

internal/lib/channel.ml
eth-arm: Done.

internal/lib/channel.ml
eth-arm: Done.

maranget:
Ok I have made a few remarks, but I am nitpicking (except for the quadratic  list construction idiom).

I am wondering whether it would be possible for tools in internal to use the functions defined in `lib/misc.ml` or not? For instance `lib/misc.ml` already includes a `list_compare` function. However it may complicate things more that we benefit from sharing code.

PR #97, opened 2020-10-29
[litmus] Add shebang to comp.sh and run.sh

### LINKED ISSUE #96: Litmus generated scripts do not have shebangs
This can cause issues when not using bash.

image

This issue can be fixed by adding a shebang to the top of `comp.sh` and `run.sh`.

BreD1810:
This is required to avoid errors in non-bash shells.

Closes #96.

### DIFF
--- a/litmus/dumpRun.ml
+++ b/litmus/dumpRun.ml
@@ -217,7 +217,7 @@ let dump_command names =
 let dump_shell names =
   Misc.output_protect
     (fun out_chan ->
-      output_line out_chan "#!/bin/bash\n" ;
+      output_line out_chan "#!/bin/sh\n" ;
       output_line out_chan "date" ;
       output_line out_chan "LITMUSOPTS=\"${@:-$LITMUSOPTS}\"" ;
       begin match Cfg.crossrun with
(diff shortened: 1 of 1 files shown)


litmus/dumpRun.ml
```diff
@@ -217,6 +217,7 @@ let dump_command names =
 let dump_shell names =
   Misc.output_protect
     (fun out_chan ->
+      output_line out_chan "#!/bin/bash\n" ;
```
maranget, line 220:
```suggestion
      output_line out_chan "#!/bin/sh\n" ;
```
Hi, I'd rather have sh in place of bash, do you object?

litmus/dumpRun.ml
```diff
@@ -261,7 +262,7 @@ let dump_shell names =
       end in
       let module RU = RunUtils.Make(O) in
       RU.report_machine out_chan ;
-      output_line out_chan "head -1 comp.sh" ;
+      output_line out_chan "sed '2q;d' comp.sh" ;
```
maranget, line 265: Why change? Is `sed` more likely to be available than `head`?

litmus/dumpRun.ml
```diff
@@ -278,6 +279,7 @@ let dump_shell_cont arch sources utils =
   let module RU = RunUtils.Make(O) in
   Misc.output_protect
     (fun chan ->
+      fprintf chan "#!/bin/bash\n" ;
```
maranget, line 282:
```suggestion
      fprintf chan "#!/bin/sh\n" ;
```

maranget (commented):
Thanks for submitting this PR.

litmus/dumpRun.ml
BreD1810: I believe the `head` command grabbed the first line of the `comp.sh` file (`gcc=...`) As this will now be the shebang, we instead want the second line, hence the move to `sed`. Not sure if I misunderstand, or if there's a better way?

maranget:
Merged, thanks.

Issue #98, opened 2020-10-30
Herd does not build with OCamlbuild


eth-arm:
PR #84 introduced a new library dependency, [Stdint](https://github.com/herd/herdtools7/blob/1fd2e73e1e690ddc315c47f54501bfbdd4891ce2/lib/dune#L30). The OCamlbuild script was not updated to match, and now Herd does not build with OCamlbuild.

maranget:
Issue hopefully solved by adding [Stdint dependency](https://github.com/herd/herdtools7/blob/48660e7986c78a3fac270d13c86588b9be82fd2c/_tags#L17) for ocamlbuild. Unfortunately, adopted solution  adds another dependency on [ocamlfind](https://github.com/herd/herdtools7/blob/48660e7986c78a3fac270d13c86588b9be82fd2c/ocb-build.sh#L27). In fact I still need ocamlbuild for two reasons:

1. I sometimes (not that often) need exception traces. I get them by running the debug byte code executables, which I build easily with `ocamlbuild herd/herd.d.byte`.
2. The web interface (directory herd-www) still relies on ocamlbuild for its build process.

I'd be happy to say goodbye to ocamlbuild if someone would find a way to perform tasks  1. and 2. without ocamlbuild.

jalglave:
Solved by https://github.com/herd/herdtools7/commit/48660e7986c78a3fac270d13c86588b9be82fd2c

PR #100, opened 2020-10-30
[all] Add `make check-deps` to ensure dependencies are installed

eth-arm:
This PR:
- Adds a `check-deps` phony target.
- Checks that `opam` is installed.
- Checks that `menhir` and `stdint` are installed.
- If `D=dune`, check that `dune` is installed.
- If `D=ocb`, check that `ocamlbuild` and `ocamlfind` are installed.
- Performs this check before `make build`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,17 +3,24 @@ D=dune
 #For building with ocamlbuild set
 #D=ocb
 
+OPAM_DEPS = menhir stdint
+
 ifeq ($(D), dune)
 	HERD = _build/install/default/bin/herd7
 	HERD_REGRESSION_TEST = _build/default/internal/herd_regression_test.exe
+
+	OPAM_DEPS += dune
 else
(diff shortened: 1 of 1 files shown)


jalglave:
hi Ethel, on first attempt when I do:
make clean; make

This says:
Package js_of_ocaml not installed; please run: opam install js_of_ocaml
make: *** [check-deps] Error 1

We use js_of_ocaml only for the web interface so it would be good if this wasn't a fatal error for most of our users. Would it be possible to separate the dependencies for the main tools, and the ones for the web interface?

Thanks in advance,
Jade

eth-arm:
I've removed `js_of_ocaml` for now; I'll have a look at the web UI build system later (along with e.g. [can we build it with Dune?](https://github.com/herd/herdtools7/issues/98)).

jalglave:
Thank you very much Ethel - I'll integrate this now.
Jade

PR #103, opened 2020-11-03
[all] Replace pre-commit `make all` with `make test`

eth-arm:
This PR:
- Changes the pre-commit `make all` for `make test`. We have tests now, so we should run them, and commits that break tests should not be allowed.
- Correspondingly triggers a `make test` if any `.litmus` or `.litmus.expected` files change.


jalglave:
Thanks Ethel!
Jade

PR #107, opened 2020-11-05
[internal] Create ExtendedBaseModules module

eth-arm:
I am sending this PR as more of a "request for discussion" / "request for advice", rather than "this code is ready to merge".

This PR:
- Creates a module, `ExtendedBaseModules`, to:
  - Backport modules and functions from future versions of OCaml to 4.05, e.g. `Int`.
  - Extend built-in modules with new features, e.g. `to_ocaml_string`, which returns an OCaml-syntax representation of the value.
  - Creates a functor `List.Make` that builds a `List` with a typed `compare` function, suitable for e.g. creating sets of lists of things.
- Adds relevant tests for these modules.
- Converts existing uses of `Test.pp_{int,string}_list` and `Test.{int,string}_list_compare` to use these new modules.
- Removes the replaced functions from `Test`.

For example, using OCaml 4.05:

```ocaml
open ExtendedBaseModules

module IntList = struct
  let compare = List.compare Int.compare
end

if IntList.compare xs [1; 2; 3] <> 0 then
    failwith "oh no"
```

I would also like to use this mechanism to backport `Option.compare` & `Option.equal` to parse catalogue shelf files in an upcoming PR.

Regarding `List.compare` specifically, the second commit adds a functor `List.Make` to mirror `Set.Make`. For example:

```ocaml
module IntList = List.Make(Int)

module IntListSet = Set.Make(List.Make(Int))
```

I plan to keep this approach limited to `internal/` for now, to experiment with this and find out what does and does not work.

Do you have opinions about this approach, for either extending built-ins, or backporting features? What do you think about the usefulness of `List.Make`? Do you think the names could be better, e.g. `open FutureFeatures` or `List.Make_compare`?

### DIFF
--- a/defs.sh
+++ b/defs.sh
@@ -9,7 +9,7 @@ NATIVE="$HERD $LITMUS $TOOLS $GEN $JINGLE"
 
 # Internal-only, not installed.
 INTERNAL="herd_regression_test.native herd_diycross_regression_test.native"
-TESTS="channel_test.native command_test.native filesystem_test.native test_test.native"
+TESTS="channel_test.native command_test.native extendedBaseModules_test.native filesystem_test.native"
 
 mk_exe () {
   D=$1
--- a/internal/lib/compare.ml
+++ b/internal/lib/compare.ml
@@ -14,46 +14,7 @@
 (* "http://www.cecill.info". We also give a copy in LICENSE.txt.            *)
 (****************************************************************************)
 
-(** Extending built-in / base modules, either to port future features into
- *  earlier versions of OCaml, or to add new functionality. *)
-
-module Int = struct
-  type t = int
-
-  let compare (x : int) (y : int) = compare x y
-
-  let to_ocaml_string i = Printf.sprintf "%i" i
-end
-
-module String = struct
-  include String
-
-  let to_ocaml_string s = Printf.sprintf "%S" s
-end
-
-module List = struct
-  include List
-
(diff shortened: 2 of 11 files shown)


jalglave:
Hi @maranget please could you let us know if you disagree strongly with that direction, or if you think it's a good direction to take? Then Ethel can carry on making progress on this.
Thanks,
Jade

maranget:
Anything that facilitates backward compatibility has my blessing. Go (in `internal/` for now)!

To tell you the awful truth, I am a bit concerned by overengineering. This is how I have proceeded up to now:

- I handle the backporting problem by avoiding new features :). When I am forced to (or when I fancy), by adding ad-hoc functions in `lib/misc.{ml,mli}`.  For instance, see `Misc.lowercase`, which I wrote because the function I needed has been renamed `lowercase_ascii`  at some point.
- When I found that some module <N> needs significant extension, I wrote a `my<M>` module (see for instance `lib/mySet.{ml,mli}`). So as regards the List.Make module: think about its use cases, what use for it do you plan? discuss writing some code near use place or a few functions in `Misc`. Perhaps it is high time to add structure?

So please,  follow your ideas, keeping complexity to an acceptable level and convince me I am too cautious. One argument that would convince me would be a design that would permit this support code, to be used outside of `internal/`. As to names I have no strong opinion, keep them as short and significant as possible.... (Yes, there is a contradiction here).

PR #109, opened 2020-11-09
Variant in test metadata

maranget:
This PR introduces "variants" in the metadata of tests. This is done by using a "Variant" field.

Writing
```
Variant=<tags>
```
Has the same effect as adding option `-variant <tags>` at the end the the command line. As a consequence, variants from the test take precedence over the variants from the command line.

### DIFF
--- a/herd/AArch64Sem.ml
+++ b/herd/AArch64Sem.ml
@@ -17,7 +17,6 @@
 module Make
     (TopConf:sig
       module C : Sem.Config
-      val precision : bool
       val dirty : DirtyBit.t
     end)
     (V:Value.S)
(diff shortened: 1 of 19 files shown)


jalglave:
Thanks Luc - I think this should go into master too.
Jade

maranget:
Hum, I think the present PR does not apply to master without conficts

jalglave:
Ok then don't worry. I will do a rebase w.r.t. master this week so will
take care of it.

Thanks,
Jade

maranget:
You are in fact right, it would have been better to base this PR upon master. However I am not sure there is an urgent need of the feature in master...

PR #116, opened 2020-11-16
Extend fault support.

maranget:
There are two extensions:
- Faults can now occur in the `locations`  field of litmus tests. As a convenient mean to observe faults with no a priori on their occurrence or absence.
- Comparison tools `mcmp7` and `mcompare7` handle faults with no label, _i.e._ `Fault(<proc>,<var>)`.


jalglave:
Thank you Luc.
Jade

PR #117, opened 2020-11-19
Kvm diy reloaded

maranget:
More KVM related features for generators
- Release (L) and Acquire (both of them, A and Q) combine with PTE writes and reads.
- Address and data dependencies to and from PTE accesses, clarify size issue.
- Observation of chains of PTE writes, either by local observers or by reading final value of PTE.


maranget:
Force pushes were here to rewrite history into 3 separate commits, nothing more.

jalglave:
Thanks Luc.
Jade

PR #119, opened 2020-11-20
Replace stdint with homegrown uint library

eth-arm:
To remove our dependency on the external `stdint` library, this PR creates a module `Uint`, to implement the relevant API from `Stdint`, using pure OCaml. It should also be extendable to 256-bit or larger Unsigned Integers when we need them.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -23,13 +23,11 @@ Requirements
 - OCaml (version >= 4.05.0)
 - dune
 - menhir (version >= 20180530)
-- stdint library (version >= 0.5.1)
-
 
 We strongly recommend to have this base software installed through the opam
 package manager.
 
-    % opam install dune menhir stdint
+    % opam install dune menhir
 
 Notice: Compilation with ocamlbuild is possible, by setting D=ocb in Makefile.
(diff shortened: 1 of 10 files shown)


jalglave:
Thank you very much Ethel, this allows us to get rid of the stdint dependency.
Jade

maranget:
I like the idea of removing a dependency very much. There is some non-idiomatic code in the Uint128 module (where you write `let f a = ... fst a... snd a` it would be more idiomatic to write `let f (hi,lo) = .... hi ... lo ...`. That said, there is no need to rewrite this.

eth-arm:
I'll send a follow-up PR that fixes that; there are some other cleanups I want to add as well.

Would the idiomatic way of expressing `let f a b = …` be `let f (a_hi, a_lo) (b_hi, b_lo) = …`?

maranget:
No please, this is not necessary. My advice is for future code. By  idiomatic, I  mean "use pattern matching more". As to your question, there is no idiomatic way to name variables... (I'd say `let f (h1,l1) (h2,l2) = ...` because I like short names for local variables, a matter of taste.

PR #121, opened 2020-11-23
Explicit markers for the absence of faults for observed faults in tools output logs.

maranget:
This permits correct final conditions for tests that describe behaviours in tables. Consider for instance two logs A an B, where A and B contain the outcome `0:X0=0 /\ fault (P0,x)`, with B containing the additional outcome `0:X0=0`. That is, the absence of `fault(P0,x)` is left implicit. Then, the comparison mechanics of `mcompare` will yield the conditon `exists 0:X0=0` to encode the outcomes present in B and absent in A. This is not correct as `0:X0=0 /\ Fault(p0,x)` does match this condition. With explicit fault absence, `mcompare` will produce the correct condition:
```
exists 0:X0=0 /\ ~Fault(P0,x)
```

Legacy logs are handled by the `sum` tools, identical outcomes up to having or not having explicit absent faults are considered equal and their number of occurrences are summed. The outcome in output is the one with explicit fault absence.

### DIFF
--- a/herd/archExtra_herd.ml
+++ b/herd/archExtra_herd.ml
@@ -127,7 +127,8 @@ module type S = sig
 
 (* Set of states *)
   type final_state = state * FaultSet.t
-  val do_dump_final_state : (string -> string) -> final_state -> string
+  val do_dump_final_state :
+    FaultAtomSet.t -> (string -> string) -> final_state -> string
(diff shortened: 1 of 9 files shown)


jalglave:
Thanks Luc.
Jade

Issue #123, opened 2020-11-30
Tracking Issue for regression-testing the Catalogue


eth-arm:
This issue exists to track & log PRs for implementing regression-testing of the Catalogue. This Issue will link to those PRs, and those PRs will link to this Issue.

I am currently doing this work, but want to split it into multiple PRs instead of a single giant PR.

eth-arm:
An initial version of this tool was added in PR #145, then removed in PR #165 because it was not providing value. I am leaving Arm, and so will not be able to finish this. That would leave an unused testing tool and a broken log-parsing library in the codebase.

I propose 3 possible plans of action:
* Change the intended behavior of `herd_catalogue_regression_test` to mirror `herd_regression_test` and `herd_diycross_test`: compare against a verbatim expected output, instead of comparing kinds. This would provide testing coverage, and would allow [kinds.ml](https://github.com/herd/herdtools7/blob/94204938d063ca9970ac1c1238176c58ba5bee23/internal/lib/kinds.ml), [kinds.mli](https://github.com/herd/herdtools7/blob/94204938d063ca9970ac1c1238176c58ba5bee23/internal/lib/kinds.mli), [log.ml](https://github.com/herd/herdtools7/blob/94204938d063ca9970ac1c1238176c58ba5bee23/internal/lib/log.ml), and [log.mli](https://github.com/herd/herdtools7/blob/94204938d063ca9970ac1c1238176c58ba5bee23/internal/lib/log.mli) to be deleted.
* Fix [log.ml](https://github.com/herd/herdtools7/blob/94204938d063ca9970ac1c1238176c58ba5bee23/internal/lib/log.ml) to extract kinds, write new kinds files, and test against those with the existing [herd_catalogue_regression_test.ml](https://github.com/herd/herdtools7/blob/94204938d063ca9970ac1c1238176c58ba5bee23/internal/herd_catalogue_regression_test.ml).
[...]

PR #124, opened 2020-11-30
[internal_lib] Add module Shelf

eth-arm:
This PR adds a module `Shelf`, to work with `shelf.py` Catalogue files using OCaml.

To do this, this PR also:
- Adds a module `Base`, to extend builtin modules, and backport modules from future versions of OCaml.
- Adds a module `Compare`, to house utilities for writing or generating `compare` functions.
- Adds a module `OcamlString`, to house utilities for writing or generating `to_ocaml_string` functions.
- Extends `Command` with `run_with_stdin_and_stdout`, to run runtime-generated Python scripts on `shelf.py` files.

This PR is a part of Issue #123, implementing regression-testing of the Catalogue.

Regarding the name `to_ocaml_string`, I also considered a name like `repr` like Python's `__repr__` ("internal representation"); what do you think? The contract of `to_ocaml_string` is that it returns a similar representation to the OCaml toplevel REPL.

### DIFF
--- a/defs.sh
+++ b/defs.sh
@@ -9,7 +9,7 @@ NATIVE="$HERD $LITMUS $TOOLS $GEN $JINGLE"
 
 # Internal-only, not installed.
 INTERNAL="herd_regression_test.native herd_diycross_regression_test.native"
-TESTS="channel_test.native command_test.native filesystem_test.native test_test.native uint_test.native"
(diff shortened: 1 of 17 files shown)


jalglave:
Thanks Ethel, this looks good to me so I'll integrate it. Regarding your question on naming: I don't mind and I can see you have documented the contract you mentioned in comments.
Thanks,
Jade

PR #125, opened 2020-11-30
Bbm primitives

maranget:
Add two Cat primitives, in order to implement one of the conditions requiring break-before-make. 

#### Using break-before-make when updating translation table entries
...
- A change of the output address (OA), if the OA of at least one of the old translation table entries and the new translation table entry is writable.
...


jalglave:
Great, thank you Luc.
Jade

PR #127, opened 2020-11-30
[build] Mirror Dune and OCamlBuild configs

eth-arm:
**Summary**: Replace manual lists in `defs.sh` with a new script that reads from Dune files, `binaries_of_dune`.

Currently, there are two places that binaries are defined:

- For Dune, they are in a given directory's `dune` file.
- For OCamlBuild, they are in `defs.sh`.

These two sources of truth have become out-of-sync before (commit e2837555c73417ac9f207c863298ccbf7890be16, PR #63), and as more tests are added to the project, the risk of error increases.

This commit makes OCB use Dune files as the single source of truth. To do this, this PR:

- Adds a module `Sexp`, to parse Dune files (and other S-expressions).
- Adds tests for `Sexp`.
- Adds a binary `binaries_of_dune` to extract executables and tests from Dune files. To avoid bootstrapping issues from the need for compilation, this binary is a script run with the OCaml top-level.
- Modifies the manual lists of binaries in `defs.sh` with shell functions that call `binaries_of_dune`.

I have tested this on my machine with OCaml 4.05 & 4.11.1, running `make test D=ocb` and `make install D=ocb`.

AFAICT, there is no way to use a module in the OCaml top-level without compiling to bytecode, so instead I have used `#use`, a textual import.

### DIFF
--- a/defs.sh
+++ b/defs.sh
@@ -1,15 +1,27 @@
 set -o errexit
 
-HERD="herd.native"
-LITMUS="litmus.native klitmus.native"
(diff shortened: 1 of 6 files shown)


jalglave:
Thanks for doing this Ethel.
Jade

PR #128, opened 2020-12-02
[test] Run `dune runtest` with --profile=release

eth-arm:
This builds the test binaries with the same profile as `make build`, which stops Dune from rebuilding them, resulting in faster pre-commits.


jalglave:
Thanks for doing this Ethel.
Jade

PR #129, opened 2020-12-03
Add free-form attributes (e.g., cacheability, shareability, etc) to the PTE

relokin:
This series of patches adds support for specifying attributes in the PTE. The motivation is to use to specify cacheability and shareability attributes. For example:

`pte_x = (oa:phy_y,af:1,db:0,dbm:0,attrs:(Normal,Inner-shareable,Inner-write-back,Outer-write-back))`

These attributes are used to annotate Access Actions and the corresponding events. They are also exposed to cat and enable rules that use these annotations.

These annotations are more of a free-form list of strings, without any specific rules of what these strings can be, to allow different architectures to express their own.

### DIFF
--- a/herd/machModelChecker.ml
+++ b/herd/machModelChecker.ml
@@ -307,6 +307,28 @@ module Make
                  "DATA", is_data_port;
                  "NDATA", (fun e -> not (is_data_port e));])) in
       let m =
+        let impl_pte_reads = E.EventSet.filter
+                               (fun e -> E.Act.is_implicit_pte_read e.E.action)
+                               (Lazy.force mem_evts) in
+        let evts_map =
+          E.EventSet.fold
+            (fun e evts_map ->
+              let pteval_v = E.read_of e in
+              let attrs =
+                let open Constant in
+                match pteval_v with
+                | Some (S.A.V.Val (PteVal v)) ->
+                   PTEVal.Attrs.as_list v.PTEVal.attrs
+                | _ -> assert false in
+              List.fold_right
+                (fun attr evts_map ->
+                  let evts_w_attr = StringMap.safe_find E.EventSet.empty attr evts_map in
+                  let evts_w_attr = E.EventSet.add e evts_w_attr in
+                  let evts_map = StringMap.add attr evts_w_attr evts_map in
+                  evts_map) attrs evts_map)
+            impl_pte_reads StringMap.empty in
+        I.add_sets m (StringMap.fold (fun k v l -> ("PTE" ^ k, lazy v) :: l) evts_map []) in
+      let m =
         I.add_sets m
           (List.map
              (fun (k,a) ->
--- a/lib/PTEVal.ml
+++ b/lib/PTEVal.ml
@@ -16,6 +16,24 @@
 
 open Printf
 
+module Attrs = struct
+  type t = StringSet.t
+
+  (* By default we assume the attributes of the memory malloc would
+     return on Linux. This is architecture specific, however, for now,
+     translation is supported only for AArch64. *)
+  let default =
+    List.fold_right StringSet.add
+      [ "Normal" ; "Inner-shareable"; "Inner-write-back"; "Outer-write-back" ]
+      StringSet.empty
+
+  let compare a1 a2 = StringSet.compare a1 a2
+  let eq a1 a2 = StringSet.equal a1 a2
+  let pp a = String.concat ", " (StringSet.elements a)
+  let as_list a = StringSet.elements a
+  let of_list l = StringSet.of_list l
+end
+
 type t = {
   oa : string;
   valid : int;
@@ -23,10 +41,11 @@ type t = {
   db : int;
   dbm : int;
   el0 : int;
+  attrs: Attrs.t;
   }
 
 (* For ordinary tests not to fault, the dirty bit has to be set. *)
-let prot_default =  { oa=""; valid=1; af=1; db=1; dbm=0; el0=1;}
+let prot_default =  { oa=""; valid=1; af=1; db=1; dbm=0; el0=1; attrs=Attrs.default; }
 let default s = { prot_default with  oa=Misc.add_physical s; }
 
 let pp_field ok pp eq ac p k =
@@ -38,12 +57,14 @@ and pp_af ok = pp_int_field ok "af" (fun p -> p.af)
 and pp_db ok = pp_int_field ok "db" (fun p -> p.db)
 and pp_dbm ok = pp_int_field ok "dbm" (fun p -> p.dbm)
 and pp_el0 ok = pp_int_field ok "el0" (fun p -> p.el0)
+and pp_attrs ok = pp_field ok (fun a -> Attrs.pp a) Attrs.eq (fun p -> p.attrs)
 
 let set_oa p s = { p with oa = Misc.add_physical s; }
 
 let is_default t =
   let d = prot_default in
-  t.valid=d.valid && t.af=d.af && t.db=d.db && t.dbm=d.dbm && t.el0=d.el0
+  t.valid=d.valid && t.af=d.af && t.db=d.db && t.dbm=d.dbm && t.el0=d.el0 &&
+    t.attrs=Attrs.default
 
 (* If showall is true, field will always be printed.
    Otherwise, field will be printed only if non-default.
@@ -51,7 +72,8 @@ let is_default t =
    (1) Fields older than el0 are always printed.
    (2) Fields from el0 (included) are printed if non-default. *)
 let do_pp showall p =
-  let k = pp_el0 false p [] in
+  let k = pp_attrs false p [] in
+  let k = pp_el0 false p k in
   let k = pp_valid showall p k in
   let k = pp_dbm showall p k in
   let k = pp_db showall p k in
@@ -71,18 +93,25 @@ let my_int_of_string s v =
     _ -> Warn.user_error "PTE field %s should be an integer" s
   in v
 
+type pte_prop =
+  | KV of (string * string)
+  | Attrs of string list
+
 let do_of_list p l =
-
-  let add_field a (s,v) = match s with
-  | "oa" -> { a with oa = v }
-  | "af" -> { a with af = my_int_of_string s v }
-  | "db" -> { a with db = my_int_of_string s v }
-  | "dbm" -> { a with dbm = my_int_of_string s v }
-  | "valid" -> { a with valid = my_int_of_string s v }
-  | "el0" -> { a with el0 = my_int_of_string s v }
-  | _ ->
-      Warn.user_error "Illegal PTE property %s" s in
-
+  let add_field a v = match v with
+    | KV (s, v) -> begin
+        match s with
+        | "oa" -> { a with oa = v }
+        | "af" -> { a with af = my_int_of_string s v }
+        | "db" -> { a with db = my_int_of_string s v }
+        | "dbm" -> { a with dbm = my_int_of_string s v }
+        | "valid" -> { a with valid = my_int_of_string s v }
+        | "el0" -> { a with el0 = my_int_of_string s v }
+        | _ ->
+           Warn.user_error "Illegal PTE property %s" s
+      end
+    | Attrs l -> { a with attrs = Attrs.of_list l }
+  in
   let rec of_list a = function
     | [] -> a
     | h::t -> of_list (add_field a h) t in
@@ -108,6 +137,8 @@ let compare =
     lex_compare (fun p1 p2 -> Misc.int_compare p1.af p2.af) cmp in
   let cmp =
     lex_compare (fun p1 p2 -> String.compare p1.oa p2.oa) cmp in
+  let cmp =
+    lex_compare (fun p1 p2 -> Attrs.compare p1.attrs p2.attrs) cmp in
   cmp
 
 let eq p1 p2 =
@@ -116,4 +147,5 @@ let eq p1 p2 =
   Misc.int_eq p1.db p2.db &&
   Misc.int_eq p1.dbm p2.dbm &&
   Misc.int_eq p1.valid p2.valid &&
-  Misc.int_eq p1.el0 p2.el0
+  Misc.int_eq p1.el0 p2.el0 &&
+  Attrs.eq p1.attrs p2.attrs
--- a/herd/BellAction.ml
+++ b/herd/BellAction.ml
@@ -122,6 +122,9 @@ end = struct
 
   let is_PA_val _ = false
 
+  (* Unimplemented *)
+  let is_implicit_pte_read _ = assert false
+
   let is_atomic a = match a with
   | Access (_,_,_,true,_,_) ->
       assert (is_mem a); true
--- a/herd/CAction.ml
+++ b/herd/CAction.ml
@@ -183,6 +183,9 @@ end = struct
 
   let is_PA_val _ = false
 
+  (* Unimplemented *)
+  let is_implicit_pte_read _ = assert false
+
         (* The following definition of is_atomic
            is quite arbitrary. *)
 
--- a/herd/action.mli
+++ b/herd/action.mli
@@ -65,6 +65,7 @@ module type S = sig
   val get_mem_dir : action -> Dir.dirn
   val get_mem_size : action -> MachSize.sz
   val is_PA_val : A.V.v -> bool
+  val is_implicit_pte_read : action -> bool
 
 (* relative to the registers of the given proc *)
   val is_reg_store : action -> A.proc -> bool
--- a/herd/machAction.ml
+++ b/herd/machAction.ml
@@ -339,6 +339,10 @@ end = struct
         end
   | Some _,Some _ -> Warn.fatal "No Amo action on pteval"
 
+  let is_implicit_pte_read = function
+  | Access (R,_,_,_,_,_,A_PTE) -> true
+  | _ -> false
+
 (* relative to the registers of the given proc *)
   let is_reg_store a (p:int) = match a with
   | Access (W,A.Location_reg (q,_),_,_,_,_,_) -> p = q
--- a/lib/PTEVal.mli
+++ b/lib/PTEVal.mli
@@ -16,6 +16,16 @@
 
 (** Abstraction of page table entry (PTE) *)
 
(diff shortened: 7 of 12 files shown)


herd/AArch64Sem.ml
```diff
@@ -345,8 +347,11 @@ module Make
           (fun (pte_v,a_pte) ma -> (* now we have PTE content *)
             (* Monad will carry changing internal pte value *)
             let ma = ma >>= fun (pte_v,_) -> M.unitT pte_v in
+            let attr = match pte_v.attr_v with
```
relokin, line 350: This where my implementation has a problem. The concrete value of pte_v.attr_v is not yet determined. To solve this, I think I need to pass the V.v value without doing anything on it and delay this check until after we have a concrete execution.

lib/symbConstant.ml
```diff
@@ -77,11 +87,17 @@ module Make(Scalar:Scalar.S) = struct
       Misc.string_eq  s1 s2 && Misc.int_eq p1 p2
   | Tag t1,Tag t2 -> Misc.string_eq t1 t2
   | PteVal p1,PteVal p2 -> PTEVal.compare p1 p2 = 0
-  | (PteVal _,(Symbolic _|Concrete _|Label _|Tag _))
-  | (Concrete _,(Symbolic _|Label _|Tag _|PteVal _))
-  | (Symbolic _,(Concrete _|Label _|Tag _|PteVal _))
-  | (Label _,(Concrete _|Symbolic _|Tag _|PteVal _))
-  | (Tag _,(Concrete _|Symbolic _|Label _|PteVal _))
+  | MemAttr l1,MemAttr l2 ->
```
maranget, line 90: Is there some assumed sorting of attributes?
If not `["A";"B"]`  and `["B";"A";] are different, if yes they are identical and the above comparison function is wrong.

In any case an appropriate comment near the definition of `MemAttr` would be useful.

lib/symbConstant.ml
```diff
@@ -77,11 +87,17 @@ module Make(Scalar:Scalar.S) = struct
...
+      List.length l1 = List.length l2 &&
+        List.fold_right2
+          (fun s1 s2 c -> c && Misc.string_eq s1 s2)
+          l1 l2 true
```
maranget, line 94: Is there some assumed sorting of attributes?
If not `["A";"B"]` and `["B";"A";]` are different, if yes they are identical and the above comparison function is wrong.

In any case an appropriate comment near the definition of `MemAttr` would be useful.

maranget: Moreover, if sorted, the list of strings should perhaps be implemented as an abstract data type, so as to easily maintain the invariant.

herd/AArch64Arch_herd.ml
```diff
@@ -145,6 +145,9 @@ module Make (C:Arch_herd.Config) (V:Value.S) =
       | I_TLBI (_,_)
           -> None
 
+    let default_memattr = V.memattrToV [ "Normal"; "Inner-write-back"; "Outer-write-back";
```
maranget, line 148: Is this default list of attributes the one of kvm-unit-tests ?

herd/AArch64Sem.ml
```diff
@@ -71,25 +71,24 @@ module Make
...
 (* Ordinary access action *)
-      let access_anexp anexp d loc v ac = Act.Access (d,loc,v,AArch64.N,anexp,quad,ac)
-      let access_ord d loc v ac = access_anexp AArch64.Exp d loc v ac
+      let access_anexp anexp d loc v ac attr =
+        Act.Access (d,loc,v,AArch64.N,anexp,quad,ac,attr)
+      let access_ord d loc v ac attr = access_anexp AArch64.Exp d loc v ac attr
 
 (* Basic read, from register *)
-      let mk_read sz an anexp loc v =
+      let mk_read sz an anexp loc v attr =
```
maranget, line 79: I do not understand why memory attributes are part of action. After all memory attributes are present in the page table entry for an access and thus their value can be retrieved from the associated PTE read.

lib/symbConstant.ml
relokin: That's a good point and I will implement it as a sorted list of strings. At some point, I think we need to add sanity checks, those sanity checks would be different for different architectures. For example, for  AArch64 a location can be either in Normal or in Device Memory, as well as defaults if we have specified a memory type, but we have other attributes e.g., cacheablility, then it should be Normal memory. Not sure if those checks fit better in the cat file or in herd.

herd/AArch64Arch_herd.ml
relokin: That's what you get when you do a malloc from user space. I thought this is a reasonable default that also maintains backwards compatibility.

herd/AArch64Sem.ml
relokin: Correct, I am not sure how that would work but I'll have a look right away.

lib/symbConstant.ml
jalglave: hi Nikos, personally I would put those sanity checks in the cat file.
Thanks,
Jade

herd/AArch64Sem.ml
maranget: Having looked a `AArch64Sem.ml` code, the current code makes sense. The difficulty resides in adding yet another argument to the `Access` constructor, we can live with it.

lib/PTEVal.ml
```diff
@@ -23,12 +23,19 @@ type t = {
...
   dbm : int;
   el0 : int;
+  attrs: string list;
   }
 
 (* For ordinary tests not to fault, the dirty bit has to be set. *)
-let prot_default =  { oa=""; valid=1; af=1; db=1; dbm=0; el0=1;}
+(* FIXME: Should we set the default attributes to what Linux would use
+   for userspace applications, in that case we would need the
+   A.default_memattr *)
```
maranget, line 32: Ideally `PTEVal.ml` should be arch dependent. That is it should consist in a functor parmetrized by some arch basic structure.
However, the A.default_parameter can be added later, _e.g._ while emitting `pte_<id>` initial write events in `init_writes*` functions in `eventsMonad.ml`.

herd/AArch64Sem.ml
relokin: We need to the value of the attributes in `herd/machModelChecker.ml` (https://github.com/herd/herdtools7/pull/129/commits/d96e760de272583da8781b8ad3b938fa17aa4528#). Is there a way an easy way to get the `Rpte` event prior to all `M` events to annotate them with attributes? If not do you think it would be worth implementing?

relokin: Yes, I thought the same, adding one more argument to Access is not making the long constructor any better. I could also turn it into a record to add some structure, but that change would be quite intrusive.

lib/PTEVal.ml
```diff
@@ -38,20 +45,25 @@ and pp_af ok = pp_int_field ok "af" (fun p -> p.af)
 and pp_db ok = pp_int_field ok "db" (fun p -> p.db)
 and pp_dbm ok = pp_int_field ok "dbm" (fun p -> p.dbm)
 and pp_el0 ok = pp_int_field ok "el0" (fun p -> p.el0)
+and pp_attrs ok = pp_field ok (fun a -> sprintf "attrs:(%s)" (String.concat ", " a))
+                    attrs_eq (fun p -> p.attrs)
```
maranget, line 49: Wel `ok=false` meaning do not print when equal to default. Something more involved is called for if we want arch-dependant defaults. This may be an extra parameter to the printing functions... Or significant refactoring to get arch-dependant pteval's.

maranget: A simpler solution would be to consider that pteval are used for AArch64 only and to delay refactoring until another architecture implements virtual memory.

lib/PTEVal.ml
```diff
@@ -108,6 +127,13 @@ let compare =
     lex_compare (fun p1 p2 -> Misc.int_compare p1.af p2.af) cmp in
   let cmp =
     lex_compare (fun p1 p2 -> String.compare p1.oa p2.oa) cmp in
+  let cmp =
+    lex_compare (fun p1 p2 -> List.fold_right2 (fun s1 s2 a ->
+                                  match a with
+                                  | 0 -> String.compare s1 s2
+                                  | _ -> a)
+                                p1.attrs p2.attrs 0)
```
maranget, line 135: This comparison function for lists of strings appears elsewhere at least once (`symbConstant.ml`). One may think about moving it to the `Misc` module and sharing code.

lib/PTEVal.ml
```diff
@@ -38,20 +45,25 @@ and pp_af ok = pp_int_field ok "af" (fun p -> p.af)
...
+                    attrs_eq (fun p -> p.attrs)
 
 let set_oa p s = { p with oa = Misc.add_physical s; }
 
 let is_default t =
   let d = prot_default in
-  t.valid=d.valid && t.af=d.af && t.db=d.db && t.dbm=d.dbm && t.el0=d.el0
+  t.valid=d.valid && t.af=d.af && t.db=d.db && t.dbm=d.dbm && t.el0=d.el0 && t.attrs=[]
+
+let is_valid t = t.valid = 1
```
maranget, line 57: If I am not wrong, this function `is_valid` is not called anywhere and is dead code.

herd/machModelChecker.ml
```diff
@@ -306,6 +306,22 @@ module Make
                  "F", E.is_barrier;
                  "DATA", is_data_port;
                  "NDATA", (fun e -> not (is_data_port e));])) in
+      let evts_map = E.EventSet.fold
+                (fun e evts_map ->
+                  let attrs = S.A.V.memattr_as_list (E.Act.memattr_of e.E.action) in
+                  List.fold_right
+                    (fun attr evts_map ->
+                      let evts_w_attr =
+                        match StringMap.find_opt attr evts_map with
```
maranget, line 315: There is a `StringMap.safe_find` _default_  _key_ _map_ function that would enable writing terser code (see `lib/MyMap.ml`).

herd/machModelChecker.ml
```diff
@@ -306,6 +306,22 @@ module Make
...
+                        | Some evts -> E.EventSet.add e evts
+                        | None -> E.EventSet.add e E.EventSet.empty
+                      in
+                      let evts_map = StringMap.remove attr evts_map in
```
maranget, line 319: This call is not useful as maps keep only one binding for a given key.

maranget:
Followup on  https://github.com/herd/herdtools7/pull/129#discussion_r556727774.

On not adding a new argument to constructor `Access`. Here is an alternative plan, which I tend to favor since it avoids the extra constructor argument.

1. Define set of events that would apply to `Rpte` events, sets are based upon the presence of the attribute in value read (or written, for generality ?). Those sets are to be defined in `machModelChecker.ml` as the attribute names are extracted from present events. Hence the technique may require exporting adequate function(s) from `machAction.ml`.
2. In cat file one can connect adequate  `Rpte` to the corresponding physical access M, as (1) M are explicit accesses (set EXP) (2) there is a non-ambiguous intrinsic relation from `Rpte` to `M`.

So considering set PTEattr for a given attribute, one should be able to compute the set Mattr of memory accesses with attribute attr in cat file as:
```
let Mattr = range ([R&Mattr]; intrinsic; [M&Exp])
```
(This is inspired by the computation of set `NFME` in a cat file by @jalglave)

Moreover the clever technique used to create the event sets (would it be Mattr or PTEattr) entails a specific problem: cat model may refer to names that have not been defined. The solution is to insert the following cat code before refering to those names:
```
let PTEattr = try PTEattr with emptyset
```

maranget (commented):
Empty review

herd/AArch64Sem.ml
maranget: I have made specific comment to address the question https://github.com/herd/herdtools7/pull/129#issuecomment-760123055.

lib/symbConstant.ml
maranget: Let us follow @jalglave  here.

relokin:
> Followup on [#129 (comment)](https://github.com/herd/herdtools7/pull/129#discussion_r556727774).
> 
> _La nuit porte conseil_
> 
> On not adding a new argument to constructor `Access`. Here is an alternative plan, which I tend to favor since it avoids the extra constructor argument.
> 
>     1. Define set of events that would apply to `Rpte` events, sets are based upon the presence of the attribute in value read (or written, for generality ?). Those sets are to be defined in `machModelChecker.ml` as the attribute names are extracted from present events. Hence the technique may require exporting adequate function(s) from `machAction.ml`.
> 
>     2. In cat file one can connect adequate  `Rpte` to the corresponding physical access M, as (1) M are explicit accesses (set EXP) (2) there is a non-ambiguous intrinsic relation from `Rpte` to `M`.
> 
> 
> So considering set `PTEattr` for a given attribute, one should be able to compute the set `Mattr` of memory accesses with attribute attr in cat file as:
> 
> ```
> let Mattr = range ([R&PTEattr]; intrinsic; [M&Exp])
> ```
> 
> (This is inspired by the computation of set `NFME` in a cat file by @jalglave)
> 

That's very cool, in some ways it can help with other open questions I had in my mind (e.g., how do we check that a PTE attribute is valid if they become events like all others, having a special prefix could work but it's not a very satisfying solution?).

I have only one concern which is that the implicit `Rpte` events should also be annotate with attributes themselves (shareability, cacheability, ...). They are like regular load instructions and their memory attributes are determined by system registers (e.g., `TCR_EL0.SH0`). I suspect that we won't want to add meta implicit events. We can also ignore this problem for now, I think the solution you're proposing is really neat, up to you.

maranget:
> That's very cool, in some ways it can help with other open questions I had in my mind (e.g., how do we check that a PTE attribute is valid if they become events like all others, having a special prefix could work but it's not a very satisfying solution?).

I do not quite understand "they" above, do you mean PTE reads ? Attributes ?

maranget:
> I have only one concern which is that the implicit `Rpte` events should also be annotate with attributes themselves (shareability, cacheability, ...). They are like regular load instructions and their memory attributes are determined by system registers (e.g., `TCR_EL0.SH0`). I suspect that we won't want to add meta implicit events. We can also ignore this problem for now, I think the solution you're proposing is really neat, up to you.

I have no idea on how implicit page table access can be controlled. At the moment I'd suggest adopting some immutable behaviour that corresponds to sensible defaults.

Explicit page table accesses are yet another story...

relokin:
> > That's very cool, in some ways it can help with other open questions I had in my mind (e.g., how do we check that a PTE attribute is valid if they become events like all others, having a special prefix could work but it's not a very satisfying solution?).
> 
> I do not quite understand "they" above, do you mean PTE reads ? Attributes ?

By "they" I meant PTE attributes. With your solution on how to deal with PTE attributes, I think it won't be as hard. This will become more clear when we start looking at the actual cat file.

> > I have only one concern which is that the implicit `Rpte` events should also be annotate with attributes themselves (shareability, cacheability, ...). They are like regular load instructions and their memory attributes are determined by system registers (e.g., `TCR_EL0.SH0`). I suspect that we won't want to add meta implicit events. We can also ignore this problem for now, I think the solution you're proposing is really neat, up to you.
> 
> I have no idea on how implicit page table access can be controlled. At the moment I'd suggest adopting some immutable behaviour that corresponds to sensible defaults.
> 
> Explicit page table accesses are yet another story...

I just had a chat with @jalglave as well and she suggested the same, we should probably ignore this for now. A pagetable walk corresponding to `Rpte_x` is far more complex and the memory attributes of that walk is just a small detail :).

maranget:
> > > That's very cool, in some ways it can help with other open questions I had in my mind (e.g., how do we check that a PTE attribute is valid if they become events like all others, having a special prefix could work but it's not a very satisfying solution?).
> > 
> > 
> > I do not quite understand "they" above, do you mean PTE reads ? Attributes ?
> 
> By "they" I meant PTE attributes. With your solution on how to deal with PTE attributes, I think it won't be as hard. This will become more clear when we start looking at the actual cat file.
> 
> > >

I do not understand why the attributes themselves become events. Given some attribute `<attr>` my suggestion was to predefine the set of events `PTE<attr>` and then to compute a set `M<attr>` in the cat file. I guess we agree on this central point and that we'll see consequences later...

lib/PTEVal.ml
relokin: ack, this is now resolved.

lib/PTEVal.ml
relokin: I happy to add a functor, I've uploaded a version where the AArch64 defaults are hardcoded  but if you prefer the  clean solution, I'm happy to give it a go.

lib/PTEVal.ml
relokin: Ack, the new version has resolved this.

herd/machModelChecker.ml
relokin: The code has changed slightly but the new version uses `StringMap.safe_find`

herd/machModelChecker.ml
relokin: Ack, I wasn't sure if the default behavior would be to keep or replace the old mapping. The code has changed a bit but the new version doesn't use `StringMap.remove`

relokin:
Ok that was easier than I thought. The new version of the code raises events `PTE<attr>` for every memory attribute in the PTE as read read by the implicit `Rpte_x`. The only unresolved issue from the previous review is the value of the default memory attributes which is now hardcoded to that for AArch64.

jalglave:
Thanks Nikos. @maranget do you object to merging this as it is? (cf what Nikos highlights: "The only unresolved issue from the previous review is the value of the default memory attributes which is now hardcoded to that for AArch64.")
Thanks in advance,
Jade

maranget:
> Thanks Nikos. @maranget do you object to merging this as it is? (cf what Nikos highlights: "The only unresolved issue from the previous review is the value of the default memory attributes which is now hardcoded to that for AArch64.")
> Thanks in advance,
> Jade

I like the recent change, let us merge. I'd also like to perform the PTE<attr> setting only in kvm mode, for a minor performance improvement in non-kvm mode. I'll commit a change after merge, so as not to delay this long-awaited merge.
Thanks to @relokin and all.

relokin:
Thank you both for the reviews!

PR #130, opened 2020-12-03
[gen] Add Hat to forbidden.conf

benbancroft:
This commit adds Hat to the relaxation list.

It was found that Hat allows for the generation
of some interesting tests that have multiple reads
of the same write, which is particurly useful if
the test also uses Amo.*


jalglave:
Thanks Ben.
Jade

PR #131, opened 2020-12-07
[litmus/lib] Add support for Neon instructions

BreD1810:
This PR:
- Adds parsing support for Neon registers and the various ways of accessing them.
- Adds parsing support for LD1-LD4, ST1-ST4 multiple/single structure, and replicate variants.
- Adds parsing support for SIMD LDR/STR and LDUR/STUR.
- Adds parsing support for SIMD MOV and MOVI.
- Adds support for generating the above instructions using the litmus tool.
- Adds catches in jingle/herd to stop warnings on build.

Note: We added a dependency for `str` for using regex to parse Neon registers with arrangement specifiers. Perhaps there is a better way of doing this without the need for a dependency?

We have also written several litmus tests to test the changes. These were implemented in our CI but removed for upstreaming. These tests are available [here](https://drive.google.com/file/d/1nJPPD_8LAi0AVBQF0jAWwmU5EQ3aL0b4/view?usp=sharing). Note that the post-index examples have empty expected outputs, matching the expected output for existing post-index instructions.


BreD1810:
We have now added support for final constraint checking for Neon registers. This means we have parts of Luke's vector patch within this PR.

One issue is that currently we have no way to initialise the contents of the vector register to 0. Neon instructions mostly affect the file width of the register, so this isn't a huge issue. However, certain move instructions and single structure loads may observe this issue. A workaround is to manually set the register contents in the litmus tests, such as:
```
MOVI V0.4S,#0
```

We have tried a few different fixes, but with no success:
- Setting the register value to zero (i.e. `register int128_t stable_v0 asm("v0") = 0`). However this appears to have no affect in either GCC of Clang.
- Adding the input of the register as 0, e.g.
```
// Assembly here
:[v0] "=w" (stable_v0)
:"[v0]" (0)
:"cc","memory"
```
However, we have observed that GCC allocates registers that are not what we asked it to use (although Clang works fine in our testing). This is a problem as the register lists in Neon instructions have to be in ascending order.

litmus/compile.ml
```diff
@@ -45,9 +45,12 @@ module Generic (A : Arch_litmus.Base)
       let pointer = CType.Pointer base
       let code_pointer = Pointer (Base "ins_t")
       let tag = Base "tag_t"
+      let array_ty sz = CType.Array ("int", sz)
+        (* is it right to assume int as the base type?*)
```
BreD1810, line 49: This assumption limits us to only having checks by using 32 bit element sizes (e.g. we can safely `V0.4S`, but not `V0.16B`). This forces the user to change their final constraint to interpret the Neon register as having 32 bit lanes.

We could infer the element sizes from the number of elements in the vector. However, this change would be Neon specific, and therefore not ideal.

maranget:
> However, we have observed that GCC allocates registers that are not what we asked it to use (although Clang works fine in our testing). This is a problem as the register lists in Neon instructions have to be in ascending order.
You may be interested by an undocumented feature of litmus: You can declare some register in metadata to be stable (`Stable=<register list>`), where `register list` is a comma separated list or register names. Then, litmus attempts to instruct the C compiler to allocates those registers as themselves. You may have to modify the feature so as to handle Neon registers.

BreD1810:
> You may be interested by an undocumented feature of litmus: You can declare some registers in metadata to be stable (`Stable=<register list>`), where `register list` is a comma separated list or register names. Then, litmus attempts to instruct the C compiler to allocates those registers as themselves. You may have to modify the feature so as to handle Neon registers.

We already have our registers set as stable using `stable_regs` in `litmus/AArch64Compile_litmus.ml`. I think this would have the same affect as the metadata change you suggested. This solved the register allocation problem generally speaking. 

However, the issue resurfaces when we try to initialise the vector registers to 0. For example, if we edit some C code produced by litmus to set the register values in the input section:
```c
  for (int _i = _size_of_test-1 ; _i >= 0 ; _i--) {
    barrier_wait(_th_id,_i,&barrier[_i]);
    register int128_t stable_v0 asm("v0");
    register int128_t stable_v1 asm("v1");
    register int128_t stable_v2 asm("v2");
asm __volatile__ (
"\n"
"#START _litmus_P0\n"
"#_litmus_P0_0\n\t"
"movi %[v0].2s,#1\n"
"#_litmus_P0_1\n\t"
"movi %[v1].2s,#1\n"
"#_litmus_P0_2\n\t"
"st2 {%[v0].2s,%[v1].2s},[%[x0]]\n"
"#_litmus_P0_3\n\t"
"ld1 {%[v2].4s},[%[x0]]\n"
"#_litmus_P0_4\n\t"
"st1 {%[v2].s}[0],[%[x1]]\n"
"#_litmus_P0_5\n\t"
"st1 {%[v2].s}[1],[%[x2]]\n"
"#_litmus_P0_6\n\t"
"st1 {%[v2].s}[2],[%[x3]]\n"
"#_litmus_P0_7\n\t"
"st1 {%[v2].s}[3],[%[x4]]\n"
"#END _litmus_P0\n\t"
:[v2] "=&w" (stable_v2),[v1] "=&w" (stable_v1),[v0] "=&w" (stable_v0)
:[x0] "r" (&_a->b[_i]),[x1] "r" (&_a->x[_i]),[x2] "r" (&_a->y[_i]),[x3] "r" (&_a->z[_i]),[x4] "r" (&_a->a[_i]),"[v2]" (0),"[v1]" (0),"[v0]" (0)
:"cc","memory"
);
  }
  stabilize_globals(0,_a);
  mbar();
  return NULL;
}
```
Compilation with GCC then gives the following error:
```
/tmp/ccTOOKyg.s: Assembler messages:
/tmp/ccTOOKyg.s:490: Error: invalid register list at operand 1 -- `st2 {v2.2s,v1.2s},[x1]'
```

maranget:
Ah, this look like a gcc bug or limitation, have you tried a recent version of gcc ? I guess that, for now, the only solution available is using Clang. As for tests one may indeed consider explicit initialisation to be the safest option.

BreD1810:
We've got the same results on GCC 10.2.0, 8.3.0 and also a version of 9 (not sure which exactly).

maranget:
Ok, did you try replacing the constraint modifier "=" by "+" (or maybe "=&" by "+"). I just discovered the existence of "+" by reading gcc. documentation. I may change something by miracle.

BreD1810:
Just tried that now. The registers allocate as we ask, but we are back to having the issue of the register not being initialised to 0 as using `+` means we have to remove the input operands. If we don't, we get the compile error:
```c
test.c:172:1: error: inconsistent operand constraints in an ‘asm’
 asm __volatile__ (
 ^~~
test.c: In function ‘P0’:
test.c:172:1: error: inconsistent operand constraints in an ‘asm’
 asm __volatile__ (
 ^~~
```

maranget:
I tried this by hand editing a simple litmus test. It is working as expected. Did you tried using the constraint "+" and the explicit initialisation of the stable_x0 variable. If you have not yet, It is worth trying.
```
  register int stable_x0 asm("x0") = 1;
asm __volatile__ (
"\n"
"#START _litmus_P0\n"
"#_litmus_P0_0\n\t"
"add %w[x0],%w[x0],#1\n"
"#END _litmus_P0\n\t"
:[x0] "+r" (stable_x0)
:
:"cc","memory"
);
```

BreD1810:
It seems that works without causing any issues with register allocation. Thank you very much for your help on that. I've pushed a new commit with changes to ensure the Neon registers are initialised to 0 when needed.

jalglave:
hi all, please can you make sure to build on the vector patch which just went in?
Thanks,
Jade

BreD1810:
We've now rebased on master. This passes our CI pipeline, but we've not done any extensive manual testing.

The litmus tests we used can still be found [here](https://drive.google.com/file/d/1nJPPD_8LAi0AVBQF0jAWwmU5EQ3aL0b4/view?usp=sharing). V34-V37 are the tests specifically including final state checking.

maranget:
@BreD1810 Hi, I am considering merging this PR #131. Is there any reason for your  tests not to be part of the PR?

BreD1810:
@maranget The tests I mentioned weren't included as they were for the litmus tool instead of the herd tool. We wrote a small test suite for the litmus tool - a script to check that litmus doesn't error when compiling them to C code, and another that compiles the C and executes it (which should be run on an AArch64 device).

These scripts are somewhat hacky and were written quickly for use in our CI. However, I can add these scripts and tests if you would like?

maranget:
> These scripts are somewhat hacky and were written quickly for use in our CI. However, I can add these scripts and tests if you would like?

Hum, Yes having the tests in the PR would be a real bonus, you can add the tests in a `litmus/tests/...` directory by imitation of `herd/tests`. The scripts probably are too specific to your installation, are they? Perhaps you can simply add a short README.md file to your test directory with directives to compile and run the tests.

It can also be interesting to include the .expected files, less so for the .expected.t files.

BreD1810:
I've now added the litmus tests and expected outputs.

I've added the scripts on a separate branch [here](https://github.com/BreD1810/herdtools7/commit/469d2806631058ba1ac261e6068c6cfe3a49246f). `litmus-tests.sh` checks that the litmus tool generates C without error, `pi-litmus.sh` compiles and executes the C, checking the expected output.

I think `pi-litmus.sh` is fine targeting `bash`, as this is include with Raspbian (everything used in the script works without additional package installs on Raspbian 64-bit). However, if the script should target more general AArch64 devices, this should probably be changed to `sh`.

I just tested `litmus-tests.sh` with dash (`/bin/sh` on debian systems) and it doesn't work. This seems to be caused by the if statement using a bashism.

Would you like me to go ahead with the README you recommended, or include/edit these scripts?

maranget:
Please do not bother with shell scripts, it is nearly impossible to write portable shell scripts, especially in a hurry. A README file that describes the files is sufficient.

The general plan will be to integrate litmus testing in @eth-arm testing infrastructure.

FYI, hat I will do for testing is concatenating all your .expected files into one big expected file, run all tests somewhere and then compare expected file and output file with the specific` mcompare7` tool. Final litmus testing procedure will probably be different.

BreD1810:
I've added the README as suggested.

maranget:
I have a problem with test V10.litmus, what is the exact meaning of instructions `LDR B0,[X0,#8]` and `STR B0,[X2]` ?

BreD1810:
The `LDR` loads SIMD register `B0` with data from `x` (pointed to by `X0`) with a 8 byte immediate unsigned offset (see C7-1836 of the Architecture Reference Manual).

The `STR` then loads the value in `BO` to memory location `y` pointed to by `x2`.

maranget:
My actual question is more about the size of register B0 and thus about the size of variables x and y.

maranget:
Ah fine, where are those `B` registers defined ? Anyway, here is my definition of `V10.litmus`
```
AArch64 V10
(* Tests load / store immediate, with offset *)
{
int64_t x;
int8_t y;
0:X0=x; 0:X2=y;
}

 P0                 ;
 MOV X1,#1          ;
 STR X1, [X0, #8]   ;
 LDR B0, [X0, #8]   ;
 STR B0, [X2]       ;
 forall (y=1)
```

BreD1810:
C1.2.5 of the Architecture Reference Manual gives details on the various forms of SIMD and FP registers, more specifically on page C1-185 for the `B` register representation.

`B0` points to register `V0`, but only referring to the lower 8 bits. We've listed them in `bvrs` in `lib/AArch64Base.ml`. The `parse_simd_reg` functions is used to parse them.

maranget:
Thanks.

maranget:
@BreD1810 I merge the PR, thanks for you work.

I still have a few minor changes to perform (in tests). I'll commit them right after the merge. Thanks again.

PR #132, opened 2020-12-07
[internal_lib] Unify Command.run* into a single Command.run

eth-arm:
The number of functions to run commands has grown to 3:
- `Command.run`,
- `Command.run_with_stdout`,
- `Command.run_with_stdin_and_stdout`.

As upcoming PR (#133) is going to need to use the stderr of a command, this is going to grow to 4, and may eventually have one for each of the full powerset of {`stdin`,`stdout`,`stderr`}.

Instead, this PR implements One True Function, `Command.run`, which takes 3 optional function parameters, one for each of `~stdin`, `~stdout`, and `~stderr`. The named arguments should also be easier to read than positional arguments after a long function name.

This PR also backports `Fun.protect` and `Option.get` from OCaml 4.08.

This PR blocks PR #133, which in-turn blocks PR #94.

### DIFF
--- a/internal/lib/command.ml
+++ b/internal/lib/command.ml
@@ -16,6 +16,9 @@
 
 (** Utilities for running commands. *)
 
+module Fun = Base.Fun
+module Option = Base.Option
+
 exception Error of string
 
 let command bin args =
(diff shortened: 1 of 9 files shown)


internal/lib/command.ml
```diff
@@ -16,36 +16,89 @@
...
-  in_f stdin ;
-  let ret = out_f stdout in
-  match Unix.close_process (stdout, stdin) with
-  | Unix.WEXITED 0 -> ret
+(* Sentinel values for the optional function parameters to [run]. *)
+let unset_in (_ : out_channel) = failwith "unset_in"
+let unset_out (_ : in_channel) = failwith "unset_out"
+let unset_err (_ : in_channel) = failwith "unset_err"
+
+let run ?stdin:(in_f=unset_in) ?stdout:(out_f=unset_out) ?stderr:(err_f=unset_err) bin args =
```
maranget, line 35: Hi Ethel,

Are you aware of another way to use optional arguments?
```OCaml
let run ?stdin:in_f ... =
    ...
    let in_fd,in_f,close_in_pipe = match in_f with
    | None -> Unix.stdin,nop,nop
    | Some in_f -> 
         let fd, o = out_pipe () in
         fd, (fun _ -> Unix.close fd ; in_f o), (fun _ -> close_out o) in
```
I find this solution more elegant, since it avoids the never called sentinel values and the `==` comparison. If interested, have a look [at section 4.1.1 in OCaml manual](https://caml.inria.fr/pub/docs/manual-ocaml/lablexamples.html#ss:optional-arguments).

eth-arm: That is much better! Thank you! :D

maranget: Not that much. Thanks, merging imminent.

PR #135, opened 2020-12-08
[litmus] Add EL0 mode for KVM AArch64

maranget:
Tests metadata may specify EL0 mode for a comma separated list of thread names:
`EL0=P<i1>,P<i2>,...`

Switch to EL0 is performed at the beginning of thread code (by `eret`
with proper setting of `sp_el0` (per cpu user stack), `elr_el1`
(next instruction after eret) and spsr_el1 (all zero).

Return to EL1 is performed by a system call `svc`. Handler code changes
saved pstate so as to fool kvm infrastructure and return to EL1. Surprisingly
it looks as if the trick is working.

Notice that stacks are different in EL0 and EL1 mode, which may
cause problems (not yet observed). Namely, thread code is in a function body.
Switch to EL0 is performed as late as possible, so as
to alleviate the risk of a reference to stack pointer crossing
this boundary.

### DIFF
--- a/lib/lexScan.mli
+++ b/lib/lexScan.mli
@@ -15,6 +15,8 @@
 (****************************************************************************)
 
 (** Miscellaneous lexers *)
+exception Error
 
 val is_num : string -> bool
 val info : string -> (string * string) option
+val procs : string -> Proc.t list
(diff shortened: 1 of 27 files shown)


maranget:
Support of currently running experiments. Merge for avoiding confusion. However, some additional work to integrate an `el0` field in pteval's would have been welcome.

PR #137, opened 2021-01-07
[jingle,litmus] Allow setting libdir with -set-libdir

eth-arm:
This PR allows setting the libdir for Jingle & Litmus binaries with a flag, `-set-libdir`.

This was added to Herd in commit d465040feb53c1363c8c63868b4b6297195d7ad9.


jalglave:
Thanks Ethel.

PR #140, opened 2021-01-12
Vector reloaded

maranget:
This PR extends #94 with semantical elements, resulting in better separation between high level locations (with unscaled array accesses) and low-level ones.

One benefit is printing high-level locations as then occur in input.


jalglave:
Thanks Luc.
Jade

PR #142, opened 2021-01-14
[herd] Adjust AArch64 CAS instruction semantics.

maranget:
Reference `CAS <Xs>, <Xt>, <Xn>`

 - in all cases all of the following apply:
   - D1: there is an Intrinsic data dependency from the Register read effect
     of Xn to the Memory read effect of the Memory Location addressed by Xn
   - C1: there is an Intrinsic control dependency from the Register read
     effect of Xs to the Register write effect of Xs
   - C2: there is an Intrinsic control dependency from the Memory read effect
     of the Memory Location addressed by Xn to the Register write effect of Xs

 - if the CAS instruction fails all of the following apply:
   - Df1: there is an Intrinsic data dependency from the Memory read effect
     of the Memory Location addressed by Xn to the Register write effect of XsD

 - if the CAS instruction succeeds all of the following apply:
   - Ds1: there is an Intrinsic data dependency from the Register read effect
     of Xs to the Register write effect of Xs
   - Ds2: there is an Intrinsic data dependency from the Register read effect
     of Xn to the Memory write effect of the Memory Location addressed by Xn
   - Ds3: there is an Intrinsic data dependency from the Register read effect
     of Xt to the Memory write effect of Xn
   - Cs1: there is an Intrinsic control dependency from the Register read
     effect of Xs to the Memory write effect of the Memory Location addressed
     by Xs
   - Cs2: there is an Intrinsic control dependency from the Memory read effect
     of the Memory Location addressed by Xn to the Memory write effect of the
     Memory Location addressed by Xn

### DIFF
--- a/herd/event.ml
+++ b/herd/event.ml
@@ -1619,13 +1619,14 @@ module Make  (C:Config) (AI:Arch_herd.S) (Act:Action.S with module A = AI) :
              wrs.intra_causality_data
              rm.intra_causality_data)
           (EventRel.union
-             (EventRel.cartesian output_rn input_rm)
-             (EventRel.cartesian output_rm input_wrs));
+             (EventRel.cartesian output_rn input_rm)    (* D1 *)
+             (EventRel.cartesian output_rm input_wrs)); (* Df1 *)
         intra_causality_control =
-        EventRel.union5
+        EventRel.union6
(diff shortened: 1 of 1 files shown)


jalglave:
Thank you Luc. I'll fix the typos in the commit message and merge.
Thanks again,
Jade

PR #144, opened 2021-01-15
[all] Remove unused OCaml files

eth-arm:
This PR removes unused OCaml files. It was tested by making the main project with `dune` and `ocamlbuild`, and also the `doc` and `herd-www` subprojects.


maranget:
Fine, thanks, merged.

PR #145, opened 2021-01-15
[herd] Add binary herd_catalogue_regression_test to test Catalogue shelves

eth-arm:
This PR adds a tool, `herd_catalogue_regression_test` that compares the output of running `herd7` with the contents of a `shelf.py` against a directory of `kinds.txt` files.

It also:
- Adds modules `Log` and `Kinds` for parsing `herd7` output logs and `kinds.txt` files respectively.
- Makes some changes to `TestHerd` to support optional arguments and multiple `.litmus` files per invocation.
- Adds some baseline `kinds.txt` files.

The tool can be used from the commandline like:

```sh
% herd_catalogue_regression_test \
    -herd-path _build/default/herd/herd.exe \
    -libdir-path ./herd/libdir \
    -kinds-dir catalogue/aarch64-MTE-mixed/kinds \
    -shelf-path catalogue/aarch64-MTE-mixed/shelf.py \
    -variant mixed \
    -variant memtag \
    test
```

In addition to `test`, it also has subcommands:
- `show`, which prints the `herd7` commands it would have run with `test`.
- `promote`, which runs `herd7` and updates the contents of `-kinds-dir` with the output.


jalglave:
Thanks @eth-arm this is great. As discussed, as a followup PR, if you could add the ability for your tool to know which arguments to use when used standalone, that would be great.
Thanks again,
Jade

PR #146, opened 2021-01-15
[herd] Adjust internal dependencies for ARM (AArch32).

maranget:
Simplify internal dependencies of the few implemented predicated instructions (ie loads, stores and moves).


jalglave:
Thanks Luc.
Jade

PR #148, opened 2021-01-19
Set internal control dependency from PTE read to db test-and-set attempt, replacing data dependency.

maranget:
It is more logical to have such an internal dependency from
page table read to db test-and-set, as the latter is attempted
when pteval bits have some values (db:0, dbm:1, valid:1) and not
attempted otherwise.


jalglave:
Thanks Luc.
Jade

Issue #149, opened 2021-01-19
make test fails with Python 3.


BreD1810:
My machine uses python 3.9.1, which causes this error with `make test`:
```
sh ./dune-build.sh $HOME
+ ./version-gen.sh /home/brad
+ dune build --profile release @all
Done: 4798/4800 (jobs: 1)
dune runtest --profile=release
  shelf_test alias internal/lib/tests/runtest (exit 2)
(cd _build/default/internal/lib/tests && ./shelf_test.exe)
Traceback (most recent call last):
  File "<stdin>", line 4, in <module>
NameError: name 'execfile' is not defined
Failed Shelf.of_file reads simple valid files: raised exception
Fatal error: exception Command.Error("Process returned error code 1")
Done: 1138/1141 (jobs: 1)make: *** [Makefile:78: dune-test] Error 1
```

It also fails in our CI running Python 2.7.

Python 3.9.1 can be supported by changing the `execfile` call (line 66 `internal/lib/shelf.ml`) to:
```ocaml
    Printf.fprintf chan "exec(open(%s).read(), m)\n" (Filename.quote (Filename.basename path)) ;
```
Note that `execfile` was [removed in Python 3.3](https://docs.python.org/3.3/whatsnew/3.0.html?highlight=execfile#builtins) so I'm not sure which Python version this is targetting?
Perhaps the targeted Python version should be documented in `install.md`?

eth-arm:
What OS & version is your machine & your CI running?

BreD1810:
My machine is running Python 2.7.16, OS is the latest version of Arch Linux.

The CI running in a docker container, which you can run locally by pulling `registry.gitlab.com/bred1810/docker-herdtools:latest`. Dockerfile is [here](https://gitlab.com/BreD1810/docker-herdtools/-/blob/master/Dockerfile).
This uses Python 2.7.16, OS is Debian 10.

eth-arm:
PR #152 adds support for both systems where `python` is Python 2 and where it is Python 3.

However, `execfile` [definitely exists in Python 2.7](https://docs.python.org/2.7/library/functions.html#execfile), so that should only fail if the Python binary that is actually being run is not actually 2.7.

BreD1810:
Ah, you're right. I misremembered the CI's error when investigating on my machine. That error appears related to #150.
PR #152 fixes this issue for me. Thank you for fixing that.

BreD1810:
Resolved by #152

PR #152, opened 2021-01-19
[internal_lib] Support parsing shelves on systems with no Python2

eth-arm:
On macOS & some Linux distributions `python` is legacy Python 2, and `python3` is Python 3. However, on other Linux distributions `python` is Python 3 and there is no Python 2.

One of the changes from Python 2 to 3 was removing `execfile()` in favor of `exec()`. This commit uses the correct one depending on Python version.

This PR fixes Issue #149.


maranget:
Hi @eth-arm, thanks for the fix. Given that @BreD1810 validates the fix, I merge.
I do not feel very comfortable with version depending code. It would be nice if such tricks could be kept to a (very low) minimum.

Issue #150, opened 2021-01-19
run_built_binary fails on linux


BreD1810:
Line 29 makes a call to `stat` using `-f`. On Linux, this displays file system status, so the following error occurs:
```bash
make test                                                                                                                   

...

internal/run_built_binary herd_catalogue_regression_test \
	-herd-path _build/install/default/bin/herd7 \
	-libdir-path herd/libdir \
	-kinds-dir catalogue/aarch64/kinds \
	-shelf-path catalogue/aarch64/shelf.py \
	test
stat: cannot read file system information for '%m %N': No such file or directory
internal/run_built_binary: line 45: Total:: command not found
make: *** [Makefile:110: test] Error 127
```

I believe the option is supposed to be for formatting, so perhaps `--format` is preferred?

The man page is available [here](https://linux.die.net/man/1/stat).

Once I apply this fix, the script still fails, this time with:
```bash
internal/run_built_binary herd_catalogue_regression_test \
	-herd-path _build/install/default/bin/herd7 \
	-libdir-path herd/libdir \
	-kinds-dir catalogue/aarch64/kinds \
	-shelf-path catalogue/aarch64/shelf.py \
	test
internal/run_built_binary: line 45: '/home/brad/gdp/repo/_build/default/internal/herd_catalogue_regression_test.exe': No such file or directory
make: *** [Makefile:110: test] Error 127
```

This file does exist, but apostrophes are not stripped from around the filename, so it cannot find the file.
I can make the test pass by changing line 45 to:
```bash
command="$(most_recent ${paths} | xargs) $@"
$command
```
I'm not the best with Bash, so perhaps there is a better way.

BreD1810:
It seems that this script also fails with the shell "dash", which I think is `/bin/sh` on Debian systems (it certainly is in our CI docker image).

Running the script gets the following error:
```bash
internal/run_built_binary: 17: internal/run_built_binary: Syntax error: "(" unexpected
make: *** [Makefile:110: test] Error 2
```

The `function` keyword is a ["bashism"](https://wiki.ubuntu.com/DashAsBinSh#function), hence breaking non-bash compatible shells like dash. Simply removing them fixes the script whilst maintaining bash support:
```bash
absolute_path() {
  readonly prev_pwd="${PWD}"
  cd "${1}"
  pwd
  cd "${prev_pwd}"
}

most_recent() {
  # List (mtime, path) pairs.
  # Sort them descending by mtime.
  # Take the first of the list.
[...]

PR #153, opened 2021-01-19
[build] Make various shell scripts more portable

eth-arm:
This PR improves the portability of various shell scripts in the repository:
- The GNU Coreutils and BSD versions of `stat` take different flags, so hopefully do the correct thing on each platform.
- Use `/bin/sh` instead of `/bin/bash`, as `/bin/bash` is not guaranteed to exist.
- Remove the `function` keyword from `run_built_binary`, as that is a "[bashism](https://wiki.ubuntu.com/DashAsBinSh#function)".

This PR fixes Issue #150.

### DIFF
--- a/dune-install.sh
+++ b/dune-install.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 
 set -eu
 
--- a/dune-uninstall.sh
+++ b/dune-uninstall.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 
 set -eu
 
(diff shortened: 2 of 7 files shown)


maranget:
@eth-arm thanks for this PR. Replacing shell scripts by python ones surely goes in the right direction as regards portability. My personal taste would favor OCaml scripts over Python ones. The script `generate_includes.ml/generate_names.ml` in `herd-www/` is an example of such an OCaml script. Nevertheless, if you feel more comfortable with Python, which will probably yield terser code, please go ahead.

eth-arm:
I would have preferred to do it with pure OCaml, however it would create a bootstrap cycle:
- To find the most recent copy of a binary, the script needs to read a file's `mtime`.
- To read `mtime` with OCaml, the script would need the `Unix` library.
- The `Unix` library is not in the toplevel and requires compilation.
- To run _that_ compiled binary would require a tool to find and run compiled binaries.

PR #158, opened 2021-01-27
Pte squared

maranget:
Handle accesses to the page table as ordinary, virtual, accesses. Caveats.
- Access to page table control is more than just page tables attributes.
- Altering the page table entry of page table pages is not supported.
Also notice that the mode is enabled dynamically by variant `-variant pte-squared`.

### DIFF
--- a/herd/AArch64Sem.ml
+++ b/herd/AArch64Sem.ml
@@ -314,12 +314,14 @@ module Make
 
          *)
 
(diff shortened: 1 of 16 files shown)


maranget:
This PR (to the KVM branch) is not perfect. It will nevertheless be merged into KVM so as to avoid maintaining yet another branch. Activation is controlled by `-variant pte2`.

PR #159, opened 2021-01-29
[herd] AArch64, do *not* interpret `SWP _,ZR,[_]` as a store.

maranget:
This PR fixes a semantic error that was lurking since the introduction
of the SWP instruction.


jalglave:
Thanks Luc - I've tested it and will merge it now.
Jade

PR #160, opened 2021-02-03
[gen] Add support for Neon

BreD1810:
This PR adds basic support for the Neon extension.
The user can use the annotation `N1` to `N4` to ask diy to emit Neon instructions.

Current limitations:
- Memory sizes are limited to `uint128`, which could be exceeded by larger instructions (e.g. `st4`). This could be updated to use a larger array type.
- The final condition generated only covers 1 non-sc case. There are several, and some of the generalised final conditions become quite complex. A rough document on the generalised conditions can be found [here](https://docs.google.com/document/d/16jE5fCFCWA1JuSffHELIPbFwOb6Ek13NaGeCtL2Y3p4/edit?usp=sharing).

### DIFF
--- a/gen/AArch64Arch_gen.ml
+++ b/gen/AArch64Arch_gen.ml
@@ -503,7 +503,9 @@ include
 
       let pp_reg = pp_reg
       let free_registers = allowed_for_symb
-      include NoSpecial
+
+      type special = reg
+      let specials = vregs
     end)
 
 end
--- a/gen/AArch64Arch_gen.ml
+++ b/gen/AArch64Arch_gen.ml
@@ -34,6 +34,7 @@ let do_tag = C.variant Variant_gen.MemTag
 let do_morello = C.variant Variant_gen.Morello
 let do_kvm = C.variant Variant_gen.KVM
 
+let do_neon = C.variant Variant_gen.Neon
 open Code
 open Printf
 
@@ -62,9 +63,10 @@ type atom_pte =
   | Read|ReadAcq|ReadAcqPc
   | Set of w_pte
   | SetRel of w_pte
+type neon_sizes = N1 | N2 | N3 | N4
 type atom_acc =
   | Plain of capa_opt | Acq of capa_opt | AcqPc of capa_opt | Rel of capa_opt
-  | Atomic of atom_rw | Tag | CapaTag | CapaSeal | Pte of atom_pte
+  | Atomic of atom_rw | Tag | CapaTag | CapaSeal | Pte of atom_pte | Neon of neon_sizes
 
 type atom = atom_acc * MachMixed.t option
 
@@ -107,6 +109,11 @@ let applies_atom (a,_) d = match a,d with
      | ReadAcqPc -> "Q"
      | Set set -> pp_w_pte set
      | SetRel set -> pp_w_pte set ^"L"
+   let pp_neon_size = function
+     | N1 -> "1"
+     | N2 -> "2"
+     | N3 -> "3"
+     | N4 -> "4"
 
    let pp_atom_acc = function
      | Atomic rw -> sprintf "X%s" (pp_atom_rw rw)
@@ -118,6 +125,7 @@ let applies_atom (a,_) d = match a,d with
      | CapaTag -> "Ct"
      | CapaSeal -> "Cs"
      | Pte p -> sprintf "Pte%s" (pp_atom_pte p)
+     | Neon n -> sprintf "N%s" (pp_neon_size n)
 
    let pp_atom (a,m) = match a with
    | Plain o ->
@@ -162,6 +170,8 @@ let applies_atom (a,_) d = match a,d with
      if do_morello then fun f r -> f CapaSeal (f CapaTag r)
      else fun _f r -> r
 
+   let fold_neon f r = f N1 (f N2 (f N3 (f N4 r)))
+
    let fold_acc_opt o f r =
      let r = f (Acq o) r in
      let r = f (AcqPc o) r in
@@ -172,6 +182,7 @@ let applies_atom (a,_) d = match a,d with
      let r = if mixed then r else fold_pte (fun p r -> f (Pte p) r) r in
      let r = fold_morello f r in
      let r = fold_tag f r in
+     let r = fold_neon (fun n -> f (Neon n)) r in
      let r = fold_acc_opt None f r in
      let r =
        if do_morello then
@@ -192,7 +203,7 @@ let applies_atom (a,_) d = match a,d with
 
    let worth_final (a,_) = match a with
      | Atomic _ -> true
-     | Acq _|AcqPc _|Rel _|Plain _|Tag|CapaTag|CapaSeal|Pte _ -> false
+     | Acq _|AcqPc _|Rel _|Plain _|Tag|CapaTag|CapaSeal|Pte _|Neon _ -> false
 
 
 
@@ -255,7 +266,7 @@ let applies_atom (a,_) d = match a,d with
    | CapaTag,None -> Code.CapaTag
    | CapaSeal,None -> Code.CapaSeal
    | (Tag|CapaTag|CapaSeal|Pte _),Some _ -> assert false
-   | (Plain _|Acq _|AcqPc _|Rel _|Atomic (PP|PL|AP|AL)),_
+   | (Plain _|Acq _|AcqPc _|Rel _|Atomic (PP|PL|AP|AL)|Neon _),_
       -> Code.Ord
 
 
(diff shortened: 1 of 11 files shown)


maranget:
Hi, expect some delays i handling this PR as we are in the process of merging the master and KVM branch.

maranget:
@BreD1810 Hi,master is now stable. You may consider updating the PR with reasonable confidence.
I have noticed that some of the the tests do not allocate sufficient space for variables. Consider test V36 for instance:
```
AArch64 V36
(* Test Neon final state checking
   LD2 single structure *)
{
x = 22;
0:X0=x;
}
 P0                         ;
 MOVI V0.4S,#0              ; (* Note issues involving register default values *)
 MOVI V1.4S,#0              ;
 MOV W1,#1                  ;
 STR W1,[X0,#4]             ;
 LD2 {V0.S,V1.S}[0],[X0]    ;
forall(0:V0.S = {22,0,0,0} /\ 0:V1.S = {1,0,0,0})
```
Declaring `x` as `int x` allocates 4 bytes, while at least 8 are required (due to the store `STR W1,[X0,#4] ` above).  As a consequence some scaffold variables can be overwritten by the test itself, resulting in programm abort.

One possible fix is declaring x as `int64_t int` or, perhaps better as `int x[4] = {22,0,0,0}`.

BreD1810:
@maranget I've rebased on master and run a few tests, and I believe the behaviour is unchanged from when I first submitted this PR.
Good catch on the above comment. There are no new tests in this PR, but the Herd Neon PR should be updated to reflect this.

maranget:
> @maranget I've rebased on master and run a few tests, and I believe the behaviour is unchanged from when I first submitted this PR.
Ok, I'll look at this PR more thoroughly.

> Good catch on the above comment. There are no new tests in this PR, but the Herd Neon PR should be updated to reflect this.
Why not submit a new PR?

maranget:
As you noticed, there is a memory size problem for tests that use qualifiers N2 and above, here is an example:
```
% /opt/PR/bin/diyone7  -arch AArch64 -variant neon N2 PodWW N2 Rfe N1 PodRR N1 Fre  -name T
% cat T.litmus
AArch64 T
"PodWWN2N2 RfeN2N1 PodRRN1N1 FreN1N2"
Generator=diyone7 (version 7.56+02~dev)
Com=Rf Fr
Orig=PodWWN2N2 RfeN2N1 PodRRN1N1 FreN1N2
{
uint128_t y; uint128_t x;

0:X0=x; 0:X1=y;
1:X0=y; 1:X1=x;
}
 P0                      | P1               ;
 MOVI V0.4S,#1           | LD1 {V0.4S},[X0] ;
 MOVI V1.4S,#1           | LD1 {V1.4S},[X1] ;
 ST2 {V0.4S, V1.4S},[X0] |                  ;
 MOVI V2.4S,#1           |                  ;
 MOVI V3.4S,#1           |                  ;
 ST2 {V2.4S, V3.4S},[X1] |                  ;
exists (1:V0.4S=1 /\ 1:V1.4S=0)
```
If I understand correctly, The `ST2` instructions will write 2x4 int's to memory, which is twice the size of an `int128_t`. Hence, running the test is dangerous, and abort can be achieved easily (see attached log).

The fix is declaring x and y as `int [8]` arrays. The following `U.litmus` test illustrates the fix:
```
AArch64 U
{
int x[8]; int y[8];

0:X0=x; 0:X1=y;
1:X0=y; 1:X1=x;
}
 P0                      | P1               ;
 MOVI V0.4S,#1           | LD1 {V0.4S},[X0] ;
 MOVI V1.4S,#1           | LD1 {V1.4S},[X1] ;
 ST2 {V0.4S, V1.4S},[X0] |                  ;
 MOVI V2.4S,#1           |                  ;
 MOVI V3.4S,#1           |                  ;
 ST2 {V2.4S, V3.4S},[X1] |                  ;
exists (1:V0.4S=1 /\ 1:V1.4S=0)
```
Tu run this test, the very last version of `litmus` (with PR #170 merged-in) is required.

Test `U.litmus` has been written by modifying `T.litmus` by hand. However, it should be relatively simple. to automatically find the right type for variables by scanning cycles.

I have other, more complex, suggestions that relate to `V` register typing and printing. In herd in litmus, types also serves the purpose of selecting the right method to print register contents. If instead of writing `1:V0.4S=1` in condition above, one writes `1:V0.S={1,1,1,1}`, one forces the type of `V0` to be `int [4]`. Issuing such array values in final conditions would be a first step toward more informative test output.

Moreover, `diy` tools provide a worth noticing option `-cond observe`. In observe mode, there is no condition and teh final content of locations listed in the location field (as for instance `locations [1:V0.S; 1:V1.S;]` are printed. Hence her are two suggestions for more informative output:

- Would it be possible for `V` registers to have int [4] as default type, instead of the current `int` ?
- Even better one can think about declaring array types for registers (in init section) ? (Which may require significant changes)
[LOG.txt](https://github.com/herd/herdtools7/files/6002385/LOG.txt)

BreD1810:
It looks like the rebase did break something - previous final states would contain `int[4]`s (see [this branch](https://github.com/bred1810/herdtools7/tree/diy-neon-old)).

I did spend some time trying to change the size in the init section based on the annotations passed, but I couldn't find a way and instead settled for setting 1 size for all Neon tests. Now with the update you mentioned I should be able to change to the size to be big enough for an N4. However, do you have any suggestions to change the type based on the annotations passed? (e.g. using an `int[4]` for `N1`, `int[8]` for `N2`...)

I'm quite busy with semester 2 already, so it might be some time before I can have a thorough look into this and the other points you mentioned.

maranget:
> It looks like the rebase did break something - previous final states would contain `int[4]`s (see [this branch](https://github.com/bred1810/herdtools7/tree/diy-neon-old)).

I do not understand what you mean by final states contains `int[4]`'s. I'll have a look


> 
> I did spend some time trying to change the size in the init section based on the annotations passed, but I couldn't find a way and instead settled for setting 1 size for all Neon tests. Now with the update you mentioned I should be able to change the size to be big enough for an N4. However, do you have any suggestions to change the type based on the annotations passed? (e.g. using an `int[4]` for `N1`, `int[8]` for `N2`...)

No precise idea yet. Tests are build from cycles (_e.g._ argument to `diyone7`). The approprate information is there.

> 
> I'm quite busy with semester 2 already, so it might be some time before I can have a thorough look into this and the other points you mentioned.

No problem, meanwhile I am adding decent default types for vector registers in `litmus`.

BreD1810:
> I do not understand what you mean by final states contains `int[4]`'s. I'll have a look

Before the rebase, an N1 version of mp would have a final condition of:
```
exists (1:V0.4S={1,1,1,1} /\ 1:V1.4S={0,0,0,0})
```
Since rebasing the array is lost and instead the final condition is:
```
exists (1:V0.4S=1 /\ 1:V1.4S=0)
```

maranget:
Ok thanks, this should be relatively easy to restore.

maranget:
> > I do not understand what you mean by final states contains `int[4]`'s. I'll have a look
> 
> Before the rebase, an N1 version of mp would have a final condition of:
> 
> ```
> exists (1:V0.4S={1,1,1,1} /\ 1:V1.4S={0,0,0,0})
> ```
> 
> Since rebasing the array is lost and instead the final condition is:
> 
> ```
> exists (1:V0.4S=1 /\ 1:V1.4S=0)
> ```

FYI, PR#174 solves a different, related issue: with option `-cond observe` the final value of registers Vn.4S will be printed as 4 integer arrays. More precisely, when location [...; Vn.4S;...] is present in test (as generated with option `-cond observe`), litmus generates code that will print the final content of register `Vn.4S` as an array of four `int`'s.

BreD1810:
@maranget Apologies for the delay. I've now rebased again, and fixed the final condition issue (should now print arrays).

I've had another look at changing the sizes of the initial condition, but haven't had much luck. It seems that even explicitly setting the size to an array that would fit up to an `N4` test would require a lot of changes (see `gen/typBase.ml`). Let me know if you have any suggestions for this.

maranget:
Hi @BreD1810, You may be interested by PR #210 that solves the problem of providing locations of sufficient sizes for Neon accesses. I am still not satisfied with the resulting code. Known shortcomings.
1. Only the first register of wide loads (such as LD2, LD3, LD4) is observed, try:
```
% diyone7 -arch AArch64 -variant neon PodWW N2 Rfe N2 PodRR Fre
```
2. Final values of memory locations are not observed, try:
```
% diyone7 -arch AArch64 -variant neon PodWW N1 Coe N1 PodWW N1 Coe N1
```
Solving item 1. looks imperative before merging, item 2. is less urgent as using option `-obs local`  is a potential workaround.

If you have some time to spare... Have a look at how loads are compiled. Having functions such as as `emit_load` and `emit_access` to return a list of registers in place of a single register may be a (heavyweight) option. I guess something with less code modification can be invented, for instance by using the state component `st` to carry additional registers...

If you don't have time, just tell me.

BreD1810:
Hi @maranget, #210 looks good. Your first point was an inherent issue with our implementation, and we ran out of time in the project to try to find a solution. I tried to explore changing `emit_load` and `emit_access` as you suggested, but it seemed to be a large change and I didn't get very far.

As for your second point, the missing memory locations appears to be something else that broke in the rebase. [This diff](https://github.com/BreD1810/herdtools7/commit/7ad6f7ff2efcd96e3a6b3348c9508054edf11429#diff-5667887a15ae894a3bfb02b2ef1ca8c8607af2ea27b847ba7f1e26c3c9029457R972) provides a fix (note that the `do_ord_writes` function is now `do_get_writes` after the KVM update). At the time I considered this a bit of a hack, and I am not really sure of how this would affect `get_pte_writes`, for example.

Unfortunately university work pretty much consumes all of my time right now (and admittedly my knowledge of the codebase isn't as good as it used to be), so it may be best for you to continue to work on this.

maranget:
Hi @BreD1810, thanks for the additional patch. I perfectly understand that you are short of time now. My plan as follows : merge this PR in, and rebase #210  on top of the new master.  That way you get complete credit for your code.

Thanks for the good work, gen probably is the less structured part of herdtools...

PR #162, opened 2021-02-09
[git] Ignore .exe and .native files

eth-arm:
This PR adds `.exe` and `.native` to the `.gitignore` file, to ignore build outputs from Dune and OCamlbuild.

This is a minor quality-of-life improvement when copying built binaries between branches. For example, when copying a modified `herd_regression_test.exe` from one branch into another that doesn't have those modifications:
```sh
% cd ~/src/herdtools/regression-test-variants
% make
% cd ~/src/herdtools/new-feature
% cp ~/src/herdtools/regression-test-variants/_build/…/herd_regression_test.exe .
% ./herd_regression_test.exe …
```

This PR also sorts the contents of the `.gitignore` file.


maranget:
Thanks, merged.

PR #161, opened 2021-02-05
[herd] Add NEON support

AymenQ:
Add support for simulating NEON instructions with the herd tool.
The command-line argument `-variant neon` can be used to enable this support.

A suite of tests is included in `herd/tests/instructions/AArch64.neon/`.

Please note that this PR currently depends on #157 (for use of the `-variant` flag with `herd_regression_test`); the appropriate commits have been cherry-picked onto this branch to ease testing, but we will of course rebase once #157 is merged.

PR #163, opened 2021-02-10
[lib] Implement shift_right (arithmetic) for Uint.Uint128

eth-arm:
This PR implements `Uint.Uint128.shift_right`, the arithmetic right shift.

This PR does not block PR #161, but it does enable `-variant neon` and `-variant morello` to pass non-Neon / non-Morello regression tests that were using arithmetic right shift, e.g. [A63.litmus](https://github.com/herd/herdtools7/blob/3e395ed199f34354c3e3f4d8d66d951e0c6273b8/herd/tests/instructions/AArch64/A63.litmus).

### DIFF
--- a/lib/uint.ml
+++ b/lib/uint.ml
@@ -216,8 +216,6 @@ module Uint128 = struct
     else
       Uint64.shift_left (snd a) (by-64), Uint64.zero
 
-  let shift_right _a _by = failwith "not implemented: shift_right"
-
   let shift_right_logical a by =
     if by < 0 then
       invalid_arg (Printf.sprintf "shift_right by negative: %i" by)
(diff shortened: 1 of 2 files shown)


lib/uint.ml
```diff
@@ -235,6 +233,15 @@ module Uint128 = struct
...
       Uint64.zero, Uint64.shift_right_logical (fst a) (by-64)
 
+  let shift_right a by =
+    let upper_bit = shift_left one (num_bits - 1) in
+    let has_msb = logand a upper_bit = upper_bit in
+    let rest = shift_right_logical a by in
+    if has_msb then
+      logor rest upper_bit
+    else
+      rest
```
maranget, line 243: Hi, if I remember correctly, arithmetic shift right replicates the sign bit. It looks as if the above code copies this most significant bit into the result without replication.

eth-arm: Oops, yes it does; good catch. Done.

maranget:
Merged, thanks @eth-arm

PR #169, opened 2021-02-15
[internal_lib] Detect Python version in Shelf module

eth-arm:
This PR changes the `Shelf` parser to automatically detect available Python versions.

It also enriches the `Command.Error` exception with an `error` record type.

### DIFF
--- a/internal/lib/shelf.ml
+++ b/internal/lib/shelf.ml
@@ -81,7 +81,11 @@ let list_of_file path key =
   in
   let lines = ref [] in
   let read_lines c = lines := Channel.read_lines c in
-  Command.run ~stdin:script ~stdout:read_lines "python" [] ;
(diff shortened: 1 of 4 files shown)


internal/lib/shelf.ml
```diff
@@ -56,6 +56,21 @@ let to_ocaml_string shelf = OcamlString.record [
...
+      try
+        Command.run ~stdout:dev_null ~stderr:dev_null p ["--version"] ;
+        true
+      with
+        Command.Error _ -> false
+    in
+    match List.find_opt exists ["python"; "python3"] with
+    | Some p -> p
+    | None -> failwith "Could not find either python or python3"
+  end
```
maranget, line 72: Nice usage of Lazy.

PR #174, opened 2021-02-18
[litmus] Type neon registers

maranget:
Varying default types for registers. For instance (AArch64), having `locations [0:V0.4S;]` will command printing the final value of vector register V0 of thread 0 as a array of four `int`s.


maranget:
May suffice for `litmus` to yield adequate output for Neon tests produced by generators with option `-cond observe`. Neon generators are introduced by PR #160.

PR #176, opened 2021-02-19
[herd] Add an AArch64 test for ASR with a negative initial value

eth-arm:
This PR adds a herd Instructions Test for `ASR` (Arithmetic Shift Right) on AArch64 to confirm that it performs _arithmetic_ and not _logical_ shifts for negative values.

### DIFF
--- a/herd/tests/instructions/AArch64/A96.litmus
+++ b/herd/tests/instructions/AArch64/A96.litmus
@@ -1,0 +1,9 @@
+AArch64 A96
+(* Tests asr (immediate), negative initial value, right shift by 2 *)
+
+{ int64_t 0:X0 = -256; }
+
+P0;
+ASR X0, X0, #2;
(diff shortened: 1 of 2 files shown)


eth-arm:
A related question: Is there a naming scheme for these tests? Is `A*` vs `L*` important?

I gave this test an arbitrary unused number, but it's not going to be sequential with the other `ASR` tests without renaming many other tests.

maranget:
> A related question: Is there a naming scheme for these tests? Is `A*` vs `L*` important?
> 
> I gave this test an arbitrary unused number, but it's not going to be sequential with the other `ASR` tests without renaming many other tests.

There  is no naming scheme, except perhaps using increasing numbers in fixed size format, which results in numerical and lexical ordering being the same. Using `A*` or `L*` is not important.

Issue #154, opened 2021-01-19
Herdtools does not build on 32-bit systems


eth-arm:
Trying to build herdtools7 on a Raspberry Pi running Raspbian Buster 32-bit fails with:

```sh
% make
sh ./dune-build.sh $HOME
+ ./version-gen.sh /home/eth
+ dune build --profile release @all
File "lib/symbConstant.ml", line 37, characters 45-56:
Error: Integer literal exceeds the range of representable integers of type int
File "lib/symbValue.ml", line 115, characters 21-32:
Error: Integer literal exceeds the range of representable integers of type int
make: *** [Makefile:32: build] Error 1
```

This seems to be because the [size of `int` is dependent on machine architecture](https://ocaml.org/releases/4.05/htmlman/libref/Pervasives.html#6_Integerarithmetic), and the [bit-manipulation in `symbConstant.ml`](https://github.com/herd/herdtools7/blob/bd4dba32189b7f3689614aab2005adbb2877b891/lib/symbConstant.ml#L37) assumes a 64-bit architecture.

maranget:
You are correct, if port to 32 bits is considered important, one may consider implementing the 'cap' field as [int64](https://caml.inria.fr/pub/docs/manual-ocaml/libref/Int64.html). However, I am wondering what is the minimal size required for this field?

eth-arm:
After asking @ambroise-arm, the Capability is [bits 128 to 95 of a 129-bit register](https://github.com/herd/herdtools7/blob/5fbe674560e0eadc8eeab9e1aa111cd69cd9e005/lib/constant.mli#L24), which does not fit into an `Int32`, so I think it should be `Int64`.

maranget:
Ok thanks, good to know.

maranget:
The issue is hopefully closed by PR #179, which implements Morello capacities as OCaml `int64` type.

PR #179, opened 2021-02-24
[all] Implement morello capacities with int64.

maranget:
The change should allow compilation on 32bits platforms,
solving issue #154.

Notice that Uint.Uint64.t may be a more natural choice.
However, this would hinder using the syntactical features
of int64.


maranget:
The change should allow compilation on 32bits platforms,
solving issue #154.

Notice that Uint.Uint64.t may be a more natural choice.
However, this would hinder using the syntactical features
of int64.

PR #182, opened 2021-03-02
[herd] KVM, no branch for direct access

maranget:
Do not emit branching events in the direct access to memory case.

When memory accesses involve page table walk, a branching event is inserted between page table read and
(physical) memory access(es). For direct accesses, such as page table access without variant `pte2`, those
branching events are superfluous.


jalglave:
Thanks Luc.
Jade

PR #183, opened 2021-03-03
[herd] C, extend the definition of control dependency.

maranget:
Control dependency is now defined as follows:
  + Given iico_data, rf from register to register (rf-reg)
    and control from read event in condition  to any event in alternative
    of an "if" construct.
  + Ctrl starts with iico from memory location, followed by
    a possibly empty sequence of iico_data, rf-reg and control, and ends in
    a control relation.

With respect to previous definition, this amounts to add "control" steps
in the middle chain of elementary relations.


jalglave:
Thanks Luc. I've added links to the LKML thread for context.
Jade

PR #184, opened 2021-03-03
[herd] Discard executions with wrong rf matching.

maranget:
Some executions fail during the solver phase because a pair
of memory write and read are not matched properly. For instance
one may match a pte write with a 32bits read. Such execution
is better discared than trigerring herd failure.

This is attempted by introducing a specific exception for
operation failure and catching it around calls to solver
in its "memory" phase. For now only masking operation failure
are handled that way.


jalglave:
Oh great! Thanks very much for this.
Jade

PR #186, opened 2021-03-04
[herd] Wrong rf matching continued

maranget:
This is a followup to PR #184. As desirable behaviour is unclear, made draft.


maranget:
One difficulty is as follows: a failing mask may mask a contradiction in equations. In such case the candidate can be safely discarded. But in other occasions the failing mask is a programming error that is better signaled by crashing herd. I still cannot make the distinction.

maranget:
PR closed, superseded by PR #188.

PR #192, opened 2021-03-10
[herd] Check execution level when accessing PTE.

maranget:
And fault when accessing PTE from EL0.
Notice that PTE flags of PTE addresses
cannot be changed at the moment, hence
`-variant precise` is mandatory for meaningful
experiments


jalglave:
Thanks Luc.
Jade

PR #195, opened 2021-03-16
[herd] Mixed align

maranget:
This PR fixes an incompatibility between arrays and `-variant mixed`. Alignment check was wrong in case when array element size is smaller than access size.


lib/machSize.mli
```diff
@@ -32,8 +32,11 @@ val get_off : sz -> sz -> int list
 (* All valid offsets for sz2 in sz1, reduced list *)
 val get_off_reduced : sz -> sz -> int list
 
-(* Smaller of two *)
 val compare : sz -> sz -> int
+
+(* Smaller of two *)
+val smaller_equal : sz -> sz -> bool
```
eth-arm, line 38: Can this be named `smaller_or_equal` or `less_than_or_equal`?

herd/tests/instructions/AArch64.mixed/M000.litmus
```diff
@@ -0,0 +1,14 @@
+AArch64 M000
+(* Does not work yet without option -debug mixed... *)
```
eth-arm, line 2: Should this work in non-mixed mode eventually, or is it fundamentally a mixed-mode test?

eth-arm, line 2: Do you mean:
```suggestion
(* Does not work yet without option -variant mixed... *)
```

maranget: This is fundamentally a mixed-size test and it would be a good thing to have tests specific to mixed-size. Unfortunately, this test is not ready yet. There is another, unrelated, shortcoming.

maranget: I do mean `-debug mixed`. Undocummented `-debug mixed`  selects a less efficient but less error-prone technique to enumerate co and rf relations.

lib/machSize.mli
maranget: Good catch, thanks.

herd/memUtils.ml
```diff
@@ -540,10 +540,15 @@ let lift_proc_info i evts =
...
+        if MachSize.less_than_or_equal sz_e sz_s then begin
+          let ncell = idx / nbytes_s and idx0 = idx mod nbytes_s in
+          0 <= ncell && ncell < array_sz
+          &&  List.exists (Misc.int_eq idx0) (MachSize.get_off sz_s sz_e)
+        end else begin
+            idx >= 0 &&
+            (let nbytes_e = MachSize.nbytes sz_e in
+             idx mod nbytes_e = 0 &&
+             (let idx_max = idx/nbytes_s + (nbytes_e/nbytes_s) in
+             idx_max <= array_sz))
```
eth-arm, line 552: The second check confirms that the index is aligned with the size of the element, and the third check confirms that the whole access will not go beyond the bounds of the whole array.

Have I read that correctly?

maranget: You are correct. Is there something too cryptic here?

eth-arm: I was checking that I had understood correctly. Looks good to me! :)

PR #198, opened 2021-03-19
[catalogue] Remove reference to removed file demo/cats/anarchy.cat

eth-arm:
This PR removes a reference to a file `catalogue/demo/cats/anarchy.cat`, which was removed in commit 14b60172beeb54d2690dea69b1bb1b9051e3c064.
This dangling reference was causing the `lint-shelves` pre-commit check to fail; this PR means that all pre-commit checks now pass.

### DIFF
--- a/catalogue/demo/shelf.py
+++ b/catalogue/demo/shelf.py
@@ -7,7 +7,6 @@ bells = [
 ]
 
 cats = [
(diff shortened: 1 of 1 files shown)


maranget:
Fine, thanks.

PR #200, opened 2021-03-20
[gen] Fix problem in init section of generated tests.

maranget:
When type is explicit (option `-type <t>`), some locations were declared more than once, which herd and litmus do not accept. The fix is not adding type to variables that are initialised explicitly.


maranget:
Naive solution does not work completely. Duplications are avoided by default types for constant is not selected.

maranget:
Closing this PR after saving useful part. Rebasing was wrong.

PR #206, opened 2021-03-29
[build] defs.sh no-longer parses duplicated dune files in _build

eth-arm:
This PR fixes a minor annoyance where running `% make test D=dune` then `% make test D=ocb` will run each OCaml test twice in OCamlbuild mode.


maranget:
Simple and effective, thanks.

PR #209, opened 2021-03-30
[herd-www] fix undefined variables in Makefile

eth-arm:
This PR fixes `make clean` for `herd-www/`:

- `JSON_SHELVES` and `JSON_SPINES` were used inconsistently, so they are
  unified to `JSON_SHELVES`, because "catalogue shelves" is the
  terminology used elsewhere in the repo.
- `WWW_LIB` and `WEB_LIB` were used inconsistently, so they were unified
  to `WWW_LIB` to match `WWW_CATALOGUE` and `WWW_PROG`.
- `OCBOPT` was used but not defined, so is now removed.


maranget:
Fine, merged, thanks.

PR #211, opened 2021-03-30
[tools] Add a flag to mdiff7 to change diff behavior, -act

eth-arm:
This PR re-enables the log intersection behavior of `mdiff7`. This used to be done with the binary `minter7`, which was removed in commit e9362dd8c2262412bd879971cc2a69c4257ccdd6.

### DIFF
--- a/tools/matrix.mli
+++ b/tools/matrix.mli
@@ -33,13 +33,6 @@ module type Config = sig
 end
 
 
-(* Goes nowhere else *)
-(*
-val one_line : string -> cell
-val one_lines : string list -> column
-val one_liness : string list list -> matrix
-*)
-
 (*****************)
 (* Simple Build  *)
 (*****************)
--- a/tools/mdiff.ml
+++ b/tools/mdiff.ml
@@ -18,6 +18,8 @@
 open Printf
 open LogState
 
+type act = Diff | Inter
+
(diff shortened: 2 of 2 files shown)


maranget:
Hi @eth-arm. I do not remember to have deleted `minter7`. It was in fact an alternative name for `mdiff7`, as illustrated by the code candidate for deletion in commit 32c56c5831d5242e6fd1c4fe872e3fdddd5c512f.

As far as I remember, `mdiff7` was first installed and then a link with name `minter7` was created that pointe do `mdiff7`. I do not remember when `minter7` disappeared, nor why. However I see no real reason to get rid of it definitively. It is perfectly possible that the installation procedure was tricky to implement on top of opam, who knows?

Perhaps a reasonable solution is to add a command line option to `mdiff7` that would perform log intersection, in place of the nasty trick of checking command name (a.k.a. Sys.argv.(0)).

eth-arm:
Done. `mdiff7` now has a flag to configure behavior:

```sh
% mdiff7                log1 log2  # Diff.
% mdiff7 -act diff      log1 log2  # Also diff.
% mdiff7 -act inter     log1 log2  # Intersect.
% mdiff7 -act intersect log1 log2  # Also intersect.
```

PR #217, opened 2021-04-12
[build] `make check-deps` no-longer requires opam

eth-arm:
The previous check-deps implementation was created to check that
`stdint` was installed, and assumed the existance of `opam`.

However, we no-longer require `stdint`, and `opam` is not required to be
installed to build herdtools, so now `make check-deps` just checks for
the presence of `ocaml`, `menhir`, and either `dune` or `ocamlbuild` &
`ocamlfind`.

Tested by applying this patch on a machine that has all the required
dependencies, but NOT `opam`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,5 @@
+.PHONY: check-deps
+
 PREFIX=$$HOME
 D=dune
 #For building with ocamlbuild set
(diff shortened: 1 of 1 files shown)


maranget:
Thanks @eth-arm. I merge. Notice that the documentation recommends opam as the easiest way to install dune, menhir, etc.

PR #218, opened 2021-04-12
[herd] Fix iico-data dependancy in CAS equal case

HadrienRenaud:
In CAS, the `rs` register provides the compared value and receives the value read from memory.
In the equal case, where the value read from memory and the one in the `rs` register are equal, the `iico-data` dependency that provided data to the write in the `rs` register was coming fromthe `rs` register, instead of coming from the value read in memory.

This probably has no impact because the two values are equal, and there is `iico-ctrl` dependencies from both events.

For a simple `CAS` instruction, before we can see the `iico-data` arrow between the events `e` and `g`:

CAS

After, the arrow is between the events `a` and `g` (the events names are the same, only the layout changes):

CAS-modified

### DIFF
--- a/herd/event.ml
+++ b/herd/event.ml
@@ -1821,7 +1821,7 @@ module Make  (C:Config) (AI:Arch_herd.S) (Act:Action.S with module A = AI) :
              wm.intra_causality_data)
           (EventRel.union4
              (EventRel.cartesian (get_output rn) input_rm)    (* D1 *)
-             (EventRel.cartesian output_rs input_wrs)         (* Ds1 *)
+             (EventRel.cartesian output_rm input_wrs)         (* Ds1 *)
              (EventRel.cartesian (get_output rn) input_wm)    (* Ds2 *)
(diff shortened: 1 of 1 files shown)


maranget:
Hi @HadrienRenaud, nice to meet your. I am a bit puzzled. The read register Xs to write register Xs data dependency originates from an interaction with @jalglave who showed me a document that clearly described the dependencies of successful CAS.

For the involved dependency, whose name is Ds1 (as in source), the document reads:
   - Ds1: there is an Intrinsic data dependency from the Read register effect
      of Xs to the Write register effect of Xs.

Has something changed since, or was the document used at that time not the right one? If so, I'll gladly merge the PR.

HadrienRenaud:
Hi @maranget,

You are right, although this behavior has surprised me, it is described as this in the spec.
Nothing has changed since the previous PR.

I'm closing this PR, sorry for the trouble.

maranget:
You are welcome. In fact your are right to find this difference of dependencies between failing and successful CAS. This difference had to be questionned.

PR #226, opened 2021-04-26
[herd] Setaf out of the box

maranget:
Draw test-and-set events for the AF bit of page table entries in a special way: out of thread and instructions boxes.


maranget:
Superseded by PR #234.

PR #235, opened 2021-05-04
[herd] Add support for assertions in cat

relokin:
Typically rules specified in cat files limit allowed executions of a
litmus test. This patch adds support for also turning the same rules
into assertions that exit the simulation with an error message.

### DIFF
--- a/lib/AST.mli
+++ b/lib/AST.mli
@@ -79,7 +79,7 @@ and binding = TxtLoc.t * pat * exp
 
 type do_test = Acyclic | Irreflexive | TestEmpty
 type test = Yes of do_test | No of do_test
-type test_type = Flagged | UndefinedUnless | Check
+type test_type = Flagged | UndefinedUnless | Check | Assert
 type app_test = TxtLoc.t * pos * test * exp * string option
 type is_rec = IsRec | IsNotRec
 
--- a/lib/interpreter.ml
+++ b/lib/interpreter.ml
@@ -168,7 +168,7 @@ module Make
 
     let check_through test_type ok = match test_type with
     | Check ->  U.check_through ok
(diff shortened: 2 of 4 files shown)


relokin:
The above patch would allow me to assert on tests like the one below where the cacheability attributes for `x` are not set correctly (`x` can be outer write-through and outer write-back at the same time):
```
AArch64 MP+dmb.sys-Nsh-WB
{
pte_x=(attrs:(Normal,Inner-write-back,Inner-write-through,Outer-write-back,Non-shareable));
pte_y=(attrs:(Normal,Inner-write-back,Outer-write-back,Non-shareable));
0:X1=x; 0:X3=y;
1:X1=y; 1:X3=x;
}
 P0          | P1          ;
 MOV W0,#1   | LDR W0,[X1] ;
 STR W0,[X1] | DMB SY      ;
 DMB SY      | LDR W2,[X3] ;
 MOV W2,#1   |             ;
 STR W2,[X3] |             ;
scopes: (Ish P0 P1)
exists
(1:X0=1 /\ 1:X2=0)```

maranget:
Hi @relokin, the feature could have been implemented on top of the existing "flag" mechanism, bay adding yet another command line option, similar to `-badflag`. However, your solution more direct and less error-prone. Merged.

Thanks.

relokin:
Thanks for the review Luc, I wasn't aware of `-badflag`. I can have a look and if you think that'd useful we can merge the two concepts.

maranget:
A first idea would be a `-assertflag <flag>`. However this is error-prone as <flag> must be the name of the flag one intend turn into an assert. It is easy to misspell the flag name. Your solution is simpler, let us keep it. Furthemore, there are so many command line option...

PR #243, opened 2021-05-28
inline variant flag propagated when specified

lukeg101:
In commit f78fb880 support was added for specifying variants inline in
litmus tests with the syntax `Variant=foo`.

This commit ensures that this variant information is propagated when
parsing in Herd, so we can use this information during simulation.
In particular that the inline variant information is parsed and propagated
to `herd/mem` where mixed checks are made.

A mixed-mode unit test is provided which has `variant=mixed` specified
so we know that this feature works when the user specifies this flag.

Kindly reviewed by @relokin 

This patch works but we were wondering if there was something cleaner we could do, in particular `check_sizes` in `herd/mem` fails the mixed check as it uses `Conf.variant`(hence this patch must set `Conf.variant` when specified inline) while the `S.A.is_mixed` flag is true in `herd/parseTest`. So there is two sources of information, but perhaps there should only be one?

What do you think? have we missed something obvious?


lukeg101:
I agree with you, `S.A.is_mixed` should reflect the variant spec. My confusion was todo with how the sources of information are propagated, which made it difficult to debug as `S.A.is_mixed` is true, but `Mem` cannot see that.

I had a look at the dependencies and found `Mem.Make` (where the variant Information is needed) is passed into `top_herd.Make` which itself is called from `parseTest` when running `top_herd.run`. Currently `TestVariant.Make` is called in `parseTest.do_from_file` but that information doesn't get forwarded to `Mem`. Perhaps `Mem` is made before `TestVariant.Make` is run?

This patch fixes this by re-running `TestVariant` before `top_herd.Make` is run, so we have up to date variant Information in `Mem` but I feel as though there might be a better single place to setup `Mem` and `S.A.is_mixed` etc... so all information is consistent rather than running `TestVariant` twice. I'm sorry if I'm overthinking it.

maranget:
There is something strange here: I do not understand why functor `TestVariant.Make` is applied twice in `parseTest.ml`.
I'd rather have module `Make` to take an extra `Conf` argument, parameter being the `Conf` module computed at the beginning of the `do_from_file` function.

lukeg101:
Fair, I will figure out how to rework my patch to do that. Thank you for your help :)

lukeg101:
Hi folks,
I did something much simpler to fix this, I moved the `TestConf.Make` construct outside of the `if Conf.check_name` condition. This appears to be a harmless change as we are not doing anything related to `check_name`, and we do not need to make `TestConf` twice

maranget:
> Hi folks,
> I did something much simpler to fix this, I moved the original `TestConf.Make` construct outside of the `if Conf.check_name` condition. This appears to be a harmless change as we are not doing anything related to `check_name`, and we do not need to make `TestConf` twice, but we get the behaviour we expect

Hum that's clever. I still find the code error prone (it was already before your fix). May I ask you a last change that would make the situation a bit better: change the name of the Conf argument of the Top functor into say TopConf. That way it will be clearer where the configuration parameters are coming from. We cannot exclude that in the future one may refer in this file  to the `.variant`  parameter. Then, the different names would suggest whether  this .variant is before (`TopConf.variant`) or after (`Conf.variant`)  having  handled the test inline variant.

maranget:
Merged, thanks.

lukeg101:
Thanks Luc, and sorry for my delayed reply - I was on holiday for the last 2 weeks :)

maranget:
Hi Luke,

Never mind. I fancied surpassing as much PR as possible before I go for holiday myself :)

PR #247, opened 2021-06-08
[herd] One read for all

maranget:
This PR implements single read page table walk. AF and DB write are performed as write of some RMW of which read is the common page table walk read. Notice that the other purpose of this table walk read is to retrieve physical address and permissions.


jalglave:
Reviewed locally. Thanks Luc.

PR #248, opened 2021-06-09
[herd] PoD set  restricted to branch conditional events

maranget:
This changes the definition of the ctrl relation in `aarch64dep.cat`.
Namely, internal "commit" events should not impact on the ctrl
dependency relation between venets generated by different instructions.


jalglave:
Thanks Luc.

PR #250, opened 2021-06-11
Litmus minor clean-ups

relokin:
Hi Luc,

These are some minor clean-ups. I wanted to remove unnecessary functions, use more of the kvm-unit-tests API and avoid some #ifdefs. Feel free reject any of them as they don't fix any bug in particular.

As a more general note, what do you think about moving some of the C code to .c file rather than generating from ocaml? We can also create different versions of some functions to avoid #ifdefs and/or avoid their dynamic generation (for example `postlude` and `postlude_kvm`).

Nikos

### DIFF
--- a/litmus/libdir/_hash.c
+++ b/litmus/libdir/_hash.c
@@ -151,11 +151,7 @@ static void hash_add(hash_t *t,log_t *key, param_t *v,count_t c,int ok) {
     h++ ;
     h %= HASHSZ ;
   }
-#ifdef KVM
-  printf("Hash table is full\n") ;
-#else
   fprintf(stderr,"Hash table is full\n") ;
-#endif
   exit(2) ;
 }
 
--- a/litmus/libdir/_instance.c
+++ b/litmus/libdir/_instance.c
@@ -104,6 +104,9 @@ static global_t global  =
 #ifdef ACTIVE
     active,
 #endif
+    0,
+    SIZE_OF_TEST, NUMBER_OF_RUN, NEXE, NOCCS,
+    0,
   };
 
 static void init_global(global_t *g,int id) {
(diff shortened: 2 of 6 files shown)


maranget:
> Hi Luc,

Hi @relokin 
> These are some minor clean-ups. I wanted to remove unnecessary functions, use more of the kvm-unit-tests API and avoid some #ifdefs. Feel free reject any of them as they don't fix any bug in particular.
> 
Such cleanup is always welcome. I have checked functionality and saw not problem. So I merge. Thanks.

> As a more general note, what do you think about moving some of the C code to .c file rather than generating from ocaml? We can also create different versions of some functions to avoid #ifdefs and/or avoid their dynamic generation (for example `postlude` and `postlude_kvm`).
> 

I have no strong opinion on this. Let us be pragmatic! As for `postlude`, if I understand correctly what you mean, some dynamic generation is mandatory as the postlude includes some parts that depend upon each test. You want to abstract this, passing test name and condition and more maybe as arguments ?

relokin:
> > Hi Luc,
> 
> Hi @relokin
> 
> > These are some minor clean-ups. I wanted to remove unnecessary functions, use more of the kvm-unit-tests API and avoid some #ifdefs. Feel free reject any of them as they don't fix any bug in particular.
> 
> Such cleanup is always welcome. I have checked functionality and saw not problem. So I merge. Thanks.
> 

Thanks!

> > As a more general note, what do you think about moving some of the C code to .c file rather than generating from ocaml? We can also create different versions of some functions to avoid #ifdefs and/or avoid their dynamic generation (for example `postlude` and `postlude_kvm`).
> 
> I have no strong opinion on this. Let us be pragmatic! As for `postlude`, if I understand correctly what you mean, some dynamic generation is mandatory as the postlude includes some parts that depend upon each test. You want to abstract this, passing test name and condition and more maybe as arguments ?

Absolutely! We will still need some dynamic generation, some will be in arguments or defines but not all functions can be static.

PR #253, opened 2021-06-16
[herd] Lift memop for DC by VA instructions

relokin:
This change adds support for translation and other memory features for
DC instructions by VA. DC instructions by set and way bypass this.

### DIFF
--- a/herd/AArch64Sem.ml
+++ b/herd/AArch64Sem.ml
@@ -236,11 +236,6 @@ module Make
         let oloc = if A.TLBI.inv_all op then None else Some loc in
         M.mk_singleton_es (Act.Inv (op,oloc)) ii
 
-(* Data cache operations *)
-      let dc_loc op loc ii =
-        let oloc = if AArch64Base.DC.sw op then None else Some loc in
-        M.mk_singleton_es (Act.DC (op,oloc)) ii
-
 (* Neon size *)
       let neon_esize r = match r with
       | AArch64Base.Vreg (_,(_,esize)) -> esize
@@ -530,7 +525,7 @@ module Make
           (fun _ -> mk_fault a ii (Some "EL0")) ii >>! B.Exit
 
 
-      let check_ptw proc dir a_virt ma _an ii mdirect mok mfault =
+      let check_ptw proc dir updatedb a_virt ma _an ii mdirect mok mfault =
 
         let is_el0  = List.exists (Proc.equal proc) TopConf.procs_user in
         let check_el0 m =
@@ -571,7 +566,7 @@ module Make
             let setbits =
               match dir with
               | Dir.W ->
-                 if hd then
+                 if hd && updatedb then
                    add_setbits
                      (m_op Op.And (is_zero ipte.af_v) (is_zero ipte.db_v))
                      "af:0 && db:0"
@@ -763,9 +758,6 @@ module Make
 (* Page tables and TLBs *)
       let do_inv op a ii = inv_loc op (A.Location_global a) ii
 
-(* Data cache operations *)
-      let do_dc op a ii = dc_loc op (A.Location_global a) ii
-
 (***************************)
 (* Various lift functions. *)
 (***************************)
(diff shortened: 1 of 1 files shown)


herd/AArch64Sem.ml
```diff
@@ -1333,6 +1325,28 @@ module Make
...
+      let dc_loc op a ii =
+        let mk_act loc = Act.DC (op,Some loc) in
+        let loc = A.Location_global a in
+        M.mk_singleton_es (mk_act loc) ii
+
+      let do_dc op rd ii =
+        if AArch64Base.DC.sw op then
+          M.mk_singleton_es (Act.DC (op, None)) ii >>! B.Next
+        else begin
+            (* TODO: The size for DC should be a cache line *)
```
maranget, line 1338: Perhaps we can rely on the assumption: different variables are allocated to different cache lines? If you need t allocated variables to the same cache line, we can use arrays and then decide whether two cells are in the same cache line or not, assuming  alignment of arrays (_i.e._ alignment according to total array size).

relokin: I think so. I am not very familiar with the specifics of mixed mode. 

In non-mixed mode: All accesses to a memory location have to be of a specific size, don't they (e.g, byte, half word, ...)? In the mixed mode then this assumption goes away. If that's the case, we can make the same assumption for DC instructions; in non-mixed mode, when we clean location `x` we use the size of all the previous or next memory instruction that accesses `x`.

In mixed mode: We have to be able to handle the case where `x` and `y` (or x[0] and x[1]) map into the same cache line, I think we have most of the mechanics in place, maybe we also need to be able to reason about allocation alignment.

maranget: I do not know yet. Anyway I'll merge this PR.

herd/AArch64Sem.ml
```diff
@@ -1333,6 +1325,28 @@ module Make
...
+            let mop _ac a = dc_loc op a ii in
+            lift_memop Dir.R
+              (fun ac ma _mv -> (* value fake here *)
+                if Access.is_physical ac then
+                  M.bind_ctrldata ma (mop ac)
```
maranget, line 1343: I wanted to merge this PR. But now I am not so sure about the control dependency. Can you please comment?

relokin: This is very similar to what we do for regular load/store instructions https://github.com/herd/herdtools7/blob/cc5ae26620ff3191c215b10c932f695d3de1c7c8/herd/AArch64Sem.ml#L890. If I understand correctly, this has to do with the branching event we add between the `Rpte` event and the corresponding memory read/write event. DC instructions similarly to loads and stores may raise a fault if the PTE is invalid.

maranget: I see that DC is assimilated to a read effect. This has some importance for the hardware update of the access and dirty bits. I'll explore this by some experiment before merging. Notice that documentation states that `DC CVAU` checks for read permission. Hence the `Dir.R` argument to `liftmem_op` looks appropriate.

relokin: The clean or clean and invalidate DC variants (DC CVAU, DC CVAC, DC CIVAC) all require read permissions. There is one exception which is the invalidate DC IVAC instruction which requires write permissions. I need to check though if it sets the dirty bit.

maranget: I have just checked on hardware that `DC CVAU` behaves as a load as regards hardware flag management...  I can check `DC IVAC`.

maranget: As far as I have tested, `DC IVAC` require write permission. However if hardware management of the dirty bit is enabled, it no longer fails  but does *not* update the dirty bit...

relokin: I think this version addresses the problem with DC IVAC, by raising a fault if it doesn't have write permissions but avoiding to update the DB if it's not set.

maranget: You think right...

maranget:
Merged, thanks.

relokin:
Thanks Luc!

PR #256, opened 2021-06-29
[lib]: Added Void C Type

lukeg101:
Kindly reviewed by @relokin 

Consider the C code fragment:
```
void *P0 (...) {
    ....
}
```
which is the (pthread friendly) format of concurrent C functions. We may
wish to generate this using diy, and so we would need to know to print a
void pointer type. This patch adds the `void` type in lib, which we can
later combine to generate such a code fragment.

### DIFF
--- a/herd/machAction.ml
+++ b/herd/machAction.ml
@@ -346,8 +346,8 @@ end = struct
     | Some (A.V.Val (PteVal v)) -> V.is_zero (V.intToV v.PTEVal.db)
     | _ -> false
 
-  let read_pteaf0 act = af0_val (read_of act)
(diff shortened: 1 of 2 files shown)


lukeg101:
Also I tried a new way of doing review here but please let me know if it mucks up the linear commit history.

I prepared the patch on the `const` branch of my fork, then submitted a PR from `const` -> `const-review:fork` on my fork. Nikos then reviewed it on my fork and I merged the PR. Following this, I submitted a PR `const-review:fork` -> `master:this_repo` to send the code here. 

If this doesn't work I can just submit a PR straight from `const` branch of my fork

maranget:
Merged, thanks.

PR #257, opened 2021-06-29
[lib]: Added parsing support for the const type

lukeg101:
Kindly reviewed by @relokin [here](https://github.com/lukeg101/herdtools7/pull/1)

Following our discussion I was unsure about whether to change `strip_volatile/atomic`. I was checking for the case where the user provides the functions malformed types like `Volatile (Volatile t)` which doesn't arise in practice, but would be good to handle against this case.

Consider the following fragment of a litmus C test:

P0 (const int* y,volatile int* x) {
  int r0 = *x;
  *y = 1;
}
which uses const qualified types. This patch adds parsing support for
const qualified types. Like volatile we do not implement semantics
for const in Herd.

Additionally we add testing infrastructure to automatically test litmus
C tests against the rc11.cat memory model, like we do with AArch64.
As such, unittests are provided for const and volatile qualified types in herd

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -110,6 +110,16 @@ test::
 
 test::
 	@ echo
+	$(HERD_REGRESSION_TEST) \
+		-herd-path $(HERD) \
+		-libdir-path ./herd/libdir \
+		-litmus-dir ./herd/tests/instructions/C \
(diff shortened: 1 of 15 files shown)


maranget:
Merged, thanks.

Additional regression tests for RC11 would be quite useful.

PR #261, opened 2021-07-30
Extend the syntax of faults to include a proposition

relokin:
Currently a fault is specified as

fault(Pn,Ln,Loc)

This patch extends the notation to include an optional proposition:

fault(Pn,Ln,Loc,Prop)

This proposition will be checked from the context of the fault handler if and when it is invoked. This proposition is not as general and can only include conditions on the values of memory variables (no registers, or faults). In the current version, we're not restoring the values the pte values and the locations in the fault's proposition have to be readable.

### DIFF
--- a/herd/AArch64Sem.ml
+++ b/herd/AArch64Sem.ml
@@ -102,9 +102,6 @@ module Make
 
       let mk_read_std = mk_read quad AArch64.N
 
-      let mk_fault a ii msg =
-        M.mk_singleton_es (Act.Fault (ii,A.Location_global a,msg)) ii
-
       let read_loc v is_data = M.read_loc is_data (mk_read v AArch64.N aexp)
 
       let read_reg is_data r ii = match r with
@@ -520,9 +517,6 @@ module Make
 
          *)
 
-      let mk_pte_fault a ma ii =
-        insert_commit_to_fault ma
-          (fun _ -> mk_fault a ii (Some "EL0")) ii >>! B.Exit
 
       let an_xpte =
         let open AArch64 in
@@ -543,7 +537,7 @@ module Make
         | NoRet|T|S -> N
 
 
-      let check_ptw proc dir updatedb a_virt ma an ii mdirect mok mfault =
+      let check_ptw proc dir updatedb a_virt ma an ii mdirect mok mfault mkfault =
 
         let is_el0  = List.exists (Proc.equal proc) TopConf.procs_user in
         let check_el0 m =
@@ -669,7 +663,9 @@ module Make
           M.choiceT cond mvirt
             (* Non-virtual accesses are disallowed from EL0.
                For instance, user code cannot access the page table. *)
-            (if is_el0 then mk_pte_fault a_virt ma ii
+            (if is_el0 then
+               insert_commit_to_fault ma
+                 (fun _ -> mkfault a_virt ii (Some "EL0")) ii >>! B.Exit
              else mdirect)
 
 (* Read memory, return value read *)
@@ -787,23 +783,23 @@ module Make
   addresses. Thus the resulting monads will possess
   extra dependencies w.r.t the simple case.
  *)
-      let lift_memtag_phy mop a_virt ma ii =
+      let lift_memtag_phy mop a_virt ma mkfault ii =
         M.delay_kont "4" ma
           (fun _ ma ->
             let mm = mop Access.PHY ma in
             delayed_check_tags a_virt ma ii
               (mm  >>! B.Next)
-              (let mfault = mk_fault a_virt ii None in
+              (let mfault = mkfault a_virt ii None in
               if C.precision then  mfault >>! B.Exit
               else (mfault >>| mm) >>! B.Next))
 
-      let lift_memtag_virt mop ma ii =
+      let lift_memtag_virt mop ma mkfault ii =
         M.delay_kont "5" ma
           (fun a_virt ma  ->
             let mm = mop Access.VIR (ma >>= fun a -> loc_extract a) in
             delayed_check_tags a_virt ma ii
               (mm  >>! B.Next)
-              (let mfault = ma >>= fun a -> mk_fault a ii None in
+              (let mfault = ma >>= fun a -> mkfault a ii None in
               if C.precision then  mfault >>! B.Exit
               else (mfault >>| mm) >>! B.Next))
 
@@ -821,16 +817,17 @@ module Make
             fun (r,_) -> M.unitT r
         else m
 
-      let lift_kvm dir updatedb mop ma an ii mphy =
+
+      let lift_kvm dir updatedb mop ma an mkfault ii mphy =
         let mfault ma a =
-          insert_commit_to_fault ma (fun _ -> mk_fault a ii None) ii
+          insert_commit_to_fault ma (fun _ -> mkfault a ii None) ii
           >>! if C.precision then B.Exit else B.ReExec in
         let maccess a ma =
           check_ptw ii.AArch64.proc dir updatedb a ma an ii
             ((let m = mop Access.PTE ma in
               fire_spurious_af dir a m) >>! B.Next)
             mphy
-            mfault in
+            mfault mkfault in
         M.delay_kont "6" ma
           (if pte2 then maccess
            else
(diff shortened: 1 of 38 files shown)


maranget:
Hi @relokin, I'am curious about the motivation of this PR.

maranget:
Hi @nikos, I have used the following test for. a preliminary check:
```
AArch64 FaultProp
{
int x=1;
int y=2;
pte_x=(valid:0);
0:X2=x;
0:X4=y;
}
 P0          ;
 LDR W0,[X2] ;
locations [y;]
forall Fault(P0,x,y=2)

(*
  - herd: ignores the proposition argument to faults.
  - litmus: value 'y' not read by fault handler
  - proposition not shown in test pretty-print.
*)
```
[FaultProp.txt](https://github.com/herd/herdtools7/files/6930226/FaultProp.txt)


Notice that I have tested yesterday version.

relokin:
Hi @maranget, sorry for taking so long to address your comments. In the end the changes to herd were much more intrusive than what I anticipated. And ofc, in the meantime master has progressed quite a bit. Before I rebase and resolve the conflicts may I ask you to have a quick look and let me know if you think this solution is in the right direction?

Right now for every memory location in the proposition of the fault, we generate explicit memory read events that will undergo translation, pretty much as if we had a load instruction to execute. In that sense, we add events that are not due to instructions but due to the requirements of the final proposition. Does that sound right or do you think we should take a completely different direction?

maranget:
Hi @relokin, the approach looks right for a start.  It is difficult for me to evaluate such relevance on a partial solution. In any case if variables are to be read from the context of the fault handler, adding explicit read events connected to the fault event seems unavoidable.

Perhaps the value read  (_i.e._ the symbolic variables that are the result of read events) can be stored into the fault event for easier later retrieval. At the moment I see no reference to those values, but I may be wrong.

relokin:
Hi Luc, thanks for the quick review! I've now rebased on top of master and have resolved a couple of minor issues I was not happy with. Not sure what you meant by partial solution but this patch now should have complete support for faults with propositions for both herd7 and litmus7.

In the approach I took the values of read events are not stored in the fault event. The read events follow the fault event with iico_relations. Then we retrieve the values of the explicit read events by looking at the instruction that is associated with them which should be the same as the one associated with the fault event. Does that make sense?

Please let me know if you have any questions/concerns on a conceptual level or the code itself.

maranget:
Hi @relokin, the code looks fine (I have looked at herd only). I still think that the values read at the time of the fault would be better included in the fault event, as part of a list of bindings for instance. This would make the function `to_fault` easy to write: https://github.com/herd/herdtools7/blob/74c84bc2ff87056f154ce52659f80c072e7fbb56/herd/machAction.ml#L308

relokin:
> Hi @relokin, the code looks fine (I have looked at herd only). I still think that the values read at the time of the fault would be better included in the fault event, as part of a list of bindings for instance. This would make the function `to_fault` easy to write:
> 
> https://github.com/herd/herdtools7/blob/74c84bc2ff87056f154ce52659f80c072e7fbb56/herd/machAction.ml#L308
> 

Thanks for the review Luc!

I will give your suggestion a go.

> Random thought: when we have variable _x_ in the proposition of a fault, do we understand it as PA(_x_) or do we have to perform the access through the page table? Without thinking too much about it I'd side with the simpler PA(_x_) option (litmus is another story).

That's a good question. I took it for granted that we would want to know the value of the virtual address [x]. But if we don't do it that way what is `PA(x)`? It's the initial physical address of `x`? Otherwise if we used the current physical address of `x` this would require a read to `PTE(x)`. 

Not related to this change, but in mind `PA(x)` is a function, so it gives the current physical address that `x` is mapped to.

maranget:
Variables are defined in the init section of a test, by their symbolic virtual addresses: x, y, etc, Then PA(_x_), although a functional notation, always is the physical address of variable _x_, regardless of what PTE(_x_) contents. BTW, the functional notation is the "new syntax" (not my idea), old notation `phy_`_x_ still works.

Inside herd PA(_x_) is symbol `Physical (`_x_`,`_o_`)` where _o_ is an offset https://github.com/herd/herdtools7/blob/7e39512869a336f5af4977808b9aea9be26d18a3/lib/constant.ml#L83

Anyway, the alternative is extracting the current value of physical addresse from PTE(_x_). I'd rather avoid this, as  not being compatible with the usual meaning of variable _x_ in test final condition: it always refer to variable _x_ regardless of PTE(_x_) modification --- _i.e._ physical access is to PA(_x_). Accordingly, litmus restores the PTE's to their initial values, before reading variables to compute the final condition. Of course the difficulty here is that the read if the variables is performed by the fault handler, that is more or less during test execution. Nevertheless I'd rather have  a stable choice for variable reference from final condition.

For litmus it probably means that extra page table entries to variables are to be created, for the fault handler to read them safely.

relokin:
> Hi @relokin, the code looks fine (I have looked at herd only). I still think that the values read at the time of the fault would be better included in the fault event, as part of a list of bindings for instance. This would make the function `to_fault` easy to write:
> 
> https://github.com/herd/herdtools7/blob/74c84bc2ff87056f154ce52659f80c072e7fbb56/herd/machAction.ml#L308
> 

hi @maranget, I am now looking to address this comment too.

My intention, was to have the read events appear in the event graph after the fault event and connected with `po` edges. Right now, I am using the `bind_crtldata` but I am looking to change it to a different bind operator that will give me a `po` edge instead. However, to get to a solution where the values read would be included in the fault event I think I would like to reverse the order of the edge. And that would mean that the fault appears after the read events in program order. Is that what you had in mind or did I misunderstand your suggestion?

maranget:
>And that would mean that the fault appears after the read events in program order. Is that what you had in mind or did I misunderstand your suggestion?

Indeed, I was suggesting that the reads are before the fault. However, I'd rather expect a data deprendency  (`iico_data`) from the read events to the fault event, and not `po`. As a matter of facts, `po` is a total order and it would be problematic to order the read events themselves.

Also notice that, provided you update the substitution function on faults, those variables would get their concrete values automatically, somehow.
https://github.com/herd/herdtools7/blob/57c24354622104b6e3f8b4ef3bd7a42f34381f82/herd/machAction.ml#L633

relokin:
But in practice what we would like to model here, is whether a read could return a certain value from a memory location `x` from the context of a fault handler. I thought that the solution where the read events follow the fault event (with a po edge) would be more representative.

maranget:
> On 4 Oct 2021, at 18:39, Nikos Nikoleris ***@***.***> wrote:
> 
> 
> But in practice what we would like to model here, is whether a read could return a certain value from a memory location x from the context of a fault handler. I thought that the solution where the read events follow the fault event (with a po edge) would be more representative.
> 
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub, or unsubscribe.
> Triage notifications on the go with GitHub Mobile for iOS or Android. 
> 

Hum,  can’t we view the fault as an exception or interrupt that depends upon the reads?

relokin:
> Hum, can’t we view the fault as an exception or interrupt that depends upon the reads?

From a semantics point of view, I think there should be nothing between the read of PTE event and the resulting fault event. It sounds though your suggestion is motivated by the fact that the implementation would be simpler much more straightforward. But I guess what we have to decide is whether this deviation from how things would execute on hardware is worthwhile.

Otherwise, would it be easy for a new operator in monad.mli to back propagate the read values to the fault event?

Or should we abandon the idea of including conditions in the fault atom in favor of a more general solution where there is a separate special thread that serves as the fault handler? Something like:

```
AArch64 FaultContext
{
int x=1;
int y=2;
pte_x=(valid:0);
0:X2=x;
F0:X4=y;
}
 P0                  | F0                  ;
 LDR W0,[X2] | LDR W0,[X4] ;
exits (fault(P0,x) /\ F0:W0=2)```

Where F0 is executed if there is a fault in P0 and after the fault event?

relokin:
My suggestion was to wrap the code in a function `codeF0()` and call it from `fault_handler`,  if the specific fault happens. In a way, that would allow this code to execute concurrently with the other threads P1, P2 without adding too much delay for book keeping.

As you suggest, it would be ideal if `F0` could be associated with a fault at specific instruction `Fault(P0,L0)`, or a process `Fault(P0)` or any fault. I think we should be able to capture that with a statement in the init section.

relokin:
This is no longer needed

PR #265, opened 2021-08-13
[herd] Annotate read of page table entry from instruction annotation.

maranget:
Instruction annotations A,Q and L are used to annotate page table walk read. Notice that this may create read events annotated with release and exclusive events annotated with acquirePC. There are no instructions that generate those annotations directly. Indeed the idea is more to identify the instruction that generates the PTW read than to annotate the PTW read itself.


jalglave:
Thanks Luc, I've reviewed and tested it locally.
Jade

PR #274, opened 2021-09-22
[herd] Fix loop

maranget:
Fix detection of insufficient loop unrolling. Pruned executions are now excluded from final display as they should (unless `-variant toofar` is specified).


relokin (approved):
I had a quick look at the code and run a couple of tests and the changes look good to me!

PR #275, opened 2021-09-23
[herd] Strict load exclusive/store exclusive

maranget:
By default herd does not yield constrained unpredictable behaviours.

As regards load exclusive/store exclusive matching pairs, the store will succeed only if addresses and access sizes match exactly.

This strict behaviour can be relaxed with `-variant ConstrainedUnpredictable`. Then store exclusive may succeed for diffrent addresses or different access sizes.


maranget:
Draft, waiting for unit tests.

herd/archExtra_herd.ml
```diff
@@ -274,12 +278,14 @@ module Make(C:Config) (I:I) : S with module I = I
 
       let reg_state_empty = RegMap.empty
 
+      type ii_env =  { regs:reg_state; lx_sz:MachSize.sz option; }
```
relokin, line 281: That's a nice and simple solution but I have one concern which has to do with our overall implementation of the Exclusives monitor.

In AArch64, a load exclusive marks the location it reads from as exclusive. A subsequent store exclusive will succeed if the location it writes to is marked as exclusive. A location marked as exclusive has to be unmarked if another core performs a store (or a successful store exclusive) to that same location [1]. This means that for every store we perform we have to check whether the location we write has been marked as exclusive and if yes clear the mark.

In herd7, we use the special ResAddr register of the core to mark a location as exclusive. This means that, every time we execute a store we should check all other cores' ResAddr and if the location there matches the current store we should clear it. Along with the ResAddr we have to check the size, to check for overlapping accesses. So I was wondering if there a different way in which the contents of ResAddr becomes a tuple (addr,sz), or we add another register ResSize which we can check together, or we move the Exclusives monitor to the inst_instance_id.

[1]: There are also some implementation defined options with regards to whether the mark is lost if an exclusive access from the same core references an unmarked memory location which could ignore for now.

herd/AArch64Sem.ml
```diff
@@ -39,6 +39,7 @@ module Make
     let kvm = C.variant Variant.Kvm
     let is_branching = kvm && not (C.variant Variant.NoPteBranch)
     let pte2 = kvm && C.variant Variant.PTE2
+    let do_cu = C.variant Variant.ConstrainedUnpredictable
```
relokin, line 42: I am wondering whether it would be better to instead raise a flag ContrainedUnpredictable and keep things simple. The problem is that there are multiple parts of the architecture that allow for ContrainedUnpredictable behaviors, not only LDRX/STRX. Also an given implementation makes a certain choice (e.g., Exclusives monitor on a cache-line granularity) which means that if the LDRX/STRX pair will succeed if some conditions are met and not unconditionally.

herd/archExtra_herd.ml
maranget: Interaction of several threads and of reservation is not handled by the means of the ResAddr register.

The ResAddr "register" is used to record the address reserved by the (po)-last LoadExclusive it does not implement the Exclusive monitors, but rather control the local part of it. More precisely, its purpose is to inform the (po)-next STX of the current reservation (if any). That is, if the effective address of the STX matches the reservation, then the store may fail or succeed, otherwise it always fails (Constrained unpredictable not allowed and access sizes ignored).

Hence, in case of matching addresses, success and failure are considered, leading to different execution candidates, which are input to the Cat interpreter. Atomicity violations, _i.e_ the situations where some write by another thread occurs between the load exclusive and the store conditional, are rejected by the model (rule is something like `empty rmw & (fr;co)`, where `rmw` includes LDX/STX  pairs of matching events.

herd/AArch64Sem.ml
maranget: The general idea is to avoid Contrained Unpredictable by default. The `-variant ContrainedUnpredictable` command line option instructs herd to take the opposite view to yield such behaviours, _as much as possible_. In the case of LDRX/STRX this means that store exclusive are possible, regardless of the reservation value (if any). I am not sure it is desirable to instruct herd about details such as granule size etc. Other Contrained Unpredictable are not handled yet and may be in the future. Hopefully, one variant will be enough...

herd/archExtra_herd.ml
relokin: oh I see, I forgot about the `empty rmw & (fr;co)` axiom. But doesn't that mean that executions where a STRX fails because of an intervening write are discarded by herd7? Meaning there are more architecturally valid states that won't be included in the output?

relokin: Ok I had another look and now I understand how it works. Sorry I was looking at only one part of the implementation and I was missing the big picture. Everything makes sense and the change looks good to me.

maranget: Well it took several years to the current implementation. I'd not call this obvious.

PR #278, opened 2021-09-30
[test] Test catalogue

maranget:
Restore catalogue regression testing. Kinds are tested.


maranget:
This PR has a dependency on PR #279, as herd should accept option `-timeout`. However, all other features should be functionnal.

PR #282, opened 2021-10-07
[litmus] Get rid of the unecessary next_tb field from ctx_t

relokin:

### DIFF
--- a/litmus/libdir/_instance.c
+++ b/litmus/libdir/_instance.c
@@ -30,9 +30,6 @@ typedef struct {
 #ifdef SEE_FAULTS
   see_fault_t f;
 #endif
-#ifdef HAVE_TIMEBASE
-  tb_t next_tb;
-#endif
   hash_t t;
   sense_t b;
   param_t p;
--- a/litmus/preSi.ml
+++ b/litmus/preSi.ml
@@ -1359,10 +1359,10 @@ module Make
               inits
         end ;
         (* Synchronise *)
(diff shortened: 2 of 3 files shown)


maranget:
I think the field is useful. Its purpose is for all the threads to agree on a common starting point used for synchronisation. Here is the original produced code:
```C
      _a->next_tb = read_timebase();
      barrier_wait(barrier);
      tb_t _tb0 = _a->next_tb;
      int _delta;
      do { _delta =  read_timebase() - _tb0; } while (_delta < _b->delay);
```
Here, all threads write into the common `next_tb`, then synchronise, then read  `next_tb`. The intend is for all therads to read the same value, the one written last. Hence the thread should leave the `do` loop about at the same time (if the `_b->delay` fields are the same for all threads, but this is another story).

The new code is at follows:
```C
     tb_t _tb0 = read_timebase();
      barrier_wait(barrier);
      int _delta;
      do { _delta =  read_timebase() - _tb0; } while (_delta < _b->delay);
```
Here each thread read the timebase before entering the barrier, reading a value likely different from other therads. Hence there is no reason for the threads to leave the `do` loop more or less at the same time.

There is one assumption here: threads read the same timebase register or at least timebase that reflect the same time. Maybe I miss something?

maranget:
Agreed it is not 100% clear that _a is a pointer to a struct of type `ctx_t` that is shared between the threads of a given test instance, while _b is a pointer to a struct of type `parg_t` that is private to each thread. Perhaps  some comment in generated code would be useful.

PR #294, opened 2021-11-01
semantics for AMO instructions in DIY

hugookeeffe:
Implementing the semantics of AMO instructions into DIY so that the post condition can consider this when determining its values

### DIFF
--- a/gen/AArch64Arch_gen.ml
+++ b/gen/AArch64Arch_gen.ml
@@ -598,24 +598,15 @@ let get_ie e = match e with
 let fold_edge f r = Code.fold_ie (fun ie r -> f (IFF ie) (f (FIF ie) r)) r
 
 
-let compute_rmw r old co =
-   begin match r with
-    | LdOp op ->
+let compute_rmw r old co = match r with
+    | LdOp op | StOp op ->
       begin match op with
         | A_ADD -> old + co
         | A_SMAX -> if old > co then old else co
         | A_SMIN -> if old < co then old else co
         | A_EOR | A_SET | A_CLR -> co
-      end
-    | StOp op ->
-      begin match op with
-        | A_ADD -> old + co
-        | A_SMAX -> if old > co then old else co
-        | A_SMIN -> if old < co then old else co
-        | A_EOR | A_SET | A_CLR -> co
-      end
+    end
     | LrSc | Swp | Cas  -> co
-  end
 
 include
     ArchExtra_gen.Make
--- a/gen/AArch64Arch_gen.ml
+++ b/gen/AArch64Arch_gen.ml
@@ -606,7 +606,7 @@ let compute_rmw r old co = match r with
         | A_SMIN -> if old < co then old else co
         | A_EOR -> old lxor co
         | A_SET -> old lor co
-        | A_CLR -> lnot (old land co)
+        | A_CLR -> old land (lnot co)
     end
     | LrSc | Swp | Cas  -> co
 
--- a/gen/AArch64Arch_gen.ml
+++ b/gen/AArch64Arch_gen.ml
@@ -604,7 +604,9 @@ let compute_rmw r old co = match r with
         | A_ADD -> old + co
         | A_SMAX -> if old > co then old else co
         | A_SMIN -> if old < co then old else co
-        | A_EOR | A_SET | A_CLR -> co
+        | A_EOR -> old lxor co
+        | A_SET -> old lor co
+        | A_CLR -> lnot (old land co)
     end
     | LrSc | Swp | Cas  -> co
 
--- a/gen/AArch64Arch_gen.ml
+++ b/gen/AArch64Arch_gen.ml
@@ -597,6 +597,26 @@ let get_ie e = match e with
 
 let fold_edge f r = Code.fold_ie (fun ie r -> f (IFF ie) (f (FIF ie) r)) r
 
+
+let compute_rmw r old co =
+   begin match r with
+    | LdOp op ->
+      begin match op with
+        | A_ADD -> old + co
+        | A_SMAX -> if old > co then old else co
+        | A_SMIN -> if old < co then old else co
+        | A_EOR | A_SET | A_CLR -> co
+      end
+    | StOp op ->
+      begin match op with
+        | A_ADD -> old + co
+        | A_SMAX -> if old > co then old else co
+        | A_SMIN -> if old < co then old else co
+        | A_EOR | A_SET | A_CLR -> co
+      end
+    | LrSc | Swp | Cas  -> co
+  end
+
 include
     ArchExtra_gen.Make
     (struct
--- a/gen/cycle.ml
+++ b/gen/cycle.ml
@@ -463,16 +463,16 @@ module CoSt = struct
 
   let get_cell st = st.co_cell
 
-  let set_cell ?(rv = -1) st n =
+  let set_cell st n =
     let e = n.evt in match e.bank with
     | Ord -> begin
        let old = st.co_cell.(0) in
        let cell = E.overwrite_value old e.atom e.v in
        let co_cell = Array.copy st.co_cell in
-       let cell2 = match n.edge.E.edge with
+       let cell2 = match n.prev.edge.E.edge with
         | E.Rmw rmw -> E.compute_rmw rmw old cell
-        | _ -> if rv = -1 then cell else rv in
-      co_cell.(0) <- cell2;
+        | _ -> cell in
+       co_cell.(0) <- cell2;
       {e with cell=co_cell;},{ st with co_cell; }
     end
     | _ -> e,st
@@ -713,7 +713,7 @@ let set_same_loc st n0 =
 
   let tr_value e v = E.tr_value e.atom v
 
-  let rec do_set_write_val ?(rv = -1) st pte_val = function
+  let rec do_set_write_val st pte_val = function
     | [] -> ()
     | n::ns ->
         begin if Code.is_data n.evt.loc then
@@ -746,7 +746,7 @@ let set_same_loc st n0 =
                    n.evt <- { n.evt with v = tr_value n.evt v; } ;
                    (* Writing Ord resets morello tag *)
                    let st = CoSt.set_co st CapaTag evt_null.ctag in
-                   let e,st = CoSt.set_cell st n ~rv:rv in
+                   let e,st = CoSt.set_cell st n in
                    n.evt <- e ;
                    do_set_write_val st pte_val ns
                 | Tag|CapaTag|CapaSeal ->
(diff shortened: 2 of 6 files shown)


gen/cycle.ml
```diff
@@ -726,7 +726,6 @@ let set_same_loc st n0 =
             let ord = CoSt.get_co st Ord in
             let ctag = CoSt.get_co st CapaTag in
             let cseal = CoSt.get_co st CapaSeal in
-            fprintf stderr "test 20 \n" ;
```
relokin: I think we might want to squash this commit to the previous one.

gen/AArch64Arch_gen.ml
```diff
@@ -597,6 +597,26 @@ let get_ie e = match e with
...
+let compute_rmw r old co =
+   begin match r with
+    | LdOp op ->
+      begin match op with
+        | A_ADD -> old + co
+        | A_SMAX -> if old > co then old else co
+        | A_SMIN -> if old < co then old else co
+        | A_EOR | A_SET | A_CLR -> co
+      end
+    | StOp op ->
```
relokin, line 610: Can you handle both LdOp and StOp together? It looks like, we handle the same operators.

gen/AArch64Arch_gen.ml
```diff
@@ -597,6 +597,26 @@ let get_ie e = match e with
...
 
+
+let compute_rmw r old co =
+   begin match r with
+    | LdOp op ->
+      begin match op with
+        | A_ADD -> old + co
+        | A_SMAX -> if old > co then old else co
+        | A_SMIN -> if old < co then old else co
+        | A_EOR | A_SET | A_CLR -> co
```
relokin, line 608: I know this is wip, but if in the end we don't add support for bitwise ops yet, maybe we could add a warning.

gen/cycle.ml
```diff
@@ -463,13 +463,20 @@ module CoSt = struct
...
 
-  let set_cell st e = match e.bank with
-    | Ord ->
-       let old = st.co_cell.(0) in
-       let cell = E.overwrite_value old e.atom e.v in
-       let co_cell = Array.copy st.co_cell in
-       co_cell.(0) <- cell ;
-       {e with cell=co_cell;},{ st with co_cell; }
+  let set_cell ?(rv = -1) st n =
+    let e = n.evt in  match e.bank with
```
relokin, line 467: Just to keep this diff to a minimum, maybe we could move this let further down where it's used. Also many of the lines below we're only changing their indentation.

gen/cycle.ml
```diff
@@ -463,13 +463,20 @@ module CoSt = struct
...
+    | Ord -> begin
+      let old = st.co_cell.(0) in
+      let cell = E.overwrite_value old e.atom e.v in
+      let co_cell = Array.copy st.co_cell in
+      match n.edge.E.edge with
```
relokin, line 472: can we simplify this to something like
```
let cell2 =
      match n.edge.E.edge with
      | E.Rmw rmw -> E.compute_rmw rmw old cell
      | _ -> if rv = -1 then cell else rv in
co_cell.(0) <- cell2;
{e with cell=co_cell;},{ st with co_cell; }
```

gen/cycle.ml
```diff
@@ -719,6 +726,7 @@ let set_same_loc st n0 =
             let ord = CoSt.get_co st Ord in
             let ctag = CoSt.get_co st CapaTag in
             let cseal = CoSt.get_co st CapaSeal in
+            fprintf stderr "test 20 \n" ;
```
relokin: you could remove this, or squash the next patch into this one.

gen/cycle.ml
```diff
@@ -741,16 +749,16 @@ let set_same_loc st n0 =
...
                    n.evt <- e ;
                    do_set_write_val st pte_val ns
                 | Tag|CapaTag|CapaSeal ->
-                   let st = CoSt.next_co st bank in
-                   let v = CoSt.get_co st bank in
-                   n.evt <- { n.evt with v = v; } ;
-                   let e,st = CoSt.set_tcell st n.evt in
-                   n.evt <- e ;
-                   do_set_write_val st pte_val ns
+                  let st = CoSt.next_co st bank in
```
relokin, line 755: I don't think there is any meaningful change here, is there? It's only indentation.

gen/edge.ml
```diff
@@ -31,6 +31,7 @@ module type S = sig
   type atom
   type rmw
 
+
```
relokin, line 34: Unintended change?

gen/cycle.ml
```diff
@@ -782,9 +790,30 @@ let set_same_loc st n0 =
...
             | Code _ ->
-               do_set_write_val st pte_val ns
+              do_set_write_val st pte_val ns
             end
-        | Some R | Some J |None -> do_set_write_val st pte_val ns
+        | Some R ->
+          begin match n.evt.loc with
+            | Data _ ->
+              let bank = n.evt.bank in
+              begin match bank with
```
relokin, line 798: just to avoid too much nesting you could also do something like
```
begin match bank, n.evt.rmw with
| Ord, true ->
....
```

relokin (commented):
Thanks for the fix Hugo!

Just a couple of comments from a quick look at the code. I will also try to give it a go later today.

gen/AArch64Arch_gen.ml
```diff
@@ -597,6 +597,26 @@ let get_ie e = match e with
...
+      end
+    | StOp op ->
+      begin match op with
+        | A_ADD -> old + co
+        | A_SMAX -> if old > co then old else co
+        | A_SMIN -> if old < co then old else co
+        | A_EOR | A_SET | A_CLR -> co
+      end
+    | LrSc | Swp | Cas  -> co
+  end
```
maranget, line 618: This `begin/end` is not rquired. ON ca write `let compute_rmw r old co = match ...` directly.

gen/AArch64Arch_gen.ml
maranget: Agreed, you can use the or-pattern `LdOp op|Stop op -> ,,,` to share the identical clauses.

gen/cycle.ml
```diff
@@ -708,7 +715,7 @@ let set_same_loc st n0 =
 
   let tr_value e v = E.tr_value e.atom v
 
-  let rec do_set_write_val st pte_val = function
+  let rec do_set_write_val ?(rv = -1) st pte_val = function
```
maranget, line 718: I do not understand the logic behind this optional argument `rv`. The state argument `st` has two components of interest, value `v` (in practice `v` is the value bound to `Ord`  in `st.map`) and component `co_cell`. Component `co_cell` (here, in the non-neon case `co_cell.(0)`) represents the contents of memory after the write is performed. As regards our extension component v still follows the sequence if integers, 0,1,2,... as before. The computation of `co_cell` changes: it becomes `set.co_cell.(0)` op `v` (op is the operation performed by the AMO operation and `v` is the operand of the instructio, available from `set_cell` as `n.evt.v`,

Hence, the function `set_cell` does not need the extra `rv` argument as the old value of memory is available as `st.cell.(0)` and the operand is available as `n.evt.v`.

Notice that further adaptation may be  needed for mixed-size mode: the calls to functions `tr_value` at lign 741/748 below and `E.overwrite_value` in `set_cell` are concern mixed-size mod and are trivial in non-mixed-size mode. But mixed-size is yet another story.

gen/cycle.ml
```diff
@@ -463,16 +463,16 @@ module CoSt = struct
...
        let cell = E.overwrite_value old e.atom e.v in
        let co_cell = Array.copy st.co_cell in
-       let cell2 = match n.edge.E.edge with
+       let cell2 = match n.prev.edge.E.edge with
         | E.Rmw rmw -> E.compute_rmw rmw old cell
-        | _ -> if rv = -1 then cell else rv in
-      co_cell.(0) <- cell2;
+        | _ -> cell in
+       co_cell.(0) <- cell2;
       {e with cell=co_cell;},{ st with co_cell; }
```
maranget: Performing a RWM operation is as follows:

1. Read old value, here `let old = st.co_cell.(0)`
2. Compute, here `let cell2 = ... E.compute_rmw old e.v`
3. Store the result, here  `co_cell .(0) <- E.overwrite_value old e.atom cell2`

The `E.overwrite` function replaces the the subpart of the memory location specified by `e.atom` with the operation result, leaving the rest unchanged. This applies only to mixed-size mode, where `e.atom` may specify some subpart such as second byte, first 16bits word, etc.

gen/cycle.ml
```diff
@@ -463,16 +463,16 @@ module CoSt = struct
...
 
-  let set_cell ?(rv = -1) st n =
+  let set_cell st n =
     let e = n.evt in match e.bank with
     | Ord -> begin
        let old = st.co_cell.(0) in
        let cell = E.overwrite_value old e.atom e.v in
        let co_cell = Array.copy st.co_cell in
-       let cell2 = match n.edge.E.edge with
+       let cell2 = match n.prev.edge.E.edge with
```
maranget, line 472: This ` .prev` is surprising.

gen/cycle.ml
```diff
@@ -787,27 +787,9 @@ let set_same_loc st n0 =
...
-                   let e,st1 = CoSt.set_cell st1 n in
-                   n.evt <- e ;
-                   let co = n.evt.cell.(0) in
-                   do_set_write_val st1 pte_val ns ~rv:co
-                | _   -> do_set_write_val st pte_val ns
-              end
-            | Code _ -> do_set_write_val st pte_val ns
-          end
-        | Some J |None -> do_set_write_val st pte_val ns
+        | Some R | Some J |None -> do_set_write_val st pte_val ns
```
maranget: A matter of style...
```suggestion
        | Some (R|J) |None -> do_set_write_val st pte_val ns
```

gen/AArch64Arch_gen.ml
```diff
@@ -604,7 +604,9 @@ let compute_rmw r old co = match r with
         | A_ADD -> old + co
         | A_SMAX -> if old > co then old else co
         | A_SMIN -> if old < co then old else co
-        | A_EOR | A_SET | A_CLR -> co
+        | A_EOR -> old lxor co
+        | A_SET -> old lor co
+        | A_CLR -> lnot (old land co)
```
maranget, line 609: Isn't it `old land (lnot co)`?

hugookeeffe: Ah yes, that is my bad. I will change that.

gen/cycle.ml
hugookeeffe: Thank you!

gen/cycle.ml
```diff
@@ -463,13 +463,18 @@ module CoSt = struct
...
+    let e = n.evt in match e.bank with
+    | Ord -> begin
        let old = st.co_cell.(0) in
        let cell = E.overwrite_value old e.atom e.v in
        let co_cell = Array.copy st.co_cell in
-       co_cell.(0) <- cell ;
-       {e with cell=co_cell;},{ st with co_cell; }
+       let cell2 = match n.prev.edge.E.edge with
+        | E.Rmw rmw -> E.compute_rmw rmw old e.v
+        | _ -> cell in
```
relokin, line 474: Shouldn't this be:
```
| _ -> e.v
```
In which case we can also get rid of line 470?

relokin (commented):
Thanks for all the changes Hugo. I have one comment but otherwise, it looks good to me and runs fine with a couple of examples I tried.

@maranget how do you typically merge? Should Hugo squash all commits into one and then we can use fast-forward merge? Or will the merge squash them automatically.

gen/cycle.ml
maranget: Hi @relokin, I usually merge with a merge commit. If commit history is complex I squash. I mean I use github "squash and merge" button. Here I do not know yet.

maranget: Hi @hugookeeffe and @relokin, I'd like to merge the PR in once the last comment by @relokin above is handled. 

Notice that, once simplified, the `set_cell` function can be adapted easily to mixed-size by using the `E.extract_value` function. For clarity, commit ba305dfc1eb3ffd57087c04e57b6def2c65eee6a performs the adaptation (also addressing @relokin's comment).

hugookeeffe: Hi @maranget, I have made the changes for the last comment in my recent commit. Thank you for the help!

maranget: Thanks, I'll merge tomorrow.

maranget:
Merge after squash, thanks.

PR #308, opened 2021-12-03
[gen] Better control for `-variant MixedStrictOverlap`

maranget:
In particular always accept identical accesses on both sides
of Rmw edges, as this is often the only possibility.


maranget:
Closed, as included in PR #309

PR #320, opened 2022-02-07
Minor fixes in the build system for generated tests in KVM mode

relokin:


maranget:
@relokin LGTM. I merge. I notice duplicated  occurrences of some gcc option `-std=gnu99` and `-Wall` in produced Makefile. It does not harm much, I guess.

PR #328, opened 2022-02-22
Memattrs fixes

relokin:
Two minor fixes for handling memory attributes.
* create PTE* events for all PTE accesses: implicit/explicit and read/write.
* Introduce one more event PTEMemAttr, a catch all event for all PTE accesses which have one more memory attributes defined.

### DIFF
--- a/herd/machModelChecker.ml
+++ b/herd/machModelChecker.ml
@@ -352,35 +352,38 @@ module Make
                  "NDATA", (fun e -> not (is_data_port e));])) in
       let m =
         if kvm then begin
+            let attrs_of_evt e =
+              let pteval_v =
+                match E.read_of e with
+                | Some _ as v -> v
+                | _ -> E.written_of e in
+              let open Constant in
+              match pteval_v with
+              | Some (S.A.V.Val (PteVal v)) ->
+                 S.A.V.Cst.PteVal.get_attrs v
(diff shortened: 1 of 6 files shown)


herd/machModelChecker.ml
```diff
@@ -372,10 +376,12 @@ module Make
                       let evts_w_attr = E.EventSet.add e evts_w_attr in
                       let evts_map = StringMap.add attr evts_w_attr evts_map in
                       evts_map) attrs evts_map)
-                impl_pte_reads StringMap.empty in
+                pte_accesses StringMap.empty in
             I.add_sets m
               (StringMap.fold
-                 (fun k v l -> ("PTE" ^ k, lazy v) :: l) evts_map [])
+                 (fun k v l ->
+                   ("PTE" ^ k, lazy v) :: ("PTEMemAttr", lazy v) :: l)
```
maranget, line 383: Hi @relokin, This code may attempt to perform the "PTEMemAttr" binding several tiimes, once per key `k`  present in `evts_map`.

relokin: Thanks for the review @maranget and sorry for the delay in addressing your comment.

Very good point, I just pushed an update which I believe addresses this.

maranget:
Looks good to me, thanks.

PR #331, opened 2022-02-23
[AArch64 cat] Pick dependencies

jalglave:
This change is a relaxation.

Arm has investigated the weakening of its dependency definitions, in particular
the notion of control dependency. In light of this investigation, Arm has
decided to keep its notion of control dependency strong, but proposes to
introduce a new notion of dependency, called "Pick dependency". Pick
dependencies allow for very targetted ordering, and it is hoped that this will
enable optimisations.

See also: https://www.spinics.net/lists/kernel/msg4028862.html
and the attached tech report entitled "The Ballad of Dependencies".

Motivation
-------------

At the moment the behaviour MP+rel+CAS-addr:
```
AArch64 MP+rel+CAS-addr
{
z=1;
0:X1=x; 0:X3=y;
1:X1=x; 1:X3=y; 1:X8=z;
}
P0           | P1               ;
MOV W0,#1    | LDR W0, [X1]     ;
STR W0,[X3]  | MOV W5, W0       ;
STLR W0,[X1] | CAS W0, W6, [X8] ;
             | LDR W0, [X8]     ;
             | EOR X0, X0, X0   ;
             | ADD X3, X3, X0   ;
             | LDR W4, [X3]     ;
exists (1:X5=1 /\ 1:X4=0)
```
is Forbidden by the architecture.
Arm would like to weaken this to allow the behaviour.

The reason for MP+rel-CAS-addr being currently Forbidden is as follows:

1. there is a Data dependency on P1 from the Memory read Effect of LDR W0,[X1]
(event c) to the Memory write Effect of CAS W0,W6,[X8] (event e)
2. there is a read-from on P1 from the Memory write Effect of CAS W0,W6,[X8]
(event e) to the Memory read Effect of LDR W0,[X8] (event f) - in other words
the Memory read Effect of LDR W0,[X8] (event f) is a Local read successor of
the Memory write Effect of CAS W0,W6,[X8] (event e)
3. there is an Address dependency on P1 from the Memory read Effect of LDR
W0,[X8] (event f) to the Memory read Effect of LDR W4,[X3] (event g).

Points 1 and 2 combined impose order between the Memory read Effect of LDR
W0,[X1] (event c) to the Memory read Effect of LDR W0,[X8] (event f), because
data; rfi is in dob. In the text of the Arm ARM this corresponds to the clause:
(*)
RW2 is a Local read successor R2 of a write W3 and there is an Address
dependency or a Data dependency from R1 to W3.

A way forward to allow this behaviour would be to decide that there is no
[...]

PR #335, opened 2022-03-02
[herd-www] Update herd www

maranget:
Update web interface and catalogue following  PR #331


maranget:
Hi @jalglave, this PR updates  two  catalogue books: aarch64 and aarch64-mixed. May I have your opinion on other books, in particular on aarch64-MTE.

PR #342, opened 2022-03-04
[herd] Add a new event set TLBUncacheable

relokin:
The Arm architecture permits TLBs to cache a PTE when it doesn't cause
a Translation fault or an Access flag fault.


relokin:
Marking this draft because it's unclear if we want to go down this path.

relokin:
We've decided that there is a better way of dealing with this in the cat file and the changes in this PR are not needed.

PR #347, opened 2022-03-15
[herd] Basic implementation of Java architecture in herd

ShuyangLiu:
Hello,

Here is a basic implementation of "java" arch for herd. Currently, it only support a limited subset of Java in Herd, not in other tools (`assert false` if other tools try to use JAVA as the architecture). 

Thank you!


maranget:
PR Merged by hand, due to internal API changes that preclude compilation.

Thanks a lot @ShuyangLiu.

PR #348, opened 2022-03-15
Segmenting large configuration files

hugookeeffe:
The purpose of this script is to segment a large configuration file and run it with diy7. This was made in order to be able to generate more tests for certain configuration files and to reduce run time for certain large configuration files.

### DIFF
--- a/tools/segment.sh
+++ b/tools/segment.sh
@@ -73,6 +73,7 @@ fi
 # create the config files
 whole=""
 for ((i=0; i<${seg_size}; i++)); do
+    cd src && rm -rf src${i} && mkdir src${i} && cd ..
     for aspect in "${array2[@]:(i*${relax_size}+1):${relax_size}}"; do
         whole+="$aspect "
     done
@@ -86,7 +87,7 @@ $ol
 
 -name Armv8-MIXED-ext-forbidden-$i-
 
--relaxlist Rfe Fre Wse Hat $whole" >./src/${filename}${i}.conf
+-relaxlist Rfe Fre Wse Hat $whole" >./src/src${i}/${filename}${i}.conf
     whole=""
 done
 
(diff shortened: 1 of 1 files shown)


maranget:
Hi @hugookeeffe, the script is useful, thanks. I ran it and notice the following problems:

- Some lines in the @all file us ends with the two characters `\` \n` (_i.e._ backslash followed by n). This prevents the file from being used by other tools. I tried `name7 @all`.
- May I suggest another organisation of tests: create sub-directories src00, src01 etc. with each directory containing its `.conf` files, the tests and its `@all` file; and a toplevel `@all` files that simply lists `src00/@all`, `src01/@all` etc.

hugookeeffe:
> Hi @hugookeeffe, the script is useful, thanks. I ran it and notice the following problems:
> 
> * Some lines in the `@all` file us ends with the two characters `\`  and `n` (_i.e._ backslash followed by n). This prevents the file from being used by other tools. I tried `name7 @all`.
I think I have solved this issue in the newest commit, let me know if it still does not work.
> * May I suggest another organisation of tests: create sub-directories src00, src01 etc. with each directory containing its `.conf` files, the tests and its `@all` file; and a toplevel `@all` files that simply lists `src00/@all`, `src01/@all` etc.
I have implemented this change in the newest commit

Let me know if there are any other changes you would like me to make

tools/segment.sh
```diff
@@ -0,0 +1,109 @@
...
+    else
+        diy7 -o ./src/src${i} -conf ./src/src${i}/${filename}${i}.conf& read -t $timeout || kill $!
+    fi
+done
+
+
+
+for ((i=0; i<${seg_size}; i++)); do
+    echo "./src/src${i}/@all" >>@all
+    echo "\n" >>@all
```
maranget, line 108: This lines introduces untimely `\` `n` sequences in the output file `@all`. Deleting it apparently solves the problem: 
```suggestion
```

maranget (changes requested):
Some leftover `\` `n` sequence still here, trying  github commit.

maranget:
Merged, thanks

PR #353, opened 2022-03-22
[herd] RISC-V Ztso extension memory model

patrick-rivos:
RISC-V defines its Ztso extension as equivalent to RVWMO, but with
three adjustments [1]:

 1. All load operations behave as if they have an acquire-RCpc annotation
 2. All store operations behave as if they have a release-RCpc annotation.
 3. All AMOs behave as if they have both acquire-RCsc and release-RCsc annotations.

As such, AMO ops have been added to the RCsc definition, and
acquire/release RCpc annotations are implemented in line with page 16 of
the 1995 paper "Shared Memory Consistency Models: A Tutorial"[2].

This is a tentative implementation and may need edits in order to
be accurate/mergeable. Feedback is welcome.

[1] https://github.com/riscv/riscv-isa-manual/blob/c9a172ff2245824b0c55b234a3bb45664394d038/src/ztso.tex#L12-L17
[2] https://www.hpl.hp.com/techreports/Compaq-DEC/WRL-95-7.pdf

### DIFF
--- a/herd/libdir/riscv-tso-defs.cat
+++ b/herd/libdir/riscv-tso-defs.cat
@@ -33,7 +33,8 @@ let AcqRel = AcqRel|Sc (* Compat *)
 let AQ = (Acq|AcqRel)
 and RL = (Rel|AcqRel)
 let AMO = try AMO with (R & W) (* Compat *)
-let RCsc = (Acq|Rel|AcqRel) & (AMO|X)
+(* All AMO ops have RCsc annotations *)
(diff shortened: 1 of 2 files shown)


maranget:
Hi @patrick-rivos, I have merged your PR by hand and closed it here. Thanks.

Your interpretation of the Ztso extension looks good to me. I have checked the Ztso model to be stronger than the standard RISCV model.

Thanks again.

Issue #10, opened 2018-09-05
option to extract litmus test as AST


mn200:
Is there any way to extract a parsed litmus test as an AST in some easy-to-parse format (sexp, JSON, *etc*)?

PR #361, opened 2022-04-11
[herd]: Print Timeout for test with -debug timeout

lukeg101:
Suppose you run `herd7 -timeout 5.0 expensive_test.litmus` and it times
out on `expensive_test`. Herd silently finishes with `expensive_test` timing out and it is
desirable to be able to have a list of timed-out tests in automated
testing. This patch prints a timeout message to standard error so we can
collect lists of tests which did not finish:
```
> herd7 -timeout 5.0 -debug timeout expensive_test.litmus
...
Timeout for expensive_test.litmus
```
============

All regression tests pass

This patch has been pre-reviewed by @relokin in https://github.com/lukeg101/herdtools7/pull/10

### DIFF
--- a/herd/debug_herd.ml
+++ b/herd/debug_herd.ml
@@ -28,6 +28,7 @@ type t = {
     pretty : bool ;
     mixed : bool ;
     files : bool ;
+    timeout : bool ;
   }
 
 let tags =
@@ -43,6 +44,7 @@ let tags =
   "pretty";
   "mixed";
   "files";
+  "timeout"
 ]
 
 let none =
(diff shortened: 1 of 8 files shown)


relokin:
Looks good to me.

One minor thing; I was just wondering whether it would be better if we printed the warning unconditionally. Right now `herd7` exits without any information.

maranget:
I am ok with the current behaviour of not pintimg the warning by default, as it preserves backward compatibility.

I am wondreing how the second (merge pull #10) made it here. Anyway I'll squash te two commits and merge, thanks.

lukeg101:
Hi Luc, Thanks for this, the merge appeared because I did a PR on my own fork from `timeout-review` to `timeout` so Nikos could pre-review the code. In the future, I will submit PRs to the herd repo from the original  branch so that merges do not occur.

PR #366, opened 2022-04-22
[herd] Safer linearisations

maranget:
Functions that compute topological orders take a `nodes`
and an `edges` argument that represent the relation to be extended
as orders.

The new code systematically replaces edges by its transitive closure,
yielding a more intuitive and less dangerous semantics.

More precisely consider `nodes = {n1,n2}` and edges = `n1->n3, n3->n2}`.
Then results are:
  + Without transitive closure: `n1->n2` and `n2->n1`
  + With transitive closure `n1->n2` only.

### DIFF
--- a/lib/innerRel.ml
+++ b/lib/innerRel.ml
@@ -600,6 +600,7 @@ and module Elts = MySet.Make(O) =
       do_aux
 
     let fold_topos_ext kont res nodes edges =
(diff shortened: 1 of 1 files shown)


jalglave:
Thanks Luc, looks good to me.
Jade

maranget:
Thanks, merged.

PR #369, opened 2022-04-28
[herd] Implement AArch64 LDP and STP instructions

maranget:


maranget:
This PR is ready. However internal dependencies of store instructions are not totally satisfactory. Given the complexity of teh task another pull request will build on this one, aiming at more simple dependencies.

PR #376, opened 2022-05-05
[lib] fix RISC-V aqrl assembly format

chenguokai:
this should enable RISC-V amo related tests to build with gcc

Previously, herdtools generate amo instructions with format like `amoswap.w.aq.rl`, this mismatches with what `gcc` expects (`amoswap.w.aqrl`). This PR fixes this inconsistency.


maranget:
Well spotted, thanks. RISC-V documentation confirms your patch.

PR #381, opened 2022-05-12
[herd] Simplify the inv-domain relation.

maranget:
In effect, the `inv-domain` relation is useful as a relation
between TLBI events and PTE accesses.

Relating TLBI events with (initial) physical addresses does
not make sense, while relating  TLBI events with virtual adresses
is useless, as no event whose location is virtual exists
in `-variant kvm` mode.

Notice that `inv-domain` does not applies to page table
accesses in mode `-variant pte2`. Anyway, as far as herd is concerned,
the TLBI instruction does not apply to PTE addresses.


jalglave:
Thanks Luc, this looks good to me. @relokin tried it also.
Jade

PR #385, opened 2022-05-23
[ARM]: Added DMB ISH to arm model

lukeg101:
Consider the following store buffering litmus test.
The final state should be forbidden under the arm model.
This patch updates the arm model so that the final state is not
observed.
```
{
0:R2=x; 0:R3=y;
1:R2=y; 1:R3=x;
}
 P0           | P1           ;
 MOV R0,#1    | MOV R0,#1    ;
 STR R0,[R2]  | STR R0,[R2]  ;
 DMB ISH      | DMB ISH      ;
 LDR R1,[R3]  | LDR R1,[R3]  ;
exists (0:R1=0 /\ 1:R1=0)
```


maranget:
Timely addition, thanks.

PR #387, opened 2022-06-01
[herd] add support for asymmetric MTE fault variant

benbancroft:
MTE provides an asymmetric mode for detecting tag check failures. In particular, when such a mode is configured, only tag checks for load operations are handled in a synchronous (precise) fashion, whereas stores happen asynchronously (imprecise).

This has implications on memory ordering, in that tag checks on loads need to abort execution on the given instruction, but the checks on stores complete asynchronously and the tag check may yet fail at some later time. This can result in different output states from a litmus test running in asymmetric mode depending on the orderings of loads/stores.

An example test that shows such properties is the following:

```
AArch64 MTE-asymmetric
{
0:X0=y:red; 0:X4=x:red; 0:X5=x:green; 0:X6=y:green;
}
 P0                 ;
 MOV W2,#1          ;
 STR W2,[X5]        ;
 STR W2,[X0]        ;
 LDR W3,[X6]        ;
 LDR W1,[X4]        ;
exists 0:X3=1 /\ 0:X1=0 /\ fault(P0,y) /\ fault (P0,x)
```

The following test is interesting, in that asymmetric mode allows the first two stores to retire before their tag check completes, but the subsequent tag checks on the final LDR of x will always give a synchronous fault. We can use the LDR on y to observe that the store has gone through successfully, but it also has another useful property in that the litmus test returns a single unique state per MTE fault mode.

The following pull requests implements the required logic to model asymmetric faults in herd7. Running the above litmus test through herd7 in the three modes (including the new asymmetric one) gives the following states:

Asymmetric (-variant memtag,asym):

```
States 1
0:X1=0; 0:X3=1; Fault(P0,x:red); Fault(P0,y:red);
Ok
Witnesses
Positive: 1 Negative: 0
```

Asynchronous (-variant memtag,async):

```
States 1
0:X1=1; 0:X3=1; Fault(P0,x:red); Fault(P0,y:red);
No
Witnesses
Positive: 0 Negative: 1
```

Synchronous (-variant memtag,sync):

```
States 1
0:X1=0; 0:X3=0; Fault(P0,y:red); ~Fault(P0,x);
No
Witnesses
Positive: 0 Negative: 1
```

As you can see, the condition clause is constructed in such a way in that the litmus test is only positive when ran in asymmetric mode.

This pull request also includes a commit to support optional semantics for the AArch64 naming of the MTE fault modes, allowing for sync/synchronous to be used in the place of the precise variant and async/asynchronous to be used in the place of imprecise. My main justification behind this is because the generic name that I came up with of loadsfatal is a bit ambiguous, whereas asymmetric at least matches the architectural name and therefore it seemed to make sense to add support for the equivalent for the other tag check modes. Happy to discuss any better ideas.

### DIFF
--- a/herd/AArch64Sem.ml
+++ b/herd/AArch64Sem.ml
@@ -800,29 +800,32 @@ module Make
   addresses. Thus the resulting monads will possess
   extra dependencies w.r.t the simple case.
  *)
-      let lift_fault mfault mm = (* This is memtag fault handling *)
+      let lift_fault mfault mm dir = (* This is memtag fault handling *)
         let open Precision in
         match C.precision with
         | Fatal -> mfault >>! B.Exit
+        | LoadsFatal -> (match dir with
+            | Dir.R | Dir.F -> mfault >>! B.Exit
+            | Dir.W -> (mfault >>| mm) >>= M.ignore >>= B.next1T)
         | Skip -> Warn.fatal "Memtag extension has no 'Skip' fault handling mode"
         | Handled ->
            (mfault >>| mm) >>= M.ignore >>= B.next1T
 
-      let lift_memtag_phy mop a_virt ma ii =
+      let lift_memtag_phy dir mop a_virt ma ii =
         M.delay_kont "4" ma
           (fun _ ma ->
             let mm = mop Access.PHY ma in
             delayed_check_tags a_virt ma ii
               (mm  >>= M.ignore >>= B.next1T)
-              (lift_fault (mk_fault a_virt ii None) mm))
+              (lift_fault (mk_fault a_virt ii None) mm dir))
 
-      let lift_memtag_virt mop ma ii =
+      let lift_memtag_virt dir mop ma ii =
         M.delay_kont "5" ma
           (fun a_virt ma  ->
             let mm = mop Access.VIR (ma >>= fun a -> loc_extract a) in
             delayed_check_tags a_virt ma ii
               (mm  >>= M.ignore >>= B.next1T)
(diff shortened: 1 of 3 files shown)


maranget:
Hi @benbancroft. I see that the `-variant asym` fault mode is also implemented for `kvm` mode. Does it make sense ? If so, litmus should also implement this variation on fault handling , which requires making the difference bwteen read and writes inside the fault handler, Otherwise, it is perhaps cleaner to fail in `lift_kvm` we mode is `-variant kvm,asym`, as we do for mode `-variant memtag,skip`.

benbancroft:
Hi @maranget, thanks for looking over the pull request. I'm not at all familiar with `kvm` mode unfortunately, I personally only use the generator and herd7 for my own use cases. The code I added in `lift_kvm` was just to suppress the compiler warning as result of the new type I introduced for `LoadsFatal`, like you say it may make more sense to just fail in such cases. I will probably look to sync up with @relokin next week to discuss. Not sure how litmus7 implements its fault handling (or if it currently has handling for async faults), but as I have already implemented support in our own litmus runner for async/asymmetric faults I may have some input on that.

maranget:
Nice thanks, @relokin's and your expertise on fault handling in litmus kvm and your litmus runner will hopefully combine. 

That said, having both herd and litmus to  on `-variant kvm,asym`  is an acceptable, even if temporary, solution if we want to merge this PR swiftly.

Also notice that litmus code is not as robust as I would have expected since it compiled without warning. I'll take care of this.

maranget:
Merge by hand, because of conflict. Thanks.

PR #389, opened 2022-06-01
[herd] Preserve instruction annotations for FAULT events

relokin:
In some cases, we need to know if the instruction that generated a
FAULT event had any annotations (for example, was a release
instruction). For this reason, this change adds the annotation
information to the Fault action. Subsequently, it makes sure that herd
considers Fault actions to compute eventsets for architecturally
defined annotations. For good measure, and since we're changing the
Fault action, this change also adds the direction (e.g., read or
write) to it. However, for now we don't do anything with it.

### DIFF
--- a/herd/machAction.ml
+++ b/herd/machAction.ml
@@ -56,7 +56,7 @@ module Make (C:Config) (A : A) : sig
 (* Atomic modify, (location,value read, value written, annotation *)
     | Amo of A.location * A.V.v * A.V.v * A.lannot * A.explicit * MachSize.sz * Access.t
 (* NB: Amo used in some arch only (e.g., Arm, RISCV) *)
-    | Fault of A.inst_instance_id * A.location * string option
+    | Fault of A.inst_instance_id * A.location * Dir.dirn * A.lannot * string option
 (* Unrolling control *)
     | TooFar of string
 (* TLB Invalidate event, operation (for print and level), address, if any.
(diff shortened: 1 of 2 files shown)


maranget:
> Subsequently, it makes sure that herd
> considers Fault actions to compute eventsets for architecturally
> defined annotations.

Hi @nikos, I do not understand the above sentence, nor found code that would compute those sets. Is there something I misunderstand here (I am not sure about the exact meaning of "Subsequently").

herd/machAction.ml
```diff
@@ -487,7 +489,8 @@ end = struct
       List.map
         (fun (tag,p) ->
           let p act = match act with
-          | Access(_,_,_,annot,_,_,_)|Amo (_,_,_,annot,_,_,_) -> p annot
+          | Access(_,_,_,annot,_,_,_)|Amo (_,_,_,annot,_,_,_)
+          | Fault (_,_,_,annot,_)-> p annot
```
relokin, line 493: If I understand correctly, this change will add `Fault` in addition to `Access` and `Amo` to `asets`.

relokin: Hi @maranget, thanks for reviewing this change!

I added a comment to the line which I think implements the change. But I might be missing something, please let me know if I do.

maranget: I have read to code too quickly. The code works as advertised, as I checked on some picture with option `-doshow L` for displaying the `L` (Release) set.

maranget: Thanks, this is the product of many rewritings. If only the whole system was like that :)

maranget:
Ready for merge, from my point of view. Will merge in a few hours if no objection.

jalglave:
Thank you both for this. Please do go ahead.
Jade

On 3 Jun 2022, at 12:39, Luc Maranget ***@***.***> wrote:

﻿

⚠ Caution: External sender


Ready for merge, from my point of view. Will merge in a few hours if no objection.

maranget:
Merged, thanks.

PR #397, opened 2022-06-15
[gen]: Replace atomic_load gen with atomic_load_explicit

lukeg101:
`atomic_load(x)` is a shortcut for
`atomic_load_explicit(x,memory_order_seq_cst)`

diy generates `atomic_load(x)` calls, however herd does not accept
`atomic_load`, erroring with a message "Macro call atomic_load in CSem"

This patch ensures diy generates `atomic_load_explict(x, seq_cst)` instead
of `atomic_load` so that there is a greater match of tests generated by
diy and tests accepted by Herd

### DIFF
--- a/gen/CCompile_gen.ml
+++ b/gen/CCompile_gen.ml
@@ -857,8 +857,6 @@ module Make(O:Config) : Builder.S
         | AddZero (e,loc) ->
             sprintf "%s + (%s & 128)"
               (dump_exp e) (dump_loc_exp loc)
-        | AtomicLoad (MemOrder.SC,loc) ->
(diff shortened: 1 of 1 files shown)


lukeg101:
Hi Luc,
Yes it ensures both explicit atomic loads and stores are generated. I'll amend the commit message to reflect this

maranget:
Observe that herd accepts `atomic_load` and `atomic_store` when given option `-macro c11.def`. This option, and others, are selected when herd is given the flag `-c11`.

I'll merge the PR, as it improves tools compatibility. However using option `-c11` is quite convenient because of the other settings, which  mostly define adequate parameters for pictures.

PR #398, opened 2022-06-15
[lib]: AArch64 fix typo for ADDR -> ADR

lukeg101:
The `ADR` instruction is pretty-printed as `ADDR`.

This patch removes the erroneous D, and updates the hashes
of existing tests that are influenced by this


lib/AArch64Base.ml
```diff
@@ -1290,7 +1290,7 @@ let do_pp_instruction m =
   | I_OP3 (v,op,r1,r2,kr, s) ->
       pp_rrkr (pp_op op) v r1 r2 kr ^ pp_barrel_shift "," s (m.pp_k)
   | I_ADDR (r,lbl) ->
```
relokin, line 1292: Thanks Luke. While you're at it, is it worth fixing this as well?

lukeg101: I'm struggling to see what is wrong on line 1291 (I checked the parser and instruction set for a candidate instruction that matches this pattern [1] too), are you referring to the name of the instruction `I_ADDR`?

[1] https://developer.arm.com/documentation/dui0802/a/A64-General-Instructions/ADD--immediate-

relokin: Yes I think `I_ADDR` has the same typo and ideally, it should be `I_ADR`

maranget:
Changing hashes is problematic, because it may prevent comparing new logs (typically herd logs) with old ones (typcally litmus logs that we have been accumulating for years).

That said, it may be no problem here as `ADR` is seldom present in our tests if it is. I need to check that before merging this PR.

lukeg101:
Hi Luc, Thanks for providing that patch. I have fixed the hash regressions and applied your patch.

lukeg101:
All fixed

maranget:
Thanks, merged

PR #399, opened 2022-06-15
[all]: AArch64 Added LDRB/H and STRB/H shifters

lukeg101:
This patch adds support for inline barrel shifters
for LDRB/H and STRB/H barrel shifters for all tools
that use these instructions.

tests provided.

### DIFF
--- a/lib/AArch64Parser.mly
+++ b/lib/AArch64Parser.mly
@@ -243,16 +243,20 @@ kr:
 | xreg { A.RV (A.V64,$1) }
 | wreg  { A.RV (A.V32,$1) }
 
-kr_shift:
+kr_shift_address:
 | k                { A.K $1, A.S_NOEXT }
 | xreg             { (A.RV (A.V64,$1)),  A.S_NOEXT }
-| wreg             { (A.RV (A.V32,$1)),  A.S_NOEXT }
 | wreg COMMA shift { (A.RV (A.V32, $1)), $3 }
 | xreg COMMA shift { (A.RV (A.V64, $1)), $3 }
 
+kr_shift:
+| kr_shift_address { $1 }
+| wreg { (A.RV (A.V32,$1)),  A.S_NOEXT }
+
+(* For address argument only *)
 kr0:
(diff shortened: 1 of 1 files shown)


maranget:
LGTM, except one important issue that originates in a long standing shortcoming  that we are going to solve now, thanks to your initiative.

herdtools accept the illegal addressing mode `[Xn,Wm]` (register indexed with 32bit register, no extension). As this mode is rejected by the assembler, litmus prints this mode as `[Xn,Wm,SXTW]`. For consistency, (!) herd does the same when printing the illegal addressing mode. Then, everything works fine: the assembler is happy, `diy` generates the correct syntax, and herd accepts the incorrect syntax (without extension) as implicitly specifying signed extension.

Let clean the dust under the carpet:
 + The parser should reject the incorrect syntax
  + diy generator should emit the signed extension

I'll suggest some patches, but more probably has to be done. For instance some `assert false` will have to be added in some places, instead of performing the silent replacement of no extension by sxtw.

maranget:
And here is a patch for rejecting `[Xn,Wm]` at parse time. This is a minimal patch that does not attempt to solve all untimely parsing of illegal constructions... As I am quite sure that there are others.
[B.patch.txt](https://github.com/herd/herdtools7/files/8920534/B.patch.txt)

lukeg101:
Hi Luc, thanks for taking a look at this! Is there anything I can do to help?

lukeg101:
Hi Luc, I'm doing this now. Do you have advice for wording of the commit for the patches you provided?
I have merged A and B into one patch and rebased it onto the original PR (I will push after I have added tests)

maranget:
> Hi Luc, I'm doing this now. Do you have advice for wording of the commit for the patches you provided?
> I have merged A and B into one patch and rebased it onto the original PR (I will push after I have added tests)

What about `[gen] command sign extension when generating "W" register indexed accesses` and `[all] reject "W" register indexed accesses at parsing` ? (English mistakes fixed, if any)

maranget:
Merged, thanks for your work and initiative.

PR #400, opened 2022-06-23
Reject list

hugookeeffe:
This Pull request implements a new functionality into diy7 to include a 'reject list'. This allows the user to specify a relaxation or combinations of relaxations to exclude from generation. The syntax to use the reject list is the same as that for the relax list.
example reject list:
`-rejectlist Amo.Swp Amo.StAdd [DpAddrdR, Amo.Cas]`

Some small tweaks need to be added to this pull request still, however the code functions as intended.


gen/alt.ml
```diff
@@ -360,11 +360,34 @@ module Make(C:Builder.S)
...
+      let rsuff = List.concat rsuff in
+      let r_first = List.hd rsuff in
+      let rec f rej= match rej with
+        | h::t -> if (List.hd h) = r_first then begin
+          let truth = check2 (List.tl rsuff) (List.tl h) in
+          if truth then f t else false
+          end
+          else f t
+        | [] -> true in
+      f rl
```
relokin, line 382: Given the above definition, I think you can do this to check the suffix across all elements in rl
```suggestion
      not (List.exists (fun rl -> is_suffix rl rsuff) rl)
```

relokin (changes requested):
hey @hugookeeffe, I think this looks good, just a few comments mostly simplifying the code a little.

And two questions, do we need to worry about: 
1. the prefix as well (gen/alt.ml:401)?
2. `prefix @ suffix`?

chacon01:
Hey @hugookeeffe, thanks for implementing this. I'm seeing a couple cases of it not working as I expect, which I believe is due to annotations modifying earlier relaxations in the cycle.

I tried
```
diy7 -arch AArch64 -nprocs 2 -size 5 -exact -relax Rfe Fre Coe PodWWPL PodRWPL PodRWAP PodRRAP -mix true -rejectlist PodRRAA
```
and of the 17 test produced by the generator, 4 included the `PodRRAA` relaxation:
* `PodRWPL RfeLA PodRRAA PodRWAP Rfe`
* `PodWWPL RfeLA PodRRAA PodRWAP Coe`
* `PodWWPL RfeLA PodRRAA PodRRAP Fre`
* `PodRWAP RfePA PodRRAA PodRWAP RfePA`

In each of these cases, it looks like `PodRRAP` combined with `PodRWAP` to form `PodRRAA PodRWAP` but the checking for the rejection didn't consider that `PodRWAP` would change the annotations of the `PodRRAP`.

maranget:
Hi @hugookeeffe, the implementation does not completely eliminate rejected relaxation. The following command should not generate the `R` test:
```
% diy7 -arch AArch64 -safe Fre,PodWW,Coe,PodWR -rejectlist [Coe,PodWR] -size 4
Generator produced 3 tests
```
Namely `R` cycle is `PodWW Coe PodWR Fre`, whcih contains the rejected sequence `Coe PodWR`.

I guess the reason is that `R` cycle is generated as the following prefixes  

1. `PodWR`
2. `PodWR Fre`
3. `PodWR Fre PodWW`
4. `PodWR Rfe PodWW Coe` that is turned into a cycle as the last and first relaxations are compatible.

The current implementation does not catch the case where the rejected sequence is partitioned between end and start of last prefix. Clearly some additional check is called for at the time the prefix is considered as a cycle..

maranget:
Well my explanation may be wrong as regards details, because the cycle is built from suffixes of increasing length and not suffixes. However the general idea looks correct: rejected sequence have  to be checked for on the complete cycle.

maranget:
Commit b78c76e8574177c3d9f634223c6b802e4e433ae7 is a fix  attempt.

hugookeeffe:
> Commit [b78c76e](https://github.com/herd/herdtools7/commit/b78c76e8574177c3d9f634223c6b802e4e433ae7) is a fix attempt.
Hi @maranget, Sorry for the late reply! I have tested your commit and it works for me so I have pushed it into the branch. 
Thank you very much for this.

In regards to the question @chacon01 raised, how could I implement a check so that annotations that are modified after generating the cycle can also be rejected?
Many thanks,
Hugo

maranget:
> In regards to the question @chacon01 raised, how could I implement a check so that annotations that are modified after generating the cycle can also be rejected?

The complete cycle, with annotations resolved is available as an argument to function `last_minute` that performs sone _"last minute"_ checks.

More precisely the argument is a list of list of edges `ess`, each sublist grouping the edges per thread, Hence the cycle itself is  easily reconstructed as  `List.flatten ess`. Then, provided `last_minute` is given the rejection list as an extra (first) argument, one can perform a complete tests of each rejected list not being a substring of the cycle. The implementation of the check would be a combination of the two checks already implemented. It may be adequate to _replace_ those two checks by the new _"last minute"_ check. Namely,  it looks cleaner and more simple that the rejection check applies to the final cycle.

maranget:
> It may be adequate to _replace_ those two checks by the new _"last minute"_ check. Namely, it looks cleaner and more simple that the rejection check applies to the final cycle.

Hum I am not sure about replacing the old checks by the new one. I now tend to keep all checks. Namely the first two checks capture something different: they allow to reject substrings with annotation unresolved such as `[Coe,L,PodWR]` for instance. In that latter case, one can reject the sequence `Coe,L,PodWR` regardless of the first an last annotations og the sequence. This is we reject `CoePL,PodWRLP`; `CoeLL,PodWRLP`; etc.

maranget:
Merged by hand and close. Thanks @hugookeeffe.

PR #401, opened 2022-07-01
[litmus] Add support for custom fault handlers

relokin:
This PR adds support for custom fault handlers in litmus. The fault handler appears in the litmus test as a separate thread. For example:

```
AArch64 MP+dmb.st+RfaultR
Variant=precise
{
[PTE(z)]=(oa:PA(z),valid:0);

0:X1=x; 0:X3=y;
1:X1=y; 1:X3=x; 1:X4=z;
}
 P0          | P1           | P1.F        ;
 MOV W0,#1   | LDR W0,[X1]  | LDR W5,[X3] ;
 STR W0,[X1] |L0:           |             ;
 DMB ST      | LDR W2,[X4]  |             ;
 MOV W2,#1   |              |             ;
 STR W2,[X3] |              |             ;
exists(1:X0=1 /\ 1:X5=0)
```

The fault handler uses the same registers as the main thread. As a result, the initialization can set registers for the fault handler and the post-condition can specify tests for the values of registers as a result of the execution of the fault handler. For the time being and due to the way kvm-unit-tests sets fault (exception) handlers, litmus can supports tests with only one fault handler.

### DIFF
--- a/lib/miscParser.mli
+++ b/lib/miscParser.mli
@@ -17,7 +17,8 @@
 (** The basic types of architectures and semantics, just parsed *)
 
 (* Processor name with optional annotations *)
-type proc = int * string list option
+type func = Main | FaultHandler
+type proc = Proc.t * string list option * func
 
 val pp_proc : proc -> string
 
--- a/herd/loader.ml
+++ b/herd/loader.ml
@@ -77,7 +77,7 @@ struct
   let load prog =
     let rec load_iter num = function
     | [] -> Label.Map.empty,[],IntMap.empty,num
-    | ((proc,_),code)::prog ->
+    | ((proc,_,_),code)::prog ->
       let addr = 1000 * (proc+1) in
       let mem,starts,rets,new_num = load_iter num prog in
       let fin_mem,start,fin_rets,fin_num = load_code proc addr mem rets new_num code in
--- a/herd/test_herd.ml
+++ b/herd/test_herd.ml
@@ -165,7 +165,7 @@ module Make(A:Arch_herd.S) =
       let proc_info =
         let m =
           List.fold_left
-            (fun m ((p,ao),_) -> match ao with
+            (fun m ((p,ao,_),_) -> match ao with
             | None -> m
             | Some ans ->
                 List.fold_left
--- a/jingle/cDumper.ml
+++ b/jingle/cDumper.ml
@@ -145,7 +145,7 @@ let extract_decl init i prog =
 
 let code init prog =
   let open CAst in
-  List.map (fun ((i,_),p) ->
+  List.map (fun ((i,_,_),p) ->
 	    let params = get_params init i in
 	    let decls =  extract_decl init i (unwrap_pseudo p)
 	    in Test { proc = i;
--- a/jingle/mapping.ml
+++ b/jingle/mapping.ml
@@ -231,7 +231,7 @@ module Make(C:Config) = struct
     (pseudo_p,env)
 
   let reg_mapping =
-    List.map (fun ((i,_),(b,_)) ->
+    List.map (fun ((i,_,_),(b,_)) ->
       (i,
        (List.map (fun (sr,tr) ->
          (Source.pp_reg sr,Target.pp_reg tr))
@@ -239,7 +239,7 @@ module Make(C:Config) = struct
 
   let addr_init =
     let open MiscParser in
-    List.fold_left (fun acc ((i,_),(b,_)) ->
+    List.fold_left (fun acc ((i,_,_),(b,_)) ->
       acc@
       (List.map (fun (sa,tr) ->
         (Location_reg(i,Target.pp_reg tr),
--- a/lib/AArch64Lexer.mll
+++ b/lib/AArch64Lexer.mll
@@ -484,6 +484,8 @@ rule token = parse
 | '#'? ('-' ? num as x) { NUM (int_of_string x) }
 | 'P' (num as x)
     { PROC (int_of_string x) }
+| 'P' (num as x) ".F"
+    { PROCFH (int_of_string x) }
 | ['w''W']'%' (name as name) { SYMB_WREG name }
 | ['x''X']?'%' (name as name) { SYMB_XREG name }
 | ['c''C']?'%' (name as name) { SYMB_CREG name }
--- a/lib/AArch64Parser.mly
+++ b/lib/AArch64Parser.mly
@@ -108,7 +108,7 @@ let check_op3 op kr =
 %token STG STZG LDG
 %token ALIGND ALIGNU BUILD CHKEQ CHKSLD CHKTGD CLRTAG CPY CPYTYPE CPYVALUE CSEAL
 %token LDCT SEAL STCT UNSEAL
-%type <(int * string list option) list * (AArch64Base.parsedPseudo) list list * MiscParser.extra_data> main
+%type <MiscParser.proc list * (AArch64Base.parsedPseudo) list list * MiscParser.extra_data> main
 %type <AArch64Base.parsedPseudo list> instr_option_seq
 
 %start  main
--- a/lib/ARMParser.mly
+++ b/lib/ARMParser.mly
@@ -51,7 +51,7 @@ semi_opt:
 
 proc_list:
 | ps=separated_nonempty_list(PIPE,PROC) SEMI
-    { List.map (fun p -> p,None) ps }
+    { List.map (fun p -> p,None,MiscParser.Main) ps }
 
 iol_list :
 |  instr_option_list SEMI
--- a/lib/CGenParser_lib.ml
+++ b/lib/CGenParser_lib.ml
@@ -167,7 +167,7 @@ module Do
             fname in
         List.map (L.macros_expand ms) in
 
-    let prog =  List.map (fun p -> (p.CAst.proc,None),expand_body p.CAst.body) prog in
+    let prog =  List.map (fun p -> (p.CAst.proc,None,MiscParser.Main),expand_body p.CAst.body) prog in
 (*
     List.iter
       (fun (p,code) ->
--- a/lib/MIPSParser.mly
+++ b/lib/MIPSParser.mly
@@ -52,7 +52,7 @@ semi_opt:
 
 proc_list:
 | ps=separated_nonempty_list(PIPE,PROC) SEMI
-  { List.map (fun p -> p,None) ps }
+  { List.map (fun p -> p,None,MiscParser.Main) ps }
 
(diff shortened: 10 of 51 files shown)


relokin:
@maranget if you would like to give it a spin, you'll need to patch `kvm-unit-tests` with this [0001-arm64-Make-default-vector-functions-weak.patch.txt](https://github.com/herd/herdtools7/files/9038725/0001-arm64-Make-default-vector-functions-weak.patch.txt). If we settle on this PR I will try to upstream it.

maranget:
Hi @relokin. I have started testing the new feature. I am leaving for a few days and will test more when I am back. For now:
  + Option `-driver C` was not working properly. Here is a commit that (hopefully) fixes the problem:  0ec521d3fcfbd38b606a19ebf7b937a2d4953af7. Feel free to modify it, if not totally appropriate.
  + Something is not working in user mode: here is a small test whose output is `0:X0=0; 0:X2=0;`, while expected output is `0:X0=0; 0:X2=1;`:
```
AArch64 User
Variant=fatal
EL0=P0
{
[PTE(z)]=(valid:0);
x=1;
z=2;
0:X1=z;
0:X3=x;
}
 P0          | P0.F        ;
L0:          |             ;
 LDR W0,[X1] | LDR W2,[X3] ;

forall 0:X0=0 /\  0:X2=1
```
Notice that we get the expected output when the line `EL0=P0` is deleted.

As regards user mode test, I do not know what is happening. I see two potential difficulties:
  + The handler code attempts to [execute at EL0] (https://github.com/herd/herdtools7/blob/0ec521d3fcfbd38b606a19ebf7b937a2d4953af7/litmus/compile.ml#L625). This does not look very natural to me.
  + If I remember well, some change of stack may be involved.

maranget:
Also I have thought a bit about having several handlers. As I understand it, a synchronous exception is handled by jumping to a specific, unique, address. We could dedicate a register to hold the thread number and the handler code would perform some indirect jump. We could also use the function `current_thread_info` and the `whoami` data structure to retrieve the thread number, as the current code does in function `fault_handler`. However, I am afraid that this solution would   impact the memory model mode than the simpler solution of reserving one register for the thread number.

relokin:
> Hi @relokin. I have started testing the new feature. I am leaving for a few days and will test more when I am back. For now:

Many thanks for reviewing this.

> 
>     * Option `-driver C` was not working properly. Here is a commit that (hopefully) fixes the problem:  [0ec521d](https://github.com/herd/herdtools7/commit/0ec521d3fcfbd38b606a19ebf7b937a2d4953af7). Feel free to modify it, if not totally appropriate.
> 

I'll have a closer look. I think the problem with `-driver C` might be more fundamental. Isn't the point of `-driver C` to compile more than one litmus test into the same binary? The limitation of this PR is that it overrides the fault handler for for `el1h_sync` exceptions. That's across the whole test. So we can have one fault handler per binary. 

>     * Something is not working in user mode: here is a small test whose output is `0:X0=0; 0:X2=0;`, while expected output is `0:X0=0; 0:X2=1;`:
> 
> 
> ```
> AArch64 User
> Variant=fatal
> EL0=P0
> {
> [PTE(z)]=(valid:0);
> x=1;
> z=2;
> 0:X1=z;
> 0:X3=x;
> }
>  P0          | P0.F        ;
> L0:          |             ;
>  LDR W0,[X1] | LDR W2,[X3] ;
> 
> forall 0:X0=0 /\  0:X2=1
> ```
> 
> Notice that we get the expected output when the line `EL0=P0` is deleted.
> 
> As regards user mode, I do not know what is happening. I see two potential difficulties:
> 
>     * The handler code attempts to execute at EL0: https://github.com/herd/herdtools7/blob/0ec521d3fcfbd38b606a19ebf7b937a2d4953af7/litmus/compile.ml#L625
>       . This does not look very natural to me.
> 
>     * If I remember well, some change of stack may be involved. Also the handler address may not be the same as the one used from EL1.

I will need to look into this, I didn't even test with EL0 but it is certainly a feature we want to support.

relokin:
> Also I have thought a bit about having several handlers. As I understand it, a synchronous exception is handled by jumping to a specific, unique, address. We could dedicate a register to hold the thread number and the handler code would perform some indirect jump. We could also use the function `current_thread_info` and the `whoami` data structure to retrieve the thread number, as the current code does in function `fault_handler`. However, I am afraid that this solution would impact the memory model mode than the simpler solution of reserving one register for the thread number.

I think there is a way around this. But I wanted to have something working as soon as possible. Architecturally, we have one `el1h_sync` handler per core. `kvm-unit-tests` sets uses the same handler, in fact the same handler across all exceptions. There is nothing that prevent us from changing it. But for that we would have to "override" the [`exceptions_init`](https://gitlab.com/kvm-unit-tests/kvm-unit-tests/-/blob/master/arm/cstart64.S#L268) function and the [`vector_table`](https://gitlab.com/kvm-unit-tests/kvm-unit-tests/-/blob/master/arm/cstart64.S#L373).

This might be still a problem for tests where the roles of cores change. The only solution to that would be to have one level of indirection with some implications to the semantics of exception entry but I think this is the best we can do.

herd/AArch64Sem.ml
```diff
@@ -840,15 +840,7 @@ module Make
...
-          insert_commit_to_fault ma (fun _ -> mk_fault a dir an ii None) ii
-          >>!
-            begin
-              let open Precision in
-              match C.precision with
-              | Fatal -> B.Exit
-              | Skip -> B.Next []
-              | Handled -> B.ReExec
-            end in
+          insert_commit_to_fault ma (fun _ -> mk_fault a dir an ii None) ii >>! B.Fault in
```
relokin, line 843: This is where we would ideally set `ELR_EL1` but then when I try and read `ELR_EL1` to use it in the subsequent `ERET` the value is not determined until the solver is invoked. @maranget, is there a way to get an early solution for these trivial cases?

maranget:
Hi @relokin, what about having two PR, one for litmus and the the other for herd?  My understanding is that the litmus PR is almost ready, with some explicit limitations : no `-driver C`, no EL0. Your thoughts?

relokin:
Yes, that's a good idea. I'll do the split now.

lib/miscParser.mli
```diff
@@ -17,7 +17,8 @@
 (** The basic types of architectures and semantics, just parsed *)
 
 (* Processor name with optional annotations *)
-type proc = int * string list option
+type func = Main | FaultHandler
```
relokin, line 20: @maranget is this the right place to declare the func type? I was between miscParser and proc.mli.

maranget: I have no strong opinion on this...

maranget:
BTW I have attempted to implement the custom fault handler for `-driver C`. The idea is to add an indirection : the code labeled by `el1h_sync` reads a memory cell, say `jump_here` and performs a branch to the address read. Each test updates the memory cell `jump_here` with the appropriate address (the one labeled `el1h_sync` in this PR). I have not succeeded.

maranget:
My attempt is here https://github.com/herd/herdtools7/tree/PR401 (last commit)

relokin:
> My attempt is here https://github.com/herd/herdtools7/tree/PR401 (last commit)

I'll have a look.



> BTW I have attempted to implement the custom fault handler for `-driver C`. The idea is to add an indirection : the code labeled by `el1h_sync` reads a memory cell, say `jump_here` and performs a branch to the address read. Each test updates the memory cell `jump_here` with the appropriate address (the one labeled `el1h_sync` in this PR). I have not succeeded.

In principle, this sounds something that should work. But I am worried about the semantics, we'll be adding a load and an additional branch.

Wouldn't it better to try and redirect the `vbar_el1` to a new `vector_table` for each test? 

Each test will define it's own vector_table:
```
  .align 11
 vector_table_test$n:
          ventry  el1t_sync                       // Synchronous EL1t
          ventry  el1t_irq                        // IRQ EL1t
          ventry  el1t_fiq                        // FIQ EL1t
          ventry  el1t_error                      // Error EL1t
 
          ventry  el1h_sync_test$n           // Synchronous EL1h
          ventry  el1h_irq                        // IRQ EL1h
          ventry  el1h_fiq                        // FIQ EL1h
          ventry  el1h_error                      // Error EL1h
 
          ventry  el0_sync_64                     // Synchronous 64-bit EL0
          ventry  el0_irq_64                      // IRQ 64-bit EL0
          ventry  el0_fiq_64                      // FIQ 64-bit EL0
          ventry  el0_error_64                    // Error 64-bit EL0
 
          ventry  el0_sync_32                     // Synchronous 32-bit EL0
          ventry  el0_irq_32                      // IRQ 32-bit EL0
          ventry  el0_fiq_32                      // FIQ 32-bit EL0
          ventry  el0_error_32                    // Error 32-bit EL0
```
and its own el1h_sync_test$n handler

```
el1h_sync_test$n:
....
```
and before we start the test we reset the `vbar_el1` with:
```
  exceptions_init:
          adrp    x4, vector_table_test$n
          add     x4, x4, :lo12:vector_table_test$n
          msr     vbar_el1, x4
          isb
          ret
```

maranget:
Changing the vector table for each test looks a much better idea. I suppose that all cores should execute `exception_init` for each test.

relokin:
> Changing the vector table for each test looks a much better idea. I suppose that all cores should execute `exception_init` for each test.

Exactly! And that would also allow us to use a different el1h_sync handler for each core. Right now, it's shared across all cores in the system.

maranget:
Hum, not easy to decide. My attempt involves quite a lot of modifications and integrating them in this PR would make it quite huge. So I guess it would be easier first to merge this PR 401, and then to elaborate. Your opinion?

relokin:
I think this PR would allow us to run many (if not the majority) of the tests we have in mind. Considering that, I would vote for addressing any concerns with this PR merge it and then look at supporting more features.

Unless ofc you don't like the overall approach we take in this PR.

maranget:
Ok I am also in favor of merging this PR. Do you think it is mergeable now or do you intend to work more more on it?

maranget:
Hi @relokin, let's merge. To sum it up here are some of the limitations of the current state:
 + One handler at most
 + Not compatible with `-driver C`
 + Not functional for EL0 threads

See more?

relokin:
> Hi @relokin, let's merge. To sum it up here are some of the limitations of the current state:
> 
>     * One handler at most
> 
>     * Not compatible with `-driver C`
> 
>     * Not functional for EL0 threads
> 
> 
> See more?

This is a good summary, I am not aware of any other limitation.

relokin:
Thanks @maranget !

PR #402, opened 2022-07-06
Fixing a divergence in alternative aarch64 cat files

artkhyzha:
Alternative aarch64 cat files are out of sync with recent changes to the main cat file related to dependencies. This PR proposes bringing them in sync by copying the relevant parts of `aarch64.cat` into `armv8-common.cat`.

An evidence of the divergence can be produced by running the tests from `catalogue/aarch64/tests` using `mixed/ec.cat`. It is going to produce negative differences w.r.t. `aarch64.cat` on three tests: `CAS+data1`, `CAS+data2`, `MP+rel+CAS-addr`.


maranget:
LGTM and also syncing the models in the  `non-mixed` directory looks approprate to me. I'd like to have @jalglave word on this though.

jalglave:
Yes please do submit a similar fix for non-mixed.
Thank you both.
Jade
________________________________

jalglave:
hi @artkhyzha have you had a chance to submit a fix for non-mixed models as well please?
thanks,
Jade

jalglave:
ping @artkhyzha on non-mixed models?

artkhyzha:
@relokin -- I have updated the EC and EGC cat files. They seem plausibly up-to-date and pass some quick tests, but still need full regressing, of course.

herd/libdir/arm-models/mixed/armv8-util.cat
relokin, line 1: It looks like this is now unused which is great since it means less divergence, should we remove the file altogether?

herd/libdir/arm-models/mixed/armv8-common.cat
```diff
@@ -39,56 +39,18 @@
...
-	| bob
-        | tob
-	| lob; lob
-
-(* Internal visibility requirement *)
-acyclic po-loc | ca | rf as internal
-
-(* Atomic: Basic LDXR/STXR constraint to forbid intervening writes. *)
-empty rmw & (fre; coe) as atomic
+let IM0 = ((IW * (M\IW)) | loc & ((W\FW) * FW))
```
relokin, line 56: I don't know if this is an intentional change but previously the loc was applied to both sides of the union operator:
```suggestion
let IM0 = loc & ((IW * (M\IW)) | ((W\FW) * FW))
```

herd/libdir/arm-models/mixed/ec.cat
```diff
@@ -40,17 +40,24 @@
 include "armv8-common.cat"
 include "sca.cat"
 
-let preorder-cb = IM0 | lob | scaob
+let preorder-cb = IM0 | local-hw-reqs | scaob
```
relokin, line 43: I am curious is there a reason why we include all of `local-hw-reqs`? Isn't `lob` the subset of `local-hw-reqs` that applies to `Exp` events only?

herd/libdir/arm-models/mixed/ec.cat
```diff
@@ -40,17 +40,24 @@
...
 let preorder-cb-lift = lift(MC,preorder-cb)
 
 with cb from linearisations(MC,preorder-cb-lift)
 ~empty cb
 
 let dcbl = delift(cb) & loc
-let rf-fwd = (W*R) & po-loc & dcbl^-1 \ intervening-write(po-loc)
-let rf-nfwd = (W*R) & dcbl \ intervening-write(dcbl | po-loc)
+let rf-fwd = (W*(Exp&R)) & po-loc & dcbl^-1 & ~intervening(W,po-loc)
+let rf-nfwd = (W*(Exp&R)) & dcbl & ~(intervening(W, (dcbl | po-loc)); [Exp & R]; po-loc?)
```
relokin, line 51: Did you qualify with `Exp` these to pass some tests? I am wondering if maybe the best thing would be to qualify all of them?

herd/libdir/arm-models/mixed/egc.cat
```diff
@@ -40,16 +40,27 @@
 include "armv8-common.cat"
 include "sca.cat"
 
-let gc-req = (W * _) | (R * _) & ((range(rfe) * _) | (rfi^-1; lob))
-let preorder-gcb = IM0 | lob & gc-req | scaob
+let gc-req = (W * _) | (R * _) & ((range(rfe) * _) | (rfi^-1; local-hw-reqs))
+let preorder-gcb = IM0 | local-hw-reqs & gc-req | scaob
```
relokin, line 44: Similar question: lob vs local-hw-reqs

herd/libdir/arm-models/mixed/ec.cat
relokin, line 51: This not equivalent to the previous formulation (due to the suffix `[Exp & R]; po-loc`, I am just marking it to make sure this is intentional.

relokin (commented):
Super thank you very much Artem. I've added a few comments let me know what you think!

herd/libdir/arm-models/mixed/armv8-common.cat
artkhyzha: As in, should be harmless, and we can keep the change or unroll it.

herd/libdir/arm-models/mixed/ec.cat
artkhyzha: As you suggested elsewhere, some clauses in `local-hw-reqs` apply to `Exp` too. (Notably, `CSE-ob`.) That is the technical reason for using `local-hw-reqs`.

Conceptually, we need to use all externally observable orderings that contribute to `ob`, but do not depend on `rf` and `co`. In that sense, I'd prefer `local-hw-reqs` over `lob | CSE-ob`.

herd/libdir/arm-models/mixed/ec.cat
artkhyzha: Good point. These qualifications are not necessary, since we do not intend to run these cat files with variants enabled. Both `R` and `W` in this case would always be `Exp`.

(If we were to run them with variants. We would have `rf` for `Exp&R`, `Imp&TTD&R`, `Imp&Instr&R` defined separately. `W` would have qualifications appropriate for the variant.)

artkhyzha: Indeed intentional: `coRR` shapes are now a part of the external visibility requirements, so appropriate compensation is necessary for external completion and external global completion.

herd/libdir/arm-models/mixed/armv8-util.cat
artkhyzha: Yes. I wonder what to do with `ev.cat` too: those are effectively old versions of `aarch64.cat`.

herd/libdir/arm-models/mixed/ec.cat
artkhyzha: Updated this by unpacking one clause into multiple -- as it is easier to explain them separately. Marking as resolved.

artkhyzha: That said, leaving `W` unqualified strikes me as future-proof, since with `-variant vmsa` Explicit Reads will be potentially reading from Hardware Updates. Leaving this unresolved in case if you prefer to add Exp everywhere or drop it everywhere, @relokin . (Should not affect regressions.)

herd/libdir/arm-models/mixed/armv8-common.cat
artkhyzha: On a second thought, having the new clause reduces the number of executions without sacrificing readability or correctness. I vote for keeping the new phrasing. Marking this resolved. (Please resurrect if disagree.)

herd/libdir/arm-models/mixed/armv8-util.cat
relokin: I think the intent might have been to have the 3 equivalent models together. But it's not worth doing the effort doing the effort to fix an older version of aarch64.cat right now.

herd/libdir/arm-models/mixed/ec.cat
relokin: I am mostly thinking what how we transliterate these clauses. Because the flags are not going to be transliterated. Let's leave this open. For now, your argument is very convincing and I agree with you it should cause any regressions.

artkhyzha:
Thank you for the feedback, @relokin . I think the cat files have now stabilized. After some testing, I am reasonably confident in them, but it would be good to regress them fully. Specifically:
    - `herd/libdir/arm-models/non-mixed/{ec,egc}.cat` should both regress on non-mixed baseline tests
    - `herd/libdir/arm-models/mixed/{ec,egc}.cat` should both regress on mixed-size baseline tests (but perhaps on non-mixed ones too?)
    
@maranget, would you have time to give those regressions a go?

maranget:
Hi @artkhyzha. Non-mixed size regression has run this week-end and is fine.  Mixed-size is running ( there was an unfortunate typo in some script). Please stay tuned.

maranget:
Mixed-size regression has completed. Everything is fine.

relokin:
@artkhyzha, if you rebase this PR, I think we can merge it.

PR #403, opened 2022-07-15
[all]: Add Load Signed Byte/Half Immediate/Register Instructions

lukeg101:
This patch adds support for the LDRSB/H AArch64 Instructions

### DIFF
--- a/herd/tests/instructions/AArch64/A24.litmus
+++ b/herd/tests/instructions/AArch64/A24.litmus
@@ -1,0 +1,11 @@
+AArch64 A24
+
+(* Tests ldrsb instruction *)
+{ int64_t 0:X0 =1; 0:X1 = x; int8_t x = 127; }
+
+P0;
+  LDRSB X0, [X1];
+  MOV X2, #1;
+  ADD X0, X0, X2;
+
+forall (0:X0=128 /\ [x]=127)
--- a/herd/tests/instructions/AArch64/A25.litmus
+++ b/herd/tests/instructions/AArch64/A25.litmus
@@ -1,0 +1,11 @@
+AArch64 A25
+
+(* Tests ldrsh instruction *)
+{ int64_t 0:X0 = 1; 0:X1 = x; int16_t x = 65535; }
+
+P0;
+  LDRSH X0, [X1];
+  MOV X2, #1;
+  ADD X0, X0, X2;
+
+exists (0:X0 = 65536)
--- a/herd/AArch64Arch_herd.ml
+++ b/herd/AArch64Arch_herd.ml
@@ -211,7 +211,7 @@ module Make (C:Arch_herd.Config)(V:Value.AArch64) =
       | I_LD4 (rs,_,_,_) | I_LD4R (rs,_,_) | I_ST4 (rs,_,_,_)
       | I_LD4M (rs,_,_) | I_ST4M (rs,_,_) ->
           Some (simd_mem_access_size rs)
-      | I_LDRBH (v,_,_,_,_) | I_LDARBH (v,_,_,_)
+      | I_LDRBH (v,_,_,_,_) | I_LDARBH (v,_,_,_) | I_LDRS (_,v,_,_)
       | I_STRBH (v,_,_,_,_) | I_STLRBH (v,_,_) | I_STXRBH (v,_,_,_,_)
       | I_CASBH (v,_,_,_,_) | I_SWPBH (v,_,_,_,_)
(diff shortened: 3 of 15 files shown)


maranget:
Hi @lukeg101. Those two instructions perform sign extension from 8/16 bits to 32/64bits. For instance if location '[x]' of size 1 byte holds -1 (or 0xFF), then the target register of instruction LDRSB will hold -1 (or 0xFFFFFFFF).

The following test illustrates this point:
```
% cat LDRSB.litmus
AArch64 LDRSB
{ 
  0:X1 = x;
  int8_t x = -1;
}
    P0                  ;
 LDRSB W0,[X1] ;
forall (0:X0=-1)
```
I have checked by running the above test with litmus: teh final value of `0:X0` is -1 as an int (and 0xFFFFFFFF as an `int64_t). However, with the code of this PR we have:
```
% herd7  LDRSB.litmus
est LDRSB Required
States 1
0:X0=255;
No
...
```

The root cause is the use of the sign extension primitive with the target size (here `Word`, _i.e._ 32bits) where the source size would be appropriate (here `Byte`, _i.e._ 8bits). See comments in code for a suggested patch.

herd/tests/instructions/AArch64/A24.litmus
```diff
@@ -0,0 +1,11 @@
...
+
+(* Tests ldrsb instruction *)
+{ 0:X0 = 1; 0:X1 = x; int8_t x = 127; }
+
+P0;
+  LDRSB X0, [X1];
+  MOV X2, #1;
+  ADD X0, X0, X2;
+
+exists (0:X0 = 129 /\ x = 0)
```
maranget, line 11: A successful final condition with a `forall` quantifier would be appreciated. For instance `forall 0:X0=128 /\ [x]=127`.

herd/tests/instructions/AArch64/A24.litmus
```diff
@@ -0,0 +1,11 @@
+AArch64 A24
+
+(* Tests ldrsb instruction *)
+{ 0:X0 = 1; 0:X1 = x; int8_t x = 127; }
```
maranget, line 4: For this test to be used by litmus without a warning, it is better to declare `0:X0` to be used as a 64bits register: `int64_t 0:X0 =1`. Alternatively, one may not declare any type (implicit type is `int`, 32 bits) and use the`W` form or register un the code.

herd/tests/instructions/AArch64/A25.litmus
```diff
@@ -0,0 +1,11 @@
+AArch64 A25
+
+(* Tests ldrsh instruction *)
+{ 0:X0 = 1; 0:X1 = x; int16_t x = 65535; }
```
maranget, line 4: Same suggestions as for test `A24`: either declare `0:X0` of type `int64_t` or use `W` registers in code.

maranget (changes requested):
I suggest the following changes
1. Fix semantics (see suggested patch)
2. Add tests that illustrate the difference between signed (_.e.g._ `LDRSB`) and unsigned (_.e.g._ `LDRB`) loads. Suggestions follow.

maranget:
Here are two tests that illustrate the difference between signed and unsigne loads:
```
AArch64 A26

(* 
{ 
  0:X1 = x;
  int8_t x = -1;
  int64_t 0:X4;
  int64_t 0:X6;
}

P0;
  LDRSB W0,[X1] ;
  MOV X4,X0     ;
  LDRB W2,[X1]  ;
  MOV X6,X2     ;
forall (0:X0=-1 /\ 0:X2=255 /\ 0:X4=4294967295 /\ 0:X6=255)
```
And
```
AArch64 A27

(* 
{ 
  0:X1 = x;
  int16_t x = -1;
  int64_t 0:X4;
  int64_t 0:X6;
}

P0;
  LDRSH W0,[X1] ;
  MOV X4,X0     ;
  LDRH W2,[X1]  ;
  MOV X6,X2     ;
forall (0:X0=-1 /\ 0:X2=65535 /\ 0:X4=4294967295 /\ 0:X6=65535)
```

maranget:
Hi @lukeg101. Thanks for this PR. Fixed sized arithmetic and processor behaviours are often tricky and it is good to check  our implementation from time to time.

This useful PR has been neglected for too long and some rebasing is necessary. If too tricky, please tell.

lukeg101:
Hi Luc, thank you for your feedback, I have addressed it, please let me know if there is something else :)

maranget (approved):

maranget:
Thanks, merged.

PR #405, opened 2022-07-15
[herd] Add support for custom fault handlers

relokin:
On AArch64 `elr_el1` determines the point at which the fault handler should return to. This means that when an instruction faults we should have an additional event that writes  a label of the current instruction or the instruction address to `elr_el1`. Then, at the end of the fault handler (or otherwise when we execute `eret`), we should jump to the instruction that `elr_el1` points to.

### DIFF
--- a/herd/AArch64Arch_herd.ml
+++ b/herd/AArch64Arch_herd.ml
@@ -217,7 +217,7 @@ module Make (C:Arch_herd.Config)(V:Value.AArch64) =
       | I_LDOPBH (_,v,_,_,_,_) | I_STOPBH (_,v,_,_,_) ->
           Some (bh_to_sz v)
       | I_NOP|I_B _|I_BR _|I_BC (_, _)|I_CBZ (_, _, _)
-      | I_CBNZ (_, _, _)|I_BL _|I_BLR _|I_RET _|I_LDAR (_, _, _, _)
+      | I_CBNZ (_, _, _)|I_BL _|I_BLR _|I_RET _|I_ERET|I_LDAR (_, _, _, _)
       | I_TBNZ(_,_,_,_) | I_TBZ (_,_,_,_) | I_MOVZ (_,_,_,_) | I_MOVK(_,_,_,_)
       | I_MOV (_, _, _)|I_SXTW (_, _)|I_OP3 (_, _, _, _, _, _)
       | I_ADR (_, _)|I_RBIT (_, _, _)|I_FENCE _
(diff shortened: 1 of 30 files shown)


herd/AArch64Arch_herd.ml
```diff
@@ -363,6 +365,12 @@ module Make (C:Arch_herd.Config)(V:Value.AArch64) =
 
         end)
 
+    type fh_reg_state = reg_state
+    type fh_v = V.v
+    let set_faulting_instr addr env =
+      let env = kill_regs [elr_el1] env in
+      set_reg elr_el1 addr env
```
maranget, line 372: I could not find where function `set_faulting_instr` is used. Leftover code?

herd/machAction.ml
```diff
@@ -49,14 +49,17 @@ module type Config = sig
...
   type action =
     | Access of Dir.dirn * A.location * A.V.v * A.lannot * A.explicit * MachSize.sz * Access.t
     | Barrier of A.barrier
-    | Commit of bool (* true = bcc / false = pred *) * string option
+    | Commit of commit_type * string option
 (* Atomic modify, (location,value read, value written, annotation *)
     | Amo of A.location * A.V.v * A.V.v * A.lannot * A.explicit * MachSize.sz * Access.t
 (* NB: Amo used in some arch only (e.g., Arm, RISCV) *)
-    | Fault of A.inst_instance_id * A.location * Dir.dirn * A.lannot * string option
+    | Fault of A.inst_instance_id * A.location * Dir.dirn * A.lannot * bool * string option
```
maranget, line 62: The boolean argument is wort a comment:
```suggestion
 (* bool (fith) argument  is true when modeling fault handler  entry *)
    | Fault of A.inst_instance_id * A.location * Dir.dirn * A.lannot * bool * string option
```

herd/semExtra.ml
```diff
@@ -447,6 +452,7 @@ type concrete =
     let is_overwritable = A.is_overwritable
     let instruction_to_value = A.instruction_to_value
 
+    let set_faulting_instr = A.set_faulting_instr
```
maranget, line 455: I could not find where this function is called.

maranget (approved):
Hi @relokin Thanks, I have made two minor remarks, I'd like to have your feedback before I merge.

herd/tests/instructions/AArch64.kvm/A005.litmus
```diff
@@ -0,0 +1,11 @@
...
+{
+[PTE(x)]=(oa:PA(x),valid:0);
+0:X1=x;
+}
+ P0          | P0.F           ;
+ LDR W0,[X1] | MOV X2,#1      ;
+L0:          | ADR X3,L0      ;
+ MOV X3,#1   | MSR ELR_EL1,X3 ;
+             | ERET           ;
+ forall(0:X2=1 /\ 0:X3=1)
```
maranget, line 11: For litmus' not to issue a warning it would be nice not to use register `X3` both as address and integer and to use the `W` form of registers when they hold integers:
```suggestion
 LDR W0,[X1] | MOV W2,#1      ;
L0:          | ADR X3,L0      ;
 MOV W4,#1   | MSR ELR_EL1,X3 ;
             | ERET           ;
 forall(0:X2=1 /\ 0:X4=1)
```

PR #413, opened 2022-07-29
[herd] Variant declaration in test override other declarations.

maranget:
Variant are first defined by command line and configuration file.
Those global declarations are overridden by any test specific
declaration `Variant=...` from test source.

Some part of herd code (specifically the `Mem` module) was unaware
of such overriding.

### DIFF
--- a/herd/parseTest.ml
+++ b/herd/parseTest.ml
@@ -45,66 +45,63 @@ module NoCheck = struct
   let check parsed= parsed
 end
 
-module Top (TopConf:Config) = struct
+module Make
+    (Conf:Config)
+    (S:Sem.Semantics)
+    (P:sig
+      type pseudo
+      val parse : in_channel -> Splitter.result ->  pseudo MiscParser.t
+    end with type pseudo = S.A.pseudo)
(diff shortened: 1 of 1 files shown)


relokin:
Incidentally, I also run into the same issue in #405 https://github.com/herd/herdtools7/pull/405/commits/e7cb55d468255d632d9b325f48105df25abb33b9

maranget:
> Incidentally, I also run into the same issue in #405 [e7cb55d](https://github.com/herd/herdtools7/commit/e7cb55d468255d632d9b325f48105df25abb33b9)

Ah! I see we have very similar solutions. I additionally extracted the `Make` out of the `Top` functor, as all references to `TopConf` have disappeared from `Make`. What should we do? BTW is your PR #405 ready for review?

relokin:
Hey @maranget, up to you, let's go with the solution that is technically better (I am happy to get rid of my commit that is).

Yes #405 is ready for review.

maranget:
Since #405 is ready, it looks easier to keep it as it is.

PR #414, opened 2022-07-30
[litmus7] Preliminary SVE support (only on the parser)

octurion:
Introduce support for SVE on litmus7, but only in terms of parsing and pretty-printing. This patch only supports the following instructions:

- Contiguous SVE loads/stores (`LD1`/`ST1`).
- Filling an SVE register with a constant (`DUP`).
- Generating a loop iteration mask by comparing a loop index and the loop end (`WHILELO`).


maranget:
Hi @octurion. Why not extend the PR for complete integration in litmus? This should be quite easy, or am I missing something? Integration in herd is another story and can be left for a future PR.

It seems that this PR is work in progress, If so it would be appropriate to convert it to a draf PR.

relokin:
@maranget should we close this PR now? It's been superseded newer PRs.

maranget:
I guess we should.

relokin:
Closing as this is superseded.

PR #417, opened 2022-08-12
Minor fixes for herd and litmus when running tests with custom fault handlers

relokin:


litmus/libdir/_instance.c
```diff
@@ -173,10 +173,8 @@ static void set_role(global_t *g,thread_ctx_t *c,int part) {
     c->ctx = NULL ;
     c->role = -1 ;
   }
-#ifdef KVM  
-#ifdef HAVE_FAULT_HANDLER
+#ifdef KVM
   set_fault_vector(c->role);
-#endif
 #endif
```
maranget, line 178: Hi @relokin. This change breaks the X86_64 implementation, which does not feature fault handlers. Commit a5990914e8130ba3f79d0a8d53f926516b786cdf fixes the problem by having litmus to generate a void `set_fault_vector` function even when fault handler management is not provided. Another option would be not to apply this commit, if there is no other reason than simplification of `_instance.c` behind it. Please choose the best option.

relokin: hey @maranget, thanks for reviewing this.

The motivation for this change was this test:
```
AArch64 LDRv0+I2V
{
 PTE(x)=(oa:PA(x),valid:0);
 x=1;

 0:X1=x;
 0:X3=PTE(x);
 0:X2=(oa:PA(x),valid:1);
}
 P0            | P0.F           ;
               | STR X2,[X3]    ;
 LDR W0,[X1]   | MRS X8,ELR_EL1 ;

locations[0:X8;]
forall(0:X0=1)
```
which doesn't compile with litmus7 from master. I wasn't aware of the problem with X86_64. Is it ok if I apply your patch https://github.com/herd/herdtools7/commit/a5990914e8130ba3f79d0a8d53f926516b786cdf  ?

maranget: Hi @relokin, you can apply the patch. I am not satisfied with this solution. However any alternative I can think of is even more contrived.

PR #419, opened 2022-09-05
[litmus] Explict handler for EL0 threads

maranget:
This PR extends the implementation of explicit handlers to EL0.

One of the difficulty to solve is for the actual handler to discriminate between a fault and the final `svc` call. In the first case the user supplied handler should be executed; while in the latter case, a simple branch to end of code should be performed, thereby leaving the exception level at EL1.

### DIFF
--- a/lib/AArch64Lexer.mll
+++ b/lib/AArch64Lexer.mll
@@ -443,13 +443,6 @@ match name with
 (* System registers *)
 | "mrs"|"MRS" -> MRS
 | "msr"|"MSR" -> MSR
-| "ctr_el0"|"CTR_EL0" -> SYSREG A.CTR_EL0
-| "dciz_el0"|"DCIZ_EL0" -> SYSREG A.DCIZ_EL0
-| "mdccsr_el0"|"MDCCSR_EL0" -> SYSREG A.MDCCSR_EL0
-| "dbgdtr_el0"|"DBGDTR_EL0" -> SYSREG A.DBGDTR_EL0
-| "dbgdtrrx_el0"|"DBGDTRRX_EL0" -> SYSREG A.DBGDTRRX_EL0
-| "Dbgdtrtx_el0"|"DBGDTRTX_EL0" -> SYSREG A.DBGDTRTX_EL0
-| "elr_el1"|"ELR_EL1" -> SYSREG A.ELR_EL1
 | _ ->
     begin match A.parse_wreg name with
     | Some r -> ARCH_WREG r
@@ -473,7 +466,11 @@ match name with
                         | 'Q' -> ARCH_QREG r
                         | _ -> assert false
                         end
-                    | None -> NAME name
+                    | None ->
+                        begin match A.parse_sysreg name with
+                        | Some r -> SYSREG r
+                        | None -> NAME name
+                        end
                     end
                 end
             end
--- a/litmus/AArch64Compile_litmus.ml
+++ b/litmus/AArch64Compile_litmus.ml
@@ -1206,8 +1206,8 @@ module Make(V:Constant.S)(C:Config) =
 
     let user_prologue =
       ["mrs %[tr0],esr_el1";
-       "lsr %[tr0],%[tr0],#26";
-       "cmp %[tr0],#21";
+       "lsr %[tr0],%[tr0],#%[esr_el1_ec_shift]";
+       "cmp %[tr0],#%[esr_el1_ec_svc64]";
        "cset %[tr0],ne";
        "lsl %[tr0],%[tr0],#2";
        "adr %[tr1],2f";
--- a/litmus/AArch64Compile_litmus.ml
+++ b/litmus/AArch64Compile_litmus.ml
@@ -1204,10 +1204,25 @@ module Make(V:Constant.S)(C:Config) =
 
     let kernel_mode = [{ empty_ins with memo="svc #471"}]
 
-    let fault_handler_prologue p =
-      map_ins ["b 1f";sprintf  "asm_handler%d:" p]
+    let user_prologue =
+      ["mrs %[tr0],esr_el1";
+       "lsr %[tr0],%[tr0],#26";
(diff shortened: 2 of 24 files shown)


maranget:
Hi @relokin, When you have time, I'd like to have your opinion on this PR. The PR implements explicit handlers for litmus from EL0. In particular effective handler code discriminates the case of a fault and of an explicit `svc` instruction by a computed branch based upon the value of the `esr_el1` register, cf commit e31f0e5

litmus/AArch64Compile_litmus.ml
```diff
@@ -1204,10 +1204,25 @@ module Make(V:Constant.S)(C:Config) =
...
     let kernel_mode = [{ empty_ins with memo="svc #471"}]
 
-    let fault_handler_prologue p =
-      map_ins ["b 1f";sprintf  "asm_handler%d:" p]
-
-    and fault_handler_epilogue code =
+    let user_prologue =
+      ["mrs %[tr0],esr_el1";
+       "lsr %[tr0],%[tr0],#26";
+       "cmp %[tr0],#21";
```
relokin, line 1210: and this to:
`"cmp %[tr0],#ESR_EL1_EC_SVC64";`

litmus/AArch64Compile_litmus.ml
```diff
@@ -1204,10 +1204,25 @@ module Make(V:Constant.S)(C:Config) =
...
+       "cset %[tr0],ne";
+       "lsl %[tr0],%[tr0],#2";
+       "adr %[tr1],2f";
+       "add %[tr1],%[tr1],%[tr0]";
+       "br %[tr1]";
+       "2:";
+       "b 1f";
```
relokin, line 1217: Can't we simplify all this to:
`     "b.eq 1f";
`
?

relokin (commented):
Some minor comments that aim to make the code simpler otherwise the code looks good to me!

litmus/AArch64Compile_litmus.ml
```diff
@@ -1204,10 +1204,25 @@ module Make(V:Constant.S)(C:Config) =
 
     let kernel_mode = [{ empty_ins with memo="svc #471"}]
 
-    let fault_handler_prologue p =
-      map_ins ["b 1f";sprintf  "asm_handler%d:" p]
-
-    and fault_handler_epilogue code =
+    let user_prologue =
+      ["mrs %[tr0],esr_el1";
+       "lsr %[tr0],%[tr0],#26";
```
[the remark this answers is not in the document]
maranget: I was afraid of cpp not substituting inside strings, I have to check.

litmus/AArch64Compile_litmus.ml
maranget: Same answer, will try.

litmus/AArch64Compile_litmus.ml
maranget: Ah I wanted to avoid a conditional branch. In fact this "computed jump" also is a dependency..

litmus/AArch64Compile_litmus.ml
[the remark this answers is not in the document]
relokin: ah yes you're right, I think you might be able to use the `xstr` macro like here: https://gitlab.com/kvm-unit-tests/kvm-unit-tests/-/blob/master/arm/selftest.c#L120 . If it's too complicated please ignore.

litmus/AArch64Compile_litmus.ml
relokin: Yes that's a good point. When I tried something similar I had the same problem.

maranget: Another solution would be to change the vector table entry before the final `svc` instruction. Some synchronisaton  may be needed. I have no idea which.

lib/AArch64Lexer.mll
```diff
@@ -450,6 +450,7 @@ match name with
 | "dbgdtrrx_el0"|"DBGDTRRX_EL0" -> SYSREG A.DBGDTRRX_EL0
 | "Dbgdtrtx_el0"|"DBGDTRTX_EL0" -> SYSREG A.DBGDTRTX_EL0
 | "elr_el1"|"ELR_EL1" -> SYSREG A.ELR_EL1
+| "esr_el1"|"ESR_EL1" -> SYSREG A.ESR_EL1
```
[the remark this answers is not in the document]
maranget: Oups I forgot to delete this. System register are now lexed with specific functions defined in `AArch64Base.ml`, like other registers.

litmus/AArch64Compile_litmus.ml
relokin: I think you have to do it at the last invocation of the fault handler (you have to be at EL1), but it might be tricky to know that last invocation is.

litmus/AArch64Compile_litmus.ml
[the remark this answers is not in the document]
maranget: Finally I have committed an attempt to use symbolic constants. In place of direct inclusion in instructions, which various levels of expansion make really tricky, I leverage gcc "i" input constraint, see commit   d50ddd3bbdbb802d5c0e6ad7edea984b5ece1f24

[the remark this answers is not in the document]
relokin: LGTM! Thanks @maranget

litmus/AArch64Compile_litmus.ml
maranget: @relokin I'd rather have  try of this alternative implementation in another PR. This PR looks ready to me. Your opinion?

relokin: Sounds good. I don't have any other comments :), please feel free to go ahead with this PR.

maranget: Thanks, just merged.

PR #420, opened 2022-09-07
Get rid of the ad-hoc installation process by upgrading to dune 3.5

kit-ty-kate:
Once https://github.com/ocaml/dune/issues/6133 is fixed this should work and does simplify the installation process

### DIFF
--- a/dune
+++ b/dune
@@ -1,4 +1,11 @@
-(ignored_subdirs (herd-www))
+(dirs (:standard \ herd-www))
 (env
  (release
 (flags (:standard -w +a-3-4-9-29-33-41-45-60-67-70))))
+
+(install
+ (dirs
+  (herd/libdir as herd)
+  (litmus/libdir as litmus)
+  (jingle/libdir as jingle))
+ (section lib_root))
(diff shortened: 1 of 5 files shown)


maranget:
> Once [ocaml/dune#6133](https://github.com/ocaml/dune/issues/6133) is fixed this should work and does simplify the installation process

Thanks @kit-ty-kate, the issue looks as if it is still open, So no emergency to merge this PR? I'd rather avoid mandatory dune upgrade. Would you change be compatible with "reasonably recent" versions of dune?

kit-ty-kate:
> So no emergency to merge this PR?

No, don't worry.

> I'd rather avoid mandatory dune upgrade. Would your change be compatible with "reasonably recent" versions of dune?

Sadly no. The new `(install (dirs ...))` feature necessary to make it work is only present in dune master / the next planned release.

maranget:
Thanks @kit-ty-kate. No need to hurry, then...

fsestini:
Superseded by https://github.com/herd/herdtools7/pull/1902

PR #421, opened 2022-09-09
[herd] Print warning when a fault occurs within a fault handler

relokin:

### DIFF
--- a/herd/mem.ml
+++ b/herd/mem.ml
@@ -776,7 +776,7 @@ let match_reg_events es =
                   (A.V.pp_v v_stored) ;
                 assert false)
           rfm csn in
-      match VC.solve csn with
+      match VC.solve ~final:false csn with
       | VC.NoSolns ->
          if C.debug.Debug_herd.solver then
            pp_nosol "register" test es rfm ;
@@ -1009,7 +1009,7 @@ let match_reg_events es =
                 loads stores cns in
             if dbg then eprintf "\n%!" ;
             (* And solve *)
-            match VC.solve cns with
+            match VC.solve ~final:true cns with
             | VC.NoSolns ->
                if C.debug.Debug_herd.solver then begin
                  let rfm = add_some_mem loads stores rfm in
@@ -1273,7 +1273,7 @@ let match_reg_events es =
                       (fun load store k -> add_mem_eqs test store load k)
                       tag_loads tag_stores cns in
                   (* And solve *)
-                  match VC.solve cns with
+                  match VC.solve ~final:true cns with
                   | VC.NoSolns -> res
                   | VC.Maybe (sol,cs) ->
                       (* Time to complete rfmap *)
--- a/herd/mem.ml
+++ b/herd/mem.ml
@@ -508,20 +508,21 @@ module Make(C:Config) (S:Sem.Semantics) : S with module S = S	=
         | Ok ((tgt_proc,code),seen) -> add_code re_exec tgt_proc proc env seen code
 
       and add_fault re_exec inst dir fetch_proc proc env seen addr nexts =
-        match env.A.fh_code with
-        | Some fh_code ->
+        match env.A.fh_code,re_exec with
+        | Some _, true ->
+           let e = "Fault inside a fault handler" in
+           EM.warncodeT e true
+        | Some fh_code, false ->
            add_code true fetch_proc proc env seen fh_code
-        | None -> begin
-            if re_exec then
-              EM.unitcodeT false
-            else
-              let open Precision in
-              match C.precision,dir with
-              | Fatal,_ | LoadsFatal,(Dir.R|Dir.F) -> EM.unitcodeT true
-              | Skip,_ -> add_code false fetch_proc proc env seen nexts
-              | Handled,_ | LoadsFatal,Dir.W ->
-                 add_next_instr true fetch_proc proc env seen addr inst nexts
-          end
+        | None, true ->
+           EM.unitcodeT false
+        | None, false ->
+           let open Precision in
+           match C.precision,dir with
+           | Fatal,_ | LoadsFatal,(Dir.R|Dir.F) -> EM.unitcodeT true
+           | Skip,_ -> add_code false fetch_proc proc env seen nexts
+           | Handled,_ | LoadsFatal,Dir.W ->
+              add_next_instr true fetch_proc proc env seen addr inst nexts
 
       and next_instr re_exec inst fetch_proc proc env seen addr nexts b =
         match b with
@@ -532,6 +533,8 @@ module Make(C:Config) (S:Sem.Semantics) : S with module S = S	=
           add_lbl re_exec true proc env seen addr lbl
       | S.B.Fault dir ->
           add_fault re_exec inst dir fetch_proc proc env seen addr nexts
+      | S.B.FaultRet lbl ->
+          add_lbl false true proc env seen addr lbl
       | S.B.CondJump (v,lbl) ->
           EM.condJumpT v
             (add_lbl re_exec (not (V.is_var_determined v)) proc env seen addr lbl)
--- a/herd/AArch64Sem.ml
+++ b/herd/AArch64Sem.ml
@@ -1718,7 +1718,7 @@ module Make
 
         | I_ERET ->
            let eret_to_addr = function
-             | M.A.V.Val(Constant.Label (_, l)) -> B.branchT l
+             | M.A.V.Val(Constant.Label (_, l)) -> B.faultRetT l
              | _ ->
                 Warn.fatal "Cannot determine ERET target" in
            let commit_eret ii =
--- a/herd/branch.ml
+++ b/herd/branch.ml
@@ -36,6 +36,8 @@ module type S = sig
     | Exit
     (* Raise Fault *)
     | Fault of Dir.dirn
+    (* Return from Fault Handler *)
+    | FaultRet of lbl
 
 (* Next instruction in sequence *)
   val nextT : t monad
@@ -48,6 +50,7 @@ module type S = sig
   val branchT : lbl ->  t monad
 (* Conditional branch *)
   val bccT : v -> lbl -> t monad
+  val faultRetT : lbl -> t monad
 end
 
 
@@ -69,6 +72,8 @@ module Make(M:Monad.S) = struct
     | Exit
     (* Raise Fault *)
     | Fault of Dir.dirn
+    (* Return from Fault Handler *)
+    | FaultRet of lbl
 
 (* Utilities *)
 
@@ -81,4 +86,5 @@ module Make(M:Monad.S) = struct
   let nextSetT r v = M.unitT (Next [r,v])
   let branchT lbl = M.unitT (Jump lbl)
   let bccT v lbl = M.unitT (CondJump (v,lbl))
+  let faultRetT lbl = M.unitT (FaultRet lbl)
 end
--- a/herd/eventsMonad.ml
+++ b/herd/eventsMonad.ml
@@ -149,6 +149,11 @@ Monad type:
       eiid_next,
       (Evt.singleton (v, [VC.Failed e], E.empty_event_structure), None)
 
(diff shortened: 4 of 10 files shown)


herd/mem.ml
```diff
@@ -508,20 +508,20 @@ module Make(C:Config) (S:Sem.Semantics) : S with module S = S	=
...
-            else
-              let open Precision in
-              match C.precision,dir with
-              | Fatal,_ | LoadsFatal,(Dir.R|Dir.F) -> EM.unitcodeT true
-              | Skip,_ -> add_code false fetch_proc proc env seen nexts
-              | Handled,_ | LoadsFatal,Dir.W ->
-                 add_next_instr true fetch_proc proc env seen addr inst nexts
-          end
+        | None, true ->
+            EM.unitcodeT false
```
relokin, line 517: @maranget, I am not sure I fully understand the significance of the `bool` argument of `unitcodeT`. Do you know how it is used?

maranget: @relokin Your interpretation of the effect of the `re_exec` parameter is correct (as far as I remember and have checked). The rationale is as follows: in some fault modes, such as the `Handled` mode, executing the faulting instruction is attempted again. There is only one attempt. This behaviour is relevant for tests that, for instance, include code to repair invalid PTE's. Notice that not all (potential) traces are explored, as we miss the traces that correspond to faulting twice or more on the same instruction . Well...

relokin:
@maranget, I am happy with this but I have gone back and forth a couple of times. Please have a look when you can and let me know what you think.

herd/mem.ml
```diff
@@ -508,20 +508,21 @@ module Make(C:Config) (S:Sem.Semantics) : S with module S = S	=
         | Ok ((tgt_proc,code),seen) -> add_code re_exec tgt_proc proc env seen code
 
       and add_fault re_exec inst dir fetch_proc proc env seen addr nexts =
-        match env.A.fh_code with
-        | Some fh_code ->
+        match env.A.fh_code,re_exec with
+        | Some _, true ->
+           W.warn "Cannot handle faults in fault handlers";
+           EM.unitcodeT false
```
maranget, line 514: I have tried this example
```
AArch64 A007
{
[x]=1;
[PTE(x)]=(oa:PA(x),valid:0);
0:X1=x;
}
 P0          | P0.F           ;
L0:          | LDR W2,[X1]    ;
 LDR W0,[X1] | ADR X3,L1      ;
L1:          | MSR ELR_EL1,X3 ;
             | ERET           ;
 locations[0:X2;]
 forall(0:X0=0)
```
Default output  (_i.e._ without option `-v`) is not very informative:
```
Test A007 Required
States 0
Ok
Witnesses
Positive: 0 Negative: 0
Condition forall (0:X0=0)
Observation A007 Never 0 0
Time A007 0.00
Hash=3244a16b7608b9fe87f0e4434431896d
```
Wouldn't it be more useful to consider that a fault in a fault handler is an error:
```suggestion
           Warn.user_error "Cannot handle faults in fault handlers"
```
Output now is:
```
Warning: File "Handler/Explicit/A007.litmus": Cannot handle faults in fault handlers (User error)
```
```
```

relokin: I am afraid that this won't work. If we change it to a `user_error` if there is any execution that raises a fault in the fault handler the simulation exits, even if that execution would be forbidden by the cat file. Take a test like the one you provided:

```
AArch64 A007-valid
{
[x]=1;
[PTE(x)]=(oa:PA(x),valid:1);
0:X1=x;
}
 P0          | P0.F           ;
L0:          | LDR W2,[X1]    ;
 LDR W0,[X1] | ADR X3,L1      ;
L1:          | MSR ELR_EL1,X3 ;
             | ERET           ;
 locations[0:X2;]
 forall(0:X0=0)
```

I get 

```
Warning: File "./A007.litmus": Cannot handle faults in fault handlers (User error)
```

maranget: I do understand: the intention behind the using `user_erro` is for the simulation to exit. That way, a critical problem is signaled : herd cannot handle this situation.

With the other option of a warning, all executions such that a fault occurs in handler are discarded. Notice that that are discarded silently if the option `-v` is not provided. Calling `Warn.warn_always` would yield a message without the verbosity option.

relokin: > I do not understand: the intention behind the using `user_error` is for the simulation to abort for the current test only. That way, a critical problem is signaled : herd cannot handle this situation. However if several tests are simulated, the remaining tests are executed.

Agreed, but if we used `user_error`, herd7 would abort for legal tests too. Take `A007-valid` for example:

```
AArch64 A007-valid
{
[x]=1;
[PTE(x)]=(oa:PA(x),valid:1);
0:X1=x;
}
 P0          | P0.F           ;
L0:          | LDR W2,[X1]    ;
 LDR W0,[X1] | ADR X3,L1      ;
L1:          | MSR ELR_EL1,X3 ;
             | ERET           ;
 locations[0:X2;]
 forall(0:X0=1)
```
The only final state that we should get is:

```
0:X0=1; 0:X2=0;
```

When I use `W.warn` I get:

```
Test A007-valid Required
States 1
0:X0=1; 0:X2=0;
Ok
Witnesses
Positive: 1 Negative: 0
Condition forall (0:X0=1)
Observation A007-valid Always 1 0
Time A007-valid 0.00
Hash=96d9fbe52d41e18619c128445155ce34
```

when I use `Warn.user_error` herd7 aborts and it doesn't produce any final states at all.

```
Warning: File "./A007-valid.litmus": Cannot handle faults in fault handlers (User error)
```

I think it's fair to say that there is no user error in `A007-valid`.

> 
> With the other option of a warning, all executions such that a fault occurs in handler are discarded. Notice that that are discarded silently if the option `-v` is not provided. Calling `Warn.warn_always` would yield a message without the verbosity option.

I agree, `warn` is not ideal because herd7 doesn't print anything without `-v`.

`Warn.warn_always` is better than 'Warn.user_error` but not ideal. For example, it would unnecessarily warn the user for `A007-valid`:
```
Warning: Cannot handle faults in fault handlers
Test A007-valid Required
States 1
0:X0=1; 0:X2=0;
Ok
Witnesses
Positive: 1 Negative: 0
Condition forall (0:X0=1)
Observation A007-valid Always 1 0
Time A007-valid 0.01
Hash=96d9fbe52d41e18619c128445155ce34
```

Ideally, we would need something such that herd7 prints a warning if an (allowed) execution includes a memory instruction inside a fault handler that raised a fault. If we warn (`warn_always`)/abort (`user_error`) if it's possible for a memory instruction inside a fault handler to fault then it's as if we warned/aborted for all tests that include memory instructions inside fault handlers.

maranget: Ah now I see, thanks for the explanation and example.

I would definitively rather have herd to flag an error when a  _valid_  execution have a fault in handler. By "valid" we mean w.r..t the model. Unfortunately,  this is not possible at the moment, without significant and risky change of herd internals.
However, I found a decent alternative: flag an error when a pre-execution have a fault in handler --- A pre-execution is a candidate execution to be checked against the model. This alternative can be implemented easily by a simple extension of the existing mechanism of delayed exception in solver (see commit 192294e47eb7d8bba7024faf93db486c7a744403). This solves the case of the **A007-valid** test, which could be part of the regression suite,

For completeness, here is an example of a test  for which an error is flagged after applying the patch, while arguably the "fault-in-handler" execution is not relevant as it is rejected by model.
```
AArch64 A008
Variant=skip
{
0:X0=x; 0:X3=y; 0:X5=z;
1:X0=y; 1:X3=x;
[PTE(z)]=(valid:0)
}
 P0          | P1          | P0.F        ;
 LDR W1,[X0] | LDR W1,[X0] | CBZ W1,L0   ;
 DMB SY      | STR W1,[X3] | LDR W6,[X5] ;
 MOV W2,#1   |             |L0:          ; 
 STR W2,[X3] |             |             ;
 LDR W4,[X5] |             |             ;
exists (0:X1=1 /\ 1:X1=1)
```
Namely, the  handler code `P0.F` attempt a load of `z` only if register `0:X`` is not zero. which is excluded by the model.

However the example is contrived and I'd rather have it  rejected than to have `A007` not to be rejected and have no outcome.

relokin: thanks for the patch @maranget, I think it's in the right direction but I have one concern. Take the following test:
```
AArch64 A009
{
 [x]=1;
 [PTE(x)]=(valid:0);
 0:X0=PTE(x);
 0:X1=(valid:1,oa:PA(x));
 1:X0=x;
}
 P0          | P1          | P1.F        ;
 STR X1,[X0] | LDR W1,[X0] | LDR W2,[X0] ;
locations[1:X1;]
exists (1:X2=1)
```
I think the ideal output from herd7 should be:
```
Warning: File "./A009.litmus": Cannot handle faults in fault handlers (User error)
Test A009 Allowed
States 2
1:X1=0; 1:X2=1;
1:X1=1; 1:X2=0;
Ok
Witnesses
Positive: 1 Negative: 1
Condition exists (1:X2=1)
Observation A009 Sometimes 1 1
Time A009 0.01
Hash=0013cab22a273b4ea8fe84264f78039b
````
with your patch I get (no final states)
```
Warning: File "./A009.litmus": Cannot handle faults in fault handlers (User error)
```
without it I get (no warning)
```
Test A009 Allowed
States 2
1:X1=0; 1:X2=1;
1:X1=1; 1:X2=0;
Ok
Witnesses
Positive: 1 Negative: 1
Condition exists (1:X2=1)
Observation A009 Sometimes 1 1
Time A009 0.01
Hash=0013cab22a273b4ea8fe84264f78039b
```
Would it be easy to adapt your patch to achieve this? Happy to look into this if we agree on how we want this to work

maranget: Hi @nikos. I think I see what you mean: let us have a warning in place of an error. The warning would occur exactly  when we get the error with my patch (_i.e._ when the execution with a fault in handler is not discarded by the solver).

To get a warning in place of an error it may suffice to imitate the "delayed failure" mechanism, so as to yield some kind of "delayed warning". It may be enough to  add a`DelayedWarning of string` constructor to the type `cnstrnt` in  file `lib/valconstraint.ml`. And then to add some cases to a variety of pattern matching constructs (follow the warnings..), as well as a function `warncodeT` similar to `failcodeT` in file `herd/eventsMonad.ml`, except that it would replace `VC.Failed e` by `DelayedWarning msg`, where `msg` is the message of the warning.

maranget: The suggestion above builds upon my patch, since I modified the handling of delayed errors, so as to be able to declare the failure before calling the solver and to raise at the end of the last call to the solver.

relokin: thanks @maranget! I just updated the PR to implement this. 

Also, I kept your patch but `failcodeT` is not used by this PR. If you think we should remove let me know. Otherwise, I think this PR is ready.

maranget: The PR indeed looks ready to me. Let us keep the `failcodeT` function, as it does not harm and suggests the existence of delayed failures.
If you agree I'd like to add a test (commit 5f6e2ba8b9be025adf0970f0ebabb98cb2670029). At the moment the test triggers the warning, However, if such warning were delayed, until after model run, no warning would be issued, because the faulting execution would not pass through the model stage.
It may be, that we implement the more subtle semantics later. I'd like to have this test as a sentinel and reminder of this change possibility. A comment in test signals this unusual situation of a regression test added to encourage a regression,,,

relokin: >  If you agree, I'd like to add a test (commit [cd073c1](https://github.com/herd/herdtools7/commit/cd073c16dcce359bca210a9ceffb9aaf2dcd41d8)). At the moment the test triggers the warning, However, if such warning were delayed until after model run, no warning would be issued, because the faulting execution would not pass through the model stage. It may be that we implement the more subtle semantics later. I'd like to have this test as a sentinel and reminder of this change possibility. A comment in test signals this unusual situation of a regression test added to encourage a regression,,,

I agree, please feel free to add this extra test. Thanks for the reviews :)

PR #425, opened 2022-09-20
[litmus] Simple implementation of explicit handlers from EL0.

maranget:
Explicit handlers for EL0 threads are slightly tricky because there is only one slot for two functionalities: execute explicit handler and return to EL1 at end of thread code. This question is solved by using a branch-to-register instruction as handler code and by setting the register according to intended target code.


relokin (approved):
Not sure if you're planning to make further changes, but LGTM!

maranget:
> Not sure if you're planning to make further changes, but LGTM!

You are correct, I have delayed this merge for too long...

PR #427, opened 2022-09-30
[herd] Avoid unecessary initialization for PTE^2 entries

relokin:
When a litmus tests provides an initial value for a PTE(PTE(x)) avoid emiting the default initialization.

### DIFF
--- a/herd/event.ml
+++ b/herd/event.ml
@@ -278,7 +278,8 @@ val same_instance : event -> event -> bool
   val same_location : event -> event -> bool
   val same_location_with_faults : event -> event -> bool
   val same_value : event -> event -> bool
-(*  val is_visible_location : A.location -> bool *)
+  val same_low_order_bits : event -> event -> bool
+  (*  val is_visible_location : A.location -> bool *)
 
 (********************************)
 (* Event structure output ports *)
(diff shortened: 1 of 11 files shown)


relokin:
I just realized that this was completely unmotivated. I've now added a small test that shows the problem. 
```
AArch64 A009
Variant=kvm,PTE2
TTHM=HA
{
 [PTE(x)]=(oa:PA(x));
 [PTE(PTE(x))]=(oa:PTE(x),af:0);
 0:X1=PTE(x);
}
 P0          ;
 LDR X0,[X1] ;
exists([PTE(PTE(x))]=(oa:PTE(x),af:0));
```
Currently, for `A009` herd would compute two states:
```
Test A009 Allowed
States 2
[PTE(PTE(x))]=(oa:PTE(x), af:0);
[PTE(PTE(x))]=(oa:PTE(x));
Ok
Witnesses
Positive: 1 Negative: 3
Condition exists ([PTE(PTE(x))]=(oa:PTE(x), af:0))
Observation A009 Sometimes 1 3
Time A009 0.02
Hash=0acbe8f9f62851813328209de067d10a
```
The only allowed final state should be `[PTE(PTE(x))]=(oa:PTE(x));` and the test should be forbidden.

maranget:
Hi @nikos, is this PR still relevant, given our recent discussion on PTE  virtual and physical addresses being distinguished?

relokin:
Maybe not, we have to introduce bigger changes to PTE2 so I can close this for now. If necessary, I will open again or include it in a future PR.

PR #434, opened 2022-10-11
[herd] Add support for AArch64 FEAT_ExS

relokin:
FEAT_ExS is a feature in AArch64 that controls the semantics of exception entry
and exception return.

### DIFF
--- a/herd/variant.ml
+++ b/herd/variant.ml
@@ -45,6 +45,8 @@ type t =
   | Instances (* Compute dependencies on instruction instances *)
   | Kvm
   | ETS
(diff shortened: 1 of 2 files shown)


maranget:
Hi @nikos. This PR only defines new variants. Is it the intend?

relokin:
Yes it is. ExS is a feature like ETS. However, if you think ETS, ExS and other AArch64 features shouldn't be new variants but we need to have a new mechanism (e.g., architectural features), I would agree. In fact, ETS, ExS possibly others as well, don't change anything in the simulation, they are just passed to the cat file.

jalglave:
thanks @relokin, I'll merge now.

PR #437, opened 2022-10-13
[herd] Add the `lob-loc` cat pre-defined relation.

maranget:
The effective addresses of herd are pairs of symbols and offset. Conventionally all memory locations reside at page start and different location reside in different pages (litmus in kvm mode follows that convention.

As a consequence the ``load order bytes'' (_i.e_ the offset from page start) are represented by the offset component of effective addresses.

### DIFF
--- a/herd/event.ml
+++ b/herd/event.ml
@@ -278,7 +278,8 @@ val same_instance : event -> event -> bool
   val same_location : event -> event -> bool
   val same_location_with_faults : event -> event -> bool
   val same_value : event -> event -> bool
-(*  val is_visible_location : A.location -> bool *)
+  val same_low_order_bits : event -> event -> bool
+  (*  val is_visible_location : A.location -> bool *)
 
 (********************************)
 (* Event structure output ports *)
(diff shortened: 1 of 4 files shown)


maranget:
> hey @maranget, I had two questions/thoughts about this PR.
> 

Hi @relokin 
> * When I first saw the name `lob-loc`, I thought it has something to do with `lob` (as in `Locally-Ordered-Before`). Would `same-low-order-bits` (@jalglave has used this before) or `same-page-offset` be a better name for this relation?

@jalglave  suggested the name `lob-loc`. If some other name is better we can change it.


> * Doesn't the current convention (for both herd7 and litmus7) mean that in non-mixed mode any `x` and `y` will by default have the same low-order bits?

You are correct (and this also holds for tests produced by litmus). However, it is quite easy to get memory locations whose low order bits are different by using arrays:
{
int x;
int y[2];
}

The low order bits of locations x and y[1]  are different (offsets values are 0 and 4 respectively).

jalglave:
hi both, yes sorry I had some hesitation on the name as well. Let's go for same-low-order-bits.
Thanks,
Jade

maranget:
Merged, thanks to all.

PR #439, opened 2022-10-13
[herd] kvm and mte combination

maranget:
This PR implement combining the memory tag extension and kvm. Result looks satisfactory, some internal dependency links are not perfect though.

As an important side benefit of the PR, dependencies are now similar in both modes. Consider the case of a simple store instruction:
```
AArch64 MTE-W
{
0:X0=x;
}
  P0         ;
 MOV W1,#1   ;
 STR W1,[X0] ;
forall [x]=1
```

The `0:R[W1]1` event (data port read of the store)  does not depend on the validity test  (PTE check or color check respectively) in both modes.

A potential follow-up would be a similar rewriting of the morello mode. Then, the `lift_mem_op` function could get rid of its `mv` argument.

### DIFF
--- a/herd/AArch64Sem.ml
+++ b/herd/AArch64Sem.ml
@@ -107,10 +107,13 @@ module Make
 
       let mk_read_std = mk_read quad AArch64.N
 
+      let has_handler ii =
+        match ii.A.env.A.fh_code with
+        | Some _ -> true
+        | None -> false
+
       let mk_fault a dir annot ii ft msg =
-        let fh = match ii.A.env.A.fh_code with
-          | Some _ -> true
-          | None -> false in
+        let fh = has_handler ii in
         M.mk_singleton_es (Act.Fault (ii,A.Location_global a,dir,annot,fh,ft,msg)) ii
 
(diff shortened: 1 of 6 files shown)


maranget:
AMO instructions are not handled properly, For instance:
```
AArch64 MTE-SWP
{
0:X0=x;
[x]=2;
}
  P0              ;
 MOV W1,#1        ;
 SWP W1,W2,[X0]      ;
forall [x]=1 /\ 0:X2=2
```
In the [execution diagram](https://github.com/herd/herdtools7/files/9778452/MTE-SWP.pdf), observe that the no `iico_control` link starts at the  "h:" branching event.

maranget:
AMO instructions now look to be  handled properly, For instance:
```
AArch64 MTE-SWP
{
0:X0=x;
[x]=2;
}
  P0              ;
 MOV W1,#1        ;
 SWP W1,W2,[X0]      ;
forall [x]=1 /\ 0:X2=2
```
[New diagram](https://github.com/herd/herdtools7/files/9795438/MTE-SWP.pdf) with `iico_control` links to all teh events that depend upon the color test.

PR #444, opened 2022-10-21
Undefined Instructions

artkhyzha:
This PR attempts to sketch out a very ad-hoc support for instruction aborts. Namely, when a "B" instruction gets overwritten by a value different from "NOP", herd7 is expected to create a fault.

The intention is to, essentially, have a flow:
* read the instruction from memory
* check if it is "valid" (currently, checks if it is NOP or B), and
** if yes, execute it
** if no, trigger a fault

The expected intrinsic orderings are: [IFETCH]; iico_data; [B]; iico_ctrl; [FAULT | the rest of the instruction].

Currently known limitation: in addition to the desired orderings, there is a redundant "iico_data" edge: [IFETCH]; iico_data; [FAULT | the rest of the instruction].

### DIFF
--- a/herd/AArch64Sem.ml
+++ b/herd/AArch64Sem.ml
@@ -163,7 +163,7 @@ module Make
 (* Fetch of an instruction, i.e., a read from a label *)
       let mk_fetch an loc v =
         let ac = Access.VIR in (* Instruction fetch seen as ordinary, non PTE, access *)
-        Act.Access (Dir.F, loc, v, an, AArch64.nexp_annot, MachSize.Word, ac)
+        Act.Access (Dir.R, loc, v, an, AArch64.nexp_annot, MachSize.Word, ac)
 
 (* Basic write, to register  *)
       let mk_write sz an anexp ac v loc =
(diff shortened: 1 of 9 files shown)


artkhyzha:
Illustrating the redundant edge. [A001-00.pdf](https://github.com/herd/herdtools7/files/9840807/A001-00.pdf)

maranget:
Hi @artkhyzha, it woud be great if you could add some litmus  test, for instance the one you used to draw the diagram above. You can easily add then in the test directory `herd/tests/instruction/AArch64.self`

maranget:
Hi @artkhyzha I think the code of the  follow-up PR #449 gets rid of the redundant `iico_data` edges. Do you think the PR is now ready or do plan more changes?

artkhyzha:
The PR449 looks good to me and indeed removes the redundant edge, no specific changes planned at this point. (Thank you for taking the lead with this while I was away!)

maranget:
I am closing this PR, as the continuation PR #449 will be merged soon. Thanks for starting this and for the clear report on the presence of redundant edges,

PR #446, opened 2022-10-21
[herd] Add different types of fault and the corresponding cat sets

relokin:
This PR adds support for a new module, FaultType where the different types of faults can be defined per architecture. In AArch64, it defines Data Aborts and Tag Check faults. Then, it adds this as a field to the fault type and the Fault action and uses that information to build the corresponding event sets in cat. In AArch64 the two added sets are EXC-DATA-ABORT and EXC-TAG-CHECK.


relokin:
@maranget I only found out that the `msg` field in `fault` and `Fault` action when I saw the `pp` function. If you prefer using `msg` to specify the fault type, I am happy to use that instead or otherwise get rid of `msg` and merge its functionality into the new module.

HadrienRenaud:
I think that the file `lib/faultType.mli` is not necessary, and if the `.mli` file is deleted, in `lib/faultType.ml`, the signature `AArch64Sig` could be deleted too. It builds on my machine.

But I do not know when we want to have `.mli` files, what is the current practice @maranget ?

relokin:
> I think that the file `lib/faultType.mli` is not necessary, and if the `.mli` file is deleted, in `lib/faultType.ml`, the signature `AArch64Sig` could be deleted too. It builds on my machine.
> 
> But I do not know when we want to have `.mli` files, what is the current practice @maranget ?

Thanks @HadrienRenaud, FWIW, I added the `.mli` file just because most modules seemed to have it. I don't have a good reason for keeping it. I am happy to remove it.

maranget:
Hum, I tend to have `.mli` files. This is the ideal place for a short comment that describe module functionality (Although I am not always following this rule..).  Moreover the `faultType` module follows a pattern found elsewhere: a signature `S` and a "do-nothing" module `No`. This sort of modules always come with a `.mli` file, which is sufficient for a reader to get the idea. This especially applies to modules in `lib/` that can be used from several tool code.

That said, having the generic `FaultType.mli` to include AArch64 specific code is a bit of a surprise to me

relokin:
> That said, having the generic `FaultType.mli` to include AArch64 specific code is a bit of a surprise to me

I was looking at `memortyTypes.mli` which did something similar. I am happy to factor the AArch64 in separate files but its code is really small at the moment.

maranget:
@relokin, touché... Let us keep it like that for now.

maranget:
I have reviewed the code and it looks good to me. I'll merge in a few hours.

PR #453, opened 2022-11-03
[tools] Remove duplicates from normalized lists of outcomes

maranget:
Oddly, the tools lexer could produce lists of outcomes with duplicates, due to the erasure of fault comments. Fix this by removing duplicates early, during normalisation.

### DIFF
--- a/tools/logState.ml
+++ b/tools/logState.ml
@@ -996,7 +996,19 @@ let exclude e t =
 (* Normalize *)
 (*************)
 
-let normalize_sts sts =  List.sort (compare_state true) sts
+(* No duplicates in normalized lists of states *)
+
+let rec  normalize_sts_uniq cmp sts = match sts with
+  | []|[_] -> sts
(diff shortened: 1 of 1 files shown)


maranget:
Hi @relokin, here is my fix for (not) handling fault types in logs. The master branch tool lexer has always ignored fault comments. As a consequence, faultypes are also ignored. However, erasing information from faults may result in duplicate lines in "final states", finally yielding odd output of the comparison tools `mcmp7` and `mcompare7`.

As a conclusion this PR makes the code more robust and I do not expect it to harm in any way. Moreover it is probably orthogonal to your improvements to tools. I'd like to merge it quite soon.

relokin:
hey @maranget, please feel free to merge. I saw this issue in the code but I didn't have a fix for it. I think we need it independently of my patch.

maranget:
Ok, let's merge.

PR #454, opened 2022-11-04
[herd] Avoid po relation between events of the same instance

relokin:

### DIFF
--- a/herd/libdir/aarch64.cat
+++ b/herd/libdir/aarch64.cat
@@ -155,7 +155,7 @@ let tlo = [R & T]; tob; loc; tob^-1; [R & T]
 let po-loc =
   let Exp = M\domain(tob) in
   (po-loc & Exp*Exp) | (po & tlo)
-acyclic po-loc | ca | rf as internal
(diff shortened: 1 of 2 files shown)


maranget:
Hi @relokin, for testing modifications of the internal check, it may be a good idea to disable candidate optimisation with option `-optace false`.

maranget:
I have tested this PR on an enlarged test base, including the new tests of PR #456.

relokin:
Many thanks for this @maranget!

Out of curiosity, I saw you added many tests as the regression. Do you think running these is enough? Or do you have more tests privately?

maranget:
> Many thanks for this @maranget!
> 
> Out of curiosity, I saw you added many tests as the regression. Do you think running these is enough? Or do you have more tests privately?

Most of the tests are from a test based that we built in collaborations with the Cambridge University team. Those are the tests we routinely  on hardware, see [here](https://cambium.inria.fr/~maranget/cats7/model-aarch64/) for links to archives of those tests. I also have a few additional tests.

maranget:
Thanks @relokin, merged.

PR #460, opened 2022-11-09
[tools] Add mcompare support for labels as values in postconditions

relokin:


maranget:
Good catch, thanks @relokin

PR #463, opened 2022-11-15
[herd] Add cat support for logical operations on variants

relokin:
This change adds support for the logical operators '&&', '||' and 'not' in statements where the condition was previously a variant. This way, where previously we could use clauses of the form:

if VARIANT then ... else ...

now, we can use clauses of the form:

if VARIANT_COND then ... else ...

where VARIANT_COND is a VARIANT or a logical operation on multiple VARIANTs using the defined operators.

### DIFF
--- a/lib/modelLexer.mll
+++ b/lib/modelLexer.mll
@@ -63,6 +63,7 @@ module LU = LexUtils.Make(O)
     (* for bell files *)
     | "instructions" -> INSTRUCTIONS
     | "default" -> DEFAULT
+    | "not" -> LNOT
     | x -> VAR x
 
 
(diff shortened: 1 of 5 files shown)


lib/modelLexer.mll
```diff
@@ -63,6 +63,8 @@ module LU = LexUtils.Make(O)
     (* for bell files *)
     | "instructions" -> INSTRUCTIONS
     | "default" -> DEFAULT
+    | "&&" -> LAND
```
maranget, line 66:
```suggestion
```
There is a difficulty here: the `do_check_keyword` function is called on lexemes that match the `name` definition (see about line 120):
```OCaml
let name  = '_' ? alpha (alpha|digit|'_' | '.' | '-')* '\''?
```
As a result, the function is not called on string `"&&"` and this lexeme has to be recognized explicitly in the main lexer as:
```OCaml
| '&&'   { f Operator "&" ; LAND }
```

relokin: Very good point! Thanks  @maranget 

Initially, I was using the keyword `and` (and `or`). I tested and convinced myself that it works. Then came to the realization that we would be better off using the same notation as ocaml for logical operators, changed them to `&&` and `||` but I completely missed this.

maranget:
The PR is now ready. I'll merge it in a few hours. I also like  OCaml-style (and C-style) logical connectors much better than the Pascal-style ones AND and OR.

PR #466, opened 2022-11-24
[AArch64] Initial formalisation of the VMSA and more

jalglave:
[AArch64] Initial formalisation of the VMSA - Deprecating ETS and introducing new feature ETS2 - Internal visibility rework - MTE relaxation

This release presents a number of changes, listed below. Those changes should be reflected in the next issue of the Arm ARM. For more details please contact memory-model@arm.com.

1. Initial formalisation of the VMSA

This version of the model introduces a number of concepts required to describe orderings due to accessing page tables. Most notable are the notions of TTD-read-ordered-before, TLBI-ordered-before, and TLBI-Coherence-after.

This version of the model also handles the notions of Access Flag and Dirty Bit, and the Hardware Management thereof.

2. Deprecating ETS and Introducing new feature ETS2

As part of the work leading to this initial formalisation of the VMSA, it transpired that the ETS feature was not giving enough guarantee to be useful to software. As a consequence, Arm deprecates ETS and introduces a new feature called ETS2 instead.

This version of the model accounts for those changes.

3. Internal visibility rework

As part of the work leading to this initial formalisation of the VMSA, Arm decided to rework the way the axioms of the model are presented. More precisely, most behaviours which were caught by Internal visibility are now caught by External visibility (including coRR), except for a couple of sanity checks on Reads-from (viz, coRW1-Exp) and Coherence-after (viz, coW{W,R}-Exp).

4. MTE relaxation

Adjacent to that Internal visibility rework is an MTE relaxation: only RW and WR pairs are now preserved, and this is reflected in hw-reqs and coWR-MTE respectively.


jalglave:
@maranget please might you (re)run all regressions to double check everything is in order?
Thanks in advance,
Jade

maranget:
> @maranget please might you (re)run all regressions to double check everything is in order? Thanks in advance, Jade

I cannot do this before this evening. Have you thought about having a kind file in the the new book aarch64-VMSA? It would be nice if we could include this book in the systematic regression test.

maranget:
Hi, as announced, we have failures on the systematic  regression  tests on MTE  (_e.g._  `make mte-test`, but there are others). Should we disable them as a temporary measure, or delay merge?

maranget:
Regression is fine

maranget:
Merged, thanks to all

PR #467, opened 2022-11-26
AArch64 VMSA rc

jalglave:

PR #470, opened 2022-11-30
[litmus] Fix for printing the name of "nop" argument.

maranget:
When NOP is given as an initial value, it gets tranalated into a "nop" argument. We manage to print it as the lowercase varsion of the instruction dump. Really a fix that won't resist adding more instructions as initial values.

This PR is inspired by PR #467 by @artkhyzha.


maranget:
Superseded by PR #469, close

PR #471, opened 2022-11-30
[AArch64 cat] ETS2 clarifications and MTE fixes after VMSA release

relokin:


jalglave:
Thanks @relokin. 
@maranget please could you regress this?

catalogue/aarch64-MTE/shelf.py
```diff
@@ -58,4 +58,1097 @@
...
+    "tests/MP+dmb.sttp+dmb.sy.litmus",
+    "tests/MP+dmb.sttp+dmb.sypt.litmus",
+    "tests/MP+dmb.sttp+dmb.sytp.litmus",
+    "tests/MP+dmb.sttp+dmb.sytt.litmus",
+    "tests/MP+dmb.sttp+dsb.sy.litmus",
+    "tests/MP+dmb.sttp+dsb.sypt.litmus",
+    "tests/MP+dmb.sttp+dsb.sytp.litmus",
+    "tests/MP+dmb.sttp+dsb.sytt.litmus",
+    "tests/MP+dmb.sttp.litmus",
+    "tests/MP+dmb.sttt+addr.litmus",
```
maranget, line 430: This test is not present in sub-directory test, nor in the kind file. We may either delete this line, or keep it. add  the test and a kind for  it. The model gives kind Allow. I attach the test for reference.
[MP+dmb.sttt+addr.litmus.txt](https://github.com/herd/herdtools7/files/10127417/MP%2Bdmb.sttt%2Baddr.litmus.txt)

maranget:
Except for the missing test local regression `make test` is fine

catalogue/aarch64-MTE/shelf.py
relokin: Good catch @maranget, I've removed it because we already have `MP+dmb.stTT+addr.litmus` and they are the same. I forced push the fix.

@jalglave the cat file passes the regressions (including ets2) even when we enumerate `tlbi-after` with no constraints (`0` instead of `hw-reqs`).

maranget:
Regressions (non-mixed, mixed, VMSA and VMSA+ETS2) are fine.

jalglave:
Great, thank you both.
Please can we merge?
Thanks in advance,
Jade

maranget:
Merged, thanks to all.

PR #473, opened 2022-12-04
[all] Rework fault recording and reporting

relokin:


maranget:
Hi @relokin, the UDF instruction makes sense even when not in self mode. So I suggest to check for faults in all modes. Modifications are minor, see this commit 0c9e1c5e3e05b3603f3633d7fbcdd2e9b53dfef0.

relokin:
Thanks for the review @maranget 

> Hi @relokin, the UDF instruction makes sense even when not in self mode. So I suggest to check for faults in all modes. Modifications are minor, see this commit [0c9e1c5](https://github.com/herd/herdtools7/commit/0c9e1c5e3e05b3603f3633d7fbcdd2e9b53dfef0).

Yes that makes sense! 



> The syntax of `IllegalInstruction` faults is redundant (_e.g._ `Fault(P0:L0,0:L0,IllegalInstruction)` as in tests A013). Wouldn't it be more natural to get rid of the second parameter, yielding: `Fault(P0:L0,IllegalInstruction)`. Of course this incurs changing the type of faults and fault atoms, but I think it is worth the trouble. I'll have a try in branch PR473, if you don't mind.

Fine by me, I was a little torn by this. I thought that made we can stick with the convention that the 1st field of the fault tuple is the label of the instruction that raised the fault (similar to reporting the PC in kvm-unit-test) and the 2nd the address of the faulting address (similar to reporting FAR_EL1 in kvm-unit-tests). However, it's clearly redundant so I would mind making on of the two optional for illegal instruction faults.

relokin:
> Hi @relokin, the UDF instruction makes sense even when not in self mode. So I suggest to check for faults in all modes. Modifications are minor, see this commit [0c9e1c5](https://github.com/herd/herdtools7/commit/0c9e1c5e3e05b3603f3633d7fbcdd2e9b53dfef0).

If we do this, we should also move the regression test to aarch64, there would be nothing more in the test that requires a variant.

maranget:
Here is my attempt for making the location (second) field of fault optional, commit eabc8f7fcfe3db85ad9b8d4e593edbf6c123e0a2.

I also attempted to handle illegal instruction faults in litmus (in kvm mode). This is not working yet. There are some elements I do not understand. Besides "udf" is not accepted as a mnemonic by the assembler. Try `.word 0`?

maranget:
> > Hi @relokin, the UDF instruction makes sense even when not in self mode. So I suggest to check for faults in all modes. Modifications are minor, see this commit [0c9e1c5](https://github.com/herd/herdtools7/commit/0c9e1c5e3e05b3603f3633d7fbcdd2e9b53dfef0).
> 
> If we do this, we should also move the regression test to aarch64, there would be nothing more in the test that requires a variant.

Indeed, I forgot to do it.

relokin:
also if you want I can pull in your two commits from https://github.com/herd/herdtools7/tree/PR473 and then fix the regression here.

maranget:
> also if you want I can pull in your two commits from https://github.com/herd/herdtools7/tree/PR473 and then fix the regression here (I can add changes to the PR473 branch)

Please do pull the two commits into your PR. Notice that litmus is still not functional.

relokin:
marking this as draft as I am working on adding support for UDF in litmus (kvm mode).

relokin:
UDF in litmus works now and it actually made me realize that UDF takes an `imm` as an argument (which is ignored by hardware but still). This PR should be ready for prime time now :)

jalglave:
Thanks very much Nikos.
Luc, please could we merge this soon?
Thanks in advance,
Jade

On 18 Feb 2023, at 01:22, Nikos Nikoleris ***@***.***> wrote:

﻿

⚠ Caution: External sender


UDF in litmus works now and it actually made me realize that UDF takes an imm as an argument (which is ignored by hardware but still). This PR should be ready for prime time now :)

maranget:
Hi @relokin, from testing I'd like, if you agree, to complete this PR with two commits.

1. The assembler `gas` on some machines does not accept the `udf` mnemonics. In such a case, emitting `.word 0` as an undefined instruction seems to work. The feature is triggered by variant `-variant NoUDF`. Commit 24f2124beabff75d96ae8f2a34b5cb373e3b1ebb.
2. In some contrived situation, function `idx_addr` was present in generated C code but not called, resulting in a `gcc` warning. Commit b4548f3fad5697176afb3cafa012668a002cb33d.

There is still a small discrepancy between `herd` and `litmus`, consider test:
```
AArch64 UDF02
Variant=fatal
{
int x;
0:X1=x;
1:X1=x;
[PTE(x)]=(af:0);
2:X1=PTE(x);
2:X0=(oa:PA(x),af:1);
}
  P0        |  P1         | P2          ;
MOV W0,#1   | LDR W0,[X1] | STR X0,[X1] ;
STR W0,[X1] | CBZ W0,L1   |             ;
            | UDF #0      |             ;
            |L1:          |             ;
exists Fault(P1,UndefinedInstruction) \/ Fault(P1,x)
```
Running the test with `herd7` we have:
```
...
States 3
  ~Fault(P1,x); ~Fault(P1);
 Fault(P1,UndefinedInstruction); ~Fault(P1,x);
 Fault(P1,x,MMU:AccessFlag);
...
```
While a machine run is as follows:
```
47    *> fault(P1,UndefinedInstruction); fault(P1,x,UndefinedInstruction);
999678*> fault(P1,MMU:AccessFlag); fault(P1,x,MMU:AccessFlag);
275   :> ~fault(P1); ~fault(P1,x);
```
Admittedly, this is a corner case. I am not very certain of what to output in the various cases that can occur here: `UDF` is executed, the read of `[x]` by `P1` fails, or no fault occurs. In any case, litmus output in the first two cases is definitively weird.

maranget:
Notice that if we make the observation of faults more precise, using labels;
```
AArch64 UDF03
Variant=fatal
{
int x;
0:X1=x;
1:X1=x;
[PTE(x)]=(af:0);
2:X1=PTE(x);
2:X0=(oa:PA(x),af:1);
}
  P0        |  P1            | P2          ;
MOV W0,#1   |L0:             | STR X0,[X1] ;
STR W0,[X1] | LDR W0,[X1]    |             ;
            | CBZ W0,L2      |             ; 
            |L1:             |             ;
            | UDF #0         |             ;
            |L2:             |             ;
exists Fault(P1:L1,UndefinedInstruction) \/ Fault(P1:L0,x)
```
Then herd and litmus outputs coincide. This is herd output:
```
Test UDF03 Allowed
States 3
  ~Fault(P1:L1); ~Fault(P1:L0,x);
 Fault(P1:L0,x,MMU:AccessFlag); ~Fault(P1:L1);
 Fault(P1:L1,UndefinedInstruction); ~Fault(P1:L0,x);
```
And this is litmus output:
```
Test UDF03 Allowed
Histogram (3 states)
1391  *> ~fault(P1:L0,x); fault(P1:L1,UndefinedInstruction);
22022 :> ~fault(P1:L0,x); ~fault(P1:L1);
976587*> fault(P1:L0,x,MMU:AccessFlag); ~fault(P1:L1);
```
Still noticing that `mcompare7` does not accept such output (but I'll manage here).

relokin:
> Hi @relokin, from testing I'd like, if you agree, to complete this PR with two commits.
> 
>     1. The assembler `gas` on some machines does not accept the `udf` mnemonics. In such a case, emitting `.word 0` as an undefined instruction seems to work. The feature is triggered by variant `-variant NoUDF`. Commit [24f2124](https://github.com/herd/herdtools7/commit/24f2124beabff75d96ae8f2a34b5cb373e3b1ebb).
> 

I am very tempted to pull in your patch and drop the NoUDF variant. Indeed gas [introduced](https://sourceware.org/pipermail/binutils/2020-April/110866.html) fairly recently. Would you be OK with litmus always emitting `.word 0`  for  `UDF #n`? ` This is the official encoding for `UDF #<imm16>`

`0 | 0 | 0 | 0 | 0 | 0 | 0 | 0 | 0 | 0 | 0 | 0 | 0 | 0 | 0 | 0 | imm16`

>     2. In some contrived situation, function `idx_addr` was present in generated C code but not called, resulting in a `gcc` warning. Commit [b4548f3](https://github.com/herd/herdtools7/commit/b4548f3fad5697176afb3cafa012668a002cb33d).
> 

Good catch, I'll add your commit to the PR, thanks!

> 
> There is still a small discrepancy between `herd` and `litmus`, consider test:
> 
> ```
> AArch64 UDF02
> Variant=fatal
> {
> int x;
> 0:X1=x;
> 1:X1=x;
> [PTE(x)]=(af:0);
> 2:X1=PTE(x);
> 2:X0=(oa:PA(x),af:1);
> }
>   P0        |  P1         | P2          ;
> MOV W0,#1   | LDR W0,[X1] | STR X0,[X1] ;
> STR W0,[X1] | CBZ W0,L1   |             ;
>             | UDF #0      |             ;
>             |L1:          |             ;
> exists Fault(P1,UndefinedInstruction) \/ Fault(P1,x)
> ```
> 
> Running the test with `herd7` we have:
> 
> ```
> ...
> States 3
>   ~Fault(P1,x); ~Fault(P1);
>  Fault(P1,UndefinedInstruction); ~Fault(P1,x);
>  Fault(P1,x,MMU:AccessFlag);
> ...
> ```
> 
> While a machine run is as follows:
> 
> ```
> 47    *> fault(P1,UndefinedInstruction); fault(P1,x,UndefinedInstruction);
> 999678*> fault(P1,MMU:AccessFlag); fault(P1,x,MMU:AccessFlag);
> 275   :> ~fault(P1); ~fault(P1,x);
> ```
> 
> Admittedly, this is a corner case. I am not very certain of what to output in the various cases that can occur here: `UDF` is executed, the read of `[x]` by `P1` fails, or no fault occurs. In any case, litmus output in the first two cases is definitively weird.

I agree the way litmus7 reports faults or their absence needs a little work.

maranget:
Hi @relokin. It looks ok to emit `.word 0` for `UDF`, as you confirm this the official encoding for `UDF #0`, which I di not really noticed. I have found another corner case: executimg `UDF` in `EL0`... (small commit 783d6b09b8600c6ccbda08b1a9d7e610d2f1c1b8).

As regards output, I am not so sure about what output should be. Hence herd output looks acceptable, as it does not contain the two weird lines:
```
47    *> fault(P1,UndefinedInstruction); fault(P1,x,UndefinedInstruction);
999678*> fault(P1,MMU:AccessFlag); fault(P1,x,MMU:AccessFlag);
...
```
However if we interpret `~fault(P1)` as "no fault at all by `P1` and `~fault(P1,x)` as no fault on location `x` access by `P1`, then
1. `fault(P1,UndefinedInstruction); ~fault(P1,x);` is fine.
2. `fault(P1,x,MMU:AccessFlag);` whithout `~Fault(P1)` is also fine, as there is a fault by `P1`.
3.  But,  `~fault(P1); ~fault(P1,x);` can be simplified as `fault(P1)` alone...

But well this a a corner case and current herd output is acceptable I guess.

litmus/libdir/_aarch64/kvm_fault_handler.c
```diff
@@ -16,9 +16,11 @@ static void fault_handler(struct pt_regs *regs,unsigned int esr) {
 }
 
 static void install_fault_handler(int cpu) {
-  install_exception_handler(EL1H_SYNC, ESR_EL1_EC_DABT_EL1,fault_handler);
+  install_exception_handler(EL1H_SYNC, ESR_EL1_EC_DABT_EL1, fault_handler);
+  install_exception_handler(EL1H_SYNC, ESR_EL1_EC_UNKNOWN, fault_handler);
 #ifdef USER_MODE
   struct thread_info *ti = thread_info_sp(user_stack[cpu]);
   ti->exception_handlers[EL0_SYNC_64][ESR_EL1_EC_DABT_EL0] = fault_handler;
+  ti->exception_handlers[EL0_SYNC_64][ESR_EL1_EC_UNKOWN] = fault_handler;
```
maranget, line 24: Sorry there is a typo here...
```suggestion
  ti->exception_handlers[EL0_SYNC_64][ESR_EL1_EC_UNKNOWN] = fault_handler;
```

maranget:
The above commit fixes a typo introduced by commit dc4c2ff04dccd8797e901e0ae2eadf44f119913a. Sorry for this.

maranget:
> @maranget, I haven't looked at `mcompare7` if you have done so already let me know, otherwise, I will have a look at it next.

The`mcompare7` tool looks functional. Using it I found another discrepancy, believe it or not:
```
AArch64 UDF01
Variant=fatal
{
int x;
0:X1=x;
1:X1=x;
[PTE(x)]=(af:0);
2:X1=PTE(x);
2:X0=(oa:PA(x),af:1);
}
  P0        |  P1         | P2          ;
MOV W0,#1   | LDR W0,[X1] | STR X0,[X1] ;
STR W0,[X1] | CBZ W0,L1   |             ;
            | UDF #0      |             ;
            |L1:          |             ;
~exists 1:X0=0 /\ Fault(P1,UndefinedInstruction)

```
Herd output:
```
% herd7 -variant vmsa UDF01.litmus
Test UDF01 Forbidden
States 3
1:X0=0;  ~Fault(P1);
1:X0=0; Fault(P1,x,MMU:AccessFlag);
1:X0=1; Fault(P1,UndefinedInstruction);
...
```
Litmus output
```
...
Test UDF01 Forbidden
Histogram (3 states)
1     :>1:X0=1; fault(P1,UndefinedInstruction);
99992 :>1:X0=0; fault(P1,MMU:AccessFlag);
7     :>1:X0=0; ~fault(P1);
...
```
Observe that  litmus output is  `fault(P1,MMU:AccessFlag)`, while herd output is `fault(P1,x,MMU:AccessFlag)` for flagging a MMU fault. I tend to think that herd is right, your opinion?

relokin:
Thanks for the review Luc!

I agree, in the case of UDF01, litmus7 would be more informative if it gave the output of herd7. I am going to have a look at it now.

maranget:
Hi @relokin, I agree with you suggestion of changing litmus output to match herd output. However..

 I was not expecting such a complexity. Something may be wrong in the way we consider printing faults. The underlying rules are unclear. More precisely, what a fault is is clear either an UndefinedInstruction perfiormed by a given thread at a given code location, or a MMU fault performed by a given thread at a given code location, on a given location and of a specific subtype. Howe those are printed in test output depending on fault "specufications" is less clear...

- If there is no fault specification for a given thread, there is no fault for this thread in test output regardless of this thread performing a fault or not.
- Faults specification may include an optional label.
- MMU fault specification have a mandatory location argument.
- The fault type is optional for MMU fault specifications.
- But the fault type for Undefined instruction does not look optional of `fault(P0)` is not a legit fault specification.

All the items above are syntactic, except the first one. The meaning of fault semantics, that is, their impact on what is printed in output is unclear.
- Negative fault output do not show fault types.  I tend to see, `"~fault(P0)` as meaning 'No fault at all by P0" and `~fault(P0,x)` as meaning no fault by `P0` on location `x`. I'd would not like to change this
- As regards MMU faults, a fault is flagged with its type event if the type of of the specification is different. In other words, the type of the specification is ignored.
- Still as regards MMU faults, if the fault is not on the specified variable and by constrast with the previous rule, there will be a negative fault (_e.g._ `~fault(P0,x)`) in output.
- The same rule as for variables applies to code labels. However if there is no label in specification there will be no label in fault output.

More to come...

relokin:
Right now litmus will only record whether any of faults in the postcondition occurred or not. I found that odd and I was in the process of rewriting the code we use to record faults in litmus7.

What I did was to record the fault info into a struct (process id, location in code/pc, location of data/FAR, type - all optional/invalid when they don't apply, except from the process id) which gets populated upon a fault regards of the postcondition of the test. It got quite a bit more complex when I considered multiple faults coming from the same Pn (assuming running in any mode other than fatal).

Then go through the postcondition and check if any of the recorded faults match any of the faults in the postcondition. I think this is more on par with what herd does and it would solve the problem. What do you think?

relokin:
>     * Negative fault output do not show fault types.  I tend to see, `"~fault(P0)` as meaning 'No fault at all by P0" and `~fault(P0,x)` as meaning no fault by `P0` on location `x`. I'd would not like to change this

Generally, I agree and I don't think we need to change this.

But also I could argue that we should have `~fault(..., fault_type)` 
* Take a test with `fault(P0,x,MMU:Translation)` in its postcondition and let's assume that we only recorded `fault(P0,x,MMU:AccessFlag)`. This means `~fault(P0,x,MMU:Translation)` for the purposes of checking the postcondition and asking the question did an access to x from P0 raise an MMU:Translation fault (the answer is no).

>     * As regards MMU faults, a fault is flagged with its type event if the type of of the specification is different. In other words, the type of the specification is ignored.
> 

You mean that litmus will print the type of fault it recorded, even if that doesn't match the fault type in the postcondition? I think that's ok, isn't it. This would ofc not validate the postcondition. 

>     * Still as regards MMU faults, if the fault is not on the specified variable and by constrast with the previous rule, there will be a negative fault (_e.g._ `~fault(P0,x)`) in output.

So if we have a MMU:Translation fault in y and not fault in x we would get `~fault(P0,x)` - I think that's ok too, isn't it?

> 
>     * The same rule as for variables applies to code labels. However if there is no label in specification there will be no label in fault output.
> 

I think, this is generally ok. Only if the litmus test happened to have a label where the fault happened it's possible to print this in the fault output. We don't do this right now but with the solution above does it.

maranget:
Hi @relokin, I like the idea to record faults completely (or as much as possible, cf. labels) when they occur and to see the specifications as patterns, The concept of a concrete fault matching a pattern is clear:  items not present in the pattern are some kind of wildcards (for instance `fault(P0,x,MMU:Permission)`matches Fault(P1,x)` but not `Fault(P0,x,MMU:AccessFlag)`. (Forr clarity fault is a concrete fault Fault is a pattern, although herd/litmus does not care).

It does not tell the whole story because faults (and the absence of faults) are present in output, as soon there is some fault "pattern" (with mandatory thread and some variable or "UndefinedInstruction" argument + optional items, _i.e. _ label and MMU: specification) in final condition. A possible rule would be to treat Undefined instruction faults and MMU faults independently (i am not so sure about it, I am elaborating)

 + If the concrete fault is an UndefinedInstruction and that the pattern filters the fault, print the fault, otherwise print a negated complete pattern (this is not what we do at the moment, we print `~Fault(P0)`).
+ If the concrete fault is a MMU fault and matches some pattern with explicit MMU type, including subtype,  then print the complete fault, otherwise, print also the fault.  I am  not very keen on printing a negated pattern, whith subtype, when the fault and the pattern differ only in subtype. Essentially because negative output predates the introduction of subtype with the meaning "no (MMU) fault" and that I associate the negative fault, (with a variable) with the absence of a (MMU) fault.
Consistently, if the pattern has a variable but no `MMU:` type, then the maching will be positive for all subtypes, ansd the complete fault will be printed.  A negative fault output means no MMU fault at all (on pattern variable, and code location if given).

Observe that there is no `Fault(P0)` pattern. If  we add it, maybe it would filter _any_ fault by `P0`, and such fault(s) would be printed completely. An output `~Fault(P0)` would mean absence of fault by `P0`.

As to print the label if the fault in output, even if not mentionned in pattern, why not.

 I am not sure that the  scheme above is totally backward coompatible. But I guess it should be backward compatible only on the 'well formed case` where we have at most one MMU fault specification per thread and no  UDF instruction nor UndefinedInstruction fault specification.

maranget:
Hum I have started to test various corner cases. Only considering MMU faults there already are differences between litmus and herd output. Now, I am hesitating between the risk of over engineering and of the benefit of a clear semantics.

So here is a suggestion: please finish your work on this PR, fixing UDF0. Let us test the PR without any  further corner case and postpone fancy corner cases. (I am accumulating fancy tests on my side)

relokin:
Thanks Luc. I agree the scope of this PR has gotten out of hand :)

I suppose you wanted to address the mismatch in the output for `UDF01` between herd7 and litmus7? I am not sure there is an easy/incremental way to do that. With the current approach, litmus7 wouldn't generate code for recording information about x (no `see_fault_P1_x` just `see_fault_P1`).

But I am happy to implement the solution I described above, where `record_fault` would save as much information about the fault in a struct. What do you think?

maranget:
Please consider implementing your solution "where `record_fault` would save as much information about the fault in a struct. ". If you do so,  it would probably be much easier to synchronise litmus and herd outputs.

relokin:
hey @maranget, I think I am relatively happy with this PR. It now does a few things at once and as we've discussed:

- Allows for a more flexible specification of faults in postconditions: `fault(PN[:lLABEL][,LOCATION][,FAULT_TYPE])`
- Both herd and litmus consider the full specification of a fault in the postcondition. For example:
 
if we have `fault(P0)` in the postcondition and an execution encounters:
`Fault(P0:L0,x,MMU:Translation)`, `Fault(P0,UndefinedInstruction)` and `Fault(P1:L1,TagCheck)` 
litmus and herd will report 
`Fault(P0:L0,x,MMU:Translation)`, `Fault(P0,UndefinedInstruction)`

if we have: `fault(P0:L0,x,MMU:Translation)` in the postcondition and an execution encounters: `Fault(P0:L0,x,MMU:AccessFlag)`
litmus and herd will report:
`~Fault(P0:L0,x,MMU:Translation)`

- Renames `IllegalInstruction` to `UndefinedInstruction`
- Adds support for the UDF instruction in herd7 and litmus7

relokin:
@maranget did you have a chance to have a look at this? Sorry I don't mean to pressure you but I am working on litmus again and I would like to know if I should work on top of this branch (if we're likely to accept this with minor modifications) or not (if we decide that this is not the right direction).

relokin:
> > @maranget did you have a chance to have a look at this? Sorry I don't mean to pressure you but I am working on litmus again and I would like to know if I should work on top of this branch (if we're likely to accept this with minor modifications) or not (if we decide that this is not the right direction).
> 
> Hi @relokin, sorry for the delay. I'll have a look as soon as possible.

Many thanks @maranget let me know if you have any questions :)

litmus/libdir/_aarch64/kvm_fault_handler.c
```diff
@@ -16,9 +45,11 @@ static void fault_handler(struct pt_regs *regs,unsigned int esr) {
...
```
maranget, line 53: Typo
```suggestion
  ti->exception_handlers[EL0_SYNC_64][ESR_EL1_EC_UNKNOWN] = fault_handler;
```

maranget:
Hi @relokin. Thanks a lot for this nice PR. I have found a typo (see above) and resolved an (easy) conflict with master.
Given the numerous commits, I'd rather have you to squash them into one commit, before I merge in the usual way.

Thanks again.

maranget:
FYI I just found a discrepancy. Stay tuned...

maranget:
Here it is, test:
```
AArch64 UDF06
Variant=fatal
EL0=P0 P1
{
int x;
0:X1=x;
1:X1=x;
1:X2=0;
2:X1=PTE(x);
2:X0=(oa:PA(x));
[PTE(x)]=(db:0);
}
  P0        |  P1         |  P2         ;
MOV W0,#1   | LDR W0,[X1] | STR X0,[X1] ;
STR W0,[X1] | CBZ W0,L1   |             ;
            | UDF #0      |             ;
            |L1:          |             ;
            | MOV W2,#1   |             ;
locations [0:X1;fault(P0); fault(P1);]
exists 1:X2=0 /\ fault(P0,x)
```

Herd output, observe that `Fault(P0,x,MMU:Permission)` occurs once in the third of the listed final states:
```
Test UDF06 Allowed
States 3
0:X1=x; 1:X2=0; Fault(P1,UndefinedInstruction); ~Fault(P0,x); ~Fault(P0);
0:X1=x; 1:X2=1;  ~Fault(P1); ~Fault(P0,x); ~Fault(P0);
0:X1=x; 1:X2=1; Fault(P0,x,MMU:Permission); ~Fault(P1);
No
Witnesses
Positive: 0 Negative: 6
Condition exists (1:X2=0 /\ fault(P0,x))
Observation UDF06 Never 0 6
Time UDF06 0.01
Hash=f2dbe6a947a1d06cd195eae1bb0454be
```
And now, litmus output, `Fault(P0,x,MMU:Permission)` is listed twice...:
```
Test UDF06 Allowed
Histogram (3 states)
5     :>0:X1=x; 1:X2=0; ~fault(P0,x); ~fault(P0); fault(P1,UndefinedInstruction);
999971:>0:X1=x; 1:X2=1; ~fault(P0,x); ~fault(P0); ~fault(P1);
24    :>0:X1=x; 1:X2=1; fault(P0,x,MMU:Permission); fault(P0,x,MMU:Permission); ~fault(P1);
No

Witnesses
Positive: 0, Negative: 1000000
Condition exists (1:X2=0 /\ fault(P0,x)) is NOT validated
Hash=f2dbe6a947a1d06cd195eae1bb0454be
```
I am not sure this discrepancy prevents merging. However, it may be easy to fix.

maranget:
Here is a simplified test with similar behaviour (the same fault occurs twice in a "final state" line of litmus output):
```
AArch64 UDF07
Variant=fatal
EL0=P0
{
int x;
0:X1=x;
1:X1=PTE(x);
1:X0=(oa:PA(x));
[PTE(x)]=(db:0);
}
  P0        |  P1         ;
MOV W0,#1   | STR X0,[X1] ;
STR W0,[X1] |             ;
locations [fault(P0);]
exists 1:X2=0 /\ fault(P0,x)
```

litmus/preSi.ml
```diff
@@ -752,42 +764,33 @@ module Make
...
+            begin
+              O.o "static int idx_addr(intmax_t *v_addr,vars_t *p) {" ;
+              begin match test.T.globals with
+              | _::_ when Cfg.is_kvm ->
+                 O.oi  "intmax_t *p_addr =" ;
+                 O.oii "(intmax_t *)((uintptr_t)v_addr & PAGE_MASK);"
+              | _ -> ()
+              end ;
+              List.iter dump_test test.T.globals ;
+              O.oi "fatal(\"Error can find symbol for faulting address\"); return -1;" ;
```
maranget, line 788: Typo:
```suggestion
              O.oi "fatal(\"Error cannot find symbol for faulting address\"); return -1;" ;
```

maranget:
"Esprit d'escalier"... Output has slightly changed, as fault types can now appear as arguments of negative faults. Here is a fix for the tool lexer: 59b55e81d049a7e241398323acab5029bb4c11ed.

maranget:
> I think the fix was easy and I implemented it in [7539af2](https://github.com/herd/herdtools7/commit/7539af2a6209c8f2ef4937ab8a3329542980e858) .

Unfortunately, we are not done yet, I am afraid. Test:
```
AArch64 UDF08
Variant=fatal
{
int x;
0:X1=x;
1:X1=x;
1:X2=0;
[PTE(x)]=(af:0);
2:X1=PTE(x);
2:X0=(oa:PA(x),af:1);
}
  P0        |  P1            | P2          ;
MOV W0,#1   |L0:             | STR X0,[X1] ;
STR W0,[X1] | LDR W0,[X1]    |             ;
            | CBZ W0,L2      |             ;
            | MOV W2,#1      |             ;
            |L1:             |             ;
            | UDF #0         |             ;
            |L2:             |             ;
forall  1:X2=1  => (Fault(P1:L1,UndefinedInstruction) /\ ~Fault(P1:L0,x))
```
Notice the condition above: if 1:X2 is one, then there must be a UDF fault. Here is litmus output:
```
Test UDF08 Required
Histogram (3 states)
240   :>1:X2=0; ~fault(P1:L0,x); ~fault(P1:L1,UndefinedInstruction);
999194:>1:X2=0; fault(P1:L0,x,MMU:AccessFlag); ~fault(P1:L1,UndefinedInstruction);
566   :>1:X2=1; ~fault(P1:L0,x); 
Ok

Witnesses
Positive: 1000000, Negative: 0
Condition forall (1:X2=1 => (fault(P1:L1,UndefinedInstruction) /\ not (fault(P1:L0,x)))) is validated
```
In the last final state line, `fault(P1:L1,UndefinedInstruction)` is not displayed, while it should. This is indeed a printing problem as there are zero negative withness....

```

**Comment by maranget** [comment]
One fix may be to re-initialize the arrray `flt_reported` before printing each line of the final state account.

**Comment by relokin** [comment]
Thanks @maranget I missed that. The fixup implements your suggestion. Again not squashing for now to make sure, it's easy to see what's changed.

**Comment by maranget** [comment]
I stop for today, I have a huge test running on kylo. I'll check this tomorrow.

File: litmus/libdir/_aarch64/kvm_fault_type.c [hunk 4]
```diff
@@ -1,34 +1,172 @@
...
+  if (data_symb != DATA_SYMB_ID_UNKNOWN)
+    printf(",%s", data_symb_name[data_symb]);
+  if (ftype != FaultUnknown)
+    printf(",%s", fault_type_names[ftype]);
+  printf(");");
+}
+
+static bool fault_reported[NTHREADS][MAX_FAULTS_PER_THREAD];
+
+void pp_log_faults_init()
```
maranget, line 103: This function must be declared static, as there may be several of them with option `-driver C`
```suggestion
static void pp_log_faults_init()
```
Then the function has to be called at least once, so as to avoid a gcc warning.

maranget:
Fix attempt, 46dd37caa53774c377d2ab7c57c1b6ed60bf208e. I am using it for a huge litmus run.

maranget:
The litmus run on kylo and d06 looks fine. I guess you can squash commits and I'll merge.

BTW, there is an interesting consequence of printing faults types in mode imprecise/handled. Consider the case where PTE(x) is initially (oa:PA(x),db:0) (hence a "user" thread write attempt to x yields a MMU:Permission fault). Now consider a concurrent "system" thread that performs BBM, _i.e._ invalidates PTE(x) by writing (valid:0) into it, and then writes (oa:PA(x)) into PTE(x). 

In handled mode, the write is attempted until success. Then the situation where MMU:Permission and MMU:Translation are recorded is  possible and observed. By contrast herd does not produce such a result, as the write is attempted twice only, while three attempts are needed to yield two faults and a success.

maranget:
One advantage of "big commits" is to facilitate bisection, at technique that some users may exploit when looking for the place where a bug has been introduced. This means that at least compilation and at best "make test" are functional after each commit. If compilations is working, we can merge. Is it the case?

maranget:
Ok I have just checked, compilation after each commit works.

Let's merge finally, @relokin, thanks a lot.

relokin:
Thanks @maranget, I tried compiling after each commit over night and didn't run into any errors much like you.

Thanks for merging this :)

PR #474, opened 2022-12-05
[herd] Introduce relation intra_causality_order

maranget:
This new intra-instruction relation corresponds to the situation when some event must precede another even in the absence of output use (data) and existence dependence (control).


maranget:
Complte regression pending

jalglave:
Thanks Luc. Please can we merge?

PR #478, opened 2022-12-07
Generalise the type of PTE locations and add support for recursive PTEs

relokin:
This PR is an attempt to unify the locations for PTEs and normal data. Previously, `herd7` treated PTE locations as special `System` locations and this PR changes that to make them regular `Physical` locations which can be mapped into `Virtual` locations too. As a result, we can now simulate tests with recursive PTEs as shown in the two regression tests. Internally, we make the convention, that the location of the PTE of `Virtual` location `x` is the `Physical` location `pte_x` (PTE(x) in the syntax of a litmus test).

Unfortunately this PR is not ready, but as I will be on leave for the next month, I wanted to share my progress. There are three issues that I am aware of:

- Spurious AF updates are not working. For example, `catalogue/aarch64-VMSA/tests/LDRaf0-HA.litmus` will show one execution instead of two, the one with the spurious update is missing.
- Simulation time has significantly increased for some tests. The problem seems to be that PTE are now regular virtual/physical locations and therefore the solver has many more possible solutions to explore.
- I haven't tested litmus

Any feedback will be very welcome!

### DIFF
--- a/herd/AArch64Sem.ml
+++ b/herd/AArch64Sem.ml
@@ -416,12 +416,12 @@ module Make
       let mextract_whole_pte_val an nexp a_pte iiid =
         (M.do_read_loc false
            (fun loc v ->
-             Act.Access (Dir.R,loc,v,an,nexp,quad,Access.PTE))
+             Act.Access (Dir.R,loc,v,an,nexp,quad,Access.PHY))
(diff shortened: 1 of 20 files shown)


relokin:
Closing this as I will need to rework and rebase.

PR #479, opened 2022-12-08
[herd] New ADDR set for address ports

maranget:
Previous definition of address ports, which are crucial to addr dependencies, was a bit too loose. The new definition considers that the address ports of an instruction are the read _register_ effects that are not data ports.

### DIFF
--- a/herd/tests/instructions/AArch64/L041.litmus
+++ b/herd/tests/instructions/AArch64/L041.litmus
@@ -1,0 +1,12 @@
+AArch64 L041
+Orig=DSB.STdWR Fre DSB.STdWR Fre
+{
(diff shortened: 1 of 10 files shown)


maranget:
Regressions are fine. That is no old regression test has changed behaviour. New tests control the small relaxation introduced,

chacon01:
Thanks Luc. This appears to fix the unusual behavior I was observing.

jalglave:
Fantastic. Luc please can we merge?
Thanks in advance,
Jade

On 9 Dec 2022, at 16:51, chacon01 ***@***.***> wrote:

﻿

⚠ Caution: External sender


Thanks Luc. This appears to fix the unusual behavior I was observing.

PR #480, opened 2022-12-08
Tests improvements

HadrienRenaud:
This is a two features PR:
 1. when executing tests (e.g. `make test`) on a machine where `python` does not exists, the test program falls back on to `python3`.
 2. it is now possible to test with variants in `herd_diy_cross_regression_test` with the option `-variant`, as the other test programs in internal.

With 2, you can now run all the regression tests with a variant (here `T93`) in a simple command:

    make test REGRESSION_TEST_MODE="-variant T93 test"


maranget:
Thanks @HadrienRenaud

PR #484, opened 2022-12-13
[AArch64 cat] adjust cat file after better redefinition of ADDR set

jalglave:
@maranget


maranget:
@jalglave Thanks, merged.

PR #490, opened 2022-12-21
[AArch64, herd] Support for DIC and IDC features

artkhyzha:
This PR sketches out support for enabling ifetch-related functionality that is indicated in the DIC and IDC bits of Cache-Type Register (CTR_EL0). These features are supposed to grant extra powers to stores: when IDC is enabled, the store does not need to be superseded by DC, and when IDC and DIC are enabled, it does not need to be superseded by DC and IC.

The objective is to be able to indicate in the syntax of litmus test that the features are enabled similarly to TTHM for the virtual memory. That is, the PR adds an option "CacheType" that can have entries "DIC" and "IDC" to enable the functiinality globally, and "Pn:DIC" and "Pn:IDC" to enable it for the `Pn`. In the current solution, enabling DIC automatically enables IDC for the specified scope.

Another objective is to be able to export to cat files the sets `DIC` (and `IDC`) that are meant to represent the subset of `W` such that DIC (IDC) is enabled in the corresponding `Pn`.

The following litmus test illustrates the use of the new option.
```
AArch64 MP.FF+dmb.st+po
CacheType=P0:DIC P0:IDC
{
0:X0=NOP; 0:X1=P1:Lself00; 0:X2=P1:Lself01;
1:X1=P1:Lself00;
}
 P0          | P1         ;
 STR W0,[X1] | Lself01:   ;
 DMB ST      |  B Lself00 ;
 STR W0,[X2] |  MOV W10,#1;
             | Lself00:   ;
             |  B L01     ;
             |  MOV W11,#1;
             | L01:       ;
exists (1:X10=1 /\ 1:X11=0)
```


artkhyzha:
Curious what your thoughts are, @jalglave and @maranget. Does this look like a reasonable syntax? Is the implementation adequate? Any feedback would be welcome!

lib/cacheType.mll
```diff
@@ -0,0 +1,91 @@
...
+  let f _ = false in
+  { dic=f; idc=f; }
+
+let filter_opt f =
+  List.fold_left
+    (fun k p -> match f p with
+    | None -> k
+    | Some x -> x::k)
+    []
+
```
maranget, line 58: This function would better be defined in `lib/misc.ml`. Then this file and `dirtyBit.mll` would call `Misc.filter_opt`.

lib/cacheType.mll
```diff
@@ -0,0 +1,91 @@
...
+let get info =
+ match
+   MiscParser.get_info_on_info
+     MiscParser.cache_type_key info
+ with
+ | None -> None
+ | Some s ->
+    try
+      let xs = all [] (Lexing.from_string s) in
+      let idcs = List.filter (function (_,(IDC|DIC)) -> true) xs
```
maranget, line 68: Function `(function (_,(IDC|DIC)) -> true) `  always returns `true`. As a consequence, `idcs` is alwys equal ro `xs`. Is this intended?

This code is apparently inspired from `dirtyBit.mll`, where `HD=> HA`. I am not sure that we have DIC => IDC. Do we?

lib/cacheType.mll
```diff
@@ -0,0 +1,91 @@
...
+          (fun _ -> true)
+        else
+          let xs =
+            filter_opt
+              (function (Some _ as p,DIC) -> p | _ -> None)
+              dics in
+          fun proc -> List.exists (Misc.int_eq proc) xs in
+      Some { dic; idc; }
+    with Error ->
+      Warn.user_error "Incorrect dirty bit managment specification '%s'" s
```
maranget, line 90: The message refers to dirty bit.

lib/cacheType.mll
artkhyzha: Oh right, of course the `idcs` could be simplified. Will fix soon.

Yes, the current intention is to have DIC => IDC.

maranget:
Hi @artkhyzha I have made a few comments. Overall the code is clean and the syntax looks fine. I would have liked better  not to define the sets in `machModelChecker.ml`. But this would probably result in a more complex implementation. So let us keep it like that.

lib/cacheType.mll
artkhyzha: Thanks, fixed now.

lib/cacheType.mll
maranget: If the function is present in all supported versions of OCaml, it is a good idea to use it in place of the Misc function. Even better, we can delete the Misc function and replace its calls by calls to the standard library function everywhere. Would you please do it?

artkhyzha: Understood, thanks for confirming! The change is already available in the last commit. (Not so many uses of these functions, actually.)

maranget:
Merged, thanks @artkhyzha

PR #492, opened 2023-01-04
[herd] Non temporal memory accesses (AArch64)

maranget:
This PR handles non-temporal memory accesses. They have a noticeable impact over the memory models, as an address dependency that ends into a non-temporal read does _not_ create order.


jalglave:
Thanks @maranget. Please could you add a test to illustrate the required change too?
thanks in advance,
Jade

maranget:
Here you are. Please see test [L070](https://github.com/herd/herdtools7/pull/492/files#diff-b6cc5dcc3ea65f2ee6f088bce2ca20436d1e8769ab313744686532a8c4c5562e) and  commit 71adffa58bcf97f3fa0a3caa039876d407875803.

PR #495, opened 2023-01-17
[all] CPYF and SET instructions, WIP.

maranget:
Work in progress


maranget:
Should be merged after #492 (or includes it...)

PR #496, opened 2023-01-17
[lib]: LDXP Pretty Printing

lukeg101:
Consider the instruction `LDXP X1, X2, [X3]`

When printed using `pp_ldxp` in `lib/AArch64Base` the first register is printed as an `W`reg when it should be printed as an `X` reg.

This patch amends pretty printing to this effect.
Side effect: Hashes in tests are updated

### DIFF
--- a/lib/AArch64Base.ml
+++ b/lib/AArch64Base.ml
@@ -1117,7 +1117,7 @@ let do_pp_instruction m =
 
   let pp_ldxp memo v r1 r2 r3 =
     pp_memo memo ^ " "
-    ^ pp_wreg r1 ^","
+    ^ (if m.compat then pp_wreg r1 else pp_vreg v r1) ^","
     ^ pp_vreg v r2 ^ ",["
     ^ pp_xreg r3 ^ "]" in
 


maranget:
See [this comment](https://github.com/herd/herdtools7/pull/500#issuecomment-1386165631) for a general comment on changing hashes. However, in this specific case, it may be quite cumbersome to apply the best solution to retain the old output for hashes while introducing the new (correct) output for pretty-printing. I have to check if the suggested change impacts our precious logs.

maranget:
Well in fact, as I have just checked, implementing the best solution of retaining the old output for hashes while correcting pretty-printing is easy. In fact a very similar problem already occurred with the `ADR` instruction which was wrongly printed as `ADDR`. The relevant code is [here](https://github.com/herd/herdtools7/blob/9c7cb7d6be30430c8eed3313e6ed2e16ce8f3ad5/lib/AArch64Base.ml#L1379).

lukeg101:
Hi Luc, 
Thanks for your feedback, I have conditionalised pretty printing on `m.compat`, and hashes are no longer effected :)

PR #497, opened 2023-01-17
[lib]: Added mkdir -p helper function

lukeg101:
This function makes a linux call to `mkdir -p` to create a directory if it does not already exist.

I use this in a later commit and am uploading this first.


maranget:
Can be useful, thanks.

PR #498, opened 2023-01-17
[lib]: Added uniq_lists to MySet

lukeg101:
Helper function takes lists of lists and gets the set of all elements

I use this function in a later commit, so I am uploading it first


maranget:
Can be useful, thanks.

PR #499, opened 2023-01-17
[lib]: Fixed constant pretty printing with offsets

lukeg101:
This patch fixes the pretty printing of tests which offset symbolic addresses.

It also adds a helper function to make a symbolic location with an offset (which is used in a later patch)

### DIFF
--- a/herd/tests/instructions/AArch64.neon/V43.litmus.expected
+++ b/herd/tests/instructions/AArch64.neon/V43.litmus.expected
@@ -1,6 +1,6 @@
 Test V43 Required
 States 1
-0:X0=x+4; [y]=1; [z]=2;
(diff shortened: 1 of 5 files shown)


maranget:
Symbolic addresses with offsets were printed as `x+o` when `o` is the numeric offset. This PR changes this output t `x[o]`.

The former notation looks better to me as the second notation more looks as a reference to an array cell rather than an address, As a consequence I am not very keen on performing this change.

lukeg101:
That is fair enough, I cannot find a good reason to keep this patch, If I remember I will comment here in the future

lukeg101:
created https://github.com/herd/herdtools7/pull/502

PR #500, opened 2023-01-17
[lib]: Prints void for C tests - makes valid C

lukeg101:
When generating C tests, this patch prepends the 'void' type which is required to compile tests that do not return data after execution.

### DIFF
--- a/herd/tests/instructions/C/C01.litmus.expected
+++ b/herd/tests/instructions/C/C01.litmus.expected
@@ -6,5 +6,5 @@ Witnesses
 Positive: 1 Negative: 0
 Condition forall (0:r0=0x1)
 Observation C01 Always 1 0
-Hash=a5df714c669cbe459e22228846733357
+Hash=f26e77b9ef2528a6e6c9757a77f150a1
 
--- a/herd/tests/instructions/C/C01.litmus.expected
+++ b/herd/tests/instructions/C/C01.litmus.expected
@@ -6,5 +6,5 @@ Witnesses
 Positive: 1 Negative: 0
 Condition forall (0:r0=0x1)
 Observation C01 Always 1 0
-Hash=f26e77b9ef2528a6e6c9757a77f150a1
+Hash=a5df714c669cbe459e22228846733357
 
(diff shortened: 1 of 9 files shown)


maranget:
Changing hashes is never to be done lightly as consequences can be important. As a matter of fact, there may exists numerous logs that refer to the hashes. This is unlikely in the case of this PR. Nevertheless, I tend not to changes hashes without a very compelling reason.

Most of the time it is possible to retain the old output for hashes only. Here it would suffice to add a boolean argument to the `dump_prog`. When true, output according to the old way. when false output following the new way. Alternatively, the `DumpCast` module  can export two functions `dump_prog` and `dump_prog_for_hash` and clients would call the appropriate function.

lukeg101:
Hi Luc, I have added a boolean `preserve_hashes` argument to `dump_prog`. Please let me know if this is ok

maranget:
Hi @lukeg101, I could not notice any difference in output with the master branch. For instance I tried `diyone7 -arch C PodWW Rfe PodRR Fre` and some `litmus7` calls on the resulting C test. Accordingly  I could not find any call to `CAst.dump_prog` with the `false` as the value of the`preserve_hashes` argument.

Hence, in its current status, the intend of this PR still unclear to me. Could you tell us more, please?

lukeg101:
Hi Luc, I use `dump_prog false ...` in a later commit to produce compilable litmus tests. Apologies this is another patch adding helper feature for a later patch. Is this ok?

maranget:
May I suggest adding this PR in from of the client  PR. That way we have complete motivation for both PR.

maranget:
Thinking twice, it may be that the client PR is huge?

lukeg101:
Hi Luc, Yes the client PR is huge, my upstream plan is as follows:
- Upstream all patches in e.g. lib, herd _before_ the client that are not strictly related to the client PR - e.g. instruction semantics, bug fixes
- The client PR (separate from lib, herd etc...)

maranget:
Ok, fine, I'll merge this PR soon.

PR #502, opened 2023-01-19
[lib]: added mk_sym_with_index helper function

lukeg101:
This patch adds a helper function used in a later commit. `mk_sym_with_index` allows us to create symbolic location `v` with a specified offset `i`


maranget:
LGTM, thanks.

PR #505, opened 2023-01-20
[lib]: Added System register parsing

lukeg101:
This patch adds lexing support for a series of system registers that I have encountered.

### DIFF
--- a/lib/AArch64Lexer.mll
+++ b/lib/AArch64Lexer.mll
@@ -463,6 +463,12 @@ match name with
 (* System registers *)
 | "mrs"|"MRS" -> MRS
 | "msr"|"MSR" -> MSR
+| "ctr_el0"|"CTR_EL0" -> SYSREG A.CTR_EL0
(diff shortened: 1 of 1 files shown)


maranget:
Hi @lukeg101, I am not sure this PR to be necessary. A mechanism to lex system registers already exists in the master branch, by calling the function `parse_sysreg` defined in file `AArch64Base.ml`, [called here](https://github.com/herd/herdtools7/blob/f570642268960828b60ea1ddad7eb261f63d30c9/lib/AArch64Lexer.mll#L490). Although I have not checked,  I see no reason for this function not to be able to lex all system registers defined in `AArch64Base.ml`.

Notice that this somehow contrived mechanism of calling an external function instead of lexing registers directly has several advantages: (1) maintain compatibility between input and output (of registers); (2) limit the size of the lexing automaton; (3) share code, as registers have to be lexed in other places (although I am not sure that this last reason applies to system registers, well).

lukeg101:
Hi Luc,
Thanks for this, I did not realise these registers had been added, It seems I have overlapped work :) 
I agree with the above and will close this PR as the patch is no longer needed

PR #507, opened 2023-01-23
[all]: Added UBFM/SBFM Support

lukeg101:


lukeg101:
This is a WIP, I'd appreciate your advice on coding style to make sure there isn't a simpler way to implement this I have missed. Thanks

maranget:
Looking at the "File changed" menu,   some spurious files with .rej and .orig extensions are present.

lukeg101:
Apologies, I missed those. Now removed.

herd/AArch64Sem.ml
```diff
@@ -2399,6 +2399,25 @@ module Make
...
+          let dec_mask = int_of_string (Printf.sprintf "0b%s" (String.concat "" bitmask)) in
+          let hex_mask = Printf.sprintf "0x%x" dec_mask in
+          let rec log2 x = match x with
+            | 1 -> 0
+            | _ -> 1 + log2 ((x+1)/2) in
+          read_reg_data sz rn ii
+          >>= fun v -> M.op1 (Op.AndK hex_mask) v
+          >>= fun v -> M.op Op.ShiftRight (V.intToV (log2 kr-ks)) v
+          >>= fun v -> write_reg rd v ii
+          >>= B.next1T
```
maranget, line 2419: I do not have time for a complete review. Here is a remark. Computation at the assembler level are in `Int64`. Hence, one can use the same technique as above, but on integers of 64 bit size. For instance `int_of_string` can be replaced by `Int64.of_string` and format `"0x%x"` by `"0x%Lx"`. However I have not checked boundary cases.

Besides, I am not so sure about using the `log2` function. Shouldn't the shift amount be `kr-ks`?

Finally shouldn't there be two cases depending upon the relative position of kr and ks?

lukeg101:
Sorry about my delay on this, I had this commit around but I thought it was complete when I submitted it - `SBFM` is difficult and is fighting back. 

In particular I have 3/4 test cases working now, however I need to find a way to sign extend a bit pattern to the length of the destination register and not simply a fixed number of bits away. `SXTW` assumes the MSB in a register is the sign bit and so we can soundly sign extend from that bit position a fixed number of bits (32 or 64). However SBFM can extract a field with a sign bit from an arbitrary point in a register, where the MSB of _that field_ is the sign bit for that field. I cannot simply use `sxtw_op` to copy the field and zero extend the remaining size of the register as the number of sign extension bits in sxtw is fixed (32 or 64) whereas for SBFM we need potentially more (if we extract a field from, for example position 30)

lukeg101:
Hi Luc, this was a tough one, but I've got the tests cases passing in both herd and litmus (testing under apple M1)

maranget:
Hi @lukeg101, we are not through with it, I am afraid. I have made two small changes (see my recent commits to your branch) and discovered one discrepancy between hardware and `herd`: the `SBFM` does not copy the high order bit of the source bitfield into the destination register. More specifically, herd sets the final value of `0:X2` and `0:X5` to `0x1`. The condition of the test gives the same values as hardware.
```
AArch64 T
{
uint32_t 0:X0=0xF;
uint32_t 0:X1;
uint32_t 0:X2;
uint64_t 0:X3=0xF;
uint64_t 0:X4;
uint64_t 0:X5;
}
    P0           ;
UBFM W1,W0,#0,#0 ;
SBFM W2,W1,#0,#0 ;
UBFM X4,X3,#0,#0 ;
SBFM X5,X4,#0,#0 ;

forall 0:X1=0x1 /\ 0:X2=0xffffffff
/\ 0:X4=0x1 /\ 0:X5=0xffffffffffffffff
```

maranget:
Hi @lukeg101, I had a little free time and could have a try at implementing the msb-extension. I am not so sure about it. A few more tests, in particular on `W` registers and `SBFM` instruction., would be great.

lukeg101:
Thanks Luc! we were working on the same thing at the same time :) I will add tests some tests

lukeg101:
apologies for the force push, I used the wrong test name in the commit message

lukeg101:
More tests added now

maranget:
I have added two tests, your opinion on merging?

lukeg101:
Those tests look good, thank you. I think this is ready to merge however I think it motivates rewriting some other instructions that alias UBFM/SBFM. The Arm ARM for SBFM says:

> "This instruction is used by the aliases ASR (immediate), SBFIZ, SBFX, SXTB, SXTH, and SXTW. See Alias conditions on page C6-1778 for details of when each alias is preferred."

This would allow us to confirm that the above is correct, but also that `ASR`, `SXTW` etc... were correct as before. This also matches the assembly output of `litmus` which prefers aliases e.g. `SBFIZ` over printing `SBFM` etc... I think this is a separate task

maranget:
I agree with you. Implementing these aliases is yet another PR...

PR #508, opened 2023-01-24
[diy7] support for stdout and cycle only output

hugookeeffe:
This PR implements support for 2 new options. 
-stdout allows the user to output litmus tests to stdout
-cycleonly allows the user to output only the cycle name of a test and not the main body.

This code was written with support from @murzinv

### DIFF
--- a/gen/config.ml
+++ b/gen/config.ml
@@ -253,7 +253,7 @@ let common_specs () =
    ("-no", Arg.String (fun s -> no := Some s),
      "<fname> do not generate tests for these cycles")::
   ("-stdout", Arg.Bool (fun b ->  stdout := b),
-    "output to stdout (default false)")::
+    "output to stdout. If Cycleonly is true, then this is implied to be true. (default false)")::
   ("-cycleonly", Arg.Bool (fun b ->  cycleonly := b),
     "output only cycle, i.e. no litmus body (default false)")::
   []
--- a/gen/config.ml
+++ b/gen/config.ml
@@ -51,6 +51,8 @@ let bell = ref None
 let scope = ref Scope.No
 let variant = ref (fun (_:Variant_gen.t) -> false)
 let rejects = ref None
+let stdout = ref false
+let cycleonly = ref false
 
 let info = ref ([]:MiscParser.info)
 let add_info_line line = match LexScan.info line with
@@ -250,6 +252,10 @@ let common_specs () =
      sprintf "<i> size of integer added to base name that yield test names (default %i)" !fmt)::
    ("-no", Arg.String (fun s -> no := Some s),
      "<fname> do not generate tests for these cycles")::
+  ("-stdout", Arg.Bool (fun b ->  stdout := b),
+    "output to stdout (default false)")::
+  ("-cycleonly", Arg.Bool (fun b ->  cycleonly := b),
+    "output only cycle, i.e. no litmus body (default false)")::
   []
 
 let numeric = ref true
--- a/gen/diy.ml
+++ b/gen/diy.ml
@@ -239,6 +239,8 @@ let () =
     let addnum = !Config.addnum
     let numeric = !Config.numeric
     let lowercase = !Config.lowercase
+    let stdout = !Config.stdout
+    let cycleonly = !Config.cycleonly
 (* Specific *)
     open Config
     let choice = !Config.mode
--- a/gen/diy.ml
+++ b/gen/diy.ml
@@ -239,7 +239,7 @@ let () =
     let addnum = !Config.addnum
     let numeric = !Config.numeric
     let lowercase = !Config.lowercase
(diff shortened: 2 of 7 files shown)


hugookeeffe:
@maranget what do you think of this PR, is there anything you think I could improve?

maranget:
> @maranget what do you think of this PR, is there anything you think I could improve?

Hi @hugookeeffe, since you are asking.. 
1. I do not see the usage of `-stdout true` alone. Wouldn't it be more appropriate to have output on standard output when `-cycleonly true` is specified? Similarly, is it useful to send the contents of the `@all` file to standard error?
2. I was also expecting `-cycleonly true` to have the same behaviour as the combination of generator and of the `mcycle7` too, For instance:
```
% /opt/PR/bin/diycross7 -arch AArch64 Pod** Rfe,Fre,Coe Pod** Rfe,Fre,Coe
Generator produced 6 tests
% mcycles7 @all
MP: PodWW Rfe PodRR Fre
S: PodWW Rfe PodRW Coe
LB: PodRW Rfe PodRW Rfe
SB: PodWR Fre PodWR Fre
R: PodWW Coe PodWR Fre
2+2W: PodWW Coe PodWW Coe
```
The difference with the combination of generator and `mycle7` being that no file is created. Notice the format above: lines that consist in a test name, followed by ':', followed by cycle is valid input for `diyone7` (with no argument, `diyone7` reads standard input),  thereby providing a convenient mean to actually generate tests from a list of cycles. The name of test `t`  name is `t.name`.

Finally, although I and not so sure about it, it may be safer to extract the cycle from `t.info` than from `t.com`. See [the code that sets `com`](https://github.com/herd/herdtools7/blob/f05eded5c22cfcbb9a71777238addf85367b1517/gen/top_gen.ml#L959) To that aim, a possible expression is `Misc.Simple.assoc_opt "Cycle" t.info`. Namely the `com` field of a test may not be  set to the cycle.

hugookeeffe:
Thanks for the reply @maranget !
1. I agree that there is no use for stdout=true on its own, however we could think of a case where we would want cycleonly=true and stdout=false, so kept both options to accommodate for this.
2. For @all its being sent to stderr so that we can still have information about what relaxations are being tested, how many tests are produced and version/revision of code. However we aren't currently consuming this information so if you think its better to leave it out I will upload a patch to remove this.
3. My understanding is that using the generator first would still produce the litmus files, and the desire with this option was that no files were created at all.

maranget:
> Thanks for the reply @maranget !
You are welcome

> 
> 1. I agree that there is no use for stdout=true on its own, however we could think of a case where we would want cycleonly=true and stdout=false, so kept both options to accommodate for this.
At the moment I could not find where the cycle output was going with option `-stdout false`, hence my remark.
> 2. For @ALL its being sent to stderr so that we can still have information about what relaxations are being tested, how many tests are produced and version/revision of code. However we aren't currently consuming this information so if you think its better to leave it out I will upload a patch to remove this.
Ok, so there is a serious reason to have this output. I understand that this applies to the initial line of `@all`, but maybe not to the list of (not-_generated files. 
> 3. My understanding is that using the generator first would still produce the litmus files, and the desire with this option was that no files were created at all.
I was not clear enough, I was only suggesting a change of output format, so as to have the same output as the combination of generator and `mcycle7`, I do understand the interest of not generating the cycles. However, if after some treatment or  selection of some cycles, one wishes to generate the tests, then the format with the a name in front of each line is appropriate as `diyone7` standard input.

hugookeeffe:
Ah ok, I misunderstood your comment about changing the output format for the -cyclesonly option. I have implemented the new format in my latest commit. 

I'm struggling to find a way to stop outputting the not generated file names, I think I am not looking in the right place. Do you have any suggestions on how to do this?

maranget:
I may suffice to add `if not Config.stdout then` in front of the code line that follows this [comment](https://github.com/herd/herdtools7/blob/728bedc8f59c32dccb440b0fefde9ad4cb7b69cc/gen/dumpAll.ml#L286).
-Also, although it is not very important, I still cannot make use of the combination of options `-cycleonly true -stdout false`. This, if I am not wrong, it may be approriate to have `-cycleonly true` to imply `-stdout true`.

hugookeeffe:
Thanks for the help @maranget , I have implemented the suggestion you mentioned to not print filenames when stdout=true.

I have also changed it so `-cycleonly true` now implies `-stdout true`

maranget:
LGTM, merged, thanks. 
I noticed that the final message "Generator produced N tests" goes to standard output. Wouldn't standard error be a better choice? In any case the PR is merged. I can change the output channel of this message in a later commit, if you agree.

hugookeeffe:
Thank you for all the help! 
Ah, I hadn't noticed that. I agree that would make more sense to go to standard error.

PR #510, opened 2023-01-25
[herd]: Fix mask for negative state comparison

lukeg101:

```
AArch64 238

{}

P0;
  MOV X0, #-1;
  MOV X1, #-2;

forall (0:X0=-1 /\ 0:X1=-2)
```
When run using herd we get the result:

```
Test 238 Required
States 1
0:X0=-1; 0:X1=-2;
No
Witnesses
Positive: 0 Negative: 1
Condition forall (0:X0=-1 /\ 0:X1=-2)
Observation 238 Never 0 1
```
The states are the same, however the question over the final state is negative.

This occurs for negative numbers, but not positive numbers.

This is because the default test type size is `int`. However storing a negative number in xreg masks type as `int64_t`, according to `mem_access_size_of_t` in `archExtra_herd`.

since `int64_t` is not an `int` the comparison check fails due to the type check in `mask_type` in `archExtra_herd`.

This patch fixes this issue by masking both values used in the compare operation so the types check out.


maranget:
I'll merge this PR because the new behaviour is the same as for `litmus`. However I'd like to insist, perhaps a bit heavily that test A238 is not the kind of test that we should write. As registers are `X` registers, and that those registers are observed at the end of the test, those should be declared as `int64_t`, like in test A239, which is already "correct" using master. In test A238, one would better use `W` registers. Further notice that litmus flags a warning when compiling this test. Said otherwise, we should not invest too much  in having test A238 to work correctly. 

As regards the risks of tests similar to A238, I do not think that not specifying "correct" types for registers is harmful, but it is for memory location. Namely this absence of types  will lead to memory accesses out of intended memory areas. So, generally I encourage everyone either to use `W` registers when types are not explicitly given (and are thus `int` by default) or to declare types.

PR #513, opened 2023-01-30
[all]: Add support for LDP/STP post-indexed write

lukeg101:


maranget:
Thanks, LGTM.

There is still a small problem with dependencies: the writeback should data-depend upon the (virtual) effective address, not on the PTE read. However, other post-indexed instructions have the same problem. Let us see this later.

PR #514, opened 2023-01-30
[all]: Added Load Literal instruction

lukeg101:
This test is similar to ADR, exept it also loads the value at the memory location as well. Test provided

### DIFF
--- a/herd/AArch64Arch_herd.ml
+++ b/herd/AArch64Arch_herd.ml
@@ -191,6 +191,7 @@ module Make (C:Arch_herd.Config)(V:Value.AArch64) =
       | I_LDPSW _ -> Some (tr_variant V32)
       | I_LDR (v,_,_,_,_) | I_LDP (_,v,_,_,_,_) | I_LDXP (v,_,_,_,_)
       | I_LDUR (v,_,_,_)  | I_LDR_P(v,_,_,_)
+      | I_LDR_L(v,_,_)
       | I_STR (v,_,_,_,_) | I_STLR (v,_,_) | I_STXR (v,_,_,_,_)
       | I_STR_P (v,_,_,_)
       | I_STP (_,v,_,_,_,_) | I_STXP (v,_,_,_,_,_)
@@ -255,6 +256,7 @@ module Make (C:Arch_herd.Config)(V:Value.AArch64) =
       | I_LDR (_,r,_,_,_)|I_LDRBH (_,r,_,_,_)
       | I_LDRS (_,_,r,_)
       | I_LDUR (_,r,_,_)
+      | I_LDR_L (_,r,_)
       | I_LDAR (_,_,r,_) |I_LDARBH (_,_,r,_)
       | I_SWP (_,_,_,r,_) | I_SWPBH (_,_,_,r,_)
       | I_STXR (_,_,r,_,_) | I_STXP (_,_,r,_,_, _) | I_STXRBH (_,_,r,_,_)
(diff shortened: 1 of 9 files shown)


relokin:
@lukeg101, I tried running A245 with litmus7 and I got

```
%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
% Results for herd/tests/instructions/AArch64/A245.litmus %
%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
AArch64 245

{
 int64_t x=1;
 int64_t 0:X0=0;
}
 P0       ;
 LDR X0,x ;

forall (0:X0=1)
Generated assembler
/usr/bin/ld: /tmp/cc6j7jdq.o: in function `P0':
A245.c:(.text+0xb8): undefined reference to `x'
/usr/bin/ld: /tmp/cc6j7jdq.o: relocation R_AARCH64_LD_PREL_LO19 against symbol `x' which may bind externally can not be used when making a shared object; recompile with -fPIC
A245.c:(.text+0xb8): dangerous relocation: unsupported relocation
collect2: error: ld returned 1 exit status
File "herd/tests/instructions/AArch64/A245.litmus" Exec of 'gcc -Wall -std=gnu99 -O2 -pthread -o /tmp/dir0a7d49.tmp/A245.exe /tmp/dir0a7d49.tmp/utils.c /tmp/dir0a7d49.tmp/litmus_rand.c /tmp/dir0a7d49.tmp/outs.c /tmp/dir0a7d49.tmp/A245.c' failed
#START _litmus_P0
	ldr x4,
```

am I doing something wrong?

Isn't `ldr_l` a macro? If yes there should be a definition somewhere shouldn't there? If it is a macro should we have support for it in herd?

lukeg101:
Hi Nikos,
Thank you for looking into this. I will answer these questions in order

> am I doing something wrong?

No this is normal behaviour for the assembler, since it cannot find the symbol `x`. This is because the test does not contain information about `x` and the symbol table doesn't either. I follow the precedent set by `ADR` which accepts labels as arbitrary symbols here. Notably `ADR x10, x` is valid in herd and implemented as `MOV`. LDR literal is the same as `ADR` but takes an extra step and loads the value at location `x` as well. What is your advice on this?

This is also a precedent that compilers/disassemblers take to, see https://godbolt.org/z/s1ja5v9Yj

I am tagging @maranget here as he may be interested in the discussion around what kind of symbols should be accepted by `ADR` and `LDR` (literal).

> Isn't ldr_l a macro?

According to section C6.2.167 of the Arm arm it doesn't appear to be a macro/pseudoinstruction but rather its own instruction. Is this what you mean by macro?

lukeg101:
For clarity - I am looking into a fix for the litmus error you found

lukeg101:
Hi Nikos, sorry about my delay, I had to take a look at some GNU extended assembly syntax to understand how to implement PC-relative literal loads using labels for data. I have a tentative working solution in litmus for `LDR` literal now, and I am implementing the templating needed in `litmus/*` to get it working for load literal. I am also ensuring it works for `ADR` instructions as they work in the same way. Thanks

lukeg101:
Hi Nikos, I am closing this commit for now as I do not have a motivation for it (ie a case for why I need it), further I need to investigate the implementation of ADRP in litmus which I do have a use case for.

PR #516, opened 2023-02-01
ASL improvements

HadrienRenaud:
Pass tests (`make test`) after each commits on 4.08


maranget:
Merged by hand, because rebasing on master introduced a non-exhaustive match. Thanks.

PR #518, opened 2023-02-01
[all]: Added C support for non-storing expressions

lukeg101:
Consider the following test:

```
C C03

{}

P0 (int *y) {
  atomic_fetch_add_explicit(y,1,memory_order_seq_cst);
}

forall y=1
```
This test cannot be parsed, since the grammar expects the result of the expression to be written to a local register (e.g. `r0 = ...`)

This patch allows the above to be parsed and simulated under herd

### DIFF
--- a/lib/dune
+++ b/lib/dune
@@ -3,7 +3,7 @@
 
 
 (menhir (modules PPCParser) (flags  --fixed-exception))
-(menhir (modules CParser) (flags  --fixed-exception --explain))
+(menhir (modules CParser) (flags  --fixed-exception))
 (menhir (modules ARMParser) (flags  --fixed-exception))
 (menhir (modules MIPSParser) (flags  --fixed-exception))
 (menhir (modules X86Parser) (flags  --fixed-exception))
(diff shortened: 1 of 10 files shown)


lib/dune
```diff
@@ -3,7 +3,7 @@
 
 
 (menhir (modules PPCParser) (flags  --fixed-exception))
-(menhir (modules CParser) (flags  --fixed-exception))
+(menhir (modules CParser) (flags  --fixed-exception --explain))
```
maranget, line 6: This look like leftover code:
```suggestion
(menhir (modules CParser) (flags  --fixed-exception))
```

maranget:
The abstract syntax and semantics are fine, minimizing code changes.

As regards the parser, I would have solved the shift/reduce conflict differently. Namely, assigning a priority to the IDENTIFIER token looks a bit dangerous, given that this token can appear in many productions. I'd rather change the grammar, so that the parsing that corresponds to the reduce case is no longer legal. This is commit 66f1a008, to be applied after this PR commit.

PR #520, opened 2023-02-02
Add (__)(u)int128(_t) parsing to CType

lukeg101:
LLVM and GCC define the types `__int128_t` and `__uint128_t`

We extend the CType parsing to support these types


maranget:
Hi @lukeg101. This PR may go far beyond a few commits if you need to implement 128 bits integers. We have to understand why this addition is useful and how much improvement of tools should be performed for this addition to   be practical. A few difficulties:
1. herd7 considers the underlying type of scalars to be `Int32`, as a result the following test will fail
```
C C05

{
int128_t x=1;
}

P0 (int128_t *x) {
 *x = 0;
}

forall ([x]=0)
```
(Notice that the problem is the same for `int64_t`...) There are other similar anomalies, for instance "negative" integers in mode `-hexa true` will reveal that all integers are considered to be 32bits.
2. My gcc does not know about types `int128_t` and `uint128_t`, but it does accept `__int128` and `__uint128`.
3. Generally speaking, litmus produces C code that does not compile on tests that effectively use 128 bit integers.
```
C C06

{
__uint128_t x;
}

P0 (__uint128_t *x) {
  *x = 1 ;
}

forall ([x]=1)
```
Attempting to compile the C code yields  errors. Here is the first one:
```
C06.c: In function ‘do_dump_outcome’:
C06.c:126:36: error: expected ‘)’ before ‘PRIu128’
   fprintf(fhist,"%-6"PCTR"%c>[x]=%"PRIu128";\n",c,show ? '*' : ':',(__uint128_t)o[x_f]);
```
(File `inttypes.h` on my machine does not define `PRIu128`) Moreover, litmus-produced C code considers that saved values fit into type `intmax_t`. I am not sure that this is the case of a value of type `__int128_t`.

lukeg101:
Hi Luc, Thanks for this, I will tackle these in steps - may take some time

lukeg101:
I agree, I am focusing on implementing the `int128` type as a pair of Int64 (much like `lib/uint128Scalar`) so that test `C05` passes, I will also look into the GCC and litmus issues you list above. I won't do any more than is needed to get this working (and my tests I have downstream)

lukeg101:
Hi Luc, I have addressed point 1 above by implementing a new variant in herd `S128` which must be specified to use signed int128 bit types. I expect this will also be the means by which I handle the `int128` type in GCC and litmus (as flags may be needed). I am looking into GCC and litmus now

maranget:
> I agree, I am focusing on implementing the `int128` type as a pair of Int64 (much like `lib/uint128Scalar`) so that test `C05` passes, 
Maybe the base scalar type for C litmus tests should be upgraded to `Int64` from `Int32`. The general idea being to be as close as possible of C code running on a modern machine.

> I will also look into the GCC and litmus issues you list above. I won't do any more than is needed to get this working (and my tests I have downstream)
litmus  should be relatively easy. However I am wondering if we need `int128_t` and `uint128_t` as `gcc` does not  define those types.

lukeg101:
> Maybe the base scalar type for C litmus tests should be upgraded to `Int64` from `Int32`. The general idea being to be as close as possible of C code running on a modern machine. 

This makes sense, I have no opinion on this however.

> litmus should be relatively easy. However I am wondering if we need `int128_t` and `uint128_t` as `gcc` does not define those types.

We have `int128_t` supported in `lib/testType`, We can parse `int128_t` as `__int128_t` to keep GCC happy and allow compatibility with any previous tools that use `int128` in herd. Does that sound ok?

lukeg101:
Thinking more about this, I agree with you, the easiest way to maintain compatibility with gcc is to not accept `int128_t` and `uint128_t` in C tests - Implementing now

lukeg101:
Hi Luc, I have now addressed all of the above points:
1. Test `C05` now passing - added `S128` variant to `herd` and `litmus`
2. Forbid `int128_t` and `uint128_t` types as they are not supported by GCC - test `C07` checks this
3. Fix `litmus` error by printing types directly using `llx`. `litmus` now warns on these types rather than failing, but `C06` passes. Whether these types are supported by the architecture running `litmus` is impl. defined - I am unsure of a general fix for this warning

maranget:
Hi @lukeg101, I have just committed a new version of file `herd/CParseTest.ml` that shares code by using a fonctor.

I met more serious problems while attempting to add unit tests for testing `lib/int128.ml` (my attempt is visible as commit a4ff90e5facf02fbc3d85101d1b0c5a69e34b7ec, you can run unit tests with `dune runtest --profile=release`)

Here are a few details on this commit, which is still work in progress. I noticed first that `leading_zeros` was failing with  a `Stack_overflow` exception on input `0x0`, tried to fix this function by calling `Uint.Uint64.leading_zeros`. See [here](https://github.com/herd/herdtools7/blob/a4ff90e5facf02fbc3d85101d1b0c5a69e34b7ec/lib/int128.ml#L95)

I then copied  the `Uint.Uint128` as the new unit tests for module `Int128` (which of course is probably wrong). There are still problems in the `Int128` module and in the unit test file `lib/tests/int128_test.ml`.

I fact I am a bit puzzled now, there is little difference between signed and unsigned 2-complement arithmetic, so it should be relatively easy to implement `Int128` and Uint.Uint128` on top of a common, shared, code. As a first approach, I'd consider that there are  differences in scanning, printing, comparisions and perhaps division.

lukeg101:
Hi Luc, I appreciate your help on this! I tried to test this yesterday and had similar problems. I will cherry pick your code, add more tests, then we can explore merging the libraries for the signed and unsigned variants.

maranget:
You are welcome. Beware I have change the commit several times. The right one is 7acd52719037be1342b55708603db58d4f87d446.

lukeg101:
Thanks for your help, Luc. I have added a few tests for the remaining functions exposed in `lib/int128.ml`, fixed a build warning, and cherry picked your patch using hash `7acd527`. Thank you for redesigning the signed/unsigned interface, it is much cleaner.

lukeg101:
Hi Luc, same here, is there anything else I can do here?

lib/CType.ml
```diff
@@ -71,6 +71,8 @@ let fmt10 = function
   | "uint32_t" -> Some (Macro  "PRIu32")
   | "int64_t" -> Some (Macro  "PRIi64")
   | "uint64_t" -> Some (Macro  "PRIu64")
+  | "int128_t" | "__int128_t" -> Some (Direct "llx")
```
maranget, line 74: Function `fmt10` is for decimal output:
```suggestion
  | "int128_t" | "__int128_t" -> Some (Direct "lld")
```

lib/CType.ml
```diff
@@ -87,6 +89,8 @@ let fmt16 = function
   | "uint32_t" -> Some (Macro  "PRIx32")
   | "int64_t" -> Some (Macro  "PRIx64")
   | "uint64_t" -> Some (Macro  "PRIx64")
+  | "int128_t" | "__int128_t" -> Some (Macro  "PRIx128")
+  | "uint128_t" | "__uint128_t" -> Some (Macro  "PRIx128")
```
maranget, line 93: We may change this to
```suggestion
  | "int128_t" | "__int128_t"
  | "uint128_t" | "__uint128_t"
       -> Some (Direct "llx")
```

lib/CType.ml
```diff
@@ -71,6 +71,8 @@ let fmt10 = function
   | "uint32_t" -> Some (Macro  "PRIu32")
   | "int64_t" -> Some (Macro  "PRIi64")
   | "uint64_t" -> Some (Macro  "PRIu64")
+  | "int128_t" | "__int128_t" -> Some (Direct "llx")
+  | "uint128_t" | "__uint128_t" -> Some (Direct "llx")
```
maranget, line 75:
```suggestion
  | "uint128_t" | "__uint128_t" -> Some (Direct "llu")
```

maranget:
Hi @lukeg101, I have made a few suggestions above. However the situation is not very satisfactory yet.  Namely the prefix `ll` (for `long long`) stands for 8 byte integer on my machine, not 16... Moreover values of final states are stored in fields of type `intmax_t` of size 8 byte on my machine. Hence half of `int128_t` integers is systematically lost. The effect is visible with this test
```
C C08
Variant=S128
{
__uint128_t x;
}

P0 (__uint128_t *x) {
  *x = 1 ;
  *x = 0 - *x;
}

forall ([x]=-1)
```
Compiled with `litmus7 -hexa ...`. Output is:
```
...
4000000*>[x]=0xffffffffffffffff;
...
```
While expected ouput is
```
[4000000*>[x]=0xffffffffffffffffffffffffffffffff;
```
That is, we expect 'f' 32 times (BTW herd is correct here)

Anyway, I would tend to merge this PR as it is, there are other such glitches that we been tolerating until now. Still we need to be aware that quite a lot of further developments are still needed if we want to handle tests that use 128 bit integers in a significant way.

lukeg101:
Hi Luc,
Thank you for this review, I appreciate your help on this! I realise and I have sent PRs that required work or were not properly motivated which put responsibility on you to spend time on this. From now on I'm going to spend more time on the patches to ensure they are well tested (in herd, litmus, do not break hashes) etc... and motivated by specific features so that we do not spend too much time in review. In the future I will try to motivate my patches with litmus tests expressing this motivation, or if I cannot discuss before I do so.
 
In this case I would like to test the compilation of int128 bit litmus tests which requires we have a type that can do atomic operations (loads, stores, fetch-op, read-modify-writes) but does not require full arithmetic. I am happy to discuss what is required and do the work as needed to 'bound' this work. 

In this instance I agree we should merge, if I realise we need other missing features I will implement them.

maranget:
Hi @lukeg101 , now I see the connection with PR #526 on CASP and more generally on "pair" instructions. A last check and I'll merge this PR.

maranget:
Hi @lukeg101, I have squashed all commits in one and merged, Thanks.

PR #522, opened 2023-02-07
[all]: Add Region parsing to C Parser

lukeg101:
Add region/scope parsing to C parser, made extra data list

This patch allows us to add region and scope information to C tests.
This also changes the extra_data parameter to store a list of data
so we can pass multiple things into the test.

For C tests, extra_data is where metadata about shared data is stored as
well as region information - we need to preserve both

### DIFF
--- a/lib/BellCheck.ml
+++ b/lib/BellCheck.ml
@@ -245,10 +245,16 @@ module Make
             code)
         parsed.MiscParser.prog in
       let parsed = { parsed with MiscParser.prog; } in
-      let test_bi = parsed.MiscParser.extra_data in
-      let test_bi = match test_bi with
-      | MiscParser.BellExtra b -> b
-      | _ -> Warn.fatal "Error getting bell information from test" in
+      let test_bi =
+        let test_bi = parsed.MiscParser.extra_data in
+        begin match (List.map
+          (function
+           | MiscParser.BellExtra b -> b
+           | _ -> Warn.fatal "Error getting bell information from test")
+          test_bi) with
+          | [] -> assert false
+          | x::_ -> x
+        end in
       begin match test_bi.BellInfo.regions with
       | Some r -> check_regions r bi
       | _ -> ()
@@ -256,7 +262,7 @@ module Make
       let st = Misc.app_opt (fun st -> check_scopes st bi)  test_bi.BellInfo.scopes in
       let test_bi = { test_bi with BellInfo.scopes=st;} in
       let () = Misc.check_opt (fun lvls -> check_levels lvls bi)  test_bi.BellInfo.levels in
-      { parsed with MiscParser.extra_data = MiscParser.BellExtra test_bi; }
+      { parsed with MiscParser.extra_data = [MiscParser.BellExtra test_bi;] }
 
     let check = match C.info with
     | None -> Misc.identity
--- a/lib/CGenParser_lib.ml
+++ b/lib/CGenParser_lib.ml
@@ -44,12 +44,12 @@ module type LexParse = sig
   val deep_lexer : Lexing.lexbuf -> token
   val deep_parser :
         (Lexing.lexbuf -> token) -> Lexing.lexbuf ->
-	  pseudo list CAst.test list
+	  pseudo list CAst.test list * MiscParser.extra_data
 
   val shallow_lexer : Lexing.lexbuf -> token
   val shallow_parser :
         (Lexing.lexbuf -> token) -> Lexing.lexbuf ->
-	  string CAst.t list
+	  string CAst.t list * MiscParser.extra_data
 
   type macro
   val macros_parser :
@@ -129,9 +129,9 @@ module Do
     let init =
       I.call_parser_loc "init"
 		      chan init_loc SL.token StateParser.init in
-    let prog =
+    let prog,_ =
       I.call_parser_loc "prog" chan prog_loc L.deep_lexer L.deep_parser in
-    let prog_litmus =
+    let prog_litmus,data_litmus =
       I.call_parser_loc "prog_litmus" chan prog_loc L.shallow_lexer L.shallow_parser in
     (* Add parameter passsing as inits *)
     let full_init =
@@ -193,7 +193,7 @@ module Do
         filter = filter;
         condition = final;
         locations = locs;
-        extra_data = MiscParser.CExtra params;
+        extra_data = (MiscParser.CExtra params)::data_litmus;
       } in
     let name  = name.Name.name in
     let parsed =
--- a/litmus/top_klitmus.ml
+++ b/litmus/top_klitmus.ml
@@ -193,7 +193,7 @@ module Top(O:Config)(Tar:Tar.S) = struct
       type token = CParser.token
       module CL = CLexer.Make(struct let debug = false end)
       let lexer = CL.token false
-      let parser = CParser.shallow_main
+      let parser lexer buf = fst (CParser.shallow_main lexer buf)
     end
     module Pseudo =
       struct
--- a/litmus/top_litmus.ml
+++ b/litmus/top_litmus.ml
@@ -336,7 +336,7 @@ end = struct
         type token = CParser.token
         module CL = CLexer.Make(struct let debug = false end)
         let lexer = CL.token false
-        let parser = CParser.shallow_main
+        let parser lexer buf = fst (CParser.shallow_main lexer buf)
       end
 
       module A' = CArch_litmus.Make(O)
--- a/tools/mlock.ml
+++ b/tools/mlock.ml
@@ -57,11 +57,13 @@ module Top(O:Config)(Out:OutTests.S) = struct
 
   let changed = ref false
 
+  exception NotChanged
+
   let not_changed name =
     if O.verbose > 0 then
       Warn.fatal "test %s unchanged, no output" name.Name.name
     else
-      raise Misc.Exit
+      raise NotChanged
 
 (**********)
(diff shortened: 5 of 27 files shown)


relokin:
@lukeg101 can you provide a bit more clarity on how you would like to use regions in c? Is there a litmus test you can show to help me understand better how you're using this feature. `C07` doesn't illustrate how you're using regions, even if we removed it we would get the same final states.

lukeg101:
Hi Nikos, I would like to use region parsing to attach properties to variables in a program in order to track whether executions violate those properties. You could call this property based testing.

For instance. There might be a region of memory marked Read Only by the kernel or compiler which fails at runtime if written to. We may check whether there exists an execution that violates the property and catch it with the `Flag` construct like how UB is caught by the C11 models in this repo. This isn't a change in outcomes as you observed, but rather a different kind of outcome, for example a runtime crash. I do have such a test but its use currently depends on patches I will upstream in the future. Is this ok?

lukeg101:
This patch is now rebased

maranget:
Hi @lukeg101. Could you motivate the change of the extra_data parameter to a list? Some example would help.

maranget:
Ok now I see. Since extra parameters now are a list, wouldn't it be simpler to get rid of the `NoExtra` constructor (defined [here](https://github.com/herd/herdtools7/blob/828a9e2f3085db5440ca2e591ec6ed6311eb6d44/lib/miscParser.ml#L179) and to define `empty_extra` (just [below](https://github.com/herd/herdtools7/blob/828a9e2f3085db5440ca2e591ec6ed6311eb6d44/lib/miscParser.ml#L183) as the empty list?

lib/BellCheck.ml
```diff
@@ -245,18 +245,21 @@ module Make
...
-      let test_bi = match test_bi with
-      | MiscParser.BellExtra b -> b
-      | _ -> Warn.fatal "Error getting bell information from test" in
+      let test_bi =
+        let test_bi = parsed.MiscParser.extra_data in
+        List.hd (List.map
+          (function
+           | MiscParser.BellExtra b -> b
+           | _ -> Warn.fatal "Error getting bell information from test")
+          test_bi) in
```
maranget, line 254: Using `List.hd` is dangerous. If for some reason (_e.g._ a bug) the list is empty, users are faced with anon specific exceptions. Here is  a more robust alternative:
```suggestion
        let test_bi = parsed.MiscParser.extra_data in
        match (List.map
          (function
           | MiscParser.BellExtra b -> b
           | _ -> Warn.fatal "Error getting bell information from test")
          test_bi) with
          | [] -> assert false
          | x::_ -> x in
```
That amounts to inlining `List.hd`.

lib/CGenParser_lib.ml
```diff
@@ -44,12 +44,12 @@ module type LexParse = sig
...
   val deep_parser :
         (Lexing.lexbuf -> token) -> Lexing.lexbuf ->
-	  pseudo list CAst.test list
+	  pseudo list CAst.test list * MiscParser.extra_data
 
   val shallow_lexer : Lexing.lexbuf -> token
   val shallow_parser :
         (Lexing.lexbuf -> token) -> Lexing.lexbuf ->
-	  string CAst.t list
+	  string CAst.t list * MiscParser.extra_data
```
maranget, line 52: I notice that both parsers return extra data.

lib/CGenParser_lib.ml
```diff
@@ -148,10 +148,10 @@ module Do
 			    (MiscParser.Location_reg (p,loc),
 			     (ty,ParsedConstant.nameToV loc))::env)
 			   env t.params)
-        init prog_litmus in
-    let procs = List.map (fun p -> p.CAst.proc) prog in
+        init (fst prog_litmus) in
```
maranget, line 151: It would be clearer to change the `let` bindings of lines 132 and 134 just above into:
```OCaml
    let prog,data =
      I.call_parser_loc "prog" chan prog_loc L.deep_lexer L.deep_parser in
    let prog_litmus,data_litmus =
      I.call_parser_loc "prog_litmus" chan prog_loc L.shallow_lexer L.shallow_parser in
```
That way we'd avoid using `fst` and `snd` and have more explicit variable names.

lib/CGenParser_lib.ml
```diff
@@ -193,7 +193,7 @@ module Do
         filter = filter;
         condition = final;
         locations = locs;
-        extra_data = MiscParser.CExtra params;
+        extra_data = [MiscParser.CExtra params; (snd prog_litmus)];
```
maranget, line 196: If I am not wrong, `(snd prog)` (or `data`) does not appear anywhere. This question the type of `deep_parser`.

lib/simpleDumper.ml
```diff
@@ -67,6 +67,16 @@ CoreDumper.S with
...
+    (*
+    let print_extra t =
+      begin match t with
+        | NoExtra|CExtra _ -> ()
+        | BellExtra bi     ->
+        Out.fprintf chan "\n%s\n" (BellInfo.pp bi)
+      end in
+    let _ = List.map print_extra t.extra_data in
+    Out.fprintf chan "%s\n" (I.dump_constr t.condition) ;
+    () *)
```
maranget, line 79: Perhaps one can simply delete those added lines?

litmus/top_litmus.ml
```diff
@@ -335,7 +335,7 @@ end = struct
         type token = CParser.token
         module CL = CLexer.Make(struct let debug = false end)
         let lexer = CL.token false
-        let parser = CParser.shallow_main
+        let parser x y= fst (CParser.shallow_main x y)
```
maranget, line 338: Slight clearer:
```suggestion
        let parser lexer buf = fst (CParser.shallow_main lexer buf)
```

litmus/top_klitmus.ml
```diff
@@ -193,7 +193,7 @@ module Top(O:Config)(Tar:Tar.S) = struct
       type token = CParser.token
       module CL = CLexer.Make(struct let debug = false end)
       let lexer = CL.token false
-      let parser = CParser.shallow_main
+      let parser x y = fst (CParser.shallow_main x y)
```
maranget, line 196: Slightly clearer:
```suggestion
      let parser lexer buf = fst (CParser.shallow_main lexer buf)
```

tools/mlock.ml
```diff
@@ -383,17 +383,23 @@ module Top(O:Config)(Out:OutTests.S) = struct
...
+          List.map
+             (fun (i,code) -> i,List.map (CBase.pseudo_map tr_ins) code)
+             t.prog in
+          if not !changed then not_changed name ;
+          let extra_data = [CExtra (tr_params pss)] in
+           { t with extra_data; prog;}
+       | NoExtra
+       | BellExtra _ -> assert false)
+      extra in
+      List.hd (t_list)
```
maranget, line 402: If I understand this code correctly, the original code computes a new program '{t with prog = tr_prog t.prog; extra_data = tr_data t.extra_data;}', although in a slightly contrived manner.  In other words the original code could  (should) have been written as follows:
```OCaml
  let tr_prog prog =
     changed := false ;
     let prog = 
       List.map
        (fun (i,code) -> i,List.map (CBase.pseudo_map tr_ins) code)
        prog in
     if not !changed then not_changed name ;
    prog in
  let tr_data data =
    match  data with
    | CExtra pss -> CExtra (tr_params pss)
    | NoExtra|BellExtra _ -> assert false in
  { t with prog = tr_prog t.prog; extra_data = tr_data t.extra_data; }
````
 Then the new code would be like : ` { t with prog = tr_prog t.prog; extra_data = List.map tr_data t.extra_data; }`

**Review by maranget — APPROVED** [review]
I have made a few remarks, mostly on programming style.

**Comment by lukeg101** [comment]
> Hi @lukeg101. Could you motivate the change of the extra_data parameter to a list? Some example would help.

Hi Luc, apologies for my delay and thank you for looking into this! Consider the following litmus test:
```
{ int64_t x[2] = {0,0} }

P0 (const atomic<__int128_t> *x) {
  __int128_t r0 = atomic_load_explicit(x,
    memory_order_seq_cst);
}

regions: x:const_qualified

exists (0:r0=0)
```
When compiled using clang or llvm it crashes on Apple M1 machines: 
```
{ _Atomic __int128 *x = 0; }

P0(const _Atomic __int128*x) {
    __int128_t y = atomic_load(x);
}
```
Unfortunately this does not result in a measurable outcome when running the compiled program under herd. To model this crash, I need to assert that the memory location is const. In other words I test a property of _executions_ of the program under test. If the source program executes ok, but the compiled program crashes then that is a runtime crash - a bug!

To specify this property over executions - I use `Bellextra` of `extra_data` to model read-only regions of memory. A property assumed to hold by users and is enforced by the compiler for const qualified data. I adapt a cat file you shall see in the future to 'Flag' violations of this property.

Unfortunately, `extra_data` also uses `CExtra` to capture the arguments of each `P` in C/C++ litmus tests.

I make `extra_data` a list so we can parse `CExtra` and `BellExtra` at the same time

lukeg101:
Hi Luc, I haven't forgotten this - I am prioritising this tomorrow :)

lukeg101:
Hi Luc, I have rebased onto main and addressed the above, after lunch I shall address the last comment:
https://github.com/herd/herdtools7/pull/522#issuecomment-1594336052

maranget:
@lukeg101, merged, thanks.

lukeg101:
Thank you very much, Luc!

PR #523, opened 2023-02-07
[lib]: Instruction Address Labels and Offsets (PC and Absolute)

lukeg101:
Commit 1: [all]: Instruction addresses parsed with section names
Consider test L056:
```
AArch64 L056
(* branch to instruction address with .section name *)

{ 0:X0 = 1; }
  P0            ;
  B .text       ;
  ADD X0, X0, #1;
.text: NOP      ;

forall (0:X0 = 1)
```
Which unconditionally branches to the .text section.

This patch adds support for section labels starting with '.' or '_'

Commit 2: [lib]: Added Hex Instruction Address Labels
Consider test `L058`:

```
AArch64 L058

(* SAME AS L046 but with hex code*)

{ 0:X0 = 1
}

  P0          ;
 B 0x8        ;
 MOV W0,#1    ;
0x8: ADD W0,W0,#1 ;
forall 0:X0=2
```

We want to jump to instruction addresses specified with hex codes - this
patch does that


herd/tests/instructions/AArch64/L058.litmus
```diff
@@ -0,0 +1,12 @@
...
+
+(* SAME AS L046 but with hex code*)
+
+{ 0:X0=1
+}
+
+  P0          ;
+ B 0x8        ;
+ MOV W0,#1    ;
+0x8: ADD W0,W0,#1 ;
```
relokin, line 11: is this a valid label?

relokin (commented):
Hey @lukeg101, reading a little on what constitutes a valid label, I found these:
* A [label](https://sourceware.org/binutils/docs/as/Labels.html) is written as a symbol immediately followed by a colon ‘:’, where a [symbol](https://sourceware.org/binutils/docs/as/Symbol-Intro.html) is one or more characters chosen from the set of all letters (both upper and lower case), digits and the three characters ‘_.$’ No symbol may begin with a digit. 
* [Local labels](https://sourceware.org/binutils/docs/as/Symbol-Names.html) are different from local symbols. Local labels help compilers and programmers use names temporarily. They create symbols which are guaranteed to be unique over the entire scope of the input source code and which can be referred to by a simple notation. To define a local label, write a label of the form ‘N:’ (where N represents any non-negative integer). To refer to the most recent previous definition of that label write ‘Nb’, using the same number as when you defined the label. To refer to the next definition of a local label, write ‘Nf’. The ‘b’ stands for “backwards” and the ‘f’ stands for “forwards”.

If that's correct, I think we can align the Lexer rules on what is a label, but a hex value wouldn't be valid. You can provide a hex value to a `b` but that will be the address of an instruction and it won't branch to the label.

In addition, if we change what is a valid label in herd and litmus, we should make sure the definition of label in `tools/lexLog_tools.mll` is also the same.

lukeg101:
Hi Nikos, I have removed the commit that enables hex addresses, and added . to the name lexer in 'tools/lexLog_tools.ml'.

lukeg101:
tests pass in herd and litmus

tools/lexLog_tools.mll
```diff
@@ -118,7 +118,7 @@ let hexanum = "0x" hexa+
 let set = '{' (' '|','|('-'?(num|hexanum)))* '}'
 let alpha = [ 'a'-'z' 'A'-'Z']
 let pteval = '(' ([' ''_'','':''('')']|alpha|digit)+ ')'
-let name = alpha (alpha|digit|'_'| '.')*
+let name = (alpha|'_'|'.') (alpha|digit|'_'| '.')*
 let label = 'L' (alpha|digit)+
```
relokin, line 122: I am a little worried about this but the problem is not this PR and some quick tests seems to work, so we can fix later.

relokin:
Can you make sure that the same rules of what constitutes a name applies in the 3 lexers? And you will need to manually rebase as there are conflicts.

lukeg101:
Hi Nikos, apologies for my delay, I have rebased the patches and added the name rules to all three levers mentioned in this review.

relokin:
One minor nit. If you rebase and if @maranget has no objection we can merge this.

lukeg101:
Thank you Nikos, I have rebased so the PR is ready, just to confirm is the nit that I needed to rebase or did I miss a comment in the code?

relokin (commented):
sorry I meant to add the comment regarding one of the tests.

PR #524, opened 2023-02-07
[diy7] output generated number of tests to stderr when option -stdout true

hugookeeffe:
all that was needed was a simple change, changing printf to print as the print variable is already defined previously to change depending on Config.stdout (line 440)


hugookeeffe:
@maranget this patch moves the line 'generator produced n tests' to stderr when -stdout true. would there be anything you would like me to add to get this merged?

maranget:
LGTM, merged, thanks.

PR #526, opened 2023-02-08
[all]: Added CASP family of instructions to AArch64

lukeg101:
This patch adds support for CASP, CASPL, CASPA, CASPA instructions

Tests are provided. It is almost working, I will continue tomorrow. Please could I get advice on the following:

1. litmus7 fails when running the provided tests as it doesn't output registers that are contiguous same-size even/odd pairs - I could not figure out how to output registers in such a fashion.
2. There is a build warning as I have not used register `rt2` yet - this is related to why test `A246` does not return the expected result.

I would appreciate any advice, as the dependencies are tricky


lukeg101:
@artkhyzha @jalglave @maranget do you have advice on the following points? Please let me know, thanks.

maranget:
Hi @luke, For point 1. you can try to declare the registers as "stable". For instance, for test `A242`, add line `Stable=X6,X7,X2,X3`. Then, the "stable"  registers will (normally...) be allocated as themselves.

lukeg101:
Thanks Luc, I was not aware of that feature! I have updated the tests and added type annotations to remove all warnings. Tests now pass when run in litmus7

I will now continue working on point 2 - to fix test A246 and the build warning

lukeg101:
Now just to fix the build warning

lukeg101:
I fixed the build warning by replacing `mv` with data reads of `rt1` and `rt2`. From my understanding, `mv` is not used but we do need to read `rt1` and `rt2`. I am thinking further to see if there is something I have missed.

herd/AArch64Sem.ml
```diff
@@ -1445,6 +1445,45 @@ module Make
...
+        lift_memop rn Dir.W true
+          (fun ac ma _ ->
+            let is_phy = Access.is_physical ac in
+             M.altT
+              (let read_mem a = do_read_mem_ret sz an aexp ac a ii
+                >>| (add_size a sz
+                >>= fun a -> do_read_mem_ret sz an aexp ac a ii) in
+               let neqp (v1,v2) (x1,x2) =
+                 M.neqT v1 x1 >>| M.neqT v2 x2
+                 >>= fun _ -> M.unitT () in
```
maranget, line 1465: This condition is too strong: it checks v1 being different from x1 _and_ v2 being different from x2, while we probably want _or_ in place of _and_:
```suggestion
                 M.op Op.Eq v1 x1 >>| M.op Op.Eq v2 x2
                 >>= fun (b1,b2) -> M.op Op.And b1 b2
                 >>= M.eqT V.zero in

```

maranget:
Moreover the second memory read is not tagged as atomic. Consider:
```
AArch64 L057
Stable=X0,X1,X2,X3
{
int x[2];
0:X4=x;
1:X4=x;
}
P0                    | P1                    ;
MOV W1,#1             | MOV W0,#0             ;
STR W1,[X4,#4]        | MOV W1,#0             ;
                      | MOV W2,#2             ;
                      | MOV W3,#2             ;
                      | CASP W0,W1,W2,W3,[X4] ;
locations [1:X0;1:X1;]
forall (x={0,1} \/ x={2,1})
```
By atomicity of the CAS. output `x={2,2}` is not legal. Either the write by P0 is performed first and the CAS fails (output is `x={0,1}`, or teh CAS first succeeds and output is `x={2,1}`. This is fixed by a simple change.

herd/AArch64Sem.ml
```diff
@@ -1445,6 +1445,45 @@ module Make
...
+               M.aarch64_cas_no is_phy ma read_rs write_rs read_mem neqp)
+              (let read_rt = read_reg_data sz rt1 ii
+                >>| read_reg_data sz rt2 ii
+               and read_mem a = rmw_amo_read sz rmw ac a ii
+                >>| (add_size a sz
+                >>= fun a -> do_read_mem_ret sz an aexp ac a ii)
```
maranget, line 1471: This read is an "atomic" one, to be paired with the corresponding write:
```suggestion
                >>= fun a ->  rmw_amo_read sz rmw ac a ii)
```

maranget:
It would be great to include the  two test of previous comments in the regression suite:
[L056.txt](https://github.com/herd/herdtools7/files/10735105/L056.txt) and
[L057.litmus](https://github.com/herd/herdtools7/files/10735104/L057.txt).

lukeg101:
Thanks Luc, I have added the tests you kindly provided and made the above changes, I also made sure to check `L056` and `L057` pass in `litmus7`

lukeg101:
Hi Luc, is there anything else I can help with on this patch?

maranget:
Thanks, merged.

PR #530, opened 2023-02-21
[gen] reject generation if -relax cannot provide cycles

hugookeeffe:
This patch rejects generation of cycle when -safe and -relax are used in a configuration file, but relaxations in -relax cannot generate cycles.

Currently, the generator still generates cycles from the safe list if this is the case.
an example of where this can occur is:
`diy7 -cycleonly true -arch AArch64 -safe "DMB.SYd** Rfe Fre Coe" -relax "[PodRR,PodWR]"`

The reason for this patch is that in our current flow, some composite relaxations are automatically generated to go into `-relax`, Occasionally junk relaxations are accidentally created (e.g. [PodRR, PodWR]), which cannot provide cycles, however the generator will still use relaxations in -safe. 

This patch introduces a fail safe to stop this behaviour.

### DIFF
--- a/gen/alt.ml
+++ b/gen/alt.ml
@@ -684,9 +684,13 @@ module Make(C:Builder.S)
       List.iter (fun r -> fprintf chan "%s\n" (pp_relax r)) rs
 
     let secret_gen relax safe reject n =
+      let r_nempty = not (relax = []) in
       let relax = expand_relaxs C.ppo relax
       and safe = expand_relaxs C.ppo safe
       and reject = expand_relaxs C.ppo reject in
+      if relax = [] then if r_nempty then begin
+        Warn.fatal "relaxations provided in relaxlist could not be used to generate cycles"
+      end ;
       if O.verbose > 0 then begin
         eprintf "** Relax0 **\n" ;
         debug_rs stderr relax ;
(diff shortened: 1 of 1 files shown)


hugookeeffe:
cc: @relokin @jalglave @maranget @chacon01

maranget:
Hi @hugookeeffe. You are right no cycle should be generated in the situation you describe. As have a slight remark, more of aesthetic order: generally I tend to avoid polymorphic equality as much as possible. I suggest the following cosmetic modification.

gen/alt.ml
```diff
@@ -684,9 +684,13 @@ module Make(C:Builder.S)
       List.iter (fun r -> fprintf chan "%s\n" (pp_relax r)) rs
 
     let secret_gen relax safe reject n =
+      let r_nempty = not (relax = []) in
```
maranget, line 687:
```suggestion
      let r_nempty = Misc.consp relax in
```

gen/alt.ml
```diff
@@ -684,9 +684,13 @@ module Make(C:Builder.S)
...
       let relax = expand_relaxs C.ppo relax
       and safe = expand_relaxs C.ppo safe
       and reject = expand_relaxs C.ppo reject in
+      if relax = [] then if r_nempty then begin
```
maranget, line 691:
```suggestion
      if Misc.nilp relax then if r_nempty then begin
```

hugookeeffe:
I think I have made all the suggested changes, let me know if any others need to be made :)

maranget:
Merged, thanks.

PR #533, opened 2023-02-27
[ASL] Add parser for v0 and use extracted Arm ARM pseudocode

HadrienRenaud:
TODO:
 - [ ] Check that tests pass at every commit
 - [ ] Remove last commit 1ac58ade8bbe5dd4b75ebebff0e0c1f29bef5a76 that should be its own pr


HadrienRenaud:
It doesn't, and I don't want to erase 1ac58ad, so I'm closing this and opening another PR.

PR #535, opened 2023-02-28
[lib] Update Misc with 4.08.1 stdlib updates

HadrienRenaud:
This is neither urgent nor very important for me, but I somehow did it.

### DIFF
--- a/lib/misc.ml
+++ b/lib/misc.ml
@@ -36,13 +36,7 @@ let dot_name name_xxx =
   let base = Filename.chop_extension name_xxx in
   base ^ ".dot"
 
-let filebase f =
-  let f = Filename.basename f in
-  let b =
-    try Filename.chop_extension f
-    with Invalid_argument _ -> f in
-  b
-
+let filebase f = Filename.basename f |> Filename.remove_extension
 
 (****************)
 (* basic utils  *)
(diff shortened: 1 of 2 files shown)


lib/misc.ml
```diff
@@ -36,77 +36,43 @@ let dot_name name_xxx =
...
-let seq_opt f = function
-  | Some x -> f x
-  | None -> None
-
-let app_opt f = function
-  | None -> None
-  | Some x -> Some (f x)
+let is_none = Option.is_none
+let is_some = Option.is_some
+let as_some = Option.get
```
relokin, line 61: This changes the behavior when its argument is None from assert to raise Invalid_argument but I think this makes more sense.

lib/misc.ml
```diff
@@ -165,79 +122,10 @@ let rec list_eq eq xs ys = match xs,ys with
...
-
-let capitalize s = match s with
-| "" -> assert false
-| _ ->
-    String.make 1 (char_uppercase s.[0]) ^
-    String.sub s 1 (String.length s-1)
+let char_uppercase = Char.uppercase_ascii
+let lowercase = String.lowercase_ascii
+let uppercase = String.uppercase_ascii
+let capitalize = String.capitalize_ascii
```
relokin, line 128: Slightly different behaviour in the case of
```
Misc.capitalize ""
```
used to assert now it returns ""
Again the new behaviour makes sense to me, just checking.

relokin (approved):
LGTM!

maranget:
Merged, thanks

PR #536, opened 2023-03-01
[lib]: fixed map_reg, fold_reg for kr instructions

lukeg101:
Consider the following test (herd/tests/instructions/AArch64/A127):
```
AArch64 A127

(* MOV value from symbolic register into concrete reg *)
{ uint64_t X0; uint64_t %symbolic_reg = x; uint64_t x }

P0;
MOV X0, %symbolic_reg;

exists (0:X0 = x)
```
This fails when run in herd as `map_reg` and `fold_reg` are not implemented for `I_MOV` when the second argument (of type `'k kr`) is a register value (`RV`).

This patch fixes this. I have added tests for this which pass in herd and litmus on an Apple M1 machine.

Additionally I considered all other AArch64 instructions that have a `kr` element in their type and fixed those so they pass in both herd7, precisely for:

`LDG`, `STZG`, `STG`, `LDP_SIMD`, `STP_SIMD`, `LDP_SIMD`, `STR`, `ALIGND`, and `ALIGNU`

The tests I added for these instructions pass in herd, however I could not run them all in litmus as it requires hardware with MTE (I do not have) and litmus doesn't support the memtag variant anyway (not related to this commit).

As far as I am able to tell - this patch is complete and tested

### DIFF
--- a/herd/tests/instructions/AArch64.neon/V57.litmus
+++ b/herd/tests/instructions/AArch64.neon/V57.litmus
@@ -1,0 +1,9 @@
+AArch64 V57
+(* Tests 128-bit Neon pair loads with post-increment - register INVALID SYNTAX*)
+{
+}
+
+P0                    ;
+LDP Q0, Q1, [X0], Q2  ;
+
+forall(0:X1 = 0)
--- a/herd/tests/instructions/AArch64.neon/V57.litmus.expected-failure
+++ b/herd/tests/instructions/AArch64.neon/V57.litmus.expected-failure
@@ -1,0 +1,1 @@
+Warning: File "./herd/tests/instructions/AArch64.neon/V57.litmus", line 7, characters 18-20: unexpected 'Q2' (in prog) (User error)
--- a/herd/tests/instructions/AArch64.neon/V58.litmus
+++ b/herd/tests/instructions/AArch64.neon/V58.litmus
@@ -1,0 +1,9 @@
+AArch64 V58
+(* Tests 128-bit Neon pair store with post-increment - register INVALID SYNTAX*)
+{
+}
+
+P0                    ;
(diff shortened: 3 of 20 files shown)


maranget:
Having at some of the instruction semantics and at morello documentation I am quite puzzled by `ALIGNS`, `ALIGNU` having a `` `k kr`` argument (and not a `` `k``  argument)

maranget:
A quick look at documentation suggest that `STG`, `LDG` and `STZG` should also have a `` `k` argument. I coudl nit figure this out for SIMD instructions, but parsing also suggests a `` `k`` argument.

lukeg101:
Hi folks, I tried to replace the `kr` type with the `k` type in the above instructions, however it has caused a cascade of unrelated changes required in other unrelated places. Please bear with me whilst I try to fix these

maranget:
I am sorry about this, the changes should be rather mechanical. Tell me if you want me to handle them.

lukeg101:
Hi Luc, I will do it apologies for my delay I am travelling. Thank you for the offer :)

maranget:
OK, so I first merge this PR and then will make another PR to fix all those `` `k kr`` arguments.

maranget:
Merged, thanks

lukeg101:
Hi Luc, sorry about my delay, I was working on it and have run into some strange issues. I will submit a PR with the requested fix.

lukeg101:
Yes I am changing the ` 'k kr` types to `'k` arguments. I was having issues with `ldp/stp_simd` parsing. I had added tests and got all other instructions parsing. This is a mechanical change, I am trying to see what I missed.

maranget:
Ok I let it to you. Do not hesitate to start a draft PR.

Issue #531, opened 2023-02-23
diy7: additional "safe" relaxations reduces number of generated cycles


chacon01:
Hi,

We're observing that some invocations of `diy7` cause fewer cycles to be generated when additional relaxations are added to the "safe" list. For example
```
$ diy7 -cycleonly true -arch AArch64 -safe "Rfe Fre [PodWR,Amo.Cas] PosRR" -relax "[DpDatadW,Rfi]" -size 5 -exact -eprocs -nprocs 2
```
produces one cycle, `PodWR Amo.Cas Rfe DpDatadW Rfi PosRR Fre` whereas
```
diy7 -cycleonly true -arch AArch64 -safe "Rfe Fre [PodWR,Amo.Cas] PosRR PodWR" -relax "[DpDatadW,Rfi]" -size 5 -exact -eprocs -nprocs 2
```
produces none (note the addition of `PodWR` in the "safe" list). My expectation would be that adding `PodWR` would have caused `diy7` to produce a (non-strict) superset of what it produced without it. I tried experimenting with different modes; this doesn't appear to occur with `-mode free` (adding `PodWR` causes a new cycle to appear instead of causing one to _not_ appear).

I see that the documentation is somewhat vague about what the `sc` mode does
> In (default) sc mode, diy7 performs some optimisation, most of which we leave unspecified.

Is there some explanation why adding relaxations to the safe list would cause cycles to no longer be generated when using `sc`?

chacon01:
Hi @maranget. I'm happy with the addition of a new mode that doesn't apply `po_safe`. I have no preference on whether this new mode would be the default instead of `sc`, though perhaps it would be better to make this new mode the default to avoid surprising users in the manner I was. I suppose other users would appreciate the documentation to describe the difference between these modes - I think the general description you provided about `po_safe` is good.

Are there any other optimizations in `sc` that might cause surprising behavior like `po_safe` that we should be aware of or consider removing in this new mode?

maranget:
Hi @chacon01. I'll add the new mode (called 'default'...) and attempt to document modes a bit more. As to your question, yes there is another subtlety : if you declare some fence candidate as safe, then some rejections of composing fence and dependencies candidate occur.  Although I cannot reproduce the effect now, there is a potential "non-monotonic" effect here.

maranget:
[...]

PR #538, opened 2023-03-01
[gen] Wider default mode for diy

maranget:
See this [comment](https://github.com/herd/herdtools7/issues/531#issuecomment-1450311869).


maranget:
This PR apparently fixed the problem described at the end if issue #531. Merged.

PR #547, opened 2023-03-23
[herd] Convert immediate branches into branches with displacements

artkhyzha:
This is a draft PR to report a possible bug and to discuss a possible fix to how herd7 might end up using two representations for immediate branches, to a label and to a pc with offset, to incorrect ends.

```
AArch64 test
{
  0:X0=P1:L0;
  0:X2=instr:"B .+4";
}
 P0          | P1     ;
 LDR W1,[X0] |  B end ;
 CMP W2,W1   | L0:    ;
 B.NE Lend   |  B L1  ;
 MOV W3,#1   | L1:    ;
Lend:        | end:   ;
locations [0:X1;0:X2]
exists (0:X3=0)
```

Arguably, the final state should always have `0:X3=1`. Namely, reading from `P1:L0` ought to create an instruction value in `W1` that is the same as `instr:"B .+4”`. This does not seem to be the case.

A possible explanation for what goes wrong starts with the initial writes. The initial write for the `P1:L0` is `B L1`, which is fine if it is never read explicitly. However, if it is read explicitly by a load instruction, it can then be compared to other values or even stored elsewhere. If one stores `B L1` into a different location, it may break how herd7 executes the test.

The PR proposes a rather ad-hoc solution: upon creating initial write events for overwriteable labels, convert immediate branches into those using offsets. This solution currently has (at least) these several flaws:
- it is fairly obscurely placed into the codebase, which might mean the fix is not as general as it should be, and
- it seems sensible at this point to check if a litmus test attempts to perform an immediate branch to the code of another thread (something litmus7 is unlikely to guarantee support for), but I conjecture that this check is best done upon loading the litmus test (in `loader.ml`).


herd/mem.ml
maranget, line 1: Hi @artkhyzha, wouldn't it be nicer if this conversion function to be called from the Loader module? As the`load_code` function builds a map from labels to code index, this should be relatively easy.

lib/AArch64Base.ml
maranget, line 1: I am not sure, but a better place for this code may be in the `AArch64Arch_herd.ml` file. Namely, this is a herd functionality.

maranget (commented):
Hi @artkhyzha. Nice idea. What about applying it systematically to all branches and even when option `-variant self` is not given ? After all, actual machine code features displacements , not explicit labels.

maranget:
> The PR proposes a rather ad-hoc solution: upon creating initial write events for overwriteable labels, convert immediate branches into those using offsets. This solution currently has (at least) these several flaws:
> 
> * it is fairly obscurely placed into the codebase, which might mean the fix is not as general as it should be, and
Inspired by your remark above, I made a few suggestions that aim at generality....


> * it seems sensible at this point to check if a litmus test attempts to perform an immediate branch to the code of another thread (something litmus7 is unlikely to guarantee support for), but I conjecture that this check is best done upon loading the litmus test (in `loader.ml`).
The best technique to jump from the code of one thread to the code of another thread, usually for a subroutine call, is using a register argument initialised with a _proc_`:`_label constant. However it may be that label argument works in litmus7 (will check).

Overall I Agree with you that `loader.ml` is the place where to translate label arguments into displacements.

herd/mem.ml
artkhyzha: Thank you for the comments, Luc -- yes, I agree it would be nicer indeed, and is worth trying: it should be less confusing to simply convert all branch instructions in the litmus test upon loading.

I will be updating the PR, and will post an update here once there is progress with this and other comments.

artkhyzha: Hi @maranget, thanks again for the comment -- the functionality has been moved to `loader.ml` now.

lib/AArch64Base.ml
artkhyzha: Yes, agreed. This code got moved into `AArch64Arch_herd.ml`, and some other bits (the default non-implementation) got moved in `archExtra_herd.ml`.

artkhyzha:
Thank you for your comments, @maranget . I have made a few changes, hopefully addressing the comments. I keep this PR in the draft status, since there is an outstanding issue I have not fixed yet.

On topic of failing with a user error on tests like `A`: the new commits deal with it by applying the conversion in all AArch64 litmus tests. The conversion performs a check that may or may not fail.

The new commits added the tests in this discussion as unit tests.

In `in `tests/instructions/AArch64.self`,  tests L023 and L025 got modified. As far as I can see, the tests used to be much like test `A`, so I corrected them to avoid failures.

Similarly, in `tests/instructions/AArch64.self`, A009, S01 and S19 were affected by the changes. As far as I can see, these tests may require a correction (due to a change in output format). **The test S21 currently fails, and I am not sure it is the test's fault. I will double check.**

maranget:
> The test S21 currently fails, and I am not sure it is the test's fault. I will double check.

This is not normal. As far as I could investigate,  the instruction `ADR X0,L0` is not executed. Instead, NOP is executed and the value of X0 is then zero...

maranget:
Ok I see a bit  more. The failure correspond to the case when herd (AArch64Sem ?) consider overwriting the initial NOP (label L0, overwritable, in principle) by `B .+8`. I think that master does not have the same problem because the overwriting is with `B L2`.... Not everything is clear yet, but the root cause is probably here.

maranget:
Ok here is a fix attempt: 670aabedb516c4a827beca710343d7ec67844880. The commit message summarizes the ideal: in the case of `S21` we reject the situation depicted above by  enumerating rf for code locations only. As a result it is discovered early that no possible pairing of code read and write can result on the initial NOP being overwritten.

artkhyzha:
Thank you for the diagnosis and the fix, Luc! With your commit, all regression tests are passing. Do you think there are any other steps to take with this PR?

maranget:
> Thank you for the diagnosis and the fix, Luc! With your commit, all regression tests are passing. Do you think there are any other steps to take with this PR?

Hi @artkhyzha. I do not see anything right now, but I still want to have a look. Meanwhile, would you resolve the test name conflict, and rebase on top of master ?. May I suggest that you choose "A" names (next free name is A210, if I am not mistaken). Namely there are several PR with L tests pending (L like Luc, and Luke...). If you think the PR is ready, do not hesitate to remove the draft status.

maranget (commented):
Hi @artkhyzha, I have made a few remarks... Additionnaly,  it would be nice if you could squash all your commits into one, and add commit 670aabe. Alternatively, you can have code modification and new tests in separate commits.

Well, the objective is for `make test` to succeed after each commit. Hence it may be easier to put everything, including commit 670aabe and tests into one single commit.

In spite of my nitpicking, nice work.

herd/AArch64Arch_herd.ml
```diff
@@ -373,6 +373,32 @@ module Make (C:Arch_herd.Config)(V:Value.AArch64) =
...
+            (pp_instruction PPMode.Ascii instr) l
+        in
+      match instr with
+      | I_B (Lbl l) -> I_B (Offset (labelmap l))
+      | I_BC (c,(Lbl l)) -> I_BC (c,(Offset (labelmap l)))
+      | I_BL (Lbl l) -> I_BL (Offset (labelmap l))
+      | I_CBNZ (v,r,(Lbl l)) -> I_CBNZ (v,r,(Offset (labelmap l)))
+      | I_CBZ (v,r,(Lbl l)) -> I_CBZ (v,r,(Offset (labelmap l)))
+      | I_TBNZ (v,r,k,(Lbl l)) -> I_TBNZ (v,r,k,(Offset (labelmap l)))
+      | I_TBZ (v,r,k,(Lbl l)) -> I_TBZ (v,r,k,(Offset (labelmap l)))
```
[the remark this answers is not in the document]
artkhyzha: That's a good catch -- thank you, @maranget!

For a minute I was not sure if this conversion is strictly necessary, so I am adding a new test `AArch64/A249` to record that it is. (By analogy with the test in the PR description.)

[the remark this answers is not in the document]
artkhyzha: Note that several  AArch64 tests fail after applying conversion to ADR: A139, L023, L050, L051. Out of these, L023 needs a trivial fix (that is, removing ADR). More detail below.

Firstly, the case of `A139`. Notice that the test uses a pointer instead of a label. I am not sure I understand the intent here, I am afraid... Is the idea that pointers are PC-relative addresses? In that case, should herd treat `pointer_to_x` as a label then (and not convert it, accordingly)?
```
AArch64 A139
(* Tests whether we can load from a pointer - symbolic location+offset *)

{ int *pointer_to_x = x; int x[2]={4,42} }

P0;
  ADR X1, pointer_to_x; (* loads pointer_to_x into X1 *)
  LDR X1, [X1];         (* loads x into X1 *)
  LDR W0, [X1, #4];     (* loads x[1] into X1 *)

exists (0:X0 = 42)
```

Secondly, the case of L050 and L051. They now fail and illustrate several points. Consider the following test:
```
AArch64 L050-truncated
{
0:X0=instr:"B.EQ .+8";
}
  P0         ;
L0:          ;
 B .+4       ;
 ADR X1,L0   ;
locations [0:X1]
```
Currently herd outputs a herd-internal numerical address here (`0:X1=10000`), which leaks an implementation detail to end-user. Furthermore, it creates problems if followed up by a memory access: loads end up being unmatched, which is what happens in `L050` and `L051`, and stores complain at a store to a non-symbolic location.

Should ADR be able to return a non-label? An approach I want to try is to make ADR return a label always, so that the result of `ADR X1,L0` is then equivalent to having `0:X1=P0:L0` in precondition.

herd/loader.ml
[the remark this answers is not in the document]
artkhyzha: Thank you, @maranget. This is resolved now, I hope? A mapping for labels to procs is used to produce the same error message:
```
# herd7 tmp/B.litmus  
Warning: File "tmp/B.litmus": Label L1 not found on P0, although used in the instruction B L1 on P0 (User error)

# herd7 tmp/A.litmus
Warning: File "tmp/A.litmus": Label L0 not found on P1, although used in the instruction B L0 on P1 (User error) 
```

artkhyzha:
Thank you again for another round of comments, @maranget ! I tried addressing them -- the subject of ADR warrants some special attention when you have a chance to have another look.

maranget:
> Thank you again for another round of comments, @maranget ! I tried addressing them -- the subject of ADR warrants some special attention when you have a chance to have another look.

Indeed, test `A139` is now failing (try `make test`). This test has been introduced by @lukeg101 quite a  long time ago. I also notice that `litmus` cannot handle it. Hence, I'd like to have @lukeg101 opinion on the usefulness of the test that features a `ADR` instruction of which argument is a global location declared in the init section of the test:
```
AArch64 A139
(* Tests whether we can load from a pointer - symbolic location+offset *)

{ int *pointer_to_x = x; int x[2]={4,42} }

P0;
  ADR X1, pointer_to_x; (* loads pointer_to_x into X1 *)
  LDR X1, [X1];         (* loads x into X1 *)
  LDR W0, [X1, #4];     (* loads x[1] into X1 *)

exists (0:X0 = 42)
```
If the test is important we can restore previous behaviour rather easily. Otherwise I am in favor of changing test outcome to refect the new behaviour (_.i.e._ remove the file `A139.litmus.expected` and add the appropriate `A139.litmus.expected-failure`.

maranget:
Error message of test `herd/test/instruction/AArch64.self/A009.litmus` has changed. The reference file `herd/test/instruction/AArch64.self/A009.litmus.expected-failure` should be changed accordingly.

lukeg101:
I note the following complication:

- The compiler may store pointers to _separate_ locations next to each other in the a data section: `bss = [ptr_to_x,ptr_to_y]`, and then emit two loads to access each. This is commonly done on Armv7 where code size matters. This motivates: test A139.

relokin:
If the functionality in this test is not needed and given that litmus7 doesn't support should we just accept that we can't handle this test yet in herd7 either? Otherwise, we could rewrite the test as follows?
```AArch64 A139
(* Tests whether we can load from a pointer - symbolic location+offset *)

{ 
  int *pointer_to_x = x; 
  int x[2]={4,42}
  0:X2=pointer_to_x; 
}

P0;
  LDR X1, [X2];         (* loads x into X1 *)
  LDR W0, [X1, #4];     (* loads x[1] into X1 *)

exists (0:X0 = 42)
```

But I am not sure this adds more coverage on top of other tests we have already.

lukeg101:
I argue the functionality is needed, please see the Armv7 example (https://godbolt.org/z/5YbPnd177). It is possible to rewrite the example, but this could possibly mean we do not support code emitted by clang.

maranget:
What we should do is still unclear. Do we have to support `A139` completely in `herd`, that is restore old behaviour? Or can we accept that this tests fails as it does in this branch?

relokin:
Up to you @maranget but ADR for data is not supported in litmus7 and it's not clear why we would need to support it (we load addresses to registers we don't provide labels to data in inline assembly.

@lukeg101 I don't think anyone's suggesting that we don't need to support offsets relative to a symbolic address (i.e., the address [Xn, #imm16]. That's fully supported already and it should stay like this.

maranget:
Let us keep `A139` as it is and accept that it fails.

@artkhyzha. Would you please update the test references to the current situation (as far as I saw `AArch64/A139` and `AArch64.self/A009` outcome has changed) and reduce the number of commits? --- one squashed commit is fine  but you may wish a slightly more detailed history.

Notice that the (good) idea to consider offsets and not labels in instructions with labels arguments  should also  apply to litmus when it prints instructions... (try test A251 with litmus :) Let us keep this for a later PR.

maranget:
I have finally squashed and merged this PR which was ready. Thanks to all.

artkhyzha:
Thanks everybody, and thanks @maranget for the feedback and the final preparations. (I've been this week so missed the discussion, sorry.)

PR #551, opened 2023-03-28
ASL now produces the correct graphs for main instructions

HadrienRenaud:
Main diffs on non ASL stuff:
 - variants now accepts all sorts of characters, now just separated by commas.
 - add a new variants for using ASL with AArch64 shared pseudocode loaded.
 - add a few combinators in `monad.mli`, either to do simple bindings, or to do low level stuff edits on the `event_structure`.
 - Refacto a bit on `eventsMonad`

Main new features on the ASL side:
 - Add a new way of executing ASL litmus tests by loading all the shared AArch64 pseudocode
 - Fix monadic bindings used in the interpreter and now get the wanted dependency graphs
 - Functions and primitives now take monads and return monads, this fixes problems while evaluating functions. Function calls have now the same semantics as would have their inlined \alpha-reduced code.
 - Introduce new tests for ASL pseudocode by passing them into a very simple architecture. This allows data- and control- dependency graphs to be tested.
 - Improve translation of executions from ASL to AArch64. This translation is currently not-tested, so I'm not very confident on its correctness for strange bits extraction.

### DIFF
--- a/lib/lexSplit.mll
+++ b/lib/lexSplit.mll
@@ -21,7 +21,7 @@ exception Error
 let blank = [' ''\n''\r''\t']
 let non_blank = [^' ''\n''\r''\t']
 let digit = ['0'-'9']
-let printable = ['0'-'9''a'-'z''A'-'Z'':']
+let printable = [^' ''\n''\r''\t'',']
 rule main = parse
 | ',' | blank+  { main lexbuf }
 | digit+ as lxm { int_of_string lxm :: main lexbuf }
--- a/Makefile
+++ b/Makefile
@@ -180,6 +180,17 @@ test-asl:
 		$(REGRESSION_TEST_MODE)
 	@ echo "herd7 ASL instructions tests: OK"
 
+test:: test-pseudo-asl
+test-pseudo-asl:
+	@ echo
+	$(HERD_REGRESSION_TEST) \
+		-herd-path $(HERD) \
+		-libdir-path ./herd/libdir \
+		-litmus-dir ./herd/tests/instructions/ASL-pseudo-arch \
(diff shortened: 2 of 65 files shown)


HadrienRenaud:
The last commit fixes some build warnings on `AArch64Sem.ml`:
 - an unused value that I deleted.
 - an unused field in `ipte`: I deleted the field. We could also have ignored the warning by adding `[@@warning "-69"]` just after the `ipte` type declaration (something like line 401)

HadrienRenaud:
In the end, I pass every test in `herd/tests/instructions/AArch64` except L043 and L045 which depends on a swap graph that I don't support. This is expected behaviour.

maranget:
Hi  @HadrienRenaud, I am in favor of squashing, if you don't mind.

HadrienRenaud:
Hi @maranget,

Do you want me to do it and rebase on top of master?

maranget:
It would be nice, at least reduce the number of commits to a few (one is fine too) and rebase.

lib/lexSplit.mll
```diff
@@ -21,7 +21,7 @@ exception Error
 let blank = [' ''\n''\r''\t']
 let non_blank = [^' ''\n''\r''\t']
 let digit = ['0'-'9']
-let printable = ['0'-'9''a'-'z''A'-'Z'':']
+let printable = [^' ''\n''\r''\t'',']
```
maranget, line 24: Are your sure? Not all chars in this set are "printable"...

HadrienRenaud: I think this regex is only used to parse variants, via the lexer `string_spaces`, so maybe I should rename it to `not_space_nor_comma`?

maranget: No need to do that...

PR #552, opened 2023-03-28
[gen] Fix diy default

maranget:
Fix diy7 default mode.


maranget:
Merged by [consensus](https://github.com/herd/herdtools7/issues/531#issuecomment-1508422154).

PR #553, opened 2023-03-28
[lib/gen]: fixed __int128 parsing

lukeg101:
Motivation: Test int128 programs generated by diy are not parsed by herd.

This patch fixes the following bugs in the int128 implementation:
- `diy7` did not parse `-type uint128_t`
- litmus tests with the initial state `_Atomic __int128 x = 0` did not parse
- `diy` generates C/C++ litmus tests using the `(u)int128_t` which not valid in LLVM/GCC. Instead we generate tests using `__(u)int128`.

A future patch will fix `herd` so that it accepts `__int128` data in AArch64 (It is conceptually a separate patch that deals with Semantics).

Tests are provided in this patch

### DIFF
--- a/herd/tests/instructions/C/C11.litmus
+++ b/herd/tests/instructions/C/C11.litmus
@@ -1,0 +1,19 @@
+C C11
+
+(* Test to ensure _Atomic __int128 x = 0; is parsed**)
+
+{__int128 y; __int128 x; _Atomic __int128 z = 0;}
+
+P0 (volatile __int128* y,volatile __int128* x) {
+  __int128 r0 = *x;
+  atomic_thread_fence(memory_order_relaxed);
+  *y = 1;
+}
+
+P1 (volatile __int128* y,volatile __int128* x) {
+  __int128 r0 = *y;
+  atomic_thread_fence(memory_order_relaxed);
+  *x = 1;
+}
+
+exists (0:r0=1 /\ 1:r0=1)
--- a/gen/typBase.ml
+++ b/gen/typBase.ml
@@ -29,6 +29,7 @@ let tags =
    "int32_t"; "uint32_t";
    "int64_t"; "uint64_t";
    "int128_t"; "uint128_t";
+   "__int128"; "__uint128";
  ]
 
 let parse s = match s with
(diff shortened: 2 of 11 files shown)


lukeg101:
This patch is ready for review and will help colleagues with a compiler question.

herd/tests/instructions/C/C09.litmus
maranget, line 1: This test is the same as C08.litmus... Delete?

herd/tests/instructions/C/C10.litmus
```diff
@@ -0,0 +1,19 @@
+C LB004
```
maranget, line 1:
```suggestion
C C10
```

herd/tests/instructions/C/C11.litmus
```diff
@@ -0,0 +1,19 @@
+C LB004
```
maranget, line 1:
```suggestion
C C11
```

herd/tests/instructions/C/C10.litmus
maranget, line 1: For what it's worth, my gcc does not know about the `__uint128_t` type

herd/tests/instructions/C/C11.litmus
maranget, line 1: For what it's worth, my version of gcc chokes on this test, with message:
```
undefined reference to __atomic_load_16
```

maranget (changes requested):
Before merging I would like the tests to be given internal names that correspond to their file name. Moreover test C09 is a duplicate of test C08.

lukeg101:
Hi Luc,
Thank you for this code review, I appreciate it. I have deleted the redundant test and renamed the others to follow the test convention.

A note on `atomic_load_16`, GCC and LLVM differ on the code generation for 128-bit atomic loads, precisely LLVM emits inlined assembly and GCC emits this function call. To ensure your GCC works you need to link in the correct atomics library, for instance https://godbolt.org/z/KEf1ncEMK.

As far as I understand we do not support function calls of this kind in the tool suite. Shall I remove C11.litmus?

herd/tests/instructions/C/C11.litmus
maranget: My remark was.. a remark. Given the experimental nature of the 128 bit integer types for herdtools7 and the work involved in a complete implementation, we can accept such limitations.

maranget (approved):
All changes performed, thanks.

maranget:
Squashed and merged. Thanks @lukeg101.

PR #554, opened 2023-03-29
[herd] Add possibility for architecture specific binary operations

HadrienRenaud:
This commit adds a variant to `Op.op` and it now takes a type argument as `Op.op1`.

To make it easier, I've added a functor in `ArchOp` to take previous `ArchOp` and make them compatible.

Difficulties came from:
 - ASL translation of `ArchOp` from `ASLValues` to `AArch64Values`
 - C and Java AST depends on `Op.op`.

It is now possible to add new Arch-dependent binary operations, e.g. bitvector concatenation in ASL.


HadrienRenaud:
TODO:
 - Fix `herd/parseTest.ml` to use JavaParseTest
 - use the same phantom types for `no_archop` and remove `convert_common_op`

maranget:

PR #557, opened 2023-04-05
[tools] Implement option to negate final condition with mlog2cond

hugookeeffe:
This commit implements the ability to negate the final condition when using mlog2cond7. 

The reason for implementing this is to catch cases where HW is completely unpredictable and breaks all the rules, such as memory/register corruption issues in HW. This would create a state that isn't included within the forbidden state of a litmus test, but we would still want to catch.

By negating the possible states from herd7 output (for forbidden tests), we would be able to catch the forbidden state as well as unpredictable states due to HW issues.

e.g.
Litmus test with following forbidden state:
`exists (1:X1=1 /\ 1:X2=1)`

has these possible states (from herd output)
`1:X1=1;1:X2=0;`
`1:X1=0;1:X2=1;`

we want as forbidden state
`~exists ((1:X1=1 /\ 1:X2=0) \/ (1:X1=0 /\ 1:X2=1))`

I tried to do this previously using recond7, however was unable to do it. If there is a way to get this result by using the current tools, please let me know!

### DIFF
--- a/tools/mlog2cond.ml
+++ b/tools/mlog2cond.ml
@@ -26,6 +26,7 @@ let acceptempty = ref false
 let hexa = ref false
 let int32 = ref true
 let faulttype = ref true
+let neg = ref false
 
 let options =
   [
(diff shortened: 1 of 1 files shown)


maranget:
Hi @hugookeeffe, doesn't option `-forall true` suffices for your need?

PR #560, opened 2023-04-17
[herd] Fix double neg fault

maranget:
Fix a glitch : some lines in outcomes with negative faults were printed more than once.

The root cause is the introduction of fault types in negative faults matching. More precisely, when the pattern contains some fault type A, and that there are two outcomes that are almost identical, except that  one has fault of type B and the other has no fault, then two identical lines with ~fault(...,A) were printed. The fix consists in discarding  the outcome with the fault of type B, as the observation of this fault is in fact not specified.

This PR  complements PR #473.

### DIFF
--- a/herd/tests/instructions/AArch64.kvm/A017.litmus
+++ b/herd/tests/instructions/AArch64.kvm/A017.litmus
@@ -1,0 +1,17 @@
+AArch64 A017
+variant=fatal
+{
+int x=1 ;
+[PTE(x)]=(valid:0);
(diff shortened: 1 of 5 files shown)


relokin (approved):
The changes look good to me Luc!

maranget:
Fine, I think I'll merge very soon,  since the fix is quite simple and the glitch quite obvious. I'll add a test though.

relokin:
Yes a test would very useful!

PR #562, opened 2023-04-18
Add support for address translation of the PC prior to ifetch

relokin:
This PR changes the type of Label and makes one of the two types of Symbolic Virtual Constants. This way a Label can be translated and mapped to a Symbolic Physical Constant. Leveraging this, herd7 when self and vmsa are both enabled, for instructions with labels, it introduces a Implicit Read event to Symbolic Virtual Label as well as a branching decision. Much like for data translation for instruction fetches can raise faults.


relokin:
There are a few things that are work in progress:
* Support in litmus (requires support for self in kvm mode first)
* herd7 should translate all instructions with labels, regardless of whether there is a register initialized to that label or not
* Do we need a new mode `vmsa-instr`? If yes, we could rename `vmsa` to `vmsa-data` make `vmsa` a variant that enables both. Or otherwise we keep `vmsa` as is and when we use both `vmsa` and `self` we enable this functionality.

jalglave:
hi @relokin is there any reason to keep this as a draft? How can we help progress on this?
Thanks in advance,
Jade

PR #563, opened 2023-04-18
[All]: Add Arm Support for Several New Instructions

lukeg101:
Motivation: I wish to test the compilation targeting Armv7, as such I add several new instructions and code improvements to bring the ARM Parser in line with the AArch64 parser. Each commit is as follows (with tests):
- Add MOVW,MOVT, BX, put instruction tokens in check_name (like AArch64Parser)
- Add LDR Offset instruction
- Add LDM Instruction Support
- Add Arm IP register support
- Add LDRD Instruction support
- Add ORR Instruction support

In each case regression tests are added and unless stated below, all pass in herd and litmus.

Thank you to Nikos for kindly letting me use his Raspberry Pi 2 to test these patches.
- Test A009 does not compile using litmus, as the Raspberry Pi 2 does not support the instruction. I cannot test this.
- Test A015 is a work in progress as it causes a core dump, and I do not understand why.

Interestingly the following tests produce strange and rather unusual behaviour on the Raspberry Pi when run under litmus. I wonder if there is a hardware bug in the implementation of `MOVT`. Consider the following litmus test:
```
ARM A012
{ 0:R0 = 0 }

P0;
  MOVT R0, #0;

forall 0:R0=0x0
```
That is, we expect the `MOVT` to move 0 into the top _move top_ 16 bits of register `R0`.

When run using litmus on the Raspberry Pi 2, we get the following strange output:
```
%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
% Results for herd/tests/instructions/ARM/A012.litmus %
%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
ARM A012

{
 0:R0=0x0;
}
 P0          ;
 MOVT R0, #0 ;

forall (0:R0=0x0)
Generated assembler
#START _litmus_P0
        movt r2,#0
Test A012 Required
Histogram (16384 states)
60    :>0:R0=0x0;
60    *>0:R0=0x4;
60    *>0:R0=0x8;
60    *>0:R0=0xc;
60    *>0:R0=0x10;
60    *>0:R0=0x14;
...
...
...
60    *>0:R0=0xfff0;
60    *>0:R0=0xfff4;
60    *>0:R0=0xfff8;
60    *>0:R0=0xfffc;
No

Witnesses
Positive: 60, Negative: 999940
Condition forall (0:R0=0x0) is NOT validated
Hash=741ae0900c1ec08d09c13e59fa1e17fe
Observation A012 Sometimes 60 999940
Time A012 194.56
```
It appears to be counting up in steps of 4 bits at a time - each execution is writing a different number to `R0`. Very strange! I did not expect such a simple MOV instruction to do this. It is not clear if the `0` has been moved into the top 16 bits of this example either. Let us try another test to see if the `MOVT` moves 1 into the top 16 bits:
```
ARM A011
{ 0:R0 = 0 }

P0;
  MOVT R0, #1;

forall 0:R0=0x10000
```
This produces the following output using Litmus on the raspberry pi 2:
```
%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
% Results for herd/tests/instructions/ARM/A011.litmus %
%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
ARM A011

{
 0:R0=0x0;
}
 P0          ;
 MOVT R0, #1 ;

forall (0:R0=0x10000)
Generated assembler
#START _litmus_P0
        movt r2,#1
Test A011 Required
Histogram (16384 states)
60    :>0:R0=0x10000;
60    *>0:R0=0x10004;
60    *>0:R0=0x10008;
...
...
...
60    *>0:R0=0x1fff4;
60    *>0:R0=0x1fff8;
60    *>0:R0=0x1fffc;
No

Witnesses
Positive: 60, Negative: 999940
Condition forall (0:R0=0x10000) is NOT validated
Hash=6c83921ab28398e13c6a9d52e8461687
Observation A011 Sometimes 60 999940
Time A011 192.14
```
So it appears that yes `1` is indeed moved into the top 16-bits of `R0`, however the lower 16 bits are still  being written two by many other executions. It is not clear to me why this is the case - this feels like a hardware bug.

Please let me know what you think.
Thanks
Luke

### DIFF
--- a/lib/ARMBase.ml
+++ b/lib/ARMBase.ml
@@ -64,12 +64,30 @@ let regs =
    R10, "R10" ;
    R11, "R11" ;
    R12, "R12" ;
+   R12, "IP" ;
    SP, "SP" ;
    LR, "LR" ;
    PC, "PC" ;
    Z, "Z" ;
    RESADDR, "RESADDR" ;
  ]
+
+(* this is used by instructions which load registers in pairs*)
+(* by specifying Rd1, and implicitly loading Rd2 - see LDRD  *)
+let next_reg = function
+  | R0 -> R1
+  | R1 -> R2 ;
+  | R2 -> R3 ;
+  | R3 -> R4 ;
+  | R4 -> R5 ;
+  | R5 -> R6 ;
+  | R6 -> R7 ;
+  | R7 -> R8 ;
+  | R8 -> R9 ;
+  | R9 -> R10 ;
+  | R10 -> R11 ;
+  | R11 -> R12 ;
+  | _ -> assert false
 
 let to_parse =
   List.filter
@@ -86,7 +104,7 @@ let parse_list =
        regs)
 
 let parse_reg s =
-  try Some (List.assoc s parse_list)
+  try Some (List.assoc (Misc.uppercase s) parse_list)
   with Not_found -> None
 
 let pp_reg r = match r with
@@ -172,13 +190,17 @@ type setflags = SetFlags | DontSetFlags
 
 type condition = NE | EQ | AL (* ALWAYS *)
 
+type increment = NO | IB
+
 type 'k kinstruction =
   | I_NOP
   | I_ADD of setflags * reg * reg * 'k
   | I_ADD3 of setflags * reg * reg * reg
+  | I_BX of reg
   | I_SUB of setflags * reg * reg * 'k
   | I_SUB3 of setflags * reg * reg * reg
   | I_AND of setflags * reg * reg * 'k
+  | I_ORR of setflags * reg * reg * 'k
   | I_B of lbl
   | I_BEQ of lbl
   | I_BNE of lbl (* Was maybeVal ??? *)
@@ -187,12 +209,18 @@ type 'k kinstruction =
   | I_CMP of reg * reg
   | I_LDR of reg * reg * condition
   | I_LDREX of reg * reg
+  | I_LDRO of reg * reg * 'k * condition
+  | I_LDM2 of reg * reg * reg * increment
+  | I_LDM3 of reg * reg * reg * reg * increment
+  | I_LDRD of reg * reg * reg * 'k option
   | I_LDR3 of reg * reg * reg * condition
   | I_STR of reg * reg * condition
   | I_STR3 of reg * reg * reg * condition
   | I_STREX of reg * reg * reg * condition
   | I_MOVI of reg * 'k * condition
   | I_MOV of reg * reg * condition
+  | I_MOVW of reg * 'k
+  | I_MOVT of reg * 'k
   | I_XOR of setflags * reg * reg * reg
   | I_DMB of barrier_option
   | I_DSB of barrier_option
@@ -239,9 +267,21 @@ let do_pp_instruction m =
     ","^ "[" ^ pp_reg rn ^ "]" in
   let ppi_rrm opcode rt rn =  ppi_rrmc opcode rt rn AL in
 
+  let pp_memom opcode = function
+  | NO -> opcode
+  | IB -> opcode ^ "IB" in
+
+  let ppi_rr_multiple opcode rt rs i =
+    pp_memom opcode i ^ " " ^ pp_reg rt ^ ", {" ^
+        (String.concat "," (List.map pp_reg rs)) ^ "}"in
+
   let ppi_rrrmc opcode rt ri rn c =
     pp_memoc opcode c^" "^pp_reg rt ^ ","^
     "[" ^ pp_reg ri ^ "," ^ pp_reg rn ^ "]" in
+
+  let ppi_rrkmc opcode rt ri k =
+    opcode ^" "^pp_reg rt ^ ","^
+    "[" ^ pp_reg ri ^ "," ^ m.pp_k k ^ "]" in
 
   let ppi_strex opcode rt rn rm c =
      pp_memoc opcode c^" "^pp_reg rt ^ ","^
@@ -261,6 +301,7 @@ let do_pp_instruction m =
   | I_SUB(s,rt,rn,v) -> ppi_rri "SUB" s rt rn v
   | I_SUB3 (s,r1,r2,r3) -> ppi_rrr "SUB" s r1 r2 r3
   | I_AND(s,rt,rn,v) -> ppi_rri "AND" s rt rn v
+  | I_ORR(s,rt,rn,v) -> ppi_rri "ORR" s rt rn v
   | I_B v -> "B " ^ pp_lbl v
   | I_BEQ(v) -> "BEQ "^ pp_lbl v
   | I_BNE(v) -> "BNE "^ pp_lbl v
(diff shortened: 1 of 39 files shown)


lukeg101:
Additionally I have tried the above by putting fences around the mov: `dmb ish; movt...; dmb ish` with no change in outcome

lib/ARMBase.ml
```diff
@@ -86,7 +104,7 @@ let parse_list =
        regs)
 
 let parse_reg s =
-  try Some (List.assoc s parse_list)
+  try Some (List.assoc (String.uppercase s) parse_list)
```
maranget, line 107: String.uppercase has long been deprecates and is no longer available in recent ocaml systems
```suggestion
  try Some (List.assoc (Misc.uppercase s) parse_list)
```

maranget:
Hi @lukeg101, for your information I have tried to compile and run your tests, using litmus7 options `-conf pi2 -ascall true`:
1. A005.exe does not compile:
```
% make A005.exe
gcc -D_GNU_SOURCE -DFORCE_AFFINITY -Wall -std=gnu99 -march=armv7-a -O2 -pthread -S A005.c
gcc -D_GNU_SOURCE -DFORCE_AFFINITY -Wall -std=gnu99 -march=armv7-a -O2 -pthread  -o A005.exe affinity.o outs.o utils.o litmus_rand.o A005.s
A005.s: Assembler messages:
A005.s:30: Error: first transfer register must be even -- `ldrd r1,r2,[r0]'
Makefile:46: recipe for target 'A005.exe' failed
make: *** [A005.exe] Error 1
```
2. I have results different from yours for the MOVT tests, for instance:
```
% ./A011.exe  -s 5k -r 2k
Test A011 Required
Histogram (1 states)
40000000*>0:R0=70535;
No

Witnesses
Positive: 0, Negative: 40000000
Condition forall (0:R0=65536) is NOT validated
Hash=6c83921ab28398e13c6a9d52e8461687
Observation A011 Never 0 40000000
Time A011 1.97
```
Notice that the final value of R0 above does not corrrespond to the value in the the final condition,
3. I could compile and run test A009:
```
% ./A009.exe -s 5k -r 2k 
Test A009 Required
Histogram (1 states)
40000000:>0:R1=2; 0:R2=3;
Ok

Witnesses
Positive: 40000000, Negative: 0
Condition forall (0:R1=2 /\ 0:R2=3) is validated
Hash=fe765959009fb1eacd7bee92c02fe8fb
Observation A009 Always 40000000 0
Time A009 4.58
```
As far as I know, the target computer is a raspberry Pi 2. (notice that I used the configuration option `-conf pi2`).
4. I can confirm that A015 causes a core dump.

litmus/ARMCompile_litmus.ml
```diff
@@ -79,6 +83,20 @@ module Make(V:Constant.S)(C:Config) =
...
+      { empty_ins with
+        memo = sprintf "%s ^o0,#%i" memo i ;
+        inputs = [] ;
+        outputs = [r1]; }
+
+    let movt r1 i =
+      let memo = "movt" in
+      { empty_ins with
+        memo = sprintf "%s ^o0,#%i" memo i ;
+        inputs = [] ;
```
maranget, line 97: The `MOVT` instruction reads its register argument, this is important for `litmus7` to output code that initialize this register:
```suggestion
        inputs = [r1] ;
```

maranget:
See above for fixing the problem of the `MOVT` instruction.

maranget:
BTW,  test `A003.litmus` has the internal name `SB004`, while `A003` would be a better name. Moreover as the test is forbidden, one may change `exists` into `~exists` in its source.

maranget:
Output of `litmus7` and of `herd7` differ on test A008:

`litmus7`  output:
```
%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
% Results for herd/tests/instructions/ARM/A008.litmus %
%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
ARM A008

{
 int x[3]={0x1,0x2,0x3};
 0:R0=x;
}
 P0                 ;
 LDM R0, {R1,R2,R3} ;

forall (0:R1=0x1 /\ 0:R2=0x2 /\ 0:R3=0x3)
Generated assembler
#START _litmus_P0
        ldm r0,{ip, r4, lr}
Test A008 Required
Histogram (1 states)
400000*>0:R1=0x2; 0:R2=0x1; 0:R3=0x3;
No

Witnesses
Positive: 0, Negative: 400000
Condition forall (0:R1=0x1 /\ 0:R2=0x2 /\ 0:R3=0x3) is NOT validated
Hash=0e45af57c624e78be7cd0fd317d7e51b
Observation A008 Never 0 400000
Time A008 0.06
```
`herd7` output:
```
Test A008 Required
States 1
0:R1=1; 0:R2=2; 0:R3=3;
Ok
Witnesses
Positive: 1 Negative: 0
Condition forall (0:R1=1 /\ 0:R2=2 /\ 0:R3=3)
Observation A008 Always 1 0
Time A008 0.00
Hash=0e45af57c624e78be7cd0fd317d7e51b
```

lukeg101:
Hi Luc, 
Thank you for your patience, I have made the changes following your advice, and using the `pi2` configuration. 

I also made some tweaks to fix GCC compile errors with a new version of GCC on Ubuntu 22.

I am still working on:
- `A015` this now fails to compile rather than giving a core dump.
- `A005` warns at compile time - gives non-sense litmus runs as a result.

I cannot test:
- I still cannot compile `A009` due to ` Error: selected processor does not support `ldmib r0,{r3,r4}' in Thumb mode` - thank you for validating that test.

Please bear with me whilst I figure out why A015 and A005 do not work.

lukeg101:
I could not figure out why `A015` does not emit a `getnop` function in `presSi.ml` I will continue to look next week.

lukeg101:
Test `A005` now builds without warning and gives the correct behaviour on a pi 2

maranget:
> could not figure out why `A015` does not emit a `getnop` function in `preSi.ml` I will continue to look next week.

This one is not an easy one. I have attempted a [fix](8340dd4b6d89d775bac2b5cc305b9cb878150321). The fix is not completely satisfactory for reasons detailed in the commit message. Moreover, the whole code looks brittle, with complex underlying logics and duplication...

maranget:
> could not figure out why `A015` does not emit a `getnop` function in `preSi.ml` I will continue to look next week.

This one is not an easy one. I have attempted this  [fix]()https://github.com/herd/herdtools7/commit/50a257d85fb343e37d93f8aa9170138489ad0f30. The fix is not completely satisfactory for reasons detailed in the commit message. Moreover, the whole code looks brittle, with complex underlying logics and duplication...  However I am quite confident the fix is sufficient for the technique of initialising registers with code labels illustrated by test `A015`.

lukeg101:
Thank you very much Luc! I will put your commit on the branch and then test

lukeg101:
Apologies I've pulled in some commits from the main branch - please bear with me whilst I fix

lukeg101:
Thanks for this Luc, I ran `A015` on the pi2 and the `c` skeleton builds but returns a seg fault when run, as the `find_ins` function, called from `prelude_size`, goes out of bounds. Presumably because it looks for a NOP at the end of the function that isn't there.

I will continue to debug this after I have tested and submitted another PR I have in the pipeline (similar code, but for x86)

maranget:
Hi @lukeg101. I cannot reproduce here: that is, I can run A015. Notice that I am using gcc option `-march=armv7-a`, since option `-march=armv7-a+fp` i snot avaiable on my system.

maranget:
For your information, I am using [this configuration file](https://github.com/herd/herdtools7/blob/master/litmus/libdir/pi2-nofp.cfg). That is, I use option `-mach pi2-nofp`

lukeg101:
Hi Luc,
I tried to run the above tests using the `nofp` configuration but I cannot compile the source tests. I fear we may spend more time trying to diagnose my GCC and OS versions than it is worth. I propose we merge If you are ok with the above code?

maranget:
Hi Like, the PR is still in draft status. Please remove the draft status if you think it is ready.

I am not at ease if `A015` segfaults on your system. Are you sure about your diagnostic, _i.e._ that the segfault originates in `find_ins` going out of bounds?

lukeg101:
Hi Luc, Thank you for your help again. Nikos kindly installed GDB so I will debug the generated files to find the source of the seg fault.
I will also make the above changes you suggest regarding `ascall` and the opt level.

lukeg101:
I can confirm the seg fault is in `find_ins`. Using GDB the stack trace points to the for loop that checks whether `nop` matches an opcode in `code0` (the assembly for `P0` in `A015`). I am comparing the C output by `L025` with that of `A015` to understand how they differ...

maranget:
Hi, you can also send me the exact product of compiling A015, so that I can try it here.

lukeg101:
emailed you, thanks :)

maranget:
I have not received this email...

lukeg101:
Thank you very much for your help Luc :) I have added the `-marm` options for the pi configs and simplified A015. All work on my Pi2 now. Thank you also Nikos for assisting with the Pi.

maranget:
Squashed and merged, thanks.

PR #564, opened 2023-04-18
[AArch64] Adjustments to the VMSA model in sync with Arm ARM J.a

artkhyzha:
This PR performs the following adjustments:
* Gives a name to a relation already included into `ob` (`haz-ob`)
* Removes several redundant endpoint constraints in relations (for instance, in `tlbi-ca`)
* Renames the relation `si` into `sca-class` and refines its use in `lob` and `ob`
* Renames occurrences of the `PTE` set into `TTD`
* Reorders endpoint constraints from, say, `R & PTE & Imp` into `Imp & TTD & R`

### DIFF
--- a/herd/libdir/aarch64.cat
+++ b/herd/libdir/aarch64.cat
@@ -52,16 +52,16 @@ include "aarch64hwreqs.cat"
 let ca = fr | co
 
 (* Observed-by *)
-let obs = ([M & Exp | R & T & Imp]; (rf | ca); [M & Exp | R & T & Imp]) & ext
+let obs = ([Exp & M | Imp & T & R]; (rf | ca); [Exp & M | Imp & T & R]) & ext
 
 (* TLBI-after *)
 include "tlbi-after.cat"
-with tlbi-after from (pte-tlbi-pairs hw-reqs)
+with tlbi-after from (ttd-tlbi-pairs hw-reqs)
 
 (* TTD-read-ordered-before *)
 let TTD-read-ordered-before =
-  tlbi-after; [TLBI]; po; [dsb.full]; po; [~(M&Imp)]
-  | (if "ETS2" then tlbi-after; [TLBI]; po; [dsb.full]; po; [M & PTE & Imp] else 0)
+  tlbi-after; [TLBI]; po; [dsb.full]; po; [~(Imp & M)]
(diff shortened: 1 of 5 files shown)


maranget:
Hi @artkhyzha. For such a PR, I think it is better to squash all commits into one signe commit:
1. For the sake of simplicity, after all all those changes are part of the same batch, resulting from the same review.
2. Some people find bugs by bisection. Such a technique is made easier when the system compiles and that tests succeeds after each commit. Although I think this is the case here (I have not checked), squashing commits facilitates checking this property.

Notice that I can squash the commits myself. However, I'd rather have you to do it, as a clear sign that the PR is ready. This also gives you the opportunity of a last look at the code...

relokin:
Would a good compromise be to do squash within a merge commit (git merge --squash)? One argument is that these three commits are independent and self contained changes and if one can reason about/revert one commit independently of the others. The benefit is that git log wouldn't be as long. The downside of merging (if ofc all commits compile and are self contained) is that it makes some other things harder: for example, what if we changed our minds about haz-ob and we want to revert this one commit but keep all other changes?

artkhyzha:
Hi both, thanks for the feedback! In principle, each commit is supposed to be well-formed, but, formally speaking, full-blown testing happened only on top of all three commits. It seems that either squashing or running all tests on intermediate commits should be the action.

Happy to act on this, but another opportunity to have a look at the code helped to identify some minor cosmetic changes we might want. Will update soon.

maranget:
Hi @relokin. I am using github merge button as much as possible. With the "squash-and-merge"  choice, there is no merge commit, thereby yielding a less informative history, as I tend to copy the "description" comment of a PR as the merge commit PR. So for reducing the numbers of commit _and_ keeping the merge commit, the one solution I know of is squashing commits before merging. Rebasing before merging also yields a more legible time-based history. Most of the time, I can do it myself easily, if the PR author did not forbid pushes to its branch, or by more complex git machinery. However I'd rather had the author to reduce commit numbers.

That said we can also keep commits separated, if the author says so.

artkhyzha:
Thanks again for the feedback, @maranget . I have finished making changes and squashed the commits into one. As per @jalglave's request, may I ask you to double-check that the overall PR regresses -- and then merge if all is fine? (There shouldn't be a problem as the recent changes were purely cosmetic, but just to be sure.)

PR #567, opened 2023-04-21
[asl] Fix build for dune<3.0

HadrienRenaud:
Questions that need answer before merging:
 - [x] Should we stop supporting dune<1.6 when it is not compatible with ocaml>=4.08 ?
 - [x] `asllib` with this patch is still not building with dune=1.6.0 for a reason that I don't really understand (see later comment).
 - [x] Fix deprecated dune usage with dune>=1.6.3

### DIFF
--- a/asllib/tests/dune
+++ b/asllib/tests/dune
@@ -8,7 +8,7 @@
   (modes native)
   (modules asltests)
   (libraries asllib test_helpers)
-  (deps (glob_files asl/required/*.asl) %{test} ../libdir/stdlib.asl)
+  (deps (glob_files asl/required/*.asl) asltests.ml ../libdir/stdlib.asl)
   (action
     (setenv ASL_LIBDIR %{project_root}/asllib/libdir
       (run %{test} asl/required))))
--- a/dune
+++ b/dune
@@ -1,4 +1,4 @@
-(ignored_subdirs (herd-www))
+(dirs :standard \ herd-www)
 (env
  (release
 (flags (:standard -w +a-3-4-9-29-33-41-45-60-67-70))))
--- a/dune-project
+++ b/dune-project
@@ -1,3 +1,3 @@
-(lang dune 1.4)
+(lang dune 1.6)
 (name herdtools7)
 (using menhir 2.0)
--- a/herdtools7.opam
+++ b/herdtools7.opam
@@ -1,8 +1,7 @@
 opam-version: "2.0"
 name: "herdtools7"
-version: "7.56+02~dev"
+version: "7.56+04~dev"
 synopsis: "The herdtools suite for simulating and studying weak memory models"
-description: ""
 maintainer: "Luc Maranget <Luc.Maranget@inria.fr>"
 authors: [
   "Luc Maranget <Luc.Maranget@inria.fr>"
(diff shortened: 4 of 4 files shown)


HadrienRenaud:
With `dune=1.6.0,ocaml=4.08.1,menhir=20220210`, I have the following error when trying to build:

```
+ dune build --profile release @all
Module Parser in directory _build/default/asllib depends on Asllib.
This doesn't make sense to me.

Asllib is the main module of the library and is the only module exposed 
outside of the library. Consequently, it should be the one depending 
on all the other modules in the library.
Module Parser0 in directory _build/default/asllib depends on Asllib.
This doesn't make sense to me.

Asllib is the main module of the library and is the only module exposed 
outside of the library. Consequently, it should be the one depending 
on all the other modules in the library.
make: *** [build] Error 1
```

At the moment, I don't have an explanation for this.

maranget:
Hi I remember I ran into the same problem. Upgrading dune solved the problem, if I remember the whole story.

HadrienRenaud:
Hi @maranget,

This is indeed why no one ran into this issue since two months when this was merged into master.
However I'd like to have something that builds for all supported version, i.e. with the versions mentioned in:
 - `dune-project`, today `lang 1.4`
 - `herdtools.opam`, today `"dune"  {>= "1.4" }` (and `"ocaml" {>= "4.05.0"}` by the way)

I'd happily drop a few version until this patch is not supported, but I'm afraid it isn't until `dune>=3.0`.

maranget:
Hum, the right opam file is the one from the opam repository:
https://github.com/ocaml/opam-repository/blob/master/packages/herdtools7/herdtools7.7.56.3/opam.

Minimal ocaml version is now 4.08.0. (and the herdtools.opam file in this repository is obsolete, and should probably be deleted).

As regards dune, are you sure we need dune >= 3.0.0 and if yes, why is this a problem?

HadrienRenaud:
Thank you for your explanations.

> Hum, the right opam file is the one from the opam repository: https://github.com/ocaml/opam-repository/blob/master/packages/herdtools7/herdtools7.7.56.3/opam.

Oh ok. I don't think I need to change the current opam-released opam file. How do I change the opam file for next release?

> Minimal ocaml version is now 4.08.0. (and the herdtools.opam file in this repository is obsolete, and should probably be deleted).

I'll push a commit on this branch/PR for deleting that file then.

> As regards dune, are you sure we need dune >= 3.0.0 and if yes, why is this a problem?

I don't think that we need dune >= 3.0.0. The goal of this PR is to fix build system for dune < 3.0.0, and make sure that it builds for all dune versions from the oldest supported (1.4 before, 1.6 after what I propose) to the newest ones.

maranget:
FYI I could compile the system with dune 2.8,2 and ocaml 4.11.1...

HadrienRenaud:
Ok, dune < 3.0.0 is an over-approximation on my part, I only tested dune 2.0 or something.

With dune 1.6.3 and ocaml 4.08.1, `make build` fails while trying to build asllib/tests/dune. I don't know why they are included in 1.6 `dune build @all` and not in 2.8.2.

HadrienRenaud:
Furthermore, I think an opam file is required by dune for building, so removing the file does not work.

Could the opam file in herdtools reflect the one that will be used in the next release? I'm happy to change it to the file that is currently in the opam repository, and apply my changes on top of this one.

maranget:
> Could the opam file in herdtools reflect the one that will be used in the next release? I'm happy to change it to the file that is currently in the opam repository, and apply my changes on top of this one.

Nice idea, please proceed, and why not in this PR.

maranget:
Thanks a lot, squashed and merged

PR #572, opened 2023-05-02
[all]: Add X86_64 Support for Several New Instructions

lukeg101:
Motivation: We wish to test the compilation of X86_64 instructions.

To do this I need to add support for several new instructions and test them with both herd and litmus.

This PR does the following:
- Rework the x86_64 parser so that `check_names` exports the names of instruction tokens so we can use it when constructing an ELF grammar.
- Add automated regression testing for the X86_64 architecture, with example
- Add the `ret` instruction to herd
- Add support for the `RIP` register
- Add support for various NOP instructions used for padding in x86 code 
- Add support for the `CS` register
- Add support for the `endbr64` instruction

Tests are added that pass in both herd and litmus (tested with a XEON gold processor).


lukeg101:
Marking as draft: I found more instructions that need adding for a large-scale experiment I am running

lukeg101:
With the exception of `A006`, all tests are passing when run on a dual CPU Xeon Gold

lukeg101:
@maranget I do not know enough about litmus and how it handles branches to figure out why test `A006` is Seg faulting (all other tests are passing). Do you have advice of where I can look? It is the only test preventing this patch set from being marked 'ready to review'

maranget:
> @maranget I do not know enough about litmus and how it handles branches to figure out why test `A006` is Seg faulting (all other tests are passing). Do you have advice of where I can look? It is the only test preventing this patch set from being marked 'ready to review'

I assume that the test crashes because of the ret instruction. You cannot test it that way. 

Basically, subroutines have to be callled from the test itself. It is easier when caller and callee are on the same thread code. See this [AArch64 test](https://github.com/herd/herdtools7/blob/71a340005a4ed9e9bc8d5f46220e53e5928c8a4d/herd/tests/instructions/AArch64/A250.litmus#L1) for an example.

herd/X86_64Sem.ml
```diff
@@ -255,10 +262,35 @@ module
...
+            "illegal argument for the indirect branch instruction %s \
+            (must be a label)" (X86_64.dump_instruction i)
+
+      let build_semantics test ii =
         let rec build_semantics_inner locked ii =
           match ii.A.inst with
           | X86_64.I_NOP -> B.nextT
+          | X86_64.I_RET as i ->
+            read_reg true X86_64.RIP ii
+            >>= do_indirect_jump test [] i
```
maranget, line 293: I am puzzled, I thought that the `RET` instruction popped the return address from the stack.

lukeg101: Apologies yes it appears to do that - so I need to add support for PUSH/POP for this to make sense in herd, and presumably work in litmus also?

lukeg101:
Hi @maranget I guarded ret here on the Telechat variant introduced in https://github.com/herd/herdtools7/pull/636. If you have no more comments, please merge this patch after the PPC patch so as to not cause build issues (it depends on the Telechat variant in that PR).

maranget:
Rebased, squashed and merged, thanks @lukeg101.

lukeg101:
thank you very much, Luc!

PR #579, opened 2023-05-16
[herd] Fix handler at el0

maranget:
Fault handler never executes at EL0.


maranget:
Merge, as fix  is useful for PR #575.

PR #583, opened 2023-05-24
Add support for variant self in KVM mode

relokin:


maranget:
I have looked a bit, and checked that the two tests you sent by mail give the same result on ODROID-C2.

I am wondering how you would consider having `-variant self` in the `-mode presi` mode. The mode "presi" is the base mode of the kvm mode. As far as I have looked this would entail allocating the memory for the code copies in a different manner: allocating  a chunk of cache-line aligned memory for each function in place of allocating specific pages with PTE change. There is no other modification that I can see for now.

It could then be interesting to compare the outcomes of  tests compiled by following the same base technique in different contexts: as user code and on top of kvm.

maranget:
As I have checked litmus in mode `-mode std -variant self` uses `mmap_exec` to allocate memory for code copies and `munmap_exec` to free it.

maranget:
Hi @relokin  I noticed that the memory for code copies is not freed. This may cause a space leak with option `-driver C`.,

Commit 1353d436639a8cd901cc36b07bcbe901bce90034 is a fix attempt.

I suggest another, more cosmetic change: do not include kvm-specific code in non-kvm output,  as implemented by commit 4e81d6ad6261211bbcc905314d2e8cb8dab9ae79.

maranget:
Hi @relokin. I'd rather merge this PR ASAP (_i.e._ without implementing `-mode presi -variant self`. Indeed this PR conflicts with PR #575, whose merge has been delayed for too long, and it looks easier to merge this PR first. 

Moreover this PR is small in contents, great in functionality and  it looks ready. Your opinion?

relokin:
I agree, this is ready, I'm working on the preSi implementation now, but it doesn't all have to go in one PR.

If you would like to merge #575 please do and I will rebase on of it. Otherwise, I am happy for this to be merged.

maranget:
Fine, let us merge this first, then I'll rebase PR #575.

maranget:
Squashed and merged, thanks.

PR #587, opened 2023-05-29
[dune] Fix `dune build` when no js_of_ocaml

HadrienRenaud:
This commit:
 1. Changes the alias `@default` of dune to simply `@install` and `internal/all`. This takes the place of the alias `@zyva` which is deleted. This doesn't change anything for people using the Makefile. The rationale behind this is:

    -  nothing else in the subdirectories needs to be built;
    - `herd-www` does not feature anything installable;
    - aliasing `@default` to `@install` was the behavior for `dune < 2.0.0` which just changed last week with #580;
    - the `make test` part that uses executables that are not installable and not in internal are built with `dune runtest`;
    - probably `make build` should not fail if `js_of_ocaml-ppx` is not installed, see 2.

 2. The preprocessing in `herd-www` with `js_of_ocaml-ppx` is now optional, which should not change anything in our configuration where `dune-build.sh` is doing the building job. Idem for the alias `@jerd`.
 3. `herdtools7.opam` now reflects the facts that:

    - `dune` is now required >= 2.7;
    - `js_of_ocaml-ppx` is an optional dependency.

 4. `defs-mini.sh` is deleted, as now `defs.sh` does not take far more time than `defs-mini.sh`. `cpdir` is moved to its only usage.

All these diffs happened while I was just trying to understand why `dune build` failed on my machine. I am not really attached to anything but I think that this is clearer than what was before.

### DIFF
--- a/defs-mini.sh
+++ b/defs-mini.sh
@@ -1,4 +1,0 @@
-set -o errexit
-
-VERSION=$(cat VERSION.txt)
-REV=$(git rev-parse HEAD 2>/dev/null || echo exported)
--- a/defs.sh
+++ b/defs.sh
@@ -1,16 +1,4 @@
 set -o errexit
 
-. ./defs-mini.sh
-
-cpdir () {
-  if [ "$#" -ne 2 ]
-  then
-    echo "Usage: cpdir <from> <to>"
-    exit 1
-  fi
-
-  local from="${1}"
-  local to="${2}"
-
-  rm -rf "${to}" && mkdir -p "${to}" && ( cd "${from}" && cp -r . "${to}" )
-}
+VERSION=$(cat VERSION.txt)
+REV=$(git rev-parse HEAD 2>/dev/null || echo exported)
--- a/dune
+++ b/dune
@@ -1,10 +1,17 @@
-(dirs :standard \ catalogue )
+(dirs :standard \ catalogue)
+
 (env
  (release
-(flags (:standard -w +a-3-4-9-29-33-41-45-60-67-70))))
+  (flags
+   (:standard -w +a-3-4-9-29-33-41-45-60-67-70))))
+
 (alias
-  (name zyva)
-  (deps (alias internal/all) (alias litmus/all) (alias herd/all)
-        (alias tools/all) (alias jingle/all) (alias gen/all)
-        (alias install)))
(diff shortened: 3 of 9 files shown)


maranget:
Hi @HadrienRenaud, there are many changes here. I am not sure that all of them are necessary. I am a bit concerned about regressions. What command was failing exactly?

Besides, if `js_of_ocaml-ppx` is an optional dependency shouldn't `js_of_caml` also be one?

HadrienRenaud:
Hi @maranget 

Probably none of them are really necessary, although changes to the opam file might be good practice.

`dune build` fails if `js_of_ocaml-ppx` is not installed, but I might be the only one using it.

> Besides, if `js_of_ocaml-ppx` is an optional dependency shouldn't `js_of_caml` also be one?
`js_of_ocaml` is a dependency of `js_of_ocaml-ppx` so no need to specify both. But if you want to make that more explicit, feel free to do so.

HadrienRenaud:
> Moreover, I have checked the absence of regression on `make all`, `make test` and `make` in `herd-www`. The only significant change is the absence of error message when `js_of_ocaml-ppx` is not installed and that one attempts `dune build @jerd`. I guess we can live with that. However it would be nice if one could check the dependency on `js_of_ocaml-ppx`from `herd-www/Makefile`. I do not know how to do that without introducing a new dependency on `ocamlfind`. Do you have any clue?

You could do that with `opam`:
```
opam list js_of_ocaml-ppx --installed --short
```
Which outputs `js_of_ocaml-ppx` if it is installed, or nothing if it is not. That commands takes a few seconds on my machine (on which opam is pretty slow, I don't know why).

You could also probably do it with `dune`, either with parsing the result of `dune installed` either by making `@jerd` not disabled if `js_of_ocaml-ppx` is not installed. In that case, the error message will be the default dune one.

I can try and implement any of those if you like.

maranget:
> [...] making `@jerd` not disabled if `js_of_ocaml-ppx` is not installed. In that case, the error message will be the default dune one.

Ir looks like the most simple and robust solution. Let us do that if you agree, remove the draft status  and merge  swiftly.

HadrienRenaud:
@maranget could you have a look and merge if you feel like it?

PR #588, opened 2023-05-30
[asl] Big bundle of work

HadrienRenaud:
This PR bundle a few commits that:

1. Implement some ASL operations as special binops which simplify ASLSem
2. Improvement into ASL execution in herd
3. Rename asli into aslseq
4. Add instrumentation to ASL interpreter
5. Add circular declarations detection for ASL
6. Add a new `CanPredictFrom` primitive to ASL in herd , which will be used for CAS semantics
7. A first brick of type-checking


herd/ASLSem.ml
```diff
@@ -160,8 +161,35 @@ module Make (C : Config) = struct
...
-    let to_int = M.op1 (Op.ArchOp1 ASLValue.ASLArchOp.ToInt)
+    let to_bv v =
+      match v with
+      | V.Val (Constant.Concrete (ASLScalar.S_BitVector _)) -> return v
+      | V.Var _ | V.Val (Constant.Concrete _) ->
+          M.op1 (Op.ArchOp1 ASLValue.ToBV) v
+      | v ->
+          let v' = V.fresh_var () in
+          M.restrict M.VC.[ Assign (v', Unop (Op.ArchOp1 ASLValue.ToBV, v)) ]
+          >>! v'
```
maranget, line 172: If seen that M.op1 takes care of most of those cases. Wouldn't the following code be functionaly equivalent:
```suggestion
    let to_bv v = M.op1 (Op.ArchOp1 ASLValue.ToBV) v
```
And if there are differences are they justified?

herd/ASLSem.ml
```diff
@@ -160,8 +161,35 @@ module Make (C : Config) = struct
...
+
+    let to_int_signed v =
+      match v with
+      | V.Val (Constant.Concrete (ASLScalar.S_Int _)) -> return v
+      | V.Var _ | V.Val (Constant.Concrete _) ->
+          M.op1 (Op.ArchOp1 ASLValue.ToIntS) v
+      | v ->
+          let v' = V.fresh_var () in
+          M.restrict M.VC.[ Assign (v', Unop (Op.ArchOp1 ASLValue.ToIntS, v)) ]
+          >>! v'
```
maranget, line 192: Same remark calling `M.op1` may do the job or .... ?

herd/ASLSem.ml
```diff
@@ -194,16 +222,13 @@ module Make (C : Config) = struct
     let choice (m1 : V.v M.t) (m2 : 'b M.t) (m3 : 'b M.t) : 'b M.t =
       M.bind_ctrl_seq_data m1 (function
         | V.Val (Constant.Concrete (ASLScalar.S_Bool b)) -> if b then m2 else m3
-        | b -> M.bind_ctrl_seq_data (to_int b) (fun v -> M.choiceT v m2 m3))
+        | b ->
+            M.bind_ctrl_seq_data (to_int_signed b) (fun v -> M.choiceT v m2 m3))
```
maranget, line 226: Why signed?

herd/ASLSem.ml
```diff
@@ -242,11 +267,25 @@ module Make (C : Config) = struct
...
+          | v -> M.op1 Op.Not v >>= M.op1 (Op.ArchOp1 ASLValue.ToBool))
       | NEG -> M.op Op.Sub V.zero
-      | NOT ->
-          fun v -> M.op1 Op.Inv v >>= M.op1 (Op.ArchOp1 ASLValue.ASLArchOp.ToBV)
+      | NOT -> (
+          function
+          | V.Val (Constant.Symbolic _) as v ->
+              let v' = V.fresh_var () in
+              M.restrict M.VC.[ Assign (v', Unop (Op.Inv, v)) ] >>! v'
+          | v -> M.op1 Op.Inv v >>= to_bv)
```
maranget, line 280: Ah, maybe you want to have an additional constraint in case `v` is a "rigid" symbol (instead of failing as it would be the case in other semantics). In such a case, this looks the right way to proceed. That is, leave most of the job to `M.op1`. If so, why not write `to_bv`, `to_int` etc. in the same style?

herd/eventsMonad.ml
```diff
@@ -42,7 +42,9 @@ and type evt_struct = E.event_structure) =
     module VC =
       Valconstraint.Make
         (struct
-          let hexa = C.hexa let debug = C.debug.Debug_herd.solver
+          let hexa = C.hexa
+          let debug = C.debug.Debug_herd.solver
+          let keep_failed_as_undetermined = C.variant Variant.ASL_AArch64
```
maranget, line 47: Why do we have  `-variant asl` and `-variant asl_aarch64`? Aren't both here for commanding the ASL implementation of the `AArch64` architecture?

lib/ASLValue.ml
```diff
@@ -2,79 +2,119 @@ module ASLConstant = SymbConstant.Make (ASLScalar) (PteVal.No) (ASLBase.Instr)
...
+  let do_op1 op cst =
+    match op with
+    | Get i ->
+        let* vec = as_concrete_vector cst in
+        List.nth_opt vec i
+    | ToIntS -> (
+        match cst with
+        | Constant.Concrete s ->
+            ASLScalar.convert_to_int_signed s |> return_concrete
+        | Constant.Symbolic _ -> Some cst
```
maranget, line 92: It looks as if all the operators behave as the identity on "rigid" symbolic values. Probably worth an explanatory comment.

maranget (commented):
I have made a few comments of varying importance.  I am looking forward to your answers and remarks.

herd/eventsMonad.ml
HadrienRenaud: The ASL_AArch64 variant instruct the ASL backend to load the AArch64 shared pseudocode as an ASL library for the litmus test. This is very useful!

I reckon that it's not very explicit. Where is the canonical place to describe variants?

herd/ASLSem.ml
HadrienRenaud: Because herd works with signed integers? It should not make any difference here but still

lib/ASLValue.ml
HadrienRenaud: Done that in cac5b6aa5659f8ccdb3597f39b96c99099c4ea1e

herd/ASLSem.ml
HadrienRenaud: Done in cac5b6aa5659f8ccdb3597f39b96c99099c4ea1e

herd/ASLSem.ml
HadrienRenaud: Done in cac5b6aa5659f8ccdb3597f39b96c99099c4ea1e

HadrienRenaud:
Have this last commit resolved your questions @maranget ?
If so we can move to merge that

herd/ASLSem.ml
maranget: OK

herd/eventsMonad.ml
maranget: The best place for documenting variants  is the interface file ``lib/variant.mli`.

I still do not understand. Assuming that  `-variant asl` and `-variant asl_aarch64` to used for executing AArch64 litmus test, why would loading the AArch64 shared pseudo-code be optional?

maranget:
> 

If so, (but I still have questions), wou;ld you please squash all commits? (I can do it myslef, but then  I appear as co-author..)

herd/eventsMonad.ml
HadrienRenaud: Then the variant was already documented.

Both variant are meant to be used in different architectures:
 - ASL+AArch64 on the ASL arch
 - ASL on the AArch64 arch

We can maybe unify them, and find a name that is explicit in both architectures.

HadrienRenaud: Using herd with ASL architecture without the AArch64 shared pseudocode is useful when studying a very small piece of ASL code to identify its semantics.

HadrienRenaud:
I would not mind having you as co-author, but it is done. For me it's okay to merge, we can resolve the question around variants another time?

So that we can move on with #594.

herd/eventsMonad.ml
maranget: > Then the variant was already documented.

:)

PR #591, opened 2023-06-01
[litmus] Only call free_pages() in kvm mode for variant self

relokin:


maranget:
Thanks a lot @relokin

PR #596, opened 2023-06-21
Add LDA/LDAEX, STL,STLEX Instructions to ARM

lukeg101:
This patch is required to assist Arm's Compiler team.

We wish to test the compilation of AArch32 code. In particular Load Acquire/Store Release (exclusive) instructions.

This patch adds `LDA{EX}` (register) and `STL{EX}` register.

Tests are provided.

This is a work in progress as:
- ~~I am testing these additions on AArch32 hardware (using litmus) as we speak.~~ Validated on Graviton 3
- ~~herd produces incorrect results under the arm.cat model (I have made changes to support acquire/release(exclusive) semantics but the outcomes of `A019` is allowed when it should be forbidden. @maranget do you have any advice on what I may have missed?~~ FIXED

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -187,6 +187,18 @@ arm-test::
 		$(REGRESSION_TEST_MODE)
 	@ echo "herd7 ARM instructions tests: OK"
 
+aarch32-test::
+	@ echo
+	$(HERD_REGRESSION_TEST) \
+		-herd-path $(HERD) \
+		-libdir-path ./herd/libdir \
+		-litmus-dir ./herd/tests/instructions/AArch32 \
+		-conf ./herd/tests/instructions/AArch32/aarch32.cfg \
+		$(REGRESSION_TEST_MODE)
+	@ echo "herd7 ARM instructions tests: OK"
+
+test::aarch32-test
+
 test::
 
 diy-test:: diy-test-aarch64
--- a/herd/ARMArch_herd.ml
+++ b/herd/ARMArch_herd.ml
@@ -15,6 +15,12 @@
 (****************************************************************************)
 
 (** Define ARM architecture *)
+module Types = struct
+  type annot =
+      A | L | X | XL | XA | N | NoRet
+  type lannot = annot
+  type explicit = Exp | NExp
+end
 
 module Make (C:Arch_herd.Config) (V:Value.S) =
   struct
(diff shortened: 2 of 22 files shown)


lukeg101:
Hi there, after my compiler colleagues validated this on a Graviton 3 (that can run AArch32 bins), and that the equivalent AArch64 programs have the same behaviour I am confident this code is correct. @maranget this is the first patch :)

maranget:
Hi @lukeg101, the default model `arm.cat` has changed, are there references for these changes?

However, if I understand, this PR is more about AArch32 than about the old ARM model, wouldn't `aarch64.cat` or a simple adaptation of it work?

lukeg101:
Hi Luc, yes it's about AArch32 predominantly which should be backwards compatible with arm.cat (and use ARMParser etc...). What is the best design here? to adapt the old model or make a new 'aarch32.cat' (I can do this)

maranget:
Well up to adding line `let TagCheck = try TagCheck with emptyset`  after line 59 in `herd/libdir/aarch64util.cat`, the model `aarch64.cat` gives the same result as `arm.cat` on the few ARM regression tests.

So I guess you can perform the above modification of `aarch64util.cat`. And then add a very simple `aarch32.cat` file that would contain the single line `include "aarch64.cat"`. It would be cleaner than changing the historical `arm.cat` file.

I am not sure that `arm.cat` and `arm32.cat` are comparable in the sense that one is weaker than the other, but I'd assume that for any practical purpose one can use any of the two.

maranget:
Unfortunately using `aarch64.cat` for ARM tests is not that straightforward, because `aarch64.cat` computes data, addr etc. dependencies by itself

maranget:
I was  not so difficult, see PR #618.

lukeg101:
Ah we were coding at same time...
- I added an aarch32 file that derives from aarch64
- Added Makefile testing for the aarch32 file
- Split out AArch32-only tests that are not supported by `arm.cat`
- Added `let TagCheck = try TagCheck with emptyset` after line 59 in `herd/libdir/aarch64util.cat`

lukeg101:
looking at your code - we also did the same thing...

maranget:
Ok, LGTM, I'll merge tomorrow. As I'd rather merge a simplified history commit, would you please squash all commit  into one (or have one commit for changing the code, and another for adding the tests).

lukeg101:
all commits squashed, changes to arm.cat/armfences.cat reverted, Thanks Luc :)

maranget:
Merged, thanks @lukeg101.

PR #597, opened 2023-06-23
[herd] Refactor DC and IC actions into a common CMO

relokin:
This patches creates an architecture independent CMO action and uses it to encapsulate the DC and IC events (for AArch64). In parallel, it cleans up the relevant eventsets for AArch64, and similarly to what herd does for barriers, every DC and IC instruction will raise an eponymous event (a dc.cvau,Xt will raise a DC.CVAU event, etc).

### DIFF
--- a/lib/AArch64Base.ml
+++ b/lib/AArch64Base.ml
@@ -418,19 +418,37 @@ let barrier_compare = compare
 module IC = struct
   type funct = I
   let pp_funct = function I -> "I"
+  let fold_funct f k = f I k
 
   type typ = ALL | VA
   let pp_typ = function | ALL -> "ALL" | VA -> "VA"
+  let fold_typ f k = k |> f ALL |> f VA
 
   type point = U
   let pp_point = function U -> "U"
+  let fold_point f k = f U k
 
   type domain = IS | NO
   let pp_domain = function IS -> "IS" | NO -> ""
+  let fold_domain f k = k |> f IS |> f NO
 
   type op = { funct:funct; typ:typ; point:point; domain:domain; }
   let ivau = { funct=I; typ=VA; point=U; domain=NO; }
   let iallu = { funct=I; typ=ALL; point=U; domain=NO; }
+
+  let fold_op f k =
+    fold_funct
+      (fun funct k ->
+        fold_typ
+          (fun typ k ->
+            fold_point
+              (fun point k ->
+                fold_domain
+                  (fun domain k -> f {funct; typ; point; domain; } k)
+                  k)
+              k)
+          k)
+      k
 
   let pp_op op =
     pp_funct op.funct ^
@@ -438,7 +456,13 @@ module IC = struct
     pp_point op.point ^
     pp_domain op.domain
 
+  let do_pp tag op = "IC" ^ tag ^ (pp_op op)
+  let pp = do_pp " "
+  let pp_dot = do_pp "."
+
   let all op = match op.typ with | VA -> false | ALL -> true
+
+  let equal = ( = )
 end
 
 module DC = struct
@@ -448,12 +472,15 @@ module DC = struct
     | C -> "C"
     | CI -> "CI"
     | Z -> "Z"
+  let fold_funct f k = k |> f I |> f C |> f CI |> f Z
 
   type typ = VA | SW
   let pp_typ = function VA -> "VA" | SW -> "SW"
+  let fold_typ f k = k |> f VA |> f SW
 
   type point = CO | U
   let pp_point = function CO -> "C" | U -> "U"
+  let fold_point f k = k |> f CO |> f U
 
   type op = { funct:funct; typ:typ; point:point; }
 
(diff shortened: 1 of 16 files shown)


relokin:
@artkhyzha this is the patchset we discussed, if you can have a look, it would be very much appreciated (ping @jalglave)!

artkhyzha:
Thank you very much, Nikos -- I have just had a quick look, and the changes look nice.
I will now try interacting with this PR via cat, and then report back and see if other comments pop up.

artkhyzha:
Hmm, I am getting build errors...
```
# make && make install
sh ./dune-build.sh $HOME/.local
+ ./version-gen.sh /Users/artkhy01/.local
+ dune build --profile release
File "herd/arch_herd.mli", line 61, characters 15-20:
61 |     module CMO:Cmo.S
                    ^^^^^
Error: Unbound module Cmo
make: *** [build] Error 1    
```
It seems something like `cmo.ml` would define the signature for CMO and its default "not-implemented-yet" implementation. Does the build work out for you without such a file, @relokin ?

relokin:
Sorry Artem my bad, I forgot to add new two files. I hope it works now.

artkhyzha:
Thank you, @relokin -- I have managed to massage the cat file into running a bunch of tests using this PR. So it works for me successfully in that sense.

lib/AArch64Base.ml
```diff
@@ -418,27 +418,57 @@ let barrier_compare = compare
 module IC = struct
   type funct = I
   let pp_funct = function I -> "I"
+  let fold_funct f k = f I k
 
   type typ = ALL | VA
   let pp_typ = function | ALL -> "ALL" | VA -> "VA"
+  let fold_typ f k =
```
artkhyzha, line 425:
```suggestion
  let fold_typ f k =
    k
    |> f ALL
    |> f VA
```
As a completely optional point: this computation could be a nice opportunity to use piping (|>). This might work for many of the new `fold_*` functions. That said, the readability/aesthetics gains with that are rather subjective; please do ignore it if it looks like encryption to you!

relokin: That's a very good point @artkhyzha, I'll change all fold_* functions to use |>. Thanks!

lib/AArch64Base.ml
```diff
@@ -418,27 +418,51 @@ let barrier_compare = compare
...
     pp_point op.point ^
     pp_domain op.domain
 
+  let do_pp tag op = "IC" ^ tag ^ (pp_op op)
+  let pp = do_pp " "
+  let pp_dot = do_pp "."
+
   let all op = match op.typ with | VA -> false | ALL -> true
+
+  let compare ic1 ic2 = compare ic1 ic2 = 0
```
HadrienRenaud, line 465: Usually `compare` returns an `int`, and not a `bool`. Should this be named `equal`?

lib/AArch64Base.ml
```diff
@@ -465,10 +492,27 @@ module DC = struct
...
+      (fun funct k ->
+        fold_typ
+          (fun typ k ->
+            fold_point
+              (fun point k -> f {funct; typ; point; } k)
+              k)
+          k)
+      k
+
+    let compare dc1 dc2 = compare dc1 dc2 = 0
```
HadrienRenaud, line 515: As above, should this be named `equal`?

lib/AArch64Base.ml
HadrienRenaud: And is there a reason why you are not using `=`? As in:
```suggestion
  let compare ic1 ic2 = ic1 = ic2
```

HadrienRenaud (approved):
Hi!

Nice refacto! I have two small comments that will be fixed easily.

I find the remaining big folds a bit ugly, but I don't have a significantly better proposal so this can go.
Other than that, I find that iterating over all the events for every CMO in `machAction.ml` is unnecessary. However, fixing this requires major refactoring so probably not for today

lib/AArch64Base.ml
relokin: I thought of naming it `equal` but hesitated, other modules seem to use `compare`, in the same way. But since you spotted this as well, I'm going to change it to `equal`. But now I am thinking do we need to define `equal`, should we just use `=`?

relokin: @HadrienRenaud let me know if using `=` seems ok. I've pushed the update and if you're happy I am going to squash all fixups.

HadrienRenaud: From my understanding:
It is a bit dangerous to use `=` on recursive types, as you are using something that is not in constant time, which is why we implement equal on complex types. Here everything is quite simple so I think we could do that.
However, if in the future we change this type to something more complicated (such as having Strings or such), we might not change those quite implicit `=`
So this solution is good for now, but the solution I think is the best as a best would be to have under the definition something more explicit like
```
let equal: op -> op -> bool  = ( = )
```

Anyway, if you think I'm just nitpicking, do squash.

relokin: No you're not nitpicking :)

I've changed to `equal` and squashed all patches together. Thanks again for the review!

relokin:
Many thanks to @artkhyzha, @HadrienRenaud and @jalglave for the reviews. @maranget I think this is ready, let me know if you would like to see any changes.

maranget:
Thanks to all.  The PR is now merged. I took the liberty to rebase on top of master before merging. This technique yield a more "linear history'. Thanks again to all.

relokin:
Thanks @maranget!

PR #603, opened 2023-07-05
[herd] AArch64, ASL, implement LDAR and STLR

maranget:
Implementation is partial. In particular `LDAPR` is not implemented.

### DIFF
--- a/herd/ASLAction.ml
+++ b/herd/ASLAction.ml
@@ -10,21 +10,34 @@ module Make (A : S) = struct
   module A = A
   module V = A.V
 
+  type annot = AcqSc|AcqPc|RelSc|Std
+
+  let pp_annot = function
+    | AcqSc -> "A"
+    | AcqPc -> "Q"
+    | RelSc -> "L"
+    | Std -> "N"
+
   type action =
-    | Access of dirn * A.location * A.V.v * MachSize.sz
+    | Access of
+        dirn * A.location * A.V.v * MachSize.sz * annot
     | TooFar of string
     | NoAction
 
-  let mk_init_write loc sz v = Access (W, loc, v, sz)
+  let mk_init_write loc sz v = Access (W, loc, v, sz,Std)
 
   let pp_action = function
-    | Access (d, l, v, _sz) ->
-        Printf.sprintf "%s%s=%s" (pp_dirn d) (A.pp_location l) (V.pp false v)
+    | Access (d, l, v, _sz,an) ->
+        Printf.sprintf "%s%s=%s%s"
+          (pp_dirn d) (A.pp_location l) (V.pp false v)
+          (match an with
+           | Std -> ""
+           | _ -> "(" ^ pp_annot an ^ ")")
     | TooFar msg -> Printf.sprintf "TooFar:%s" msg
     | NoAction -> ""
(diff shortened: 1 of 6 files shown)


herd/libdir/asl-pseudocode/implementations.asl
jalglave, line 1: this should be "ConstrainedUnpredictable" (typo)

jalglave: Also it's unclear to me why return FALSE all the time?

maranget: The name of the ASL function is indeed `ConstrainUnpredictableBool` in ASL shared code. As to the return value, I assumed that a value of `FALSE` would yield an error for non-aligned accesses, which I found adequate.

jalglave: Ok, thanks @maranget!

herd/ASLAction.ml
```diff
@@ -10,21 +10,34 @@ module Make (A : S) = struct
   module A = A
   module V = A.V
 
+  type annot = AcqSc|AcqPc|RelSc|Std
```
HadrienRenaud, line 13: Can you add a comment that explain those symbols, and say what they refer to? Because at the moment I don't understand what is written here.

herd/ASLSem.ml
```diff
@@ -435,13 +449,26 @@ module Make (C : Config) = struct
...
+          Warn.fatal "Arity error for function %s." name
+
     let return_one ty = (Some ty, fun body args -> return [ body args ])
     let return_zero = (None, Fun.id)
 
     (* Primitives *)
     let extra_funcs ii_env =
       let open AST in
       let with_pos = Asllib.ASTUtils.add_dummy_pos in
+      let bool = T_Bool |> with_pos in
```
HadrienRenaud, line 471: Can I suggest the name `boolean`? That's how it's written in ASL

However `d` just under could be renamed to `integer` also, and `bv_64` to `bits_64` to be coherent so maybe for another time ...

herd/libdir/asl-pseudocode/implementations.asl
```diff
@@ -172,14 +176,19 @@ begin
     return FALSE;
 end
 
+func write_memory(addr::bits(64),size::integer,v:bits(size))
```
HadrienRenaud, line 179: What is the purpose of this function if you still have it as a primitive?

herd/libdir/asl-pseudocode/implementations.asl
```diff
@@ -188,13 +197,19 @@ begin
   };
 end
 
+func read_memory(addr::bits(64),size::integer) => bits(size)
```
HadrienRenaud, line 200: Idem for this one.

HadrienRenaud (approved):
Hi @maranget,

Thank you for that work.

I'm wondering if we should add a test to ensure that the semantics of Acquire/Release are well propagated to the aarch64 level? For example an `MP+rel+rel` in `herd/tests/instructions/AArch64.ASL`

maranget:
Closed, superseded by PR #645.

PR #604, opened 2023-07-06
[herd,asl] ASL integers as arbitrary precision integers

maranget:
Introduce an new dependency upon the Zarith library.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -23,11 +23,12 @@ Requirements
 - OCaml (version >= 4.08.0)
 - dune
 - menhir (version >= 20180530)
+- zarith
 
 We strongly recommend to have this base software installed through the opam
 package manager.
 
-    % opam install dune menhir
+    % opam install dune menhir zarith
 
 Make sure to run `eval $(opam config env)` to make tools available in your PATH.
 
--- a/asllib/bitvector.ml
+++ b/asllib/bitvector.ml
@@ -214,6 +214,26 @@ let to_int64_unsigned (length, data) =
   let _, data = remask (63, data) in
   to_int64_raw (64, data)
 
+let to_z_unsigned (sz,data) =
+  if sz=0 then Z.zero
+  else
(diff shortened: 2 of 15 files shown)


asllib/Interpreter.ml
```diff
@@ -251,86 +251,90 @@ module Make (B : Backend.S) (C : Config) = struct
...
-        let* v = eval_expr env e in
-        eval_pattern env e v p
+      | E_Concat es ->
+         prod_map (eval_expr env) es >>= B.concat_bitvectors |: Rule.EConcat
+      | E_Tuple _ -> fatal_from e @@ Error.NotYetImplemented "tuple construction"
+      | E_Unknown ty -> base_value_of_type env ty
+      | E_Pattern (e, p) ->
+         let* v = eval_expr env e in
+         eval_pattern env e v p in
+    if false then
```
HadrienRenaud, line 334: This adds a monadic bind at every expression evaluated, could we instead do something like:
```ocaml
  let _debug_eval_expr m =
    let* v = m in
    let () =
      Format.eprintf "@[<3>Eval@ @[%a@] ->@ %s@]@."
        PP.pp_expr e (B.debug_value v)
    in
    return v

  let rec eval_expr (env : env) (e : expr) : B.value B.m =
     (if false then _debug_eval_expr else Fun.id)
     @@
     match e.desc with
     ...
```

HadrienRenaud (approved):
Hi @maranget 

Looks good to me, though we might want to say in `herdtools.opam` that we are dependent on zarith.

asllib/Interpreter.ml
maranget: Yes it would be great. Given the important changes of the interpreter in the concurrent PR #598, I'd rather revert this change for now.

maranget:
> Hi @maranget
> 
> Looks good to me, though we might want to say in `herdtools.opam` that we are dependent on zarith.

Done, thanks.

PR #618, opened 2023-08-16
[herd] Introduce the `aarch32.cat` model

maranget:
This model is the default AArch64 mode `aarch64.cat`, slightly adapted to for ARM code.


maranget:
Closed, superseded by PR #596

PR #631, opened 2023-08-29
[asl,herd] Fix  the ADD/SUB immediate instructions implementation

maranget:
This PR introduces a call to the  `A.is_valid` function just after test parsing, thereby providing the ability to check instructions beyond syntactic correctness early. For instance `ADD W0,WZR,#1` is illegal and is checked that way.

 + Instruction with negative constants are aliases. For instance,   `ADD W0,W0,#-1` is executed as `SUB W0,W0,#1`.
  + Using register `ZR` in some positions is not possible, as the encoded instruction reserve this place for the stack pointer. For instance `ADD W0,WZR,#1` is in fact illegal, was executed as `ADD W0,SP,#1` and was working by miracle.

### DIFF
--- a/lib/AArch64Base.ml
+++ b/lib/AArch64Base.ml
@@ -1981,7 +1981,7 @@ let get_next =
 (* Check instruction validity, beyond parsing *)
 let mask12 = (1 lsl 12)-1
 
-let is_12bits k = k land mask12 = k
+let is_12bits_unsigned k = k land mask12 = k
 
 let is_valid i =
   match i with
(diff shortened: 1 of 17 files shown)


HadrienRenaud (approved):
Looks good to me!

I just have a question on a signed immediate, that might need a comment.

lib/AArch64Base.ml
```diff
@@ -1976,6 +1978,24 @@ let get_next =
...
+
+let is_12bits k = k land mask12 = k
+
+let is_valid i =
+  match i with
+  | I_OP3 (_,(ADD|SUB|ADDS|SUBS),_,ZR,K _,_)
+  | I_OP3 (_,(ADD|SUB),ZR,_,K _,_)
+    -> false
+  | I_OP3 (_,(ADD|SUB|ADDS|SUBS),_,_,K k,(S_NOEXT|S_LSL (0|12)))
+    -> is_12bits (abs k)
```
HadrienRenaud, line 1992: If I understand correctly, the Arm ARM says that it should be unsigned? Is that because later on, `ADD` of a negative immediate will be converted in a `SUB`? Or do we want to remove the `abs` here?

maranget: Your understanding is correct: using either ADD or SUB the immediate constant size is 12 bits positive + sign.

PR #639, opened 2023-09-04
Add support for DC CVAU and IC IVAU as endpoints into diy generator

hugookeeffe:
these new endpoints can be chosen by using D/I instead of R/W.

The reason behind wanting this change is to allow more fine-grain control of DC CVAU and IC IVAU in order to be able to use relaxations that more closely follow the cat file for ifetch.

This change does introduce the issue that when wildcards are used in diy7 (*) for directions, many more relaxations will be considered. I would suggest implementing an argument that removes D/I from wildcards in order to remove this issue; what do others think?


gen/code.ml
```diff
@@ -109,7 +111,7 @@ let seq_sd sd1 sd2 =
 let fold_ie f r = f Ext (f Int r)
 let fold_sd f r = f Diff (f Same r)
 let do_fold_extr withj f r =
-  let r = f (Dir W) (f (Dir R) (f Irr r)) in
+  let r = f (Dir W) (f (Dir R) (f (Dir D) (f (Dir I) (f Irr r)))) in
   if withj then f (Dir J) r
   else r
```
artkhyzha, line 116:
```suggestion
  let all_dirs = [Dir W; Dir R; Dir J; Dir D; Dir I] in
  let dirs = List.filter (function Dir J -> withj | -> true) all_dirs in
  List.fold_right f dirs (Irr r)
```
Just as a minor nitpick-y suggestion: as the number of the direction values grows, it might make sense fold a list here. (Please ignore if you dislike the readability of this or double-check it compiles if you like it.)

relokin:
Thanks @hugookeeffe for this.

A few points before I start looking at the code:

1. Can you suggest a couple of examples of what litmus tests we could generate with this PR. I was trying to generate a test with various examples, but with no success. For example:

```
diyone -arch AArch64 -variant self PosW{DC.CVAU} Pod{DC.CVAU}W Rfe PodRR Ifre
```

2. There are many compilation warnings, most due to partial matches. Could you have a look a fix them?

hugookeeffe:
The latest patches have fixed the compilation warnings and included the ability to generate tests where an ifetch relaxation is followed or preceded by a relaxation including DC CVAU or IC IVAU

With the latest commit, it is expected that the following tests can be generated with these arguments, including the test @relokin mentioned: 

1.
```
diyone7 -arch AArch64 -variant self PosW{DC.CVAU} Pod{DC.CVAU}W Rfe PodRR Ifre
	AArch64 A
	"PosW{DC.CVAU} Pod{DC.CVAU}W Rfe PodRR Ifre"
	Com=Rf Ifr
	Orig=PosW{DC.CVAU} Pod{DC.CVAU}W Rfe PodRR Ifre
	{
	0:X0=NOP; 0:X1=P1:Lself00; 0:X3=x;
	1:X0=x;
	}
	 P0          | P1          ;
	 STR W0,[X1] | LDR W1,[X0] ;
	 DC CVAU,X1  | Lself00:    ;
	 MOV W2,#1   | B L00       ;
	 STR W2,[X3] | MOV W2,#2   ;
	             | B L01       ;
	             | L00:        ;
	             | MOV W2,#1   ;
	             | L01:        ;
	exists (1:X1=1 /\ 1:X2=1)
```

2.
```
diyone7 -arch AArch64 "PosW{DC.CVAU} DSB.ISHs{DC.CVAU}{IC.IVAU} DSB.ISHd{IC.IVAU}W Rfe DpCtrlIsbdR Ifre" -moreedges true -variant Self
	AArch64 A
	"PosW{DC.CVAU} DSB.ISHs{DC.CVAU}{IC.IVAU} DSB.ISHd{IC.IVAU}W Rfe DpCtrlIsbdR Ifre"
	Com=Rf Ifr
	Orig=PosW{DC.CVAU} DSB.ISHs{DC.CVAU}{IC.IVAU} DSB.ISHd{IC.IVAU}W Rfe DpCtrlIsbdR Ifre
	{
	0:X0=NOP; 0:X1=P1:Lself00; 0:X3=x;
	1:X0=x;
	}
	 P0          | P1           ;
	 STR W0,[X1] | LDR W1,[X0]  ;
	 DC CVAU,X1  | CBNZ W1,LC00 ;
	 DSB ISH     | LC00:        ;
	 IC IVAU,X1  | ISB          ;
	 DSB ISH     | Lself00:     ;
	 MOV W2,#1   | B L01        ;
	 STR W2,[X3] | MOV W2,#2    ;
	             | B L02        ;
	             | L01:         ;
	             | MOV W2,#1    ;
	             | L02:         ;
	exists (1:X1=1 /\ 1:X2=1)
```
3.
```
diyone7 -arch AArch64 -moreedges true -variant Self "PodW{DC.CVAU} Pos{DC.CVAU}W Coe PodWR Fre"
AArch64 A
"PodW{DC.CVAU} Pos{DC.CVAU}W Coe PodWR Fre"
Com=Co Fr
Orig=PodW{DC.CVAU} Pos{DC.CVAU}W Coe PodWR Fre
{
0:X1=x; 0:X2=y;
1:X1=y; 1:X2=x;
}
 P0     | P1     ;
 MOV W0,#1  | MOV W0,#2  ;
 STR W0,[X1] | STR W0,[X1] ;
 DC CVAU,X2 | LDR W3,[X2] ;
 MOV W3,#1  |       ;
 STR W3,[X2] |       ;
exists ([y]=2 /\ 1:X3=0)
```
	
This PR proposes a solution to generating tests with more fine-grain control of how DC CVAU and IC IVAU can be included in tests. Without this PR, the tests 1 and 3 included above cannot be generated with the current relaxations.

One limitation of this PR is that it implements new endpoints that are architecture-specific but are defined generally in the code with this PR. Further PRs could implement architecture-specific endpoints.

gen/code.ml
hugookeeffe: that definitely looks a lot better, I will try implement this in a subsequent commit

hugookeeffe: I have made the change; it seems to work for my tests but let me know if you see anything wrong. Is it ok to resolve this comment?

artkhyzha: Thank you, @hugookeeffe -- of course, totally okay to resolve.

hugookeeffe:
The latest commit ensures that if the next edge after an ifetch edge doesn't contain one of the new directions and has the same location, it will not generate a cycle. 
This same logic cannot be applied for edges before the ifetch as we would still have ambiguous Code/Data locations.

I think this temporary fix could be considered enough for this PR as it allows the generation of all motivating tests and still blocks incorrect tests from being generated. If anyone disagrees please let me know.

PR [649](https://github.com/herd/herdtools7/pull/649/files ) attempts to apply the Code locations using annotations, allowing us to remove the ambiguity of applied Code/Data locations, fixing the above-mentioned issue. (This PR is currently in draft stage and the fix is not working correctly yet)

PR #640, opened 2023-09-05
[asl] Small various features

HadrienRenaud:
This PR implements:

 - [x] HadrienRenaud/herdtools7#29
 - [x] HadrienRenaud/herdtools7#31
 - [x] HadrienRenaud/herdtools7#32
 - [x] HadrienRenaud/herdtools7#33
 - [x] HadrienRenaud/herdtools7#30
 - [x] HadrienRenaud/herdtools7#34

### DIFF
--- a/asllib/Interpreter.ml
+++ b/asllib/Interpreter.ml
@@ -189,22 +189,7 @@ module Make (B : Backend.S) (C : Config) = struct
       | D_TypeDecl (name, _, _) ->
           name
     in
-    let use =
-      let use_e e acc = use_e acc e in
-      let use_ty _ty acc = acc (* TODO *) in
-      fun d ->
-        match d.desc with
-        | D_GlobalStorage { initial_value = Some e; ty = Some ty; _ } ->
-            ISet.empty |> use_e e |> use_ty ty
-        | D_GlobalStorage { initial_value = None; ty = Some ty; _ } ->
-            ISet.empty |> use_ty ty
-        | D_GlobalStorage { initial_value = Some e; ty = None; _ } ->
-            ISet.empty |> use_e e
-        | D_GlobalStorage _ -> ISet.empty
-        | D_TypeDecl (_, ty, s) -> use_ty ty (ISet.of_option s)
-        | D_Func _ ->
-            ISet.empty (* TODO: pure functions that can be used in constants? *)
-    in
+    let use d = use_constant_decl ISet.empty d in
     let process_one_decl d =
       match d.desc with
       | D_GlobalStorage { initial_value; name; ty; _ } ->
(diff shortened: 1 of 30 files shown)


HadrienRenaud:
@maranget Can we merge this one before it becomes too big? 

If not, tell me what I should do to get it merged? I can squash if needs be

asllib/Interpreter.ml
```diff
@@ -485,6 +470,68 @@ module Make (B : Backend.S) (C : Config) = struct
...
+            if i = n then prev
+            else
+              let* v' = B.get_index i v in
+              let* here = in_values v' ty' in
+              loop (succ i) (and' prev here)
+          in
+          loop 0 m_true
+      | T_Named _ -> assert false
+    in
+    B.choice (in_values v ty) (return true) (return false)
```
maranget, line 533: This function is unused...

asllib/tests/asl/required/subtypes-with.asl
```diff
@@ -0,0 +1,39 @@
+type rec1 of record {
+  a: integer,
+  b: integer,
```
maranget, line 3: Optional ',' after last field type definition is allowed (not in ASL document)

asllib/tests/asl/required/subtypes-with.asl
```diff
@@ -0,0 +1,39 @@
...
+type rec2bis subtypes rec1 with { 
+  c: integer,
+};
+
+func main() => integer
+begin
+  let a = rec2 {
+    a = 1,
+    b = 2,
+    c = 3
```
maranget, line 19: Optional ',' after the last field definition is _not_ allowed. Perhaps we could allow this optional comma everywhere? Notice that the ASL document does not allow those terminal ','.

maranget: Well thinking twice, following the official grammar is the best option. If we do not we encourage syntax that other tools may not accept. So the current parser is fine.

This does not mean not lobbying for a more consistent syntax,. A minor issue though.

maranget:
Some of the  most involved constructs, that is, concatenation of lhs and  neseted bitfields with types would deserve at least one example in regression test suite.

Appart from this minor suggestion (and teh more important `is_val_of_type`  being unused), LGTM.

asllib/tests/asl/required/subtypes-with.asl
HadrienRenaud: Good catch! I somehow managed not to commit it...

It should be fixed with the last pushed version.

I also integrated your example, see the error message in `asllib/tests/asl.t/run.t`. Do you think I should test more on this feature?

asllib/Interpreter.ml
HadrienRenaud: Good catch! I somehow managed not to commit it...

It should be fixed with the last pushed version.

I also integrated your example, see the error message in `asllib/tests/asl.t/run.t`. Do you think I should test more on this feature?

HadrienRenaud:
@maranget thank you for your review!

Last commit adds a few tests and a few bug fixes (see especially the bug in `storage.ml`)

I will need to rebase on #651 which will create a lot of conflicts ...

maranget:
> I will need to rebase on #651 which will create a lot of conflicts ...

You mean that we wait for this rebase before merging?

HadrienRenaud:
> > I will need to rebase on #651 which will create a lot of conflicts ...
> 
> You mean that we wait for this rebase before merging?

Depending on which you merge first, I will have to rebase the other one before it can get merged. And in the end, I will have to resolve the same conflicts so it really doesn't really matter for me which way we're doing this.

maranget:
I have rebased on top of master (which has PR #651 merged-in), Tests pass.

jalglave:
Thank you. Please let's merge.
Jade

maranget:
Merged, thanks  @HadrienRenaud and @jalglave.

PR #643, opened 2023-09-05
[aarch64,cat] a fix to addr+HU orderings to restore intent

artkhyzha:
The AArch64 cat model is being fixed to restore intent for the hardware requirement arising from a combination of address dependencies and hardware updates.

Illustrative tests and changes to the kinds file included.


artkhyzha:
Hi @maranget, -- @jalglave and I have prepared several fixes to the model in this PR. May I ask you to fully regress it? If nothing comes up, hopefully, we could consider merging.

Thanks in advance!

artkhyzha:
Hi @maranget -- will you have a chance to regress these changes? Hopefully, they could be merged at that point.

herd/libdir/aarch64hwreqs.cat
```diff
@@ -145,6 +145,9 @@ let rec local-hw-reqs =
   | [Exp & M]; po-loc; [TLBUncacheable & FAULT]
   | [Exp & R]; addr; [TLBI | DC.CVAU | IC]
   | [Exp & R]; ctrl; [HU]
+  | [Exp & R]; addr; [HU]
+  | [Exp & R]; addr; [Exp]; po; [HU]
```
relokin, line 149: Based on the kinds, it looks like these could also be:
```suggestion
  | [Exp & R]; addr; [HU & DB]
  | [Exp & R]; addr; [Exp]; po; [HU & DB]
```
I guess the *af tests are allowed anyway because of the spurious updates. Is there any test we can come up with that can tell the difference between the formulations of these two clauses?

artkhyzha: Hi Nikos, I agree about the kinds -- and indeed it seems unlikely that a test taking advantage of the ordering to `HU & AF` currently exists. Presumably, whenever there exists a cycle involving an effect from `HU & AF`, it can only be followed up by two kinds of edges: those arising from DSB and reads-from/coherence. In the former case, the dependency does not matter, and in the latter there will always exist a spurious update creating an equivalent observation.

Based on this informal and possibly incomplete logic, I'd conjecture that there is no test that currently can tell the difference between, for instance, `[Exp & R]; addr; [HU]` and `[Exp & R]; addr; [HU & DB]`. Happy to go with the more precise phrasing too.

relokin: I see thank you this very helpful. My question was whether there are any tests we can come up to show why this is necessary for [AF]. But your argument seems convincing.

Please leave the cat file as-is, I think they are necessary to forbid executions that we don't want to allow (although there are other executions with spurious updates that lead to the same final state). Otherwise the AF tests are also allowed with `-maxphantom 0`

artkhyzha:
To advance the merging of this PR, I compared the output of `herd7` the VMSA tests before and after this PR. Seemed to yield no difference.

relokin (approved):
I think this is ready. We will merge this by the end of this week. If you have any objections or need more to review this please let us know (@jalglave, @maranget).

jalglave:
Yes I’m happy with this.
Thank you both.
Jade

On 6 Nov 2023, at 22:03, Nikos Nikoleris ***@***.***> wrote:

﻿

⚠ Caution: External sender


@relokin approved this pull request.

I think this is ready. We will merge this by the end of this week. If you have any objections or need more to review this please let us know ***@***.***<https://github.com/jalglave>, @maranget<https://github.com/maranget>).

relokin:
Rebased on top of master and merged, thanks @artkhyzha!

PR #645, opened 2023-09-06
[asl] release acquire

maranget:
Implement release and acquire properties on loads (`LDAR` and `LDAPR`), stores (`STLR`), swaps (`SWPA`, `SWPL`, `SWPAL`) and compare and swaps (`CASA`, `CASL`, `CASAL`).

### DIFF
--- a/herd/AArch64Sem.ml
+++ b/herd/AArch64Sem.ml
@@ -23,6 +23,7 @@ module Make
     module ConfLoc = SemExtra.ConfigToArchConfig(C)
     module AArch64 = AArch64Arch_herd.Make(ConfLoc)(V)
     module Act = MachAction.Make(ConfLoc)(AArch64)
+    module Annot = AArch64Annot
     include SemExtra.Make(C)(AArch64)(Act)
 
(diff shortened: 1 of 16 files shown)


herd/AArch64Sem.ml
```diff
@@ -563,7 +566,7 @@ module Make
           (fun _ -> mk_fault (Some a) dir an ii ft (Some "EL0"))  ii >>! B.Exit
 
       let an_xpte =
```
HadrienRenaud, line 568: Maybe `an_xpte` and `an_pte` could go into `AArch64Annot.ml`?

HadrienRenaud (approved):
LGTM

herd/AArch64Sem.ml
maranget: Would those function also be called AArch64ASLSem.ml?

maranget: Thinking twice, I'd rather keep those functions here, close to their calling sites.

HadrienRenaud: All right

PR #647, opened 2023-09-07
[herd] Fix intra-instruction dependencies for CAS

HadrienRenaud:
This PR edits the intra-instruction dependencies from the `CAS` semantics, on the case where the comparaison is successful (ok case). This changes:

1. the intra-instruction data dependency to the Write Register Event, that was previously starting from the Read Register Event of the value to be compared, is now starting from in two new executions either the Read Register Event of the value to be compared, either from the Read Memory Event.
2. the intra-instruction control dependency from the Read Register Event of the value to be compared to the Write Register Event is now deleted.
3. the intra-instruction control dependency from the Read Memory Event to the Write Register Event is now only present on the execution where the intra-instruction data dependency in 1. is coming from the Read Register Event.

To justify those changes, we add litmus tests to test those dependencies:
1. `MP+rel+CAS-ok-RsRs-addr` tests the existence of an execution where there is no intra-instruction data dependency from the Read Register Event of the value to be compared to the Write Register Event.
2. `MP+rel+CAS-ok-MRs-addr` tests the existence of an execution where there is no intra-instruction data dependency from the Read Memory Event to the Write Register Event.
3. `MP+rel+CAS-ok-bothRs-addr` tests the absence of any execution where there is neither the intra-instruction data dependency from the Read Register Event of the value to be compared to the Write Register Event nor the intra-instruction data dependency from the Read Memory Event to the Write Register Event.
4. `LB+rel+CAS-ok-RsRs-addr` tests the existence of an execution where there is no intra-instruction control dependency from the the Register Read Event of the value to be compared to the Write Register Event.
5. `LB+rel+CAS-ok-MRs-addr` tests the presence of an intra-instruction control dependency between the Memory Read Event to the Write Register Event.

A lof of tests have impacted logs, because of the number of execution being changed by this. No other test in the catalogue change of kind.

Here are the new graphs for the CAS (two executions means two graphs for the same instruction):
aarch64_cas-ok dot 2
aarch64_cas-ok dot

---

Furthermore, we are removing an intra-instruction order dependency from the `SWP` and other atomic operation instruction (e.g. `LDADD`). This change should be immaterial. The new graph for `SWP` is:

aarch64_swap_diff dot

For completeness, we have added the test `LB+SWP-RsRt-addr+rel` to ensures the absence of any dependency between the two registers events in `SWP`.


jalglave:
Thanks @HadrienRenaud this looks good.
@maranget please could you regress this?
Thanks in advance,
Jade

PR #650, opened 2023-09-08
[asl] Implement load-reserve store conditional

maranget:
To complete PR #645.

### DIFF
--- a/asllib/Interpreter.ml
+++ b/asllib/Interpreter.ml
@@ -39,6 +39,7 @@ let _dbg = false
 module type S = sig
   module B : Backend.S
 
+  val run_env : (AST.identifier * B.value) list -> B.ast -> B.value B.m
   val run : B.ast -> B.value B.m
   val run_typed : B.ast -> StaticEnv.env -> B.value B.m
 end
@@ -193,7 +194,11 @@ module Make (B : Backend.S) (C : Config) = struct
   (* Global env *)
   (* ---------- *)
 
-  let build_global_storage eval_expr base_value =
+  let build_global_storage env0 eval_expr base_value =
+    let names =
+      List.fold_left
+        (fun k (name,_) -> ISet.add name k)
+        ISet.empty env0 in
     let def d =
       match d.desc with
       | D_Func { name; _ }
@@ -205,7 +210,7 @@ module Make (B : Backend.S) (C : Config) = struct
       let use_e e acc = use_e acc e in
       let use_ty _ty acc = acc (* TODO *) in
       fun d ->
-        match d.desc with 
+        match d.desc with
         | D_GlobalStorage { initial_value = Some e; ty = Some ty; _ } ->
             ISet.empty |> use_e e |> use_ty ty
         | D_GlobalStorage { initial_value = None; ty = Some ty; _ } ->
@@ -220,31 +225,29 @@ module Make (B : Backend.S) (C : Config) = struct
     let process_one_decl d =
       match d.desc with
       | D_GlobalStorage { initial_value; name; ty; _ } ->
-          fun env_m ->
-            let*| env = env_m in
-            let* v =
-              match (initial_value, ty) with
-              | Some e, _ -> eval_expr env e
-              | None, None -> fail_initialise d name
-              | None, Some t -> base_value env t
-            in
-            let* () =
-              match name with
-              | "PSTATE" ->
-                 (* No event at all for PSTATE, translated to reg NZCV.
-                    Moeover, initialisation of NZCV is from
-                    previous instructions, like all registers. *)
-                 return ()
-              | _ ->
-                 B.on_write_identifier name Scope_Global v in
-            IEnv.declare_global name v env |> return
+         fun env_m ->
+           if ISet.mem name names then
+             env_m
+           else
+             let*| env = env_m in
+             let* v =
+               match (initial_value, ty) with
+               | Some e, _ -> eval_expr env e
+               | None, None -> fail_initialise d name
+               | None, Some t -> base_value env t in
+             let* () =
+               match name with
+               | "PSTATE"|"RESADDR" -> return ()
+               | _ ->
+                  B.on_write_identifier name Scope_Global v in
+             IEnv.declare_global name v env |> return
       | _ -> Fun.id
     in
     dag_fold def use process_one_decl
 
   (** [build_genv static_env ast primitives] is the global environment before
       the start of the evaluation of [ast]. *)
-  let build_genv eval_expr base_value (static_env : StaticEnv.env)
+  let build_genv env0 eval_expr base_value (static_env : StaticEnv.env)
       (ast : B.primitive AST.t) =
     let funcs = IMap.empty |> build_funcs ast in
     let () =
@@ -259,7 +262,12 @@ module Make (B : Backend.S) (C : Config) = struct
       in
       { global; local = empty_scoped Scope_Global }
     in
-    let*| env = build_global_storage eval_expr base_value ast (return env) in
+    let env =
+      List.fold_left
+        (fun env (name,v) -> IEnv.declare_global name v env)
+        env env0 in
+    let*| env =
+       build_global_storage env0 eval_expr base_value ast (return env) in
     return env.global
 
   (* Bind Environment *)
@@ -994,8 +1002,8 @@ module Make (B : Backend.S) (C : Config) = struct
         List.init length (Fun.const v) |> B.create_vector
 
 
-  let run_typed (ast : B.ast) (static_env : StaticEnv.env) : B.value m =
-    let*| env = build_genv eval_expr_sef base_value static_env ast in
+  let run_typed_env env (ast : B.ast) (static_env : StaticEnv.env) : B.value m =
+    let*| env = build_genv env eval_expr_sef base_value static_env ast in
     let*| res = eval_func env "main" dummy_annotated [] [] in
     match res with
     | Normal ([ v ], _genv) -> read_value_from v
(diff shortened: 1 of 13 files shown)


asllib/Interpreter.mli
```diff
@@ -22,6 +22,7 @@
 module type S = sig
   module B : Backend.S
 
+  val run_env : (AST.identifier * B.value) list -> B.ast -> B.value B.m
```
HadrienRenaud, line 25: Can we put some documentation here? The name is not really explicit

herd/ASLSem.ml
```diff
@@ -385,6 +392,19 @@ module Make (C : Config) = struct
...
+    let checkprop m_prop =
+      let* vprop = m_prop in
+      M.assign (vbool true) vprop >>! []
+
+    let somebool _ =
+      let v = V.fresh_var () in
+      let mbool b =
+        let vb = V.Val (Constant.Concrete (ASLScalar.S_Bool b)) in
+        M.assign v vb >>= fun () -> M.unitT vb  in
+      choice (M.unitT v) (mbool true) (mbool false)
```
HadrienRenaud, line 406: Is it exactly what the ASL syntax `UNKNOWN: boolean` does?

herd/ASLSem.ml
```diff
@@ -385,6 +392,19 @@ module Make (C : Config) = struct
     (* Primitives and helpers                                                 *)
     (**************************************************************************)
 
+    let vbool b =  V.Val (Constant.Concrete (ASLScalar.S_Bool b))
+
+    let checkprop m_prop =
+      let* vprop = m_prop in
+      M.assign (vbool true) vprop >>! []
```
HadrienRenaud, line 399: Is that different from the ASL syntax `assert prop`?

herd/libdir/asl-pseudocode/implementations.asl
```diff
@@ -1,3 +1,111 @@
...
+var SuccessVA : boolean ;
+
+func AArch64_IsExclusiveVA
+(address :: bits(64), processorid :: integer, size :: integer) => boolean
+begin
+  // Try both possibilties: write or not write
+  SuccessVA = SomeBoolean();
+  let reserved = RESADDR;
+  // If write succeeds then effective address and reservation coincide.
+  if SuccessVA then CheckProp(address == reserved); end
```
HadrienRenaud, line 61:
```suggestion
  if SuccessVA then assert address == reserved; end
```

asllib/Interpreter.ml
```diff
@@ -1009,16 +1017,21 @@ module Make (B : Backend.S) (C : Config) = struct
         in
         Error.fatal_unknown_pos (Error.UncaughtException msg)
 
+  let run_typed ast env = run_typed_env [] ast env
+
   (** Main entry point for the Interpreter, [run ast primitives] type-annotate
       [ast], build a global environment and then evaluate the "main" function
       in it. *)
```
HadrienRenaud, line 1024: I think we can delete those at this point

herd/AArch64ASLSem.ml
```diff
@@ -445,62 +473,30 @@ module Make (TopConf : AArch64Sig.Config) (V : Value.AArch64ASL) :
...
 
-    let symb_values_tbl : (string, V.v) Hashtbl.t = Hashtbl.create 17
-    let symb_values_tag = "AArch64-HashedValue"
+    let tr_cst tr =
+      Constant.map tr
+        (fun _ -> Warn.fatal "Cannot translate PTE")
+        (fun _ -> Warn.fatal "Cannot translate instruction")
+
+    let aarch64_to_asl_bv = function
+      | V.Var v -> ASLS.A.V.Var v
```
HadrienRenaud, line 482: Will there be a problem if the symb counter on ASL values is not increased and we have symbolic values with the same number?
I think we should use fresh_var and keep track of why ASL value equals which AArch64 value, or make the counter of symbolic values common for all value types.

herd/AArch64ASLSem.ml
```diff
@@ -445,62 +473,30 @@ module Make (TopConf : AArch64Sig.Config) (V : Value.AArch64ASL) :
...
-          (* The initial value for PSTATE given bellow
-           * overrides the one of the "patches.asl" file,
-           *  see `herd/ASLSem.ml`.
-           *)
-          ::(MiscParser.Location_reg (proc,pstate),tc)
-          ::init_gregs
-        | None -> init_gregs
-        in
-        let prog =
+      let init = [] in
```
HadrienRenaud, line 498: Why do we still have this?
And why did you move the construction of init after?

asllib/Interpreter.mli
maranget: Of course, will do.

herd/libdir/asl-pseudocode/implementations.asl
```diff
@@ -1,3 +1,111 @@
...
+//
+// It is always safe to return TRUE which will check the physical address only.
+
+var SuccessVA : boolean ;
+
+func AArch64_IsExclusiveVA
+(address :: bits(64), processorid :: integer, size :: integer) => boolean
+begin
+  // Try both possibilties: write or not write
+  SuccessVA = SomeBoolean();
```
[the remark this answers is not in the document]
maranget: I was not aware of  UNKOWN. `SomeBoolean()` returns an alternative monad `M.altT`. with one alternative returning true and the other returning false. I have to check that `UNKNOWN` does the same...

herd/ASLSem.ml
maranget: No idea!

herd/AArch64ASLSem.ml
maranget: Yes it can be a problem if two different variables are the same.  This can be the case, I am afraid.

herd/AArch64ASLSem.ml
maranget: Aha, If we construct init before calling `ASLTH.build`  we  have to turn AArch64ASL values into `ParsedConstant.v`'s. When we do it after, we translate AArch64ASL values into ASL values, which are basically the same!

herd/ASLSem.ml
maranget: I have not checked, but I assume this is different: `checkprop` produces an equation that will be contradictory if vprop is `FALSE`; while `ASSERT` raises an exception that may have less controlled consequences.

asllib/Interpreter.ml
maranget: Delete comment?

herd/AArch64ASLSem.ml
maranget: A new commit 5d08b1cc82e379a11a717cacc7f0bb131f2d95db intend to fix this potential problem.

herd/AArch64ASLSem.ml
HadrienRenaud: Oh smart!

herd/ASLSem.ml
HadrienRenaud: After checking, in ASLSem, for the moment, `UNKNOWN: _t` produces a fresh variable. As you have a condition just in front of it, it will create an altT with the two possibilities.

maranget: Do you mean that we can write `SomeBoolean()` using `UNKNOW`? Something like
```
func SomeBoolean() => boolean
begin
  return if UNKNOWN:boolean then TRUE else FALSE end;
end
```
(Have not tried).

HadrienRenaud: This would indeed have the same effect as writing it as a primitive.

But actually, what I was suggesting is not to implement any function `SomeBool` and just use `UNKNOWN: boolean` instead of calling the function.

herd/ASLSem.ml
HadrienRenaud: Maybe we should implement assertions like this then?

asllib/Interpreter.ml
HadrienRenaud: Yes: they are not alongside their function as you put run_env between them, and do not have the correct arguments anymore.

herd/ASLSem.ml
maranget: I am not sure we should. Assertions are here to check for errors. We probably want errors to be flagged as such and not to disappear silently.

herd/ASLSem.ml
maranget: This is doable for `UNKNOWN:boolean` but not for `UNKNOWN:integer`. So we would have a significant discrepancy of semantics. This would be disturbing for users, or?

herd/ASLSem.ml
HadrienRenaud: Okay then.

herd/ASLSem.ml
HadrienRenaud: I don't understand the parallel with integers, we are not implementing `SomeInteger` here?

For clarification I am not suggesting to change the semantics of `UNKNOWN: boolean` to match those of `SomeBoolean`.

maranget: I have tried to replace the call to `SomeBoolean()` by:
```
SuccessVA =
    if UNKNOWN:boolean then TRUE else FALSE;
```
It does not work. Lacking time to experiment, I keep the current solution with a specific `SomeBoolean()` primitive. Please do not hesitate to post a follwoup, even after merging, which is imminent.

maranget:
Hi @HadrienRenaud I have merged the PR. Thanks for your review. I could see how to use the existing `UNKNOWN::boolean` constructing in place of the new primitive `SomeBoolean()`. Please elaborate if you see how to proceed. Thanks again.

herd/ASLSem.ml
HadrienRenaud: Okay, a shame but not too costly. It'd be good to investigate why using `UNKNOWN:boolean` does not work there.

PR #658, opened 2023-09-13
[herd,asl] Fix transmission of data ports from ASL to AArch64

maranget:
The fix is simple: the showing of event set `AArch64_DATA` was missing in file `asl.cat`.

Notice that "showing" sets and relations is a flexible mean to transmit them from ASL to AArch64.

Also notice that `asl.cat` already showed the identity _relation_ over the same event set as `AArch64_DATA`. We leave it, as it apparently does not harm.

### DIFF
--- a/herd/libdir/asl.cat
+++ b/herd/libdir/asl.cat
@@ -83,7 +83,7 @@ show asl_iico_ctrl, asl_iico_data, asl_rf_reg
 show iico_ctrl, iico_data, rf, rf-reg
 *)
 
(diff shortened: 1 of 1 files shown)


HadrienRenaud (approved):
That's a very short fix!
I can imagine that the debugging session was not as short ...

Of course LGTM

maranget:
Thanks @HadrienRenaud, everything was in place and I got the idea quite easily (a neat idea BTW). Merged

PR #661, opened 2023-09-14
[herd,asl] Fix asl cas

maranget:
This PR fixes a bug whose root cause was partial oversight of  requested size by memory reads.

### DIFF
--- a/asllib/Interpreter.ml
+++ b/asllib/Interpreter.ml
@@ -498,7 +498,17 @@ module Make (B : Backend.S) (C : Config) = struct
     let or' prev here = prev >>= B.binop BOR here in
     let rec in_values v ty =
       match (Types.get_structure (IEnv.to_static env) ty).desc with
-      | T_Real | T_Bool | T_Enum _ | T_String | T_Bits _ | T_Int None -> m_true
+      | T_Real | T_Bool | T_Enum _ | T_String | T_Int None -> m_true
+      | T_Bits (BitWidth_SingleExpr e, _) ->
+          let* v' = eval_expr_sef env e and* v_length = B.bitvector_length v in
+          B.binop EQ_OP v_length v'
+      | T_Bits (BitWidth_ConstrainedFormType ty_length, _) ->
+          let* v_length = B.bitvector_length v in
+          in_values v_length ty_length
+      | T_Bits (BitWidth_Constraints cs, _) ->
+          let* v_length = B.bitvector_length v in
+          let ty_length = T_Int (Some cs) |> add_pos_from ty in
+          in_values v_length ty_length
       | T_Int (Some constraints) ->
           let fold prev = function
             | Constraint_Exact e ->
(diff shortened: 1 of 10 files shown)


HadrienRenaud:
Small questions on this issue:

- does this happen also for memory writes?
- does this code generates 64 bits accesses even if it was requested a 32 bit?

I will investigate next week.

maranget:
As for reads, the memory read event will be tagged with its size (as a value of type `MachSize.sz`), the ASLValue associated to the location will be returned and passed as an argument to the function `to_bv`. As a result, if the value is not a bitvector some bitvector _of size 64_ is returned, if the value is a bitvector it is left unchanged. Also notice that, in this patch, `PhysMemRead` slices the bitvector with argument `[width-1:0]` where width is the number of bits accessed.

For writes I have not really checked. It seems that a bitvector of the requested size is witten. So there may be a risk of a slicing attempt with argument [63..0] on a bitvector of smaller size...

maranget:
Bingo, with test `SIZES.litmus`:
```
AArch64 SIZES
{
int64_t x;
0:X1=x;
}
  P0          ;
MOV W0,15     ;
STR W0,[X1]   ;
LDR X2,[X1]   ;
forall 0:X2=15
```
To run the test, the mixed-size check has to be disabled:
```
herd7 SIZES.litmus -variant dontcheckmixed -variant asl
Warning: File "SIZES.litmus": Illegal operation Slice[63, 62, 61, 60, 59, 58, 57, 56, 55, 54, 53, 52, 51, 50, 49, 48, 47, 46, 45, 44, 43, 42, 41, 40, 39, 38, 37, 36, 35, 34, 33, 32, 31, 30, 29, 28, 27, 26, 25, 24, 23, 22, 21, 20, 19, 18, 17, 16, 15, 14, 13, 12, 11, 10, 9, 8, 7, 6, 5, 4, 3, 2, 1, 0] on '00000000000000000000000000001111' (User error)
```

maranget:
Finally, it would probably be cleaner for the primitive `read_memory_gen` to return a bitvector of the requested size (comforming with its type before this PR) by introducing a primitive operation `ToBV of int`. This would not change much though.

HadrienRenaud:
In fact, I have just seen that the following test does NOT fail:
```
func test(size: integer) begin
  let - = Zeros(4) as bits(size);
end

func main () => integer begin
  test(4);
  test(3);
  return 0;
end
```
I somehow missed that runtime checks could happen on bitvectors too..

maranget:
Correct, and some extra operation to get the size of bitvectors may be necessary.

HadrienRenaud:
After 32216d1b720441332c24c850db9940a126d28bff on branch `asl-dev`, runtime checks for bitvectors are implemented in `herd` and in `aslref`, and the test mentioned just above does raise an error.

However, I don't really know how you want the ToBV to be parametrized, so I'll wait before fixing those.

maranget:
Hi @HadrienRenaud, has your fix been merged? If not, why not cherry-pick it here or make it a PR of its own?

HadrienRenaud:
Thanks @maranget 

Is there anything else we want to do before merging?

maranget:
Merged, thanks @HadrienRenaud. Notice that I have integrated all the changes from commit  32216d1, including teh disabling of optimisation performed [here](https://github.com/herd/herdtools7/commit/32216d1b720441332c24c850db9940a126d28bff#diff-349fb88b3b0bb5565cba1f54118d05fb6e7d5ebd37038104e79978f7a5380615)

HadrienRenaud:
Great 🎉 !! Thank you @maranget

PR #662, opened 2023-09-15
[fix] [asl] Rename aslseq into aslref

HadrienRenaud:
`grep aslseq` does not return anything anymore.


jalglave:
thanks @HadrienRenaud.

PR #663, opened 2023-09-15
[asl] Arrays indexed by enumerations and short-circuit evaluation

HadrienRenaud:
This PR depends on #662.


HadrienRenaud:
@jalglave After this PR, we can consider that the interpreter is feature complete for an alpha version

jalglave:
thanks for this work @HadrienRenaud.

PR #664, opened 2023-09-16
[AArch64,ARM]: Indirect Branches working for GCC tests.

lukeg101:
Fix ARM indirect branch to handle rets

      Consider the following litmus test - emitted by GCC:
      ```
      ARM A024

      Variant=telechat
      { [P1_r0]=0;[x]=0;[y]=0;
        %P0_x=x;%P0_y=y;
        %P1_P1_r0=P1_r0;
        %P1_x=x;%P1_y=y }
      (*****************************************************************)
      (* Compiler:                                                     *)
      (* arm-linux-gnueabi-gcc-10 -c -g -O2 -march=armv7-a --std=c11 -fno-section-anchors*)
      (*****************************************************************)
        P0               |  P1                   ;
         MOV R0,#2       |   LDR R2,[%P1_y]      ;
         MOV R1,#1       |   CMP R2,#1           ;
         DMB ISH         |   BEQ L0x5c           ;
         STR R0,[%P0_x]  |  L0x48:               ;
         STR R1,[%P0_y]  |   STR R2,[%P1_P1_r0]  ;
         BX LR           |   BX LR               ;
                         |  L0x5c:               ;
                         |   DMB ISH             ;
                         |   STR R2,[%P1_x]      ;
                         |   B L0x48             ;

      exists ([x]=2 /\ P1_r0=1)
      ```
      The RET is used in two places to jump around, and eventually exit.
      Currently this test does not work - we fix indirect branches to enable.
      this

This patch fixes this for RETs in AArch64 and BX in ARM. Guarded by the telechat variant of herd, introduced in a prior work

### DIFF
--- a/herd/AArch64Sem.ml
+++ b/herd/AArch64Sem.ml
@@ -2026,10 +2026,12 @@ module Make
            match v with
            | M.A.V.Var(_) as v ->
               let lbls = get_exported_labels test in
-              if Label.Full.Set.is_empty lbls then
-                Warn.fatal
-                  "Could find no potential target for indirect branch %s \
-                   (potential targets are statically known labels)" (AArch64.dump_instruction i)
+              if Label.Full.Set.is_empty lbls  then begin
+                if C.variant Variant.Telechat then M.unitT () >>! B.Exit
+                else
+                  Warn.fatal "Could find no potential target for indirect branch %s \
+                    (potential targets are statically known labels)" (AArch64.dump_instruction i)
+                end
               else
                 commit_bcc ii
                 >>= fun () -> B.indirectBranchT v lbls bds
--- a/herd/ARMSem.ml
+++ b/herd/ARMSem.ml
@@ -56,22 +56,6 @@ module
       | M.A.V.Val (Concrete i) -> Some (B.Addr (M.A.V.Cst.Scalar.to_int i))
       | _ -> None
 
-    let do_indirect_jump test bds i v =
-      match  v2tgt v with
-      | Some tgt -> M.unitT (B.Jump (tgt,bds))
-      | None ->
-         match v with
-         | M.A.V.Var(_) as v ->
-            let lbls = get_exported_labels test in
-            if Label.Full.Set.is_empty lbls then
-              Warn.fatal
-                "Could find no potential target for indirect branch %s \
-                 (potential targets are statically known labels)" (ARM.dump_instruction i)
-            else
-              B.indirectBranchT v lbls bds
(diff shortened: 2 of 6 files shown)


herd/AArch64Sem.ml
```diff
@@ -2026,10 +2026,11 @@ module Make
...
               let lbls = get_exported_labels test in
-              if Label.Full.Set.is_empty lbls then
-                Warn.fatal
-                  "Could find no potential target for indirect branch %s \
-                   (potential targets are statically known labels)" (AArch64.dump_instruction i)
+              if Label.Full.Set.is_empty lbls && C.variant Variant.Telechat then
+                M.unitT () >>! B.Exit
+              else if Label.Full.Set.is_empty lbls then
+                  Warn.fatal "Could find no potential target for indirect branch %s \
+                    (potential targets are statically known labels)" (AArch64.dump_instruction i)
```
maranget, line 2033: Wouldn't  the following coding be more direct?
```suggestion
              if Label.Full.Set.is_empty lbls  then begin
                if C.variant Variant.Telechat then M.unitT () >>! B.Exit
              else
                  Warn.fatal "Could find no potential target for indirect branch %s \
                    (potential targets are statically known labels)" (AArch64.dump_instruction i)
              end
```

herd/ARMSem.ml
```diff
@@ -85,6 +69,24 @@ module
...
+      | Some tgt -> M.unitT (B.Jump (tgt,bds))
+      | None ->
+         match v with
+         | M.A.V.Var(_) as v ->
+            let lbls = get_exported_labels test in
+              if Label.Full.Set.is_empty lbls && C.variant Variant.Telechat then
+                M.unitT () >>! B.Exit
+              else if Label.Full.Set.is_empty lbls then
+                  Warn.fatal "Could find no potential target for indirect branch %s \
+                    (potential targets are statically known labels)" (ARM.dump_instruction i)
```
maranget, line 83: Again one can avoid testing `Label.Full.Set.is_empty lbls` twice

maranget:
Hi @lukeg101, I do not understand what you mean by " The RET is used in two places to jump around, and eventually exit.". Specifically I do not understand "jump around".

lukeg101:
Apologies Luc, by that `RET`s can be used anywhere inside a compiled program in practice, rather than just at the end. Implementing `RET` as `Exit` is sufficient for the Telechat work :)

maranget:
Squashed and merged, thanks @lukeg101

PR #670, opened 2023-09-22
[herd] Add variant lse2 which implements the Armv8-A FEAT_LSE2

relokin:
The new lse2 variant changes the default semantics for pair instructions and ensure that their explicit belong to the same atomicity class.

Note that for now, herd doesn't perform any checks on the alignment of the data locations accessed by the pair instruction.

In addition, it changes the definition of the sca-class relation, to exclude the id relation. This change doesn't make any changes to the model as all uses of sca-class are optional and thefore the id subset of the old sca-class relation is unnecessary.


jalglave:
Thanks @relokin this looks good.
Jade

herd/variant.mli
```diff
@@ -53,6 +53,8 @@ type t =
   | ETS2 (*New feature introduced after deprecating ETS*)
 (* AArch64: Enhanced Exception Synchronization - FEAT_ExS *)
   | ExS | EIS | EOS
+(* AArch64: Large Scale Extensions 2 - FEAT_LSE2 *)
```
HadrienRenaud, line 56:
```suggestion
  (* AArch64: Large System Extensions version 2 - FEAT_LSE2 *)
```

herd/variant.ml
```diff
@@ -55,6 +55,8 @@ type t =
   | ETS2 (*New feature introduced after deprecating ETS*)
 (* AArch64: Enhanced Exception Synchronization - FEAT_ExS *)
   | ExS | EIS | EOS
+  (* AArch64: Large Scale Extensions 2 - FEAT_LSE2 *)
```
HadrienRenaud, line 58: I think the name is Large System Extension ... (see [here](https://developer.arm.com/downloads/-/exploration-tools/feature-names-for-a-profile))
```suggestion
  (* AArch64: Large System Extensions version 2 - FEAT_LSE2 *)
```

HadrienRenaud (commented):
Looks good to me!

You probably have looked into the case of misaligned accesses where the ASL code says the atomicity is "implementation defined"?

herd/variant.mli
relokin: Thanks @HadrienRenaud! Indeed my bad.

herd/variant.ml
relokin: I've actually removed this. It doesn't make sense to have comments here when we have them in the mli file, does it?

relokin:
Thank you both for the review!

relokin:
> Looks good to me!
> 
> You probably have looked into the case of misaligned accesses where the ASL code says the atomicity is "implementation defined"?

I haven't and with a quick grep I couldn't find it, could you give me a pointer?

HadrienRenaud:
> > Looks good to me!
> > You probably have looked into the case of misaligned accesses where the ASL code says the atomicity is "implementation defined"?
> 
> I haven't and with a quick grep I couldn't find it, could you give me a pointer?

In [`MemSingle`](https://developer.arm.com/documentation/ddi0602/2023-06/Shared-Pseudocode/aarch64-functions-memory?lang=en#AArch64.MemSingle.read.4), there is the following case (I don't understand when it is triggered):
```
        // Misaligned accesses within 16 byte aligned memory but
        // not Normal Cacheable Writeback are Atomic
        atomic = boolean IMPLEMENTATION_DEFINED "FEAT_LSE2: access is atomic";
```

relokin:
> > > Looks good to me!
> > > You probably have looked into the case of misaligned accesses where the ASL code says the atomicity is "implementation defined"?
> > 
> > 
> > I haven't and with a quick grep I couldn't find it, could you give me a pointer?
> 
> In [`MemSingle`](https://developer.arm.com/documentation/ddi0602/2023-06/Shared-Pseudocode/aarch64-functions-memory?lang=en#AArch64.MemSingle.read.4), there is the following case (I don't understand when it is triggered):
> 
> ```
>         // Misaligned accesses within 16 byte aligned memory but
>         // not Normal Cacheable Writeback are Atomic
>         atomic = boolean IMPLEMENTATION_DEFINED "FEAT_LSE2: access is atomic";
> ```

Thank you! Ah yes, I've seen that, I thought you were referring to the asllib code in herdtools.

Not sure, but from what I can gather, it says that the architecture doesn't mandate that the access is single-atomic but an implementation can implement it as atomic.

jalglave:
hi @relokin is there any reason to hold this off? How can I help to get this in?
Thanks in advance,
Jade

relokin:
> hi @relokin is there any reason to hold this off? How can I help to get this in? Thanks in advance, Jade

Happy to merge this, only one note, if one used -variant lse2, LDP/STP would be single-copy atomic regardless of the alignment of their operand. I think that's ok for now, and we can still merge this.

jalglave:
Thanks @relokin would you be able to add a little warning to the user then?
Thanks in advance,
Jade

PR #673, opened 2023-09-25
[litmus] Limit pointer to code (presi/kvm mode)

maranget:
At the moment pointer to code is possible only for architecture with fixed-size instructions. For those, the size of instructions is defined in the file `instruction.h`.

Hence consider that no pointer to code can be saved or handled in any way when the file `instruction.h` does not exist.


maranget:
Bug label added, as X86_64 tests with some label yielded invalid C code in mode `-variant presi`. The PR fixes this,

PR #682, opened 2023-09-28
Neon update

murzinv:
Hi everyone,

Recently I've been playing with generator coupled with `-variant neon` and observed few issues with that. I did an attempt to fix them. Most of the code is related to the way we perform Neon load - it seems that it is beneficial to reduce data in Neon vector registers horizontally down to single Neon vector register and reduce that Neon vector register horizontally  down to Neon scalar register which we can transfer to general purpose register. All that require new instructions ADDV and FMOV (to general) and corresponding adjustment for tracking value in generator. Neon store also have to be updated, it needs to expand data in general purpose register into Neon vector register - that's done with new DUP instruction. 

Additionally I'd like to note, that because there seems have not had complaints about `-variant neon` it probably has very lightweight user base, so I'm suggesting to rework Neon annotations to make some room for future additions.

Cheers
Vladimir


murzinv:
@hugookeeffe

gen/cycle.ml
```diff
@@ -839,7 +839,7 @@ let set_same_loc st n0 =
                       CoSet.set_cell has a case for pairs.
                       However increment of current value is by 2 *)
                    let cell = CoSt.get_cell st in
-                   assert (Array.length cell=2) ;
+                   assert (Array.length cell>=2) ;
```
hugookeeffe, line 842: why >= 2? should pairs not always have cells of length 2?

hugookeeffe (commented):
Hi Vladimir,

Thank you for the PR! Overall it looks good to me. I just had a couple of clarifying questions about things I was unsure about. 

If any of my questions don't make sense let me know and I can elaborate.

gen/AArch64Arch_gen.ml
```diff
@@ -60,17 +60,21 @@ module Mixed =
 let bellatom = false
 module SIMD = struct
 
-  type atom = N1|N2|N3|N4
+  type atom = Ne1|Ne2I|Ne3I|Ne4I
```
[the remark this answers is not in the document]
murzinv: `I` stands for `interleave`. Basically, currently `-variant neon` exposes 4 variants of instructions which maps onto annotaions
- LD1 -> N1
- LD2 -> N2
- LD3 -> N3
- LD4 -> N4

`LD2`, `LD3` and `LD4` are interleave versions and have hard requirement on number of registers. At the same time `LD1` is non-interleave (contiguous) and can have variable number of registers. However, current implementation tend to correlate number of registers used by instruction with interleave pattern and unnecessary put limitation on `LD1` having just single register. I have plans to allow `LD1` more than one register and map them on annotation w/o `I` like
- LD1 {reg} -> Ne1
- LD1 {reg,reg} -> Ne2
- LD1 {reg,reg,reg} -> Ne3
- LD1 {reg,reg,reg,reg} -> Ne4

All that could go in separate PR of cause. The reason I renamed annotation in this PR is to reduce user visible time window between _**everything works with old annotations**_ and _**things are broken due to rename in annotations, we already get used to old one, why have you changed them?**_ . Currently we know that things do not work well anyway, so making them work and rename annotations at the same time looks like right thing to do. OTOH, if there is desire to postpone rename for the next PR, I would not object.

gen/AArch64Arch_gen.ml
```diff
@@ -100,6 +104,14 @@ module SIMD = struct
       else reg r 0::regs (r+1) in
     regs 0
 
+  let reduce vec =
+    match vec with
+    | [] -> 0
+    | _ -> let sum = List.map2 (+) in
+           let init = [0;0;0;0] in
+           let v = (List.fold_right sum vec init) in
+           List.fold_right (+) v 0
```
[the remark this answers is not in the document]
murzinv: Ahh, did not know about that. I hope I got it right

```  let reduce vec =
    match vec with
    | [] -> 0
    | _ -> Misc.fold_cross vec (fun v k -> List.fold_right (+) v k) 0
```

gen/AArch64Compile_gen.ml
```diff
@@ -69,14 +80,14 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
...
-        | N2 -> call_rec N1 st
-        | N3 -> call_rec N2 st
-        | N4 -> call_rec N3 st in
+        | Ne2I -> call_rec Ne1 st
+        | Ne3I -> call_rec Ne2I st
+        | Ne4I -> call_rec Ne3I st in
       fun n st ->
         let (r,rs),st = get_reg_list n st in
-        (r,rs),A.set_friends r rs st
+        (r,rs),st
```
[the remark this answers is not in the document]
murzinv: I vaguely remember running in some assertion while I've been exploring different options for how specific neon test could look like. However, I gave it a quick try and I do not see any issues, so I'll keep it in next round.

gen/cycle.ml
murzinv: Correct, yet with Neon cells usually larger than that. IIUC, for pairs cell length must be at least 2, so when paired with Neon which has cell length of 4 and even more we still satisfy requirement for pairs, but current assert condition limit us with upper boundary.

gen/AArch64Arch_gen.ml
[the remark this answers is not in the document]
hugookeeffe: that makes sense. I have no issue with changing the names already, just wanted to make sure I understood why the changes are being made

gen/AArch64Arch_gen.ml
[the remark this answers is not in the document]
hugookeeffe: a fold cross should do both of the folds, so you shouldn't need another fold normally. I think you actually need to use Misc.fold_cross_gen because misc.fold_cross appends a list but you want to do an addition.

i think the following should work, but I haven't tested it. If it doesn't work let me know and I am happy to figure it out :)

Misc.fold_cross_gen (+) 0 vec (+) 0

lib/AArch64Parser.mly
```diff
@@ -649,6 +649,10 @@ instr:
 | DUP vreg COMMA xwr
   { let v,r = $4 in
     I_DUP ($2 , v, r) }
+| FMOV xwr COMMA scalar_regs
```
murzinv, line 655: It seems that it needs to be
``` 
| FMOV wreg COMMA hreg
   { I_FMOV_TG (V32, $2, VSIMD16, $4) }
| FMOV xreg COMMA hreg
   { I_FMOV_TG (V64, $2, VSIMD16, $4) }
| FMOV wreg COMMA sreg
   { I_FMOV_TG (V32, $2, VSIMD32, $4) }
| FMOV xreg COMMA dreg
   { I_FMOV_TG (V64, $2, VSIMD64, $4) }
```

lib/AArch64Parser.mly
```diff
@@ -646,6 +646,9 @@ instr:
 | STUR scalar_regs COMMA LBRK xreg k0_opt RBRK
   { let v,r = $2 in
     I_STUR_SIMD (v, r, $5, $6) }
+| ADDV scalar_regs COMMA vreg
```
murzinv, line 649: It seems that it needs to be
```
| ADDV breg COMMA vreg
  {  I_ADDV (VSIMD8, $2, $4) }
| ADDV hreg COMMA vreg
  {  I_ADDV (VSIMD16, $2, $4) }
| ADDV sreg COMMA vreg
  {  I_ADDV (VSIMD32, $2, $4) }
```

gen/AArch64Arch_gen.ml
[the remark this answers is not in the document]
murzinv: Hmm, it seems that `Misc.fold_cross` works differently to my original proposal. 
Before `Misc.fold_cross`
```
 diyone7 -arch AArch64 -variant neon "Rfe PodRW Rfe DpDatadWNe2P"
AArch64 A
"Rfe PodRW RfePNe2 DpDatadWNe2P"
Com=Rf Rf
Orig=Rfe PodRW RfePNe2 DpDatadWNe2P
{
int y[8];

0:X0=x; 0:X3=y;
1:X0=y; 1:X3=x;
}
 P0          | P1                      ;
 LDR W1,[X0] | LD1 {V0.4S, V1.4S},[X0] ;
 MOV W2,#1   | ADD V0.4S,V0.4S,V1.4S   ;
 STR W2,[X3] | ADDV S2,V0.4S           ;
             | FMOV W1,S2              ;
             | EOR W2,W1,W1            ;
             | ADD W2,W2,#1            ;
             | STR W2,[X3]             ;
exists (0:X1=1 /\ 1:X1=1)
 ```
 Note, `1:X1=1` which looks resalable since we only store value '1' 
 After change to `Misc.fold_cross`
 ```
  diyone7 -arch AArch64 -variant neon "Rfe PodRW Rfe DpDatadWNe2P"
AArch64 A
"Rfe PodRW RfePNe2 DpDatadWNe2P"
Com=Rf Rf
Orig=Rfe PodRW RfePNe2 DpDatadWNe2P
{
int y[8];

0:X0=x; 0:X3=y;
1:X0=y; 1:X3=x;
}
 P0          | P1                      ;
 LDR W1,[X0] | LD1 {V0.4S, V1.4S},[X0] ;
 MOV W2,#1   | ADD V0.4S,V0.4S,V1.4S   ;
 STR W2,[X3] | ADDV S2,V0.4S           ;
             | FMOV W1,S2              ;
             | EOR W2,W1,W1            ;
             | ADD W2,W2,#1            ;
             | STR W2,[X3]             ;
exists (0:X1=1 /\ 1:X1=4)
```
Note, now `1:X1=4` which seems impossible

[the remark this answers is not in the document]
murzinv: I think it can be as simple as
```
let reduce vec = List.fold_right (+) (List.flatten vec) 0
```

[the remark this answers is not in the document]
hugookeeffe: Perfect! That looks good to me

hugookeeffe:
Going through the code after your updates, all my comments have been addressed. I would be happy for this PR to move out of draft.

gen/AArch64Compile_gen.ml
```diff
@@ -1705,11 +1715,10 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
               Some rB,init,cs@lift_code [ldr_mixed rB rA MachSize.S128 0; gctype rB rB],st
           | R,Some (CapaSeal,Some _) -> assert false
           | R,Some (Neon n,None) ->
-              let c = sxtw r2 rd in
               let r3,st = next_reg st in
-              let cs0,st = calc0_gen csel st V64 r3 r2 in
-              let rB,init,cs,st = LDN.emit_load_idx n st p init loc r3 in
-              Some rB,init,Instruction c::pseudo cs0@cs,st
+              let cs0,st = calc0_gen csel st vdep r3 r2 in
```
murzinv, line 1709: This has a bug of lost dependency, we do not need to call `calc0_gen` here and simply do
```
let rB,init,cs,st = LDN.emit_load_idx n vdep st p init loc r2 in
Some rB,init,pseudo cs0@cs,st
```

gen/AArch64Compile_gen.ml
```diff
@@ -2036,8 +2029,9 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
               None,init,cs2@cs@lift_code [str_mixed MachSize.S128 0 r2 rA],st
           | Some (CapaSeal,Some _) -> assert false
           | Some (Neon n,None) ->
-             let init,cs,st = STN.emit_store_reg n st p init loc [r2] in
-             None,init,cs2@cs,st
+             let rA,init,st = U.next_init st p init loc in
+             let init,cs,st = STN.emit_store_dep n r2 st init rA e.C.v in
+             None,init,cs,st
```
murzinv, line 2034: I've lost cs2 here :(

murzinv:
@hugookeeffe thanks a lot for having a look! As you might seen I found couple of issues which are fixed in the resent update. Changes definitely deserve extensive testing, yet, as you've suggested, I'm moving it out of draft.

maranget:
Thanks @murzinv, what a swift reaction!

hugookeeffe:
@murzinv looks good to me! I think it just needs to be rebased again first

relokin:
@murzinv sorry for taking so long to review this. Could you rebase it once again on top of master? I'll try to review this tomorrow.

murzinv:
> @murzinv sorry for taking so long to review this. Could you rebase it once again on top of master? I'll try to review this tomorrow.

@hugookeeffe done

relokin (dismissed):
This looks good to me but I think there are still some problems with the support for this MOV variant. Consider the following litmus test:

```
AArch64 A
{
 0:X0=1;
}
 P0                    ;
 B END                 ;
 MOV V0.S[0], W0       ;
END:                   ;
 exists (0:V0.4S={1,0,0,0})
```
More generally, I don't know why litmus doesn't use the Arm neon types and intrinsics. For example, one could have used the type `int32_4_t` instead of the custom `int_4_t` and `vdupq_n_s32(0)` to initialise vector registers.

relokin:
@murzinv do you have any unittests we could add for the newly added in instructions for herd? At least one test for DUP, FMOV, ADDV and for the new variants of LD1 and ST1.

relokin:
> > @murzinv do you have any unittests we could add for the newly added in instructions for herd? At least one test for DUP, FMOV, ADDV and the new variants of LD1 and ST1.
> 
> I can add those. I assume they need to go under `herd/tests/instructions/AArch64.neon/`?

Ideally yes, otherwise just paste them here so we have some cases to look into while reviewing the code. Thank you!

maranget:
Hi @murzinv, I have updated test reference files and rebased this PR on master.  I noticed that `litmus` fails on test `V17.litmus`.

litmus/AArch64Compile_litmus.ml
```diff
@@ -832,11 +834,44 @@ module Make(V:Constant.S)(C:Config) =
...
+          | V32 -> add_w
+          | V64 -> add_q
+          | V128 -> assert false)
+        [r1;]) @ (add_128 [r2;])}
+
+    let mov_simd_ve r1 i1 r2 i2 =
+      { empty_ins with
+        memo = sprintf "mov %s[%i], %s[%i]" (print_simd_reg "o" 0 0 r1) i1 (print_simd_reg "i" 0 1 r2) i2;
+        inputs = [r1;r2]; (* r1 is intentionally in 'inputs', see comment for reg_class_stable *)
+        outputs = [];
```
maranget, line 874: Something is wrong here:  the code snippet `(print_simd_reg "o" 0 0 r1)` refers to the first of output registers, while the list of output registers is empty (line 874).

Moreover I could not understand the comment.

maranget: It would be safer to have `r1` in both lists, as, if I understand well `r1` is both read and written by the instruction.

litmus/AArch64Compile_litmus.ml
```diff
@@ -832,11 +834,44 @@ module Make(V:Constant.S)(C:Config) =
...
+          | V32 -> add_w
+          | V64 -> add_q
+          | V128 -> assert false)
+        [r1;]) @ (add_128 [r2;])}
+
+    let mov_simd_ve r1 i1 r2 i2 =
+      { empty_ins with
+        memo = sprintf "mov %s[%i], %s[%i]" (print_simd_reg "i" 0 0 r1) i1 (print_simd_reg "i" 0 1 r2) i2;
+        inputs = [r1;r2]; (* r1 is intentionally in 'inputs', see comment for reg_class_stable *)
+        outputs = [];
```
maranget, line 874: It would be safer to include `r1` in both lists, as, if I am correct, the instruction reads and writes `r1`:
```suggestion
        inputs = [r1;r2]; (* r1 is intentionally in 'inputs', see comment for reg_class_stable *)
        outputs = [r1];
```
The same change may apply to similar situations.

litmus/skel.ml
```diff
@@ -413,6 +413,8 @@ module Make
           if have_timebase then O.f "#define DELTA_TB %s" delta
         end ;
         O.o "/* Includes */" ;
+        let module I = ObjUtil.Insert(Cfg) in
+        I.insert_when_exists O.o "intrinsics.h" ;
```
maranget, line 417: The functor `Objutil.Insert` has already been applied above and the result is in scope as `Insert`:
```suggestion
        Insert.insert_when_exists O.o "intrinsics.h" ;
```
 Moreover the same line  should be added in the file `litmus/preSi.ml`  (line 416) for the command-line option `-mode presi` (and `-mode vmsa`) to work.

maranget:
The  [tests V64](https://github.com/herd/herdtools7/files/13820813/V64.litmus.txt) does not give the same result for herd and litmus:
```
AArch64 V64
(* Tests initialisation of neon registers (try litmus) *)
{
    uint64_t x[4] = { 255,255,255,255 };
    uint64_t* 0:X0 = &x;

    uint64_t 0:X2 = 2;
    uint64_t 0:X3 = 3;
    uint128_t 0:V0.2D;
    uint128_t 0:V1.2D;
}

P0               ;
MOV V0.D[1], X2  ;
MOV V1.D[0], X3  ;

STP Q0, Q1, [X0] ;

forall(x[0] = 0 /\ x[1] = 2 /\ x[2] = 3 /\ x[3] = 0)
```
Herd result, which I guess is the right one:
```% herd7 V64.litmus -variant neon
Test V64 Required
States 1
x[0]=0; x[1]=2; x[2]=3; x[3]=0;
Ok
Witnesses
Positive: 1 Negative: 0
Condition forall (x[0]=0 /\ x[1]=2 /\ x[2]=3 /\ x[3]=0)
Observation V64 Always 1 0
Time V64 0.01
Hash=0a4bf3e001a5bd32fceaf3651e759f70
```
Litmus result (compilation with option `-mach vougeot`):
```
 ./V64.exe -s 1k -r 10
Test V64 Required
Histogram (1 states)
40000 *>x[0]=255; x[1]=2; x[2]=3; x[3]=0;
No

Witnesses
Positive: 0, Negative: 40000
Condition forall (x[0]=0 /\ x[1]=2 /\ x[2]=3 /\ x[3]=0) is NOT validated
Hash=0a4bf3e001a5bd32fceaf3651e759f70
Observation V64 Never 0 40000
Time V64 0.00
```
Looking at the C file `V64.c`, one sees that register `v0` is not initialised.

maranget:
As far as I could check, the root cause resides in the simd instruction not  listing the registers they write in the `outputs` fileld. As a result., liveness analysis get confused and some resisters are not initialised.

litmus/AArch64Compile_litmus.ml
maranget: > OTOH, here we want to convince **compiler** to do what we want, so having `r1` on `outputs` might be safer but I need to check effects. Can we postpone such change until later?

Maybe we can consider that copying something to some vector elements without changing the others amounts to read the unchanged elements to re-write them?

maranget:
Hi @murzinv, I had to change a few tests for them to be accepted by litmus. Changes are in commit 1f8da1ae86b9c62337dcd802614422e90c74628b. Would you please consider checking my changes and cherry-pick the commit here?

maranget:
Please change the commit as you wish.

BTW, if the PR is ready would you squash all commits into a few, or one.

The reason is twofold:
  1. We want `make test`  to succeed after every commit.
  2. Simpler history.
 
The simplest way to keep authorship is for you to perform the squash.

maranget:
Hi @murzinv, I finally merged manually, having changed the reference file of test V46. Thanks a a lot for your work and patience.

PR #683, opened 2023-09-28
[herd] Restore architectural intent for AArch64 SWP/CAS that return to ZR

relokin:
Much like LD<op>, SWP and CAS instructions that return to ZR (e.g., WZR), are not regarding as doing a read for the purposes of a DMB.LD. In addition, such instructions do not load with acquire semantics. This commit restores the intent in the implementation of the semantics of SWP and CAS in herd7 following discussions with the architects.

In addition, this commit adds tests to the catalogue that demonstrate the semantics of LDADD/SWP/CAS when they to WZR and when they return to any other general purpose register.


jalglave:
Thanks @relokin for doing this. Please could you amend the commit title and message to indicate that we’re restoring intent following discussions with architects?
Thanks in advance,
Jade

jalglave:
@relokin I would have thought that what was required was a change to the cat file to lift the restriction on dmb.ld, rather than modifying the semantics of NoRet, which is correct I think. What do you think?
thanks,
Jade

relokin:
> @relokin I would have thought that what was required was a change to the cat file to lift the restriction on dmb.ld, rather than modifying the semantics of NoRet, which is correct I think. What do you think? thanks, Jade

I think the cat file is already correct, it works well for LD<op> instructions that return to ZR, because of this clause:
```
(* Barrier-ordered-before *)
let bob = ...
        | [R \ NoRet]; po; [dmb.ld]; po
...
```
The problem was that the semantics of SWP and CAS where not considering correctly the special case of returning to ZR. And so prior to the patch, for this test:
```
AArch64 SWP-WZR
{
 x=1;
 0:X1=2;
 0:X3=x;
}
P0               ;
 SWP X1,XZR,[X3] ;
locations[0:X1]
exists(x=1)
```
herd7 would generate a regular `R` event:
image
and after this patch, it generates a `R & NoRet` event:
image

jalglave:
Ah super! I agree, thanks Nikos!

On 29 Sep 2023, at 20:04, Nikos Nikoleris ***@***.***> wrote:

﻿

⚠ Caution: External sender


@relokin<https://github.com/relokin> I would have thought that what was required was a change to the cat file to lift the restriction on dmb.ld, rather than modifying the semantics of NoRet, which is correct I think. What do you think? thanks, Jade

I think the cat file is already correct, it works well for LD instructions that return to ZR, because of this clause:

(* Barrier-ordered-before *)
let bob = ...
        | [R \ NoRet]; po; [dmb.ld]; po
...


The problem was that the semantics of SWP and CAS where not considering correctly the special case of returning to ZR. And so prior to the patch, for this test:

AArch64 SWP-WZR
{
 x=1;
 0:X1=2;
 0:X3=x;
}
P0               ;
 SWP X1,XZR,[X3] ;
locations[0:X1]
exists(x=1)


herd7 would generate a regular R event:
[image]<https://user-images.githubusercontent.com/224592/271549537-21bf8251-24ad-49bf-b741-1ae0f3cc60f9.png>
and after this patch, it generates a R & NoRet event:
[image]<https://user-images.githubusercontent.com/224592/271549810-7d30fe58-a019-4104-93bd-6f0603b66402.png>

maranget:
Hi @relokin, @jalglave. I like the approach of this PR. Can we merge?

relokin:
Hey Luc, thanks for the review! Yes, this PR is ready to be merged.

PR #686, opened 2023-10-02
[herd] Delay errors and warning until after filtering by model.

maranget:
That way errors and warning from execution candidates rejected by the model do not show up.


maranget:
I guess this PR is an answer to PR #684 and issue #646. After digging in old code, I tend to think that errors have never been delayed that long for no specific reason.

More precisely, commit 94dd2ac introduced the (unsolved) equations, including `Warn` or `Failed` "_constraints_" as [an argument `_cs` to the function `model_kont`](https://github.com/herd/herdtools7/blob/94dd2ac1dfb22092413240814bb460f4a87bd4a1/herd/top_herd.ml#L394) in `herd/top_herd.ml` , but did nothing of it.

I am tempted to merge this PR.

herd/top_herd.ml
```diff
@@ -2,7 +2,7 @@
 (*                           the diy toolsuite                              *)
 (*                                                                          *)
 (* Jade Alglave, University College London, UK.                             *)
-(* Luc Maranget, INRIA Paris-Rocquencourt, France.                          *)
```
relokin, line 5: You might want to revert this before merging.

herd/mem.ml
```diff
@@ -1117,7 +1117,7 @@ let match_reg_events es =
           PP.show_es_rfm test es rfm ;
           prerr_endline "Unsolvable system"
         end ;
-      assert (rfmap_is_cyclic es rfm);
+      assert (true || rfmap_is_cyclic es rfm);
```
relokin, line 1120: should we remove this assert, or did you it intentionally?

relokin (commented):
The change makes sense to me. I have two minor comments.

herd/top_herd.ml
maranget: Thanks!

artkhyzha:
LGTM. Out of curiousity: what prompts the change in the unit tests? (Specifically, the need to specify the "DontCheckMixed" variant.)

maranget:
Hi @artkhyzha. The tests contain (at least!) two fatal errors: (1) the masking attempt of a symbolic locations and (2) accesses of different sizes of the same location. Before this PR the specific error about illegal masking was raised at the end of the second solving phase and error (1) was found first. After this PR the checking of (1) has moved beyond the point where (2) is checked (point (2) is after equation solving, while (1) is after model application).

However it is possible to disable the checking of (2), with `-variant DontCheckMixed`, thereby restoring previous behaviour.

artkhyzha:
Ah, I see, thank you for the explanation!

maranget:
Merged, thanks @relokin, @artkhyzha

PR #690, opened 2023-10-04
[asl] Small adjustments to ease ASL program generation

HadrienRenaud:
This PR fixes:
 - some pretty-print are outdated;
 - move the return type of a function into the local environment to ease the generation of code that returns the correct type.

This PR comes after #694, because I need the fix to `herdtools7.opam`.


jalglave:
Moving ASLRef.

PR #691, opened 2023-10-06
[herd,aarch64] ignore VMSA semantics for explicit accesses to labels

artkhyzha:
This PR aims to prevent creating of implicit TTD effects when building the semantics for loads and stores to instruction locations.

```
AArch64 MP.FR+dmb.st+po
{
TTD(x)=(oa:PA(x),valid:0);
0:X0=NOP; 0:X1=P1:Lself; 0:X4=(oa:PA(x),valid:1); 0:X5=TTD(x);
          1:X1=P1:Lself; 1:X3=x;
}
P0           | P1            ;
STR X4,[X5]  | Lself:        ;
DMB ST       |   B Lnext     ;
STR W0,[X1]  |   MOV W9,#1   ;
             | Lnext:        ;
             |   LDR W2,[X3] ;
exists 1:X9=1 /\ fault(P1,x)
```

Currently it cannot be run with `herd7`, because the VMSA semantics for  `STR W0,[X1]` will create effects that `machModelChecker.ml` will then incorrectly process. This PR aims to disable the VMSA semantics for such accesses, which will allow studying straightforward interactions between the ifetch and the vmsa extensions to the memory model.

### DIFF
--- a/herd/AArch64Sem.ml
+++ b/herd/AArch64Sem.ml
@@ -1074,13 +1074,16 @@ module Make
               fire_spurious_af dir a m) >>= M.ignore >>= B.next1T)
             mphy
             mfault in
-        M.delay_kont "6" ma
-          (if pte2 then maccess
-           else
+        M.delay_kont "6" ma (
+          if pte2 then maccess
+          else
              fun a ma ->
              match Act.access_of_location_std (A.Location_global a) with
-             | Access.VIR|Access.PTE -> maccess a ma
-             | ac -> mop ac ma >>= M.ignore >>= B.next1T)
+             | Access.VIR|Access.PTE when not (A.V.is_instrloc a) ->
+                 maccess a ma
+             | ac ->
+                 mop ac ma >>= M.ignore >>= B.next1T
+        )
 
       let lift_morello mop perms ma mv dir an ii =
         let mfault msg ma mv =
--- a/lib/symbValue.ml
+++ b/lib/symbValue.ml
@@ -554,6 +554,10 @@ module
 
   let is_virtual_v v =  if is_virtual v then one else zero
 
+  let is_instrloc v = match v with
+  | Val c -> Constant.is_label c
+  | Var _ -> false
+
   let is_instr_v =
     function
     | Val (Constant.Instruction _) -> one
--- a/lib/value.mli
+++ b/lib/value.mli
@@ -97,6 +97,8 @@ module type S =
       val is_virtual : v -> bool
       val as_virtual : v -> string option
 
+      val is_instrloc : v -> bool
+
       val op1 : op1_t -> v -> v
       val op : op_t -> v -> v -> v
       val op3 : Op.op3 -> v -> v -> v -> v


artkhyzha:
Not being sufficiently familiar with the herd7's treatment of the VMSA semantics, interested in @relokin's opinion. There are two changes to the `herd/AArch64Sem.ml`: in `lift_kvm` and `lift_memop`. Experimentally, I see that just one of them is sufficient. (And really I ended up with two just because I was imitating the flow for `-variant pte2`.) Do you agree that just one of the two insertions is okay?

My hunch is that inserting the check I want into `lift_memop` might be the preferred sufficient option.

herd/AArch64Sem.ml
```diff
@@ -1092,7 +1103,16 @@ module Make
                 fun ma _a ->
                 mop Access.PHY ma
                 >>= M.ignore >>= B.next1T in
-            lift_kvm dir updatedb mop ma an ii mphy
+            if self then
```
relokin, line 1106: Do you need to check for self? I am thinking that if there is a memory instruction to a location then we better be using self, otherwise we shouldn't get to this point at all.

herd/AArch64Sem.ml
```diff
@@ -1020,13 +1020,24 @@ module Make
...
+                   (mop Access.VIR ma >>= M.ignore >>= B.next1T)
+                   (maccess a ma)
+              | ac -> mop ac ma >>= M.ignore >>= B.next1T
+          else
              fun a ma ->
              match Act.access_of_location_std (A.Location_global a) with
              | Access.VIR|Access.PTE -> maccess a ma
-             | ac -> mop ac ma >>= M.ignore >>= B.next1T)
+             | ac -> mop ac ma >>= M.ignore >>= B.next1T
+        )
```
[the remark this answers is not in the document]
artkhyzha: I pushed something, but not sure if it's what you meant... Would you instead suggest something like:
```
        M.delay_kont "6" ma (
          if pte2 then maccess
          else
            fun a ma ->
              let ac = Act.access_of_location_std (A.Location_global a) in
              match a with
              | A.V.Val (Constant.Label _) ->
                mop Access.VIR ma >>= M.ignore >>= B.next1T
              | _ ->
                if Access.is_physical ac then
                else
                  maccess a ma
        )
 ```

**Review by relokin — APPROVED** [review]
Thanks for the changes. LGTM!

**Comment by artkhyzha** [comment]
To drive this PR towards a merge, I'd like to make sure it regresses on handwritten tests (importantly, the `-variant vmsa` ones). And it actually doesn't! Should have checked right away...
```
herd7 -variant vmsa catalogue/aarch64-VMSA/tests/D15347-M-load-shoot+addr.litmus 

Fatal: File "catalogue/aarch64-VMSA/tests/D15347-M-load-shoot+addr.litmus" Adios
Fatal error: exception SymbValue.Make(Cst)(ArchOp).Undetermined
```
I'll share an update once the PR is ready.

**Comment by artkhyzha** [comment]
As a matter of a quick update: I ran herd7 on the VMSA tests, to regress against the master -- seems to work for me now (after a speedy fix for something).

File: herd/AArch64Sem.ml [hunk 3]
```diff
@@ -1020,13 +1020,19 @@ module Make
...
              match Act.access_of_location_std (A.Location_global a) with
-             | Access.VIR|Access.PTE -> maccess a ma
-             | ac -> mop ac ma >>= M.ignore >>= B.next1T)
+             | Access.VIR|Access.PTE ->
+                if A.V.is_instrloc a then
+                 (* Ignore VMSA when loading/storing to code locations. *)
+                 (mop Access.VIR ma >>= M.ignore >>= B.next1T)
+                else
+                 (maccess a ma)
+             | ac -> mop ac ma >>= M.ignore >>= B.next1T
```
relokin, line 1034:
```suggestion
             | Access.VIR|Access.PTE when not (A.V.is_instrloc a) ->
                 maccess a ma
             | ac -> 
                 mop ac ma >>= M.ignore >>= B.next1T
```
Isn't better to retain the Access type when we don't perform translation?

herd/AArch64Sem.ml
```diff
@@ -1020,13 +1020,19 @@ module Make
...
+          if pte2 then maccess
+          else
              fun a ma ->
              match Act.access_of_location_std (A.Location_global a) with
-             | Access.VIR|Access.PTE -> maccess a ma
-             | ac -> mop ac ma >>= M.ignore >>= B.next1T)
+             | Access.VIR|Access.PTE ->
+                if A.V.is_instrloc a then
+                 (* Ignore VMSA when loading/storing to code locations. *)
+                 (mop Access.VIR ma >>= M.ignore >>= B.next1T)
```
relokin, line 1031: Is there any reason we change the access type here? See also my suggestion below.

relokin (commented):
One more comment but otherwise I think this ready. @artkhyzha if you have a look at this comment, we can merge this by the end of this week. If you have objections or need more time to review this please speak up (@jalglave, @maranget)

herd/AArch64Sem.ml
artkhyzha: No specific reason. It occurred to me an access to an instruction location should not have the `Access.PTE` type, but there is no reason to enforce it here.

herd/AArch64Sem.ml
artkhyzha: Thank you for the suggestion! I like how this looks, so I'd like to commit this.

relokin:
Merged into master thanks @artkhyzha!

PR #700, opened 2023-10-19
cat to config tool

hugookeeffe:
This PR introduces a new tool that allows the user to generate a list of relaxations from a cat file. This tool can be used by defining what let statement to generate relaxations for with `-let <name>`.
The tool can also generate a tree view of the expanded cat file

### DIFF
--- a/gen/mcat2config.ml
+++ b/gen/mcat2config.ml
@@ -3,6 +3,29 @@
 
 open Printf
 
+(*
+Bugs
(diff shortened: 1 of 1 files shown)


relokin:
@hugookeeffe do you have any updates for this PR?

hugookeeffe:
hi @relokin I have pushed my updates to the PR. I have listed the tasks I haven't completed at the top. I have also included some notes on further work, however, this is not considered in scope for this PR.

PR #701, opened 2023-10-23
[herd] New debug switch `-debug exception`

maranget:
This switch disables failure delaying and some exceptions handlers. As a result, exception backtraces are more informative.

**NB**: Exception backtraces are obtained by setting `OCAMLRUNPARAM=b` before herd command.


maranget:
Hi all, this PR is simple to code, non-intrusive. Moreover it introduces a feature that helps a lot for finding the source of exception, something I have been needing for long. So I merge it...

PR #705, opened 2023-10-26
[gen] assign write values in oneloc starting from start of com chain

hugookeeffe:
This patch attempts to improve the generation of post-condition for oneloc tests in the generator. Currently some tests generated with diy will generate an incorrect post-condition. One noteable example is CoRR - 
```
AArch64 A
"Rfe PosRR Fre"
Com=Rf Fr
Orig=Rfe PosRR Fre
{
0:X1=x;
1:X0=x;
}
 P0          | P1          ;
 MOV W0,#1   | LDR W1,[X0] ;
 STR W0,[X1] | LDR W2,[X0] ;
exists (1:X1=1 /\ 1:X2=1)
```

This patch aims to fix this by ensuring that oneloc tests start generating the post condition from the start of a chain of a communication edges, ensuring they are all taken into account in the post condition.

### DIFF
--- a/gen/cycle.ml
+++ b/gen/cycle.ml
@@ -612,7 +612,7 @@ let is_rmw d e = match d with
 | W -> is_rmw_edge e.prev.edge
 | J -> is_rmw_edge e.edge
 
-
+let is_com_rmw n0 = E.is_com n0.edge || is_rmw_edge n0.edge
 
 let remove_store n0 =
   let n0 =
@@ -922,16 +922,23 @@ let set_same_loc st n0 =
               m.prev.evt.loc <> m.evt.loc &&
               m.next.evt.loc = m.evt.loc) n in
         split_by_loc m
-      with Not_found -> try
-        let m =
-          find_node
-            (fun m -> match m.prev.edge.E.edge with
-            | E.Fr _|E.Rf _|E.Ws _|E.Leave _|E.Back _
-            | E.Hat|E.Rmw _|E.Irf _|E.Ifr _ -> true
-            | E.Po _|E.Dp _|E.Fenced _|E.Insert _|E.Store|E.Node _ -> false
-            | E.Id -> assert false) n in
-        split_one_loc m
-      with Exit -> Warn.fatal "Cannot set write values" in
+      with
+      | Not_found ->
+        (*check if node is preceded by a non com/rmw node and is itself a com/rmw node*)
+        let to_com_rmw n0 = not (is_com_rmw n0.prev) && is_com_rmw n0 in
+        fold (fun n0 _ -> if E.is_id n0.edge.E.edge then assert false) n ();
+        try
+          (* check for R ensures that we start on Fr or Rmw if possible*)
+          let m = find_node (fun m -> to_com_rmw m && m.evt.dir = Some R) n in
+          split_one_loc m
+        with Not_found -> try
+          (* The previous search failed. This search will return the W node from
+             which an Rf edge starts, provided that the previous edge is not a
+             communication or a Rmw edge *)
+          let m = find_node (fun m -> to_com_rmw m) n in
+          split_one_loc m
+        with Not_found -> Warn.fatal "cannot set write values"
+      | Exit -> Warn.fatal "cannot set write values" in
     let initvals = set_all_write_val nss in
     nss,initvals
 


gen/cycle.ml
```diff
@@ -927,6 +927,12 @@ let set_same_loc st n0 =
...
             | E.Fr _|E.Rf _|E.Ws _|E.Leave _|E.Back _
+            | E.Hat|E.Rmw _|E.Irf _|E.Ifr _ -> false
+            | E.Po _|E.Dp _|E.Fenced _|E.Insert _|E.Store|E.Node _ -> true
+            | E.Id -> assert false
+            &&
+            match m.edge.E.edge with
+            | E.Fr _|E.Rf _|E.Ws _|E.Leave _|E.Back _
             | E.Hat|E.Rmw _|E.Irf _|E.Ifr _ -> true
             | E.Po _|E.Dp _|E.Fenced _|E.Insert _|E.Store|E.Node _ -> false
             | E.Id -> assert false) n in
```
relokin, line 938: Can't we do?
```suggestion
            (fun m -> E.is_com m.edge && not (E.is_com m.prev.edge))
```
But as I wrote in the issue you opened, I think the logic should apply if the current edge is com and the next or the previous is not com.

relokin (commented):
Do we have a way to test this? Can we produce a conf file based on the axioms `coRW1-Exp`, `coRW1-MTE`, `coWW-Exp`, `coWR-Exp` and `coWR-MTE` and check that the test that diy7 would produce are all forbidden?

gen/cycle.ml
relokin: Also I just realised that here we consider `Rmw` as a com edge but `is_com` would return `false`. I think `is_com` is correct and here there is a bug, can you confirm this?

hugookeeffe: I think that considering rmw is correct here, as we do not want to allow the case where we start calculating the write values at the end of an rmw node. There is also another issue with the change you suggested, in that it doesn't `assert false` for E.id, which I believe is necessary.

considering these two points, I'm not sure we can make the code much simpler; we can use is_rmw for rmw however we don't have an equivalent for Id, so we might still need to use the pattern matching. However, making such a function is quite trivial, so if you think it makes the code cleaner I can implement this.

In regards to extending the logic to check if the next edge is not com, as mentioned on the github issue, I don't think this would be right as it would allow us to start from the middle of a communication edge chain.

hugookeeffe:
The following simple configuration file was tested and succesfully generates tests. From what I can tell the tests generated seem correct

```
-arch AArch64
-safe Pos** Rfe Fre
-name corr_issue
-stdout true
-size 8
-oneloc
-mode uni
```

relokin:
hey @hugookeeffe, I tried generating tests with the configuration file you provided out of 9 tests that diy7 produced 2 are allowed. The simpler of the two is the following:

```
AArch64 corr_issue004
"PosWW Rfe PosRR Fre"
Cycle=Rfe PosRR Fre PosWW
Relax=
Safe=Rfe Fre PosWW PosRR
Generator=diy7 (version 7.56+03)
Com=Rf Fr
Orig=PosWW Rfe PosRR Fre
{
0:X1=x;
1:X0=x;
}
 P0          | P1          ;
 MOV W0,#2   | LDR W1,[X0] ;
 STR W0,[X1] | LDR W2,[X0] ;
 MOV W2,#1   |             ;
 STR W2,[X1] |             ;
exists ([x]=1 /\ 1:X1=1 /\ 1:X2=1)
```

Would it be easy to fix this as well?

hugookeeffe:
@relokin my latest patch fixes this issue. What I implemented was an extra clause ensuring that the first direction started on is a read if possible. The reason for this is that starting on a read (i.e. Fre/i) ensures that the write value always comes later in the calculation than the read, hence it should always give a forbidden cycle.

I have tested this on an extended configuration file that generates ~2500 tests, and it manages to get all but 15 to be forbidden.

The tests which aren't forbidden all have the same pattern (Ld/StAdd, RfiPA). The reason these aren't being calculated completely is because they don't fit into the case i described above, hence it starts by default at the beginning of any rmw edge. This could be fixed by making sure to start after an rmw edge, or (perhaps better) att he end of a Hat if there is one. I have included an example of a not-forbidden test and the extended configuration file below.



```
AArch64 corr_issue2308
"Amo.StAdd RfiPA HatAP PosRW Rfe"
Cycle=Rfe Amo.StAdd RfiPA HatAP PosRW
Relax=
Safe=Hat Rfe PosRW [Amo.StAdd,RfiPA]
Generator=diy7 (version 7.56+03)
Com=Co Rf
Orig=Amo.StAdd RfiPA HatAP PosRW Rfe
{
0:X0=x;
1:X0=x;
}
 P0            | P1          ;
 MOV W1,#1     | LDR W1,[X0] ;
 STADD W1,[X0] | MOV W2,#2   ;
 LDAR W2,[X0]  | STR W2,[X0] ;
exists ([x]=2 /\ 0:X2=1 /\ 1:X1=1)
```


```
-arch AArch64
-safe Pos** Rfe Fre Hat [Amo.LdAdd, RfiPA] [Amo.StAdd, RfiPA] DpCtrlIsbsR
-name corr_issue
-stdout true
-size 8
-oneloc
-mode uni
```

relokin:
Thanks @hugookeeffe, I've added some suggestions to simplify some bits of your patch. Feel free to ignore if something doesn't make sense.

For `corr_issue2308`, I would think that the first write we can resolve without doubt would be the STR in P1. It's not about the previous or the current edge, but it's just that both the destination (x) and the value (2) of this write event are set. Starting from that we can continue and process the following Rfe, etc. What do you think? Can we look for such nodes and start from them if they exist?

gen/cycle.ml
```diff
@@ -922,16 +922,20 @@ let set_same_loc st n0 =
...
+      | Not_found ->
+        (*check if node is preceded by a non com/rmw node and is itself a com/rmw node*)
+        let to_com_rmw n0 = not (is_com_rmw n0.prev) && is_com_rmw n0 in
+        fold (fun n0 _ -> if E.is_id n0.edge.E.edge then assert false) n ();
+        try
+          (* check for R ensures that we start on Fr or Rmw if possible*)
+          let m = find_node (fun m -> to_com_rmw m && m.evt.dir = Some R) n in
+          split_one_loc m
+        with Not_found -> try
+          let m = find_node (fun m -> to_com_rmw m) n in
```
relokin, line 935:
```suggestion
          (* The previous search failed. This search will return the W node from
             which an Rf edge starts, provided that the previous edge is not a
             communication or a Rmw edge *)
          let m = find_node (fun m -> to_com_rmw m) n in
```

relokin (commented):
Thanks, Hugo one minor change otherwise this is ready.

Can you please squash your commits into one and rebase on top of master?

hugookeeffe:
thanks for the comment @relokin. I have implemented your suggestion and squashed the commits and rebased on top of master. If anything seems wrong let me know and I will try and fix it :)

relokin:
Thanks @hugookeeffe, can I ask that you also add some description to the commit? Something like the first message in this conversation would be very helpful.

Otherwise LGTM, I am planing to merge this by Friday, unless there are any concerns or someone needs more time to review this.

relokin:
Merged into master, thanks @hugookeeffe!

PR #709, opened 2023-10-31
[herd,asl] Fix addition of symbolic constant and offset

maranget:
There were two problems, fixed as follows:
  + Slicing a symbolic constant with position [63] is now accepted and returns the bitvector `0`.
  + Converting a Z.t to an int is now performed by
     - Computing remainder by 2^64,
     - Followed by sign adjustement.


maranget:

PR #710, opened 2023-10-31
[asllib] Fix: add global constants to the global runtime environment.

maranget:
Global constants were not accessible from the interpreter. This (simple) fix consists in copying constants from the typing environment into global storage.

 An alternative and more sophisticated fix would be to scan the constant part of the global environment as part of identifier search. This would involve more code but would probably be (slightly) more efficient at runtime.

### DIFF
--- a/asllib/Interpreter.ml
+++ b/asllib/Interpreter.ml
@@ -217,6 +217,17 @@ module Make (B : Backend.S) (C : Config) = struct
       { global; local = empty_scoped Scope_Global }
     in
     let env =
+      (*
+       * Should we put constants in storage, as we do here?
+       * Alternative would be to consider constants in env.ml.
+       *)
+      IMap.fold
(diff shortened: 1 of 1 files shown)


maranget:
Indeed, PR #711 apparently no longer requires this fix. This probably due to something having changed either in ASL code or in the typer. However, I'd tend to keep the fix, as improving robustness. Your opinion?

maranget:
In fact I do not understand everything, At some point constants declared in enumeration type were not replaced by their integer value and herd was failing. Apparently, this is no longer the case. Hence this PR looks useless, or am I missing something?

maranget:
In other words, do we still an examples of a test such that:

1. The test fails with herd or aslref compiled from the master branch.
2. This PR fixes the problem.

maranget:
I see this PR does not look very useful. Let us close it.

PR #711, opened 2023-10-31
[herd,asl] Implement branch immediate instructions

maranget:
- [x] Non-conditional (`B` instruction)
- [x] Compare with zero (`CBZ`, `CBNZ`)
- [x]  Branch conditional (`B.cond`)

### DIFF
--- a/herd/AArch64ASLSem.ml
+++ b/herd/AArch64ASLSem.ml
@@ -159,8 +159,10 @@ module Make (TopConf : AArch64Sig.Config) (V : Value.AArch64ASL) :
       let open Asllib.AST in
       let with_pos desc = Asllib.ASTUtils.add_dummy_pos desc in
       let ( ^= ) x e =
-        S_Decl (LDK_Let, LDI_Var (x, None), Some e) |> with_pos
-      in
+        S_Decl (LDK_Let, LDI_Var (x, None), Some e) |> with_pos in
+      let ( ^^= ) x e =
+        let le_x = LE_Var x |> with_pos in
+        S_Assign (le_x, e, V1) |> with_pos in
       let lit v = E_Literal v |> with_pos in
       let liti i = lit (L_Int (Z.of_int i)) in
       let litb b = lit (L_Bool b) in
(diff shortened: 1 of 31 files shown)


maranget:
As a reminder, the current implementation does _not_ emit commit (a.k.a. branching) events for `CBZ`/`CBNZ`.

maranget:
> Hi @maranget 

Hi @HadrienRenaud Hadrien
> I had a few minor comments that I'm going to forget as you've already merged this PR, but one is a bit bigger:

Sorry for this I was eager to merge this PR, even not perfect (!) because it is  connected to the on-going PR #786, The aim being to have control dependencies in AArch64/ASL.

> Shouldn't we always perform the write to `_PC` on begining of an instruction? And always hide it in AArch64?
> 
> Pros:
> 
> * The translated decode written in `AArch64ASLSem` would not have lines that do not correspond to lines in the decode as it is present in the manual;
> * Future uses of `_PC` would be already present;
> * This write to `_PC` would be part of the normal model for executing an instruction, and so we wouldn't treat any instruction differently.

I am not sure that non-branching instructions should manipulate PC. As a matter of fact and if I am not wrong, ASL code of non-branching do not read nor write PC. Moreover the PC increment performed for non-branching instructions seems to be implicit. Said otherwise, I did not find this increment in ASL code, but I have not tried very hard.

> Cons:
> 
> * We have an extra event for all other instructions (one over ~500).
> 
> Let me know what you think of this, Hadrien
> 
> Among other remarks:
> 
> 1. declaration of `_PC` is probably not allowed with the declaration of getters/setters

I do not understand.


> 2. Why not use `Set.S.find_last` for the fold to get the `PC` setting?

You are right, I tend to overlook some of library functions, I'll make a quick PR for this.


> 3. I find a bit weird to add a register in `AArch64` just to use as a translation step for ASL, but if you think it is clear that this is not actually used in any `AArch64Sem` setting, it's fine for me.

Well, I have added this PC register without thinking much about alternatives. Thinking twice I think I wanted the PC accesses to appear in the final AArch64 diagrams.

maranget:
> I am not sure that non-branching instructions should manipulate PC. As a matter of fact and if I am not wrong, ASL code of non-branching do not read nor write PC. Moreover the PC increment performed for non-branching instructions seems to be implicit. Said otherwise, I did not find this increment in ASL code, but I have not tried very hard.


Well looking at machine code for `B .+4` (branch to myself) the actual offset is zero. This suggests that PC is incremented early in the instruction execution sequence and explains

1. Why the ASL code of non-branching instructions do not write PC.
2.  why the ASL code of conditional  branches do not update PC in the branch non-taken case --- which the patched `BranchNotTaken` function in `patches.asl` does.

Hence it would be more realistic to assign the address of the next instruction to PC before executing an instruction instead of the address of the instruction itself, as **herd** code does at the moment. This would of course mean to subtract four from the assembly level offsets of the branching instructions. I'll have try.

maranget:
> > 3\. Why not use `Set.S.find_last` for the fold to get the `PC` setting?
> 
> You are right, I tend to overlook some of library functions, I'll make a quick PR for this.

I have tried, the code that uses `Set.S.find_last' is a bit longer and more contrived as one has to extract the written value. So using `fold`  is acceptable.

maranget:
> 1. declaration of _PC is probably not allowed with the declaration of getters/setters

Ah I see, In fact the _PC variable, the _PC getter and the _PC setter are all defined in `implementation.,asl`. I have removed the variable definition and it seems to work. See PR #790.

PR #713, opened 2023-11-02
[herd] Minor user-interface suggestions for herd

artkhyzha:
This PR implements two very simple safeguards, both ifetch-specific in nature.

Firstly, whenever there is a label in the initialization list, raise a user error if it is not actually present in the code of the litmus test. This already is marked as an error if there is a load or a store with such an undefined location, but the following test, for instance, is still possible:

```
AArch64 AK004
{
0:X5=P0:Lself00;
}
 P0         ;
 MOV X4,X5  ;
locations [0:X4]
```

To remove a possibility of confusing errors in litmus tests, this PR inserts a simple check into `test_herd.ml` to ensure all labels in the precondition are actually defined.

Secondly, whenever there is a store into a label, raise a user error and inform the user that they may want to use `-variant ifetch`.

```
AArch64 AK005
{
0:X0=NOP; 0:X1=P0:Lself00;
}
 P0          ;
 STR W0,[X1] ;
 Lself00:    ;
  B Lself01  ;
 MOV X2, #1  ;
 Lself01:    ;
exists (0:X2=0)
```

```
Warning: File "./herd/tests/instructions/AArch64/AK005.litmus": Store to P0:Lself00 requires ifetch functionality. Please use `-variant ifetch` as an argument to herd7 to enable it. (User error)
```

PR #714, opened 2023-11-03
[herd] Add support for AArch64 TFSR_ELx in async MTE

relokin:
This PR adds support for register `TFSR_ELx` for async MTE. When an async fault happens due to a tag mismatch we set the 0 bit of register `TFSR_ELx`. Note that we define `TFSR_ELx` which is a generic name for the registers `TFSR_EL01` `TFSR_EL1` etc. This is intentional and in order to avoid being specific about the exception level that the test runs in. For the test:

```
AArch64 B004
Variant=async
{
 0:X1=x:red
}
 P0          ;
 MOV W0,#1   ;
 STR W0,[X1] ;
forall([x]=1 /\ 0:TFSR_ELx=1)
```

The semantics are shown in this graph:

image

### DIFF
--- a/herd/AArch64Sem.ml
+++ b/herd/AArch64Sem.ml
@@ -1014,8 +1014,7 @@ module Make
           | (Handled,_)|(LoadsFatal,Dir.W) ->
              fun ma ->
              let set_tfsr = write_reg AArch64Base.tfsr V.one ii in
-             let (>>) = M.bind_ctrl_first_outputs in
-             let ma = ma >> (fun a -> (set_tfsr >>| mfault) >>! a) in
+             let ma = ma >>*== (fun a -> (set_tfsr >>| mfault) >>! a) in
              mm ma >>! B.Next []
           | Skip,_ ->
              fun _ ->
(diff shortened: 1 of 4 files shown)


relokin:
@jalglave this is the patch we discussed. I had another look at the Arm ARM:

> FEAT_MTE_ASYNC provides hardware support for asynchronously accumulating Tag Check
Faults into the TFSRE0_EL1 or TFSR_ELx register, rather than causing a synchronous exception.

Which seems to indicate that there is a TagCheck Fault so I decide to keep the event for now, but it is very easy to remove it. My dilemma was that the Arm ARM indicates that it happens, but I don't think there is a way to observe it other than reading the `TFSR_ELx` register.

jalglave:
hi @relokin I think we resolved your dilemma when chatting correct? If yes please do go ahead and integrate this.
Thanks,
Jade

jalglave:
Hi @relokin as discussed, let's get this in and we'll modify it later if needed.
Thanks,
Jade

relokin:
Thank you! I will merge this before the end of the day and after #713. If anyone needs more time to review this, please speak up :)

PR #717, opened 2023-11-16
[herd] Add branching events for AArch64 CAS instructions

relokin:
This PR adds Branching events to CAS instructions. A CAS instruction writes to memory only if its condition is satisfied. As such a Branching event is added prior with a control dependency to the potential Write memory event.


maranget:
Hi @relokin, as far as I remember Bcc is for actuel branch conditional, while Pred is for intra-instruction decisions. If I am not wrong there are no real distinction. However, again as far as I remember, Bcc _ is alone while Pred always point to some event by `iico_control`.

 In short, `Commit (Pred _)` looks appropriate.

relokin:
@jalglave can you have a look? This PR implements the following semantics:

```
AArch64 CAS-success
{
 x=0;
 0:X2=x;
 0:X0=0; 0:X1=1;
}
 P0             ;
 CAS W0,W1,[X2] ;
```

image
image

```
AArch64 CAS-fail
{
 x=2;
 0:X2=x;
 0:X0=0; 0:X1=1;
}
 P0             ;
 CAS W0,W1,[X2] ;
```
image
image

I haven't added a switch to optimise this but now I am thinking that maybe we should defer this until we find a case where this performance optimization is need and it's not only theoretical.

relokin:
I will merge this later today, please speak up if you have any objections or need more time to review.

PR #720, opened 2023-11-28
[herd, asl] Add a partial po field to account for ASL

HadrienRenaud:
The main goal of this pr can be summarized by the following test:
```
ASL test-po
{}

func main () => integer begin
  let y = 0;
  let x = y + y; // Here

  return 0;
end
```

We want the two reads to `y` to not be related by `po`. This is not possible with herd as of today, because the program order is considered a total order on the effects of a thread.

This PR introduce the possibilities for an architecture to use a partial program order.

This is done by adding a field inside the `event_structure` graph, and adding new combinators that edit specially this fields.

That results in the following graph being produced for the above test:
test-po dot
We can see that there is no `partial_po` arrow between _b_ and _c_ when there is a `po` one.

### DIFF
--- a/herd/event.ml
+++ b/herd/event.ml
@@ -198,6 +198,11 @@ val same_instance : event -> event -> bool
   with type elt0 = event
   and module Elts = EventSet
 
+  module EventTransRel : InnerTransRel.S
+    with type elt = event
+     and module Set = EventSet
+     and module Rel = EventRel
+
   type event_rel = EventRel.t
 
   val debug_rel : out_channel -> event_rel -> unit
@@ -206,7 +211,8 @@ val same_instance : event -> event -> bool
       procs : A.proc list ;
       events : EventSet.t ;                     (* really a set *)
       speculated : EventSet.t ;                 (* really a set *)
-      po : EventSet.t * EventRel.t; (* A forest: roots first + partial order *)
+      po : EventSet.t * EventRel.t; (* speculated po represented as a forest: roots first + partial order *)
+      partial_po : EventTransRel.t;
       intra_causality_data : EventRel.t ;       (* really a partial order relation *)
       intra_causality_control : EventRel.t ;    (* really a partial order relation *)
       intra_causality_order : EventRel.t ; (* Just order *)
@@ -322,6 +328,10 @@ val same_instance : event -> event -> bool
   val para_output_right :
     event_structure -> event_structure -> event_structure option
 
+  (* Partial po sequencing, output in second argument *)
+  val para_po_seq_output_right :
+    event_structure -> event_structure -> event_structure option
+
 (* Sequence memory events, otherwise parallel composition *)
   val seq_mem :
     event_structure -> event_structure -> event_structure option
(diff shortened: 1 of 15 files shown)


lib/innerRel.ml
```diff
@@ -536,9 +538,7 @@ and module Elts = MySet.Make(O) =
 
     let _all_topos1 nodes edges =
       do_all_topos nodes
-        (filter
-           (fun (e1,e2) -> Elts.mem e1 nodes && Elts.mem e2 nodes)
-           edges)
+        (filter_elts (fun e -> Elts.mem e nodes) edges)
```
maranget, line 541: _all_topos1 (and do_all_topos above) is dead code. Perhaps it is time to delete?

herd/event.ml
```diff
@@ -198,14 +198,21 @@ val same_instance : event -> event -> bool
   and module Elts = EventSet
 
   type event_rel = EventRel.t
+  type event_trans_rel = { input: EventSet.t; rel: EventRel.t; output: EventSet.t }
```
maranget, line 201: As there are other **input** and **output** record fields in this file, compilation with dune **release** profile yields a lot of warnings. I am in favor of renaming those fields as  **trans_input** and **trans_output** for instance.
Alternatively:
```suggestion
module Trans = struct
  type t = { input: EventSet.t; rel: EventRel.t; output: EventSet.t }
end
```
Then the fields are `Trans.input`, `Trans.output`, etc.

herd/event.ml
```diff
@@ -1084,7 +1121,39 @@ module Make  (C:Config) (AI:Arch_herd.S) (Act:Action.S with module A = AI) :
          if EventSet.is_empty out2 then Some out
          else Some out2
 
-(* Sequential composition *)
+    let recompute_event_trans_rel events rel =
```
maranget, line 1124: Just wondering, why is this function useful?

relokin:
@HadrienRenaud, may I ask because I think I run into a similar issue. Is `asl-po` meant to be a way to order intra-instruction Effects?

lib/InnerTransRel.mli
```diff
@@ -0,0 +1,45 @@
+(** Implements a transitive reflexive relation. *)
```
maranget, line 1: I am puzzled, are the represented relation reflexive?

If I do not miss something, the last function `to_transitive` does not produce a reflexive relation.

lib/InnerTransRel.ml
```diff
@@ -0,0 +1,91 @@
...
+
+  let union6 t1 t2 t3 t4 t5 t6 =
+    {
+      input = Set.union6 t1.input t2.input t3.input t4.input t5.input t6.input;
+      rel = Rel.union6 t1.rel t2.rel t3.rel t4.rel t5.rel t6.rel;
+      output =
+        Set.union6 t1.output t2.output t3.output t4.output t5.output t6.output;
+    }
+
+  let unions = function [] -> empty | h :: t -> List.fold_left union h t
```
maranget, line 88: Dichotomic implementation :)
```suggestion
    let rec unions2 k = function
  | [] -> k
  | [r] -> r::k
  | r1::r2::rs -> unions2 (union r1 r2::k) rs
  
  let rec unions = function
  | [] -> empty
  | [r] -> r
  | rs -> unions (unions2 [] rs)
```

maranget:
Hi @HadrienRenaud LGTM.  The conflict in ASLSem will be better resolved by you.

maranget:
@HadrienRenaud. Merged, thanks a lot.

PR #722, opened 2023-11-29
[tools] miaou! A cat that writes in English.

maranget:
Simple usage: `miaou7 -show <name> <catfile>` will translate the definition of `<name>`  from model `<catfile>` into LaTeX.

More advanced usage: the output will use LaTeX commands to pretty-print event set and relation definitions. For instance the **po** relation corresponds to the the command `\newcommand{\po}[2]{#1 is program-order before #2}`. Such commands can be defined in a specific file, say `catdefs.tex`. Then, the command-line option `-tex catdefs,tex` yields the following benefits:
  1. Commands outputed by `miaou7` and not defined are flagged; and
  2. The number of parameters of commands tells if the command correspond to an event set (one argument) or to a relation (two arguments), thereby guaranteeing  appropriate formatting of definitions.

Be aware that relation names and command differ: non-alphabetical characters are deleted from Cat names to produce legal LaTeX command names; a few Cat names undergo special treatment (_e.g._ `dmb.full` -> `\DMBFULL`, `L` -> `\REL`,...).


maranget:
Merged, thanks @jalglave

PR #723, opened 2023-11-29
[AArch64 cat] Revamp definition of same-loc relation

relokin:
1. Two Memory Effects E1 and E2 are related with same-loc, if the E1 is to the same location as E2.

2. If either E1 or E2 is an MMU Fault Effect where the Output Address is known (AccessFlag or Permission Fault) then E1 and E2 are related with same-loc if the low-order-bits of the address associated with the Fault Effect are the same as the low-order-bits of the Memory Effect and the associated Implicit TTD Reads have the same Output Address.

3. If either E1 or E2 is an MMU Fault Effect where the Output Address is not known (Translation Fault), then E1 and E2 are related with same-loc, if the low-order-bits of the address associated with the Fault Effect are the same as the low-order-bits of the Memory Effect.


relokin:
Thanks @jalglave, I am fixing a glitch with the kinds and I will merge soon.

relokin:
I think this is now ready. The only thing I wanted to check with you @jalglave and @maranget is if you're ok with having all tests from `catalogue/aarch64-VMSA/tests` in `catalogue/aarch64-VMSA/shelf.py` because I am not sure why it was chosen to include only a small subset initially. Is there any benefit in not having them?

artkhyzha:
(Not a blocker for this PR by any means, just as an observation.)

Looking at the changes: it's a good idea to have VMSA tests as a part of `make test`, but it's interesting to consider what assurances that provides. The Makefile sets `TIMEOUT=16.0`. Do many tests violate that timeout?

relokin:
> (Not a blocker for this PR by any means, just as an observation.)
> 
> Looking at the changes: it's a good idea to have VMSA tests as a part of `make test`, but it's interesting to consider what assurances that provides. The Makefile sets `TIMEOUT=16.0`. Do many tests violate that timeout?

Yes many would timeout but maybe that's fine as a default. If you know that you want to run as many as possible then you can override TIMEOUT from your cmdline

make vmsa-test TIMEOUT=600

with a timeout of 10min, I think it was only Marc03 that missed the mark :)

I was a bit reluctant to add them to `make test`, since it is invoked by the `pre-commit` hook. Even with the default small timeout, I wouldn't want to wait for so long after every commit (even with the current make test, most of the times I have to admit I do `git commit --no-verify`).

relokin:
Having said that nothing prevents us from introducing another target say make test-all and have all tests included. I think that's actually a good idea.

jalglave:
hi all, please could we get this in asap to help progressing other works which need the latest cat file?
I would welcome a new PR for make test all, but we shouldn't be blocking this one please.
Thanks in advance,
Jade

relokin:
I agree with @jalglave. I will merge this now.

That is not to say I don't agree with the desire for `make test-all`. I will open a PR with `make test-all` and we can decide how we'll redistribute what tests we run with `make test` (more frequent use) and `make test-all` (less frequent use).

PR #724, opened 2023-12-01
[Makefile] Add a test-all target

relokin:
The Makefile used to offer an option to run all tests. As we added more tests, the overhead for running all tests becomes non-trivial. More notably, the vmsa-test and ets2-test targets include more litmus tests than any other. In addition, the fact that make test is invoked in the precommit hook makes it crucial that make test terminates as quickly as possible. This commit adds another target test-all which is inclusive of all tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,6 +65,7 @@ check-deps::
 # Tests.
 TIMEOUT=16.0
 
+test-all:: test
 test:: | build
 
 test:: dune-test
@@ -378,6 +379,7 @@ mte-test:
 		$(REGRESSION_TEST_MODE)
 	@ echo "herd7 catalogue aarch64-MTE tests: OK"
 
+test-all:: vmsa-test
 vmsa-test:
 	@ echo
 	$(HERD_CATALOGUE_REGRESSION_TEST) \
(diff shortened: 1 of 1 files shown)


relokin:
Yes, it's rather easy for the catalogue tests. We would have to create separate `shelf.py` files for these three use-cases (e.g., `shelf.py` for the www, `test-shelf.py` for `make test` and `test-all-shelf.py` for `make test-all`). I guess for most catalogues, it might not make sense to separate them. But this would allow us to add some tests from the VMSA and ETS2 catalogues to the `test` target (and leave all test to `test-all`).

maranget:
Hi @jalglave and @relokin do we wait for the program above to be completed or can we merge now?

maranget:
What about merging this PR now. The feature looks useful.

jalglave:
Yes please.

maranget:
Some of the tests are failing after rebase. I guess @relokin you'll see the same. In any case we cannot merge as is.

maranget:
Ok I think I found the (stupid but not too harmful) bug. I'll make a separate PR for it.

maranget:
Fix is PR #827.

maranget:
Hi @relokin. The fix PR has just been merged. Please consider rebasing and merging

PR #730, opened 2023-12-13
[all] Implement AArch64 "reverse bytes" instructions

maranget:
Implement the REV instruction, as well as its variants REV16, REV32, REV64. LItmus test example:
```
AArch64 L094
{
uint32_t 0:X1=0x87654321;
uint32_t 0:X2;
uint64_t 0:X3=0xfedcba87654321;
uint64_t 0:X4;
}
  P0       ;
 REV W2,W1 ;
 REV X4,X3 ;

forall 0:X2=0x21436587 /\ 0:X4=0x21436587badcfe00
```


maranget:
We canot use  `patch.asl` here, because the call to assignment `Elem` is from the instruction ASL file. I suggest waiting for ASL V1 that will necessarily correct this (the prototype does).

PR #737, opened 2023-12-16
[asl] Various works

HadrienRenaud:
This PR implements 3 things:
 - [x] Fix the Makefile that would break on `make all`.
 - [x] Remove bit-vector types without explicit widths.
 - [x] Hide names of stdlib locals, so that global variable declared in an ASL program cannot interfere with the locals declared in stdlib.

### DIFF
--- a/asllib/ASTUtils.ml
+++ b/asllib/ASTUtils.ml
@@ -60,6 +60,7 @@ let add_pos_from_st pos desc =
 let add_pos_from pos desc = { pos with desc }
 let with_pos_from pos { desc; _ } = add_pos_from pos desc
 let map_desc f thing = f thing |> add_pos_from thing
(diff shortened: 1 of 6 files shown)


maranget:
Hi @HadrienRenaud, could you please provide an example of how user declared _global_ variables could interfere with (stdlib) _local_ variables?

maranget:
Ah ok, the root cause is that one cannot give the name of a global variable to a local one as in:
```
func f(z:integer) => integer
begin
  return z;
end

var z : integer;
```

maranget:
As for the point "Remove bit-vector types without explicit widths" , I see from the code that, in the type of bitvectors the specific `bits_constraint` is replaced by the more general (?) `expr`.  The connection of this replacement with   "Remove bit-vector types without explicit widths"  is unclear to me, or am I missing something?

PR #739, opened 2023-12-18
[asldoc] Progress on SFor and PSingle

HadrienRenaud:


maranget:
Merged into asldoc2, thanks @HadrienRenaud.

PR #743, opened 2023-12-20
[asl] (fix) correctly propagate ASL po

HadrienRenaud:
ASL partial po was not propagated to AArch64+ASL, which caused iico_order to be empty.
This fixes the SWP AArch64.ASL regression tests.

### DIFF
--- a/lib/InnerTransRel.ml
+++ b/lib/InnerTransRel.ml
@@ -17,6 +17,7 @@ module type S = sig
   val union6 : t -> t -> t -> t -> t -> t -> t
   val unions : t list -> t
   val to_transitive_rel : t -> Rel.t
+  val to_implicitely_transitive_rel : t -> Rel.t
 end
 
 module Make (O : MySet.OrderedType) :
@@ -101,4 +102,6 @@ module Make (O : MySet.OrderedType) :
     | li -> unions2 [] li |> unions
 
   let to_transitive_rel t = Rel.transitive_closure t.rel
+
+  let to_implicitely_transitive_rel t = t.rel
 end
--- a/lib/InnerTransRel.mli
+++ b/lib/InnerTransRel.mli
@@ -40,6 +40,9 @@ module type S = sig
 
   val to_transitive_rel : t -> Rel.t
   (** [to_transitive_rel t] is the InnerRel representation of t. *)
+
+  val to_implicitely_transitive_rel : t -> Rel.t
+  (** [to_implicitely_transitive_rel t] is a non-transitive representation of t. *)
 end
 
 module Make : functor (O : MySet.OrderedType) ->
--- a/lib/innerRel.mli
+++ b/lib/innerRel.mli
@@ -32,6 +32,7 @@ module type S =  sig
     val succs : elt0 -> map -> Elts.t
     val add : elt0 -> elt0 -> map -> map
     val subrel : map -> map -> bool
+    val exists_path : (elt0 * elt0) -> map -> bool
     val to_map : t -> map
     val of_map : map -> t
   end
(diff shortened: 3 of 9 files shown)


lib/InnerTransRel.ml
```diff
@@ -17,6 +17,7 @@ module type S = sig
   val union6 : t -> t -> t -> t -> t -> t -> t
   val unions : t list -> t
   val to_transitive_rel : t -> Rel.t
+  val to_implicitely_transitive_rel : t -> Rel.t
```
maranget, line 20: Function  `to_implicitely_transitive_rel` is not used in this PR.
```suggestion
```

lib/InnerTransRel.ml
```diff
@@ -101,4 +102,6 @@ module Make (O : MySet.OrderedType) :
     | li -> unions2 [] li |> unions
 
   let to_transitive_rel t = Rel.transitive_closure t.rel
+
+  let to_implicitely_transitive_rel t = t.rel
```
maranget, line 106:
```suggestion
```

lib/InnerTransRel.mli
```diff
@@ -40,6 +40,9 @@ module type S = sig
 
   val to_transitive_rel : t -> Rel.t
   (** [to_transitive_rel t] is the InnerRel representation of t. *)
+
+  val to_implicitely_transitive_rel : t -> Rel.t
+  (** [to_implicitely_transitive_rel t] is a non-transitive representation of t. *)
```
maranget, line 45:
```suggestion
```

maranget:
This PR brings correctness, which is good --- necessary in fact. But it also brings an important performance regression. The the about x10 slowdown originates in the transitive closure of `partial_po` being of significant size --- about 180k elements for the test [`SWP.litmus`](https://github.com/herd/herdtools7/blob/master/herd/tests/instructions/AArch64.ASL/SWP.litmus).

Previous performance can be restored by _not_ computing the transitive closure of `partial_po` and by using a specific primitive where this relaton is used (for computimg some intersection). Commit pending.

maranget:
Aslo notice that `asl_po` (another name for `partial_po` also appears in the acyclicity  test
```
acyclic (aarch64_intrinsic | asl_po) as asl_determinism
```
And here, `asl_po` not being transitive does not change the test result.

lib/innerRel.mli
```diff
@@ -32,6 +32,7 @@ module type S =  sig
     val succs : elt0 -> map -> Elts.t
     val add : elt0 -> elt0 -> map -> map
     val subrel : map -> map -> bool
+    val mem_transitive : (elt0 * elt0) -> map -> bool
```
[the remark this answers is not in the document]
maranget: Good point, perhaps we can also change the name of this function. What about "exists_path" ?

maranget:
Thanks  @HadrienRenaud, any remark remaining?

maranget:
I merged before seeing your message... Some comments.

> * Is there a way to document primitives in cat? Here `partial_po` and `inter_transitive`?
No idea, A  much cleaner way to integrate "implicitly transitive" relations would be an additional kind of values for the interpreter, some thing like `Transitive if E.EventRel`. Then an ordinary intersection could be computed by using `exists_path` in the case when one of the argument is an implicitly   transitive.

> * Should the interpreter interprete `a+ ∩ b` as `inter_transitive (a, b)` depending on some heuristics?
I guess no.

> * Should the definition of `partial_po` not be arch dependant? For the moment it is _implicitely_ transitive only for aarch64+asl, and not for ASL nor any other arch.
If the ASL arch is using `asl.cat`, I  guess it should also be given the implicitly transitive `partial_po` relation.

PR #747, opened 2023-12-28
[ASL] Fix implementation of underconstrained integers

HadrienRenaud:
This PR fixes some unsoundness in the type system: previously two expressions with under-constrained domains would have the same domain. This makes the type-system less permissive.
For example in the following test the value `3` has the type `integer {6}`:
```
func foo {N} (x: bits(N), i: integer {N})
begin
  assert Len(x) == i;
end

func bar {M} (x: bits(M))
begin
  foo (x, 3); // 3 is assigned to integer {M} !!!
end

func main () => integer
begin
  bar ('101010'); // calls with M = 6
  return 0;
end
```

This PR:
 - [x] adds internal types for implicitly typed function parameters;
 - [x] changes the domains of under-constrained integers to a purely syntactic one: `N` and `M` have different domains because `"N" != "M"`;
 - [x] the domains of well-constrained integers is changed in accordance, and default to syntactic domains: the domain of `integer {N + 1}` is different from `integer {5}` because `"N + 1" != "5"`;
 - [x] add some tests for those features, that before this PR exhibited values out of their domains;
 - [x] Store variable name with `uid` in `UnderConstrained` types
 - [x] Add function that does the `Domain.int_set` inclusion and add todo.

### DIFF
--- a/asllib/AST.mli
+++ b/asllib/AST.mli
@@ -295,13 +295,13 @@ type subprogram_type =
   | ST_Getter
   | ST_Setter
 
-type 'p subprogram_body = SB_ASL of stmt | SB_Primitive of 'p
+type subprogram_body = SB_ASL of stmt | SB_Primitive
 
-type 'p func = {
+type func = {
   name : identifier;
   parameters : (identifier * ty option) list;
   args : typed_identifier list;
-  body : 'p subprogram_body;
+  body : subprogram_body;
   return_type : ty option;
   subprogram_type : subprogram_type;
 }
@@ -324,14 +324,14 @@ type global_decl = {
 (** Global declaration type *)
 
 (** Declarations, ie. top level statement in a asl file. *)
-type 'p decl_desc =
-  | D_Func of 'p func
+type decl_desc =
+  | D_Func of func
   | D_GlobalStorage of global_decl
   | D_TypeDecl of identifier * ty * (identifier * field list) option
 
-type 'p decl = 'p decl_desc annotated
+type decl = decl_desc annotated
 
-type 'p t = 'p decl list
+type t = decl list
 (** Main AST type. *)
 
 (* -------------------------------------------------------------------------
--- a/asllib/ASTUtils.ml
+++ b/asllib/ASTUtils.ml
@@ -556,8 +556,8 @@ let rec subst_expr substs e =
   | E_Unknown _ -> e.desc
   | E_Unop (op, e) -> E_Unop (op, tr e)
 
-let dag_fold (def : 'p AST.decl -> identifier) (use : 'p AST.decl -> ISet.t)
-    (folder : 'p AST.decl -> 'a -> 'a) (ast : 'p AST.t) : 'a -> 'a =
+let dag_fold (def : AST.decl -> identifier) (use : AST.decl -> ISet.t)
+    (folder : AST.decl -> 'a -> 'a) (ast : AST.t) : 'a -> 'a =
   let def_use_map =
     List.fold_left
       (fun def_use_map d ->
@@ -600,35 +600,6 @@ let scope_compare s1 s2 =
   | Scope_Local (n1, i1), Scope_Local (n2, i2) ->
       let n = Int.compare i1 i2 in
       if n != 0 then n else String.compare n1 n2
-
-let no_primitive (ast : 'p t) : 'q t =
-  let one d =
-    let here = add_pos_from d in
-    match d.desc with
-    | D_GlobalStorage g -> D_GlobalStorage g |> here
-    | D_TypeDecl (a, b, c) -> D_TypeDecl (a, b, c) |> here
-    | D_Func { body = SB_Primitive _; _ } -> assert false
-    | D_Func
-        {
-          body = SB_ASL s;
-          args;
-          name;
-          return_type;
-          subprogram_type;
-          parameters;
-        } ->
-        D_Func
-          {
-            body = SB_ASL s;
-            args;
-            name;
-            return_type;
-            subprogram_type;
-            parameters;
-          }
-        |> here
-  in
-  List.map one ast
 
 let rec is_simple_expr e =
   match e.desc with
@@ -743,7 +714,7 @@ let rename_locals map_name ast =
     | LDI_Tuple (ldis, t) ->
         LDI_Tuple (List.map map_ldi ldis, Option.map map_t t)
   and map_body = function
-    | SB_Primitive _ as b -> b
+    | SB_Primitive as b -> b
     | SB_ASL s -> SB_ASL (map_s s)
   and map_func f =
     let map_args li = List.map (fun (name, t) -> (map_name name, map_t t)) li in
--- a/asllib/Interpreter.ml
+++ b/asllib/Interpreter.ml
@@ -38,9 +38,9 @@ let _dbg = false
 module type S = sig
   module B : Backend.S
 
-  val run_env : (AST.identifier * B.value) list -> B.ast -> B.value B.m
-  val run : B.ast -> B.value B.m
-  val run_typed : B.ast -> StaticEnv.env -> B.value B.m
+  val run_env : (AST.identifier * B.value) list -> AST.t -> B.value B.m
+  val run : AST.t -> B.value B.m
+  val run_typed : AST.t -> StaticEnv.env -> B.value B.m
 end
 
 module type Config = sig
@@ -58,7 +58,6 @@ module Make (B : Backend.S) (C : Config) = struct
 
   module EnvConf = struct
     type v = B.value
-    type primitive = B.primitive
 
     let unroll = C.unroll
   end
@@ -227,7 +226,7 @@ module Make (B : Backend.S) (C : Config) = struct
   (** [build_genv static_env ast primitives] is the global environment before
       the start of the evaluation of [ast]. *)
   let build_genv env0 eval_expr base_value (static_env : StaticEnv.env)
-      (ast : B.primitive AST.t) =
+      (ast : AST.t) =
     let funcs = IMap.empty |> build_funcs ast in
     let () =
       if _dbg then
@@ -268,6 +267,21 @@ module Make (B : Backend.S) (C : Config) = struct
       | Throwing (v, g) -> Throwing (v, g) |> return |> Fun.const)
 
   let ( let*^ ) = bind_env
+
+  (* Primitives handling *)
+  (* ------------------- *)
+  let primitive_runtimes =
+    List.to_seq B.primitives
+    |> Seq.map
+         AST.(fun ({ name; subprogram_type = _; _ }, f) -> (name, f))
+    |> Hashtbl.of_seq
+
+  let primitive_decls =
+    List.map (fun (f, _) -> D_Func f |> add_dummy_pos) B.primitives
+
+  let () =
+    if false then
+      Format.eprintf "@[<v 2>Primitives:@ %a@]@." PP.pp_t primitive_decls
 
   (*****************************************************************************)
   (*                                                                           *)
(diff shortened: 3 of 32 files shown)


asllib/AST.mli
```diff
@@ -33,6 +33,12 @@
 type position = Lexing.position
 type 'a annotated = { desc : 'a; pos_start : position; pos_end : position }
 
+type identifier = string
+(** Type of local identifiers in the AST. *)
+
+type uid = int
+(** Unique identifiers *)
```
Roman-Manevich, line 40: (** Identifiers for uniquely naming function parameters *)

asllib/ASTUtils.ml
```diff
@@ -506,12 +512,24 @@ exception FailedConstraintOp
 
 let constraint_binop op =
   let do_op c1 c2 =
-    match (c1, c2) with
-    | Constraint_Exact e1, Constraint_Exact e2 ->
+    match (c1, c2, op) with
```
Roman-Manevich, line 515: nit: you can first match the a MINUS op and use it to reverse a c2 range constraint. Then, you only have to handle the PLUS case.

asllib/Interpreter.ml
```diff
@@ -38,9 +38,9 @@ let _dbg = false
 module type S = sig
   module B : Backend.S
 
-  val run_env : (AST.identifier * B.value) list -> B.ast -> B.value B.m
-  val run : B.ast -> B.value B.m
-  val run_typed : B.ast -> StaticEnv.env -> B.value B.m
+  val run_env : (AST.identifier * B.value) list -> AST.t -> B.value B.m
```
Roman-Manevich, line 41: Out of curiosity: why did you move from B.ast to AST.t?
I understand you're dropping the primitive types, but why?

asllib/Interpreter.ml
```diff
@@ -518,11 +531,12 @@ module Make (B : Backend.S) (C : Config) = struct
     let or' prev here = prev >>= B.binop BOR here in
     let rec in_values v ty =
       match (Types.get_structure (IEnv.to_static env) ty).desc with
-      | T_Real | T_Bool | T_Enum _ | T_String | T_Int None -> m_true
+      | T_Real | T_Bool | T_Enum _ | T_String | T_Int UnConstrained -> m_true
+      | T_Int (UnderConstrained _) -> fatal_from loc Error.TypeInferenceNeeded
```
Roman-Manevich, line 535: What does this error mean in this context?
I guess it's hard for me to follow the intention of the code. Some comments wouldn't hurt.

asllib/Native.ml
```diff
@@ -226,226 +225,141 @@ module NativeBackend = struct
...
+      | li -> Error.fatal_unknown_pos @@ Error.BadArity (name, 1, List.length li)
+
+    let round_down = wrap_real_to_int "RoundDown" floor
+    let round_up = wrap_real_to_int "RoundUp" ceiling
+    let round_towards_zero = wrap_real_to_int "RoundTowardsZero" truncate
+
+    let primitives =
+      let t_bits e = T_Bits (e, []) |> add_dummy_pos in
+      let e_var x = E_Var x |> add_dummy_pos in
+      let p ?(parameters = []) ~args ?returns name f =
```
Roman-Manevich, line 321: 'p' is not an informative name for a function.
How about make_primitive?

asllib/AST.mli
```diff
@@ -185,8 +188,16 @@ and int_constraint =
   | Constraint_Range of (expr * expr)
       (** In the range of these two statically evaluable values.*)
 
-and int_constraints = int_constraint list
-(** The int_constraints represent the union of the individual constraints.*)
+(** The int_constraints constraints the integer type to a certain subset.*)
+and int_constraints =
+  | UnConstrained  (** The normal, unconstrained, integer type. *)
```
Roman-Manevich, line 193: nit: unconstrained is a single word so perpahps use Unconstrained as a constructor?

asllib/PP.ml
```diff
@@ -159,9 +159,10 @@ and pp_slice_list f = pp_comma_list pp_slice f
 
 and pp_ty f t =
   match t.desc with
-  | T_Int None -> pp_print_string f "integer"
-  | T_Int (Some int_constraint) ->
-      fprintf f "@[integer {%a}@]" pp_int_constraints int_constraint
+  | T_Int UnConstrained -> pp_print_string f "integer"
+  | T_Int (WellConstrained cs) ->
+      fprintf f "@[integer {%a}@]" pp_int_constraints cs
+  | T_Int (UnderConstrained _) -> fprintf f "an under-constrained integer"
```
Roman-Manevich, line 165: Wouldn't it be helpful to know the parameter's name (especially if there's most than one)?

asllib/Typing.ml
```diff
@@ -476,7 +487,7 @@ module Annotate (C : ANNOTATE_CONFIG) = struct
     match (Types.get_structure env t).desc with T_Bits _ -> true | _ -> false
 
   let t_bool = T_Bool |> __POS_OF__ |> add_pos_from_pos_of
-  let t_int = T_Int None |> __POS_OF__ |> add_pos_from_pos_of
+  let t_int = T_Int UnConstrained |> __POS_OF__ |> add_pos_from_pos_of
```
Roman-Manevich, line 490: T_Int UnConstrained => integer'

asllib/Typing.ml
```diff
@@ -578,42 +592,44 @@ module Annotate (C : ANNOTATE_CONFIG) = struct
...
+                      check_true' (constraints_is_strict_positive cs2)
                   | SHL | SHR ->
                       (* assert cs2 non-negative *)
-                      check_true' (constraints_is_non_negative cs2) ()
-                  | _ -> ()
+                      check_true' (constraints_is_non_negative cs2)
+                  | _ -> fun () -> ()
+                in
+                let cs =
+                  best_effort UnConstrained (fun _ ->
```
Roman-Manevich, line 629: If I understand correctly, UnConstrained here is just a dummy value. If so, it is confusing since upon a first read it looks like you're creating an unconstrained integer from a bunch of constraints. Can you perhaps replace it with unit to make it less confusing?

asllib/Typing.ml
```diff
@@ -642,38 +659,28 @@ module Annotate (C : ANNOTATE_CONFIG) = struct
...
-    *)
-    (* TODO: incomplete. *)
   (* Begin CanAssignTo *)
-  let can_assign_to env s t =
-    let s_struct = Types.get_structure env s
-    and t_struct = Types.get_structure env t in
-    match (s_struct.desc, t_struct.desc) with
-    | T_Int (Some []), T_Int (Some []) -> false
-    | _ -> Types.type_satisfies env t s |: TypingRule.CanAssignTo
+  let can_assign_to env s t = Types.type_satisfies env t s |: TypingRule.CanAssignTo
```
Roman-Manevich, line 683: Nice. Perhaps we can remove the whole notion of can-assign-to and the related rules?

asllib/Typing.ml
```diff
@@ -1740,28 +1753,31 @@ module Annotate (C : ANNOTATE_CONFIG) = struct
     (* Begin SFor *)
     | S_For (id, e1, dir, e2, s') ->
         let t1, e1' = annotate_expr env e1 and t2, e2' = annotate_expr env e2 in
-        let+ () = check_structure_integer s' env t1 in
-        let+ () = check_structure_integer s' env t2 in
+        let struct1 = Types.get_structure env t1 |> Types.to_well_constrained
```
Roman-Manevich, line 1756: Since you use the "Types.get_structure env t1 |> Types.to_well_constrained" combination more than a few times, perhaps add a new function to reuse? Perhaps something like "get_well_constrained_structure"?
Actually, would it work if Typing.get_structure already used to_well_constrained internally?

asllib/Typing.ml
```diff
@@ -1945,19 +1964,23 @@ module Annotate (C : ANNOTATE_CONFIG) = struct
       List.fold_left one_arg env3 f.args
     in
     (* Resolve dependently typed identifiers in the result type. *)
+    let env5 = env4 in
+    (*
```
Roman-Manevich, line 1968: How come you dropped this environment update?

asllib/types.ml
```diff
@@ -327,22 +377,27 @@ module Domain = struct
 
   let compare _d1 _d2 = assert false
 
-  let int_set_is_subset is1 is2 =
+  let syntax_is_subset env is1 is2 =
+    (* TODO: improve for basic cases. *)
```
Roman-Manevich, line 381: perhaps at least check if the sets of constraints are equal instead of list equality?

asllib/types.ml
```diff
@@ -327,22 +377,27 @@ module Domain = struct
...
+    constraints_equal env is1 is2
```
Roman-Manevich, line 382: perhaps at least check if the sets of constraints are equal instead of list equality?

asllib/types.ml
```diff
@@ -774,9 +839,11 @@ let rec base_value loc env t =
       let e = normalize env e in
       E_Call ("Zeros", [ e ], []) |> add_pos_from t
   | T_Enum li -> IMap.find (List.hd li) env.global.constants_values |> lit
-  | T_Int None | T_Int (Some []) -> !$0
-  | T_Int (Some (Constraint_Exact e :: _))
-  | T_Int (Some (Constraint_Range (e, _) :: _)) ->
+  | T_Int UnConstrained | T_Int (UnderConstrained _) -> !$0
```
Roman-Manevich, line 842: How come the base value of an underconstrained integer is 0? I thought the LRM says there is no base value for underconstrained integers.

asllib/types.mli
```diff
@@ -75,6 +75,13 @@ val get_structure : env -> ty -> ty
 (** The structure of a type is the primitive type that can hold the same
     values. *)
 
+val under_constrained_ty : identifier -> ty
+(** From a declared variable, builds an under-constrained integer. *)
+
+val to_well_constrained : ty -> ty
+(** Transform an under-constrained type into a well-constrained integer equal
```
Roman-Manevich, line 82: ... and leave unconstrained integers as are.

herd/ASLSem.ml
```diff
@@ -550,131 +526,132 @@ module Make (C : Config) = struct
...
+        f )
+
+    let p0 name f =
+      let f = function
+        | [] -> f ()
+        | _ :: _ -> Warn.fatal "Arity error for function %s." name
+      in
+      build_primitive name f
+
+    let p0r name ~returns f =
```
Roman-Manevich, line 547: Consider adding a single documentition line for each of the pXY functions.

asllib/Native.ml
```diff
@@ -226,226 +225,141 @@ module NativeBackend = struct
...
-    | [ v ] -> mismatch_type v [ default_t_bits ]
-    | li -> Error.fatal_unknown_pos @@ Error.BadArity ("SInt", 1, List.length li)
+  module Primitives = struct
+    let return_one v = return [ return v ]
+
+    let uint = function
+      | [ NV_Literal (L_BitVector bv) ] ->
+          L_Int (Bitvector.to_z_unsigned bv) |> nv_literal |> return_one
+      | [ v ] -> mismatch_type v [ integer' ]
+      | li ->
```
Roman-Manevich, line 236: We have a lot of changes that are due to restructuring of lines that essentially amount to whitespacing.
Perhaps we should apply ocamlformat to all our files?
That would ensure a consistent standard and avoid all those pesky differences.

asllib/Typing.ml
Roman-Manevich: Okay, let's go with `get_well_constrained_structure` then.

asllib/types.ml
Roman-Manevich: An underconstrained integer has only one value it can take - the one given by the dynamic environment. So, either add a dynamic environment as another parameter or raise an exception.
In any case, we need to define the intended semantics.

asllib/Typing.ml
Roman-Manevich: So, why not remove the commented out code?

Roman-Manevich (approved):
Looks good to me.

Issue #20, opened 2019-08-14
Soundness issue with an AArch64 litmus test


michaliskok:
Hello,

I have the following variation of message passing, where the first read of the second thread is performed `N` times (see [attached](https://github.com/herd/herdtools7/files/3503243/mp-original.txt)).

While I would expect it to have `N + 1` executions, herd only reports 3, for different values of `N`. Is this a bug or there is something wrong with the test case?

Thank you,
Michalis

maranget:
This is a herd bug. Thanks a lot for reporting it. The bug is now corrected in the master branch,development version, version 7.54+02(dev).

 Notice that the bug went unnoticed because most of our tests, if not all of them, use the CBZ or CBNZ instruction in place of the CMP / BC sequence.

Thanks again,

--Luc

PR #763, opened 2024-01-23
[herd] Implement Aarch64 instruction stz2g

maranget:
Adding the instruction revealed some shortcomings of the current MTE implementation. In particular, it is no longer realistic to assume that a complete array will never span over more than one granule, as the STZ2G instruction sets two granules.

The implementation is correct only in mixed-size mode, it is then an error to execute STZG  (resp. STZ2G)  on a memory location whose size is strictly less than one (resp. two) granules.

For instance, with the current granule size of 16 bytes, **herd** will fail on the following test:
```
AAArch64 TooSmall
Variant=mixed,mte
{
int x=1;
0:X1=x:green;
0:X2=x:red;
}
  P0           ;
 STZG X2,[X1]  ;
forall x=0
```
We have:
```
% herd7 TooSmall.litmus
Warning: File "TooSmall.litmus": Unaligned or out-of-bound access: x, 16 bytes (User error)
```

To reach granule size one has to declare an array:
```
AArch64 LargeEnough
Variant=mixed,mte
{
int64_t x[3] = {1,2,3};
0:X1=x:green;
0:X2=x:red;
}
  P0           ;
 STZG X2,[X1]  ;
forall x={0,0,3}
``` 
Now, the test is accepted,  yielding the final values of zero for `x[0]` and `x[1]`.

For backward compatibility `STZG` still operates in non-mixed mode, with the expected result of zeroing non-array locations, which is correct assuming that:
1. the size of non-array locations is less than one granule, and
2. different locations occupy different granules.

Finally, **herd** will fail, flagging an error, on tests that both have some array location and some S`TZG` instruction, regardless of whether `STZG` touches some array or not.

### DIFF
--- a/herd/eventsMonad.ml
+++ b/herd/eventsMonad.ml
@@ -1378,7 +1378,7 @@ Monad type:
       let is_tagloc a =
         let open Constant in
         match a with
-        | V.Val (Symbolic (System (TAG,_))) -> true
+        | V.Val (Symbolic (TagAddr _)) -> true
         | _ -> false
 
       let is_pteloc a =
@@ -1392,6 +1392,19 @@ Monad type:
         | V.Val (Constant.Label _) -> true
         | _ -> false
 
+(*
+ * Add init writes for tag addresses.
+ * A symbolic location has its own tag address, hence
+ * for a given x, tag(x) is initialised, except for arrays
+ * that may span over several granules.
+ * The task of initialising the appropriate tag addresses is
+ * facilitated by array init being implemented as the explicit
+ * initialisations of its elements.
+ * For instance for int t[5], we have int-sized initial writes for
+ * t+0, t+4, t+8, t+12, t+16, yielding the two tag addresses
+ * tag(t) and t(t+16). Namely, with a granule size of 4 int's
+ * the first four items have the same tag address tag(x).
+ *)
       let add_inittags env =
         let glob,tag =
           List.fold_left
@@ -1407,22 +1420,22 @@ Monad type:
             | A.Location_reg _ -> p)
             ([],[]) env in
         let tag_set = V.ValueSet.of_list tag in
+        let glob_set = V.ValueSet.of_list glob in
+        let glob_set =
+          V.ValueSet.filter
+            (fun a ->  not (is_pteloc a || is_instrloc a))
+            glob_set in
+        let s = V.ValueSet.map (fun a -> V.op1 Op.TagLoc a) glob_set in
         let env =
-          List.fold_left
-            (fun env a ->
-              if not (is_pteloc a) && not (is_instrloc a) then
-              begin
-                let atag =  V.op1 Op.TagLoc a in
-                if V.ValueSet.mem atag tag_set then env
-                else begin
-                  if dbg then
-                    eprintf "Tag %s for %s defaulting\n"
-                      (V.pp_v atag) (V.pp_v a) ;
-                  (A.Location_global atag,V.Val (Constant.default_tag))::env
-                end
-              end
-              else env)
-            env glob in
+          V.ValueSet.fold
+            (fun atag env ->
+              if V.ValueSet.mem atag tag_set then env
+              else begin
+                if dbg then
+                  eprintf "Tag %s defaulting\n" (V.pp_v atag) ;
+                (A.Location_global atag,V.Val (Constant.default_tag))::env
+             end)
+            s env in
         env
 
       let morello_init_tag s v eiid =
@@ -1482,10 +1495,10 @@ Monad type:
             | A.Location_global (V.Val (Symbolic (System (PTE,s)))) ->
                 let v = expand_pteval loc v in
                 (loc,v)::env,(virt,StringSet.add s pte)
-            | A.Location_global (V.Val (Symbolic (System (TAG,s)))) ->
-              let s = V.pp_v (V.Val (Symbolic (Physical (s,0)))) in
-              let loc = A.Location_global (V.Val (Symbolic (System (TAG,s)))) in
-              (loc,v)::env,maps
+            | A.Location_global (V.Val (Symbolic (TagAddr (VIR,s,o)))) ->
+               let loc =
+                  A.Location_global (V.Val (Symbolic (TagAddr (PHY,s,o)))) in
+               (loc,v)::env,maps
             | A.Location_global (V.Val (Symbolic (Physical _|Virtual _))) ->
                 Warn.user_error "herd cannot handle initialisation of '%s'"
                   (A.pp_location loc)
--- a/herd/machAction.ml
+++ b/herd/machAction.ml
@@ -96,12 +96,12 @@ end = struct
   let access_of_constant cst =
     let open Constant in
     match cst with
-    | Symbolic (Virtual _) -> VIR
-    | Symbolic (Physical _) -> PHY
+    | Symbolic (Virtual _) -> Access.VIR
+    | Symbolic (Physical _) -> Access.PHY
+    | Symbolic (TagAddr _) -> Access.TAG
     | Symbolic (System ((PTE|PTE2),_)) -> Access.PTE
     | Symbolic (System (TLB,_)) -> Access.TLB
-    | Symbolic (System (TAG,_)) -> Access.TAG
-    | Label _ -> VIR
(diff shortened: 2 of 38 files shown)


maranget:
Hi @relokin, this PR much elaborates from your work on MTE (In particular PR #695). If you have some time...

herd/tests/instructions/AArch64.MTE/L00.litmus
```diff
@@ -0,0 +1,14 @@
...
+{
+int x[6]={1,2,3,4,5,6};
+0:X1=x;
+0:X2=x:red;
+}
+ P0              ;
+ STZG  X2,[X1]   ;
+ MOV W0,#7       ;
+ STR W0,[X2]     ;
+ STR W0,[X1,#20] ;
```
relokin, line 12: It's a little weird that we allow an unchecked STR to `x` when previously we had a checked STR `x`. It makes sense to allow unchecked locations and checked locations but mixed uses of the same location is not very realistic. But this might not be in the scope for this PR.

relokin: Looking at L02 did you mean this to be:
```suggestion
 STR W0,[X2,#20] ; 
```

herd/tests/instructions/AArch64.MTE/L03.litmus
```diff
@@ -0,0 +1,15 @@
+AArch64 L03
+(* Non-mixed test: STGZ store zero in x,
```
relokin, line 2:
```suggestion
(* Non-mixed test: STZG store zero in x,
```

herd/tests/instructions/AArch64.MTE/L05.litmus
```diff
@@ -0,0 +1,16 @@
+AArch64 L05
```
relokin, line 1: It would be really good if herd7 could ignore the size of STZG for L03 but throw an error for L05. But I am not sure this is possible. So I guess you're suggesting that STZG should require mixed mode don't you?

herd/eventsMonad.ml
```diff
@@ -1392,6 +1392,19 @@ Monad type:
         | V.Val (Constant.Label _) -> true
         | _ -> false
 
+(*
+ * Add init writes for tag addresses.
+ * A symbolic location has its own tag addresse, hence
```
relokin, line 1397:
```suggestion
 * A symbolic location has its own tag address, hence
```

relokin (commented):
This is very good thanks Luc! I've done a quick review added a couple of comments and I will try and test it tomorrow as well.

herd/tests/instructions/AArch64.MTE/L00.litmus
maranget: I do not understand your remarks on "unchecked" STR. Which STR are checked or unchecked ?

As to `STR W0,[X1,#20]`, this is what I intended. As I understand it, both `STR` succeeds the first one `STR W0,[X2]` stores into the granule whose color has changed to `:red`, while the second `STR W0,[X1,#20]` sores into the next granule, whose color has not changed and hence stayed `:green`. Assuming a granule size of 16 bytes.

herd/tests/instructions/AArch64.MTE/L03.litmus
maranget: Thanks @relokin, nice catch.

herd/tests/instructions/AArch64.MTE/L05.litmus
maranget: You mean throw an error when the target location of `STZG` is an array in non-mixed mode?

herd/eventsMonad.ml
maranget: Thanks

herd/tests/instructions/AArch64.MTE/L00.litmus
relokin: Ah sorry, do we assume a default tag (color) for symbolic addresses as well? Meaning x is assumed to be x:green? I was under the impression we only had default values for tag locations (tag(x)).

MTE doesn’t require that all memory instructions perform a tag check. There are cases where the instruction doesn’t perform a tag check (for example stg doesn't check or a load y where y is mapped as Normal Memory and not TaggedNormal doesn’t check either). So I thought, so without checking the code, that if a symbolic address didn’t specify a colour it would signal to herd to slide the tag check.

maranget: As I understand, all locations have a tag and this tag is green by default. In other words, all locations have the TaggedNormal attribute. Do we understand each other?

relokin: OK, I think I understand now. In `-variant mte`:
* If tag(x) is not explicitly initialised then,  `herd7` will initialise to `:green"
* If a register is initialised with a symbolic location which doesn't specify a tag (for example `0:X0=x`), `herd7` will assume that a default tag `:green` for this symbolic address (i.e. `0:X0=x:green`).
* All locations are assumed to be mapped as `TaggedNormal`
and now I also understand why `L03` is correctly specified.

herd/tests/instructions/AArch64.MTE/L05.litmus
relokin: I think so yes. I think it would be useful to allow herd7 to work without enabling `-variant mixed` in cases like `L03` after all, `L03` is not incorrect (the user didn't check `x[1]` they're only checking `x` aka `x[0]`). But `L05` is wrong so they can't trust the result and they should enable `-variant mixed`. What do you think? Is it possible?

herd/machAction.ml
```diff
@@ -96,12 +96,12 @@ end = struct
   let access_of_constant cst =
     let open Constant in
     match cst with
-    | Symbolic (Virtual _) -> VIR
-    | Symbolic (Physical _) -> PHY
+    | Symbolic (Virtual _) -> Access.VIR
+    | Symbolic (Physical _) -> Access.PHY
+    | Symbolic (TagAddr _) -> Access.TAG
```
relokin, line 101: I don't fully understand why we have the Access field/type. But now that a TagAddr can be Virtual or Physical can we also get rid of Access.TAG?

herd/tests/instructions/AArch64.MTE/L05.litmus
maranget: Your solution implies looking at several places in the execution: (1) apply `STZG` to an array location and (2) reading some array cell different from the first.

I'd favor a solution that would look at the `STZG` write only. Then if we can check that STZG base address is an array address, we can do this only when the address is knows (ie  is read from initial state)
1. Fail with a message that suggests using `-variant mixed`, or
2. Fill all the array cells with zero, (up to granule size ?) 

Also, if the address is not known 1. also apply.

A more simple but less expressive solution would be to fail, still suggesting `-variant mixed` when we both have some array and `STZG` instruction in the same test.

herd/machAction.ml
maranget: The Access tag is used to restrict rf matching when addresses are not known. Matching of writes and reads is performed for every compatible writes and read, compatibility being defined as having the same access field (more or less) : https://github.com/herd/herdtools7/blob/0b1fb5931d417f1fb326ff7f9adba665d5677947/herd/mem.ml#L854

relokin: Ok and since we don't want to mix accesses by LDR/STR and other regular memory instructions with LDG/STG and implicit tag reads we need this. I see thanks!

herd/tests/instructions/AArch64.MTE/L05.litmus
relokin: > A more simple but less expressive solution would be to fail, still suggesting -variant mixed when we both have some array and STZG instruction in the same test.

I think that's a very good middle ground. But in any case don't worry too much if it's too hard we can fix in a subsequent PR.

maranget: I have finally implemented this middle ground, as it is relatively straightforward.

maranget:
> LGTM. Very nice change, looking forward to see it merged!

@relokin, merging on its way. Thanks for your review.

PR #777, opened 2024-02-02
[all] Split precision into MTE precision and Fault handling

relokin:
Prior to this patch, the tools were using a unified concept of Precision which was used to decide how to act on an MTE tag mismatch (raise an exception or set the TFSR_ELx) and how to handle a fault (exception). This patch splits the precision into two different arguments, the MTE precision and the fault handling mode.

### DIFF
--- a/herd/tests/diycross/AArch64.MTE/MTE.cfg
+++ b/herd/tests/diycross/AArch64.MTE/MTE.cfg
@@ -1,4 +1,4 @@
-variant memtag
+variant memtag,async
 #All tests should be allowed when model is relaxed.
 skipchecks external
 graph cluster
@@ -23,4 +23,3 @@ unshow dd,intrinsic
 labelbox true
 doshow fr,co,rmw
 unshow dmb.sy,dmb.st,dmb.ld,ca,dsb.sy,isb
-
--- a/lib/fault.ml
+++ b/lib/fault.ml
@@ -169,3 +169,48 @@ module Make(A:I) =
             | r -> r
         end)
   end
+
+module Handling : sig
+  type t =
+    | Handled
+    | Fatal
+    | Skip
+
+  val default : t
+  val tags : string list
+  val parse : string -> t option
+  val pp : t -> string
+
+  val is_fatal : t -> bool
+  val is_skip : t -> bool
+end
+=
+struct
+  type t =
+    | Handled
+    | Fatal
+    | Skip
+
+  let default = Fatal
+
+  let tags =  ["handled"; "fatal"; "faultToNext"; ]
+
+  let parse = function
+    | "imprecise"|"handled" -> Some Handled
+    | "precise"|"fatal" -> Some Fatal
+    | "faulttonext"|"skip" -> Some Skip
+    | _ -> None
+
+  let pp = function
+    | Handled -> "handled"
+    | Fatal -> "fatal"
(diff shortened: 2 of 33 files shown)


relokin:
Hey @maranget this is a change we've discussed in the past, if you have some time to have a look it would be great!

herd/tests/diycross/AArch64.MTE/MTE.cfg
```diff
@@ -1,4 +1,4 @@
-variant memtag
+variant memtag,async,handled
```
maranget, line 1: I am wondering with `async` as MTE mismatch control, wouldn't the fault specification be irrelevant?

maranget:
I found no test for `asymmetric`, what about this one?
```
AArch64 B007
Variant=mte,asym,fatal
{
 0:X1=x:red;
 1:X1=x:red;
}
 P0          |  P1         ;
 LDR W0,[X1] | MOV W0,#2   ;
 MOV W2,#1   | STR W0,[X1] ;
             | MOV W2,#1   ;
forall(0:X2=0 /\ 1:X2=1)
```

herd/tests/diycross/AArch64.MTE/MTE.cfg
relokin: That's a good point, maybe it doesn't make sense. I am happy to remove handled.

However, maybe it's worth pointing out that this PR changes the default behaviours. Previously, the default precision was handled. This meant asynchronous MTE (no exceptions on tag mismatches) and handled (retry the instruction) for MMU faults. I've changed the defaults to synchronous MTE and fatal for MMU faults, as I think these defaults are more frequently used.

The primary reason for above change was to ensure that the previously default behaviour is retained.

maranget: > I've changed the defaults to synchronous MTE and fatal for MMU faults, as I think these defaults are more frequently used.

I agree that the new defaults are better. My local test suite needs a few adaptations. But again I agree.

lib/fault.ml
```diff
@@ -169,3 +169,48 @@ module Make(A:I) =
...
+  val is_skip : t -> bool
+end
+=
+struct
+  type t =
+    | Handled
+    | Fatal
+    | Skip
+
+  let default = Handled
```
maranget, line 194: It looks that the default is still handled., while, according to our conversation, it is intended to change to fatal. Is this an oversight?

maranget:
I like this PR, the situation is much better now, with a clean separation of tag mismatch and fault handling.

lib/fault.ml
maranget: If the change of default impacts the regression test suite, we can change reference files. I do not expect much in the outer world to depend on such a cryptic default behaviour.

Looking back, this default behaviour was adopted because this this is the default behaviour for VM faults observed on top of kvm-unit-tests. I guess this is a rather standard hardware behaviour. However I have been surprised in several occasion and, in the context of our experiments, `fatal` looks a better choice.

relokin:
I've added one more commit with the fixes just to make reviewing easier, but if you're happy with it, I will squash it and then if we're happy with PR overall, we can merge.

maranget:
I am happy, please proceed.

maranget:
Merged, thanks @relokin.

PR #779, opened 2024-02-06
[all] Pull Request methodology

Roman-Manevich:
I'm hoping that we can agree on a methodology for using git/github to publish new features. Here is my attempt.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,29 @@ When adding a new pre-commit check, please run Pre-Commit manually first:
 
     % pre-commit run --all-files
 
+### Pull Request Methodology
+We would like to, as much as possible, maintain a linear git history, have pull
+requests reviewed, and finally merged responsibly. Therefore, we ask all
+contributors to follow the following procedure:
+1. Create a git branch: 'git checkout -b feature_name'
+2. Implement new feature/change.
+3. Commit with meaningful messages:
+   'git commit -m "[feature] one line title followed by detailed description"'
+4. Test the feature and make sure all relevant tests pass, including
+   'make test'.
+5. Push 'git push'
+6. Rebase often, repeating 2--6 as much as needed:
+  - 'git checkout master'
+  - 'git pull'
+  - 'git checkout feature_name'
+  - 'git rebase'
+  - Resolve all conflicts.
+  - 'git rebase --continue'
+  - 'git push --force'
+7. Create a pull request and ask for reviews.
+8. Address reviewer comments and make sure all issues are resolved.
+9. Once the pull request is accepted, let **the repo admin** merge.
+
 License
 =======
 
--- a/README.md
+++ b/README.md
@@ -55,27 +55,37 @@ When adding a new pre-commit check, please run Pre-Commit manually first:
     % pre-commit run --all-files
 
 ### Pull Request Methodology
-We would like to, as much as possible, maintain a linear git history, have pull
-requests reviewed, and finally merged responsibly. Therefore, we ask all
-contributors to follow the following procedure:
-1. Create a git branch: 'git checkout -b feature_name'
-2. Implement new feature/change.
-3. Commit with meaningful messages:
-   'git commit -m "[feature] one line title followed by detailed description"'
-4. Test the feature and make sure all relevant tests pass, including
-   'make test'.
-5. Push 'git push --set-upstream origin feature_name'
-6. Rebase often, repeating 2--6 as much as needed:
-  - 'git checkout master'
-  - 'git pull'
-  - 'git checkout feature_name'
-  - 'git rebase'
-  - Resolve all conflicts.
-  - 'git rebase --continue'
-  - 'git push --force'
-7. Create a pull request and ask for reviews.
-8. Address reviewer comments and make sure all issues are resolved.
-9. Once the pull request is accepted, let **the repo admin** merge.
+We would like, as much as possible, to maintain a linear git history,
+since it is simple, readable, and easier to bisect in case of bugs.
+We would also like to have pull requests reviewed, and finally merged
+responsibly.
+Therefore, our recommendation to all contributors is to follow the following
+procedure:
+1.  Create a [fork](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/fork-a-repo)
+    of `herdtools7` named, for example, `upstream` and set `herdtools7` as your remote.
+2.  Create a git branch: `git checkout -b feature`
+3.  Implement new feature/change.
+4.  Commit with meaningful messages:
+    `git commit -m "[tag] one line descriptive title, followed by detailed description"`
+    The tag(s) should refer to the component that the feature relates to, e.g.,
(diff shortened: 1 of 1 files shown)


relokin (commented):
Thanks @Roman-Manevich!

This is a very good starting point. This is very similar to the flow I would use but before we recommend that everyone adopts a flow that allows a linear history we need a wider consensus.

README.md
```diff
@@ -54,6 +54,29 @@ When adding a new pre-commit check, please run Pre-Commit manually first:
 
     % pre-commit run --all-files
 
+### Pull Request Methodology
+We would like to, as much as possible, maintain a linear git history, have pull
```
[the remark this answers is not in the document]
maranget: Hi @Roman-Manevich ans @relokin. I think the same, the reason why are:
1. Simplicity
2. Readability
3. Easier bisect search in case of bugs

maranget:
Hi all, thanks for this initiative. I'd like to be clear about what history we have in the end:

A linear history, with PR begin merged one after the other, with explicit merge message, without any overlap. This can be checked as `git lg` with the following lines in one's .gitconf file
```
[alias]
        lg = log --color --graph --pretty=format:'%Cred%h%Creset -%C(yellow)%d%Creset %s %Cgreen(%cr) %C(bold blue)<%an>%Creset' --abbrev-commit
```
We also want the tests (`make test`) to succeed after every commit (again to facilitate bisecting). A relatively light technique to achieve this is to squash the PR commit into one or a few ones (another technique is to run `make all test` after each commit, which is arguably heavier).

Hence, once the PR is ready, that is once merge has been agreed, the PR authors can proceed as follows, from their branch (if I am not wrong):
1. Rebase on top of master with `git rebase master`.
2. Squash commits into one or a few with `git rebase -i`.
3. Force push with `git push --force`.

It is to be noticed that the merger can do the same, but this means pushing to the author's clone and slightly alters authorship.

Finally, the merger has to fill the merge message, it is often convenient to copy the initial message of the conversation, which usually describes the effect and purpose of the PR.

relokin:
Thanks for this @maranget.  I agree with everything just some questions/comments.
 
> We also want the tests (`make test`) to succeed after every commit (again to facilitate bisecting). A relatively light technique to achieve this is to squash the PR commits into one or a few ones (another technique is to run `make all test` after each commit, which is arguably heavier).

I think there is still value in keeping multiple commits in a PR if:
* Each of the commit in the PR passes `make test` as you pointed out, and
* Each commit implements a logical change.
Some PRs implement large features and being able to split into different logical changes, I think helps. In fact, it's way easier for the reviewer or anyone trying to understand a PR, if the reviewing load can be split in reasonably sized meaningful changes. And there is no downside, if one wants to see only merge commits, they can do so with `git lg --merges`. So squashing is not only reducing information.

On the other hand, sometime this is not information worth keeping. If the commit history in the PR is not clean, with fixes on top of each, I would argue that these sequence of changes is not worth keeping in the project's history, they only add noise and squashing is the right strategy.

> It is to be noticed that the merger can do the same, but this means pushing to the author's clone and slightly alters authorship.

I agree it's not ideal, but when I squash commits together, I retain the Author information. git automatically changes the Commit field.

> Finally, the merger has to fill the merge message, it is often convenient to copy the initial message of the conversation, which usually describes the effect and purpose of the PR.

That's a very good point. I will start doing it.

README.md
```diff
@@ -54,6 +54,39 @@ When adding a new pre-commit check, please run Pre-Commit manually first:
...
     % pre-commit run --all-files
 
+### Pull Request Methodology
+We would like, as much as possible, to maintain a linear git history,
+since it is simple, readable, and easier to bisect in case of bugs.
+We would also like to have pull requests reviewed, and finally merged
+responsibly.
+Therefore, our recommendation to all contributors is to follow the following
+procedure:
+1.  Create a [fork](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/fork-a-repo)
```
artkhyzha, line 64: Step 6 as I understand it assumes that the remote pf the fork is called "origin" and does not rely on there being a name for the fork. Step 7 as I understand it assumes that "upstream" is indeed 'herd/herdtools7" on Github.

So, maybe this could be something like "Create a fork and clone it. Add a remote for the original repository, call it upstream"?

README.md
```diff
@@ -54,6 +54,39 @@ When adding a new pre-commit check, please run Pre-Commit manually first:
...
+4.  Commit with meaningful messages:
+    `git commit -m "[tag] one line descriptive title, followed by detailed description"`
+    The tag(s) should refer to the component that the feature relates to, e.g.,
+    `[herd]`, `[litmus]`, etc. Generally, the tags correlate to the sub-folders.
+    Also, `[all]` applies to changes that effect several components.
+5.  Test the feature and make sure all relevant tests pass, including
+    `make test`.
+6.  Assuming `origin` is your clone of `herdtools7`, push with
+    `git push --set-upstream origin feature`
+7.  Rebase often, repeating 2--6 as much as needed:
```
artkhyzha, line 77: Strictly speaking not necessary for the sequence of steps here, but I wonder if it is useful to keep origin's master branch up to date with upstream's master branch. Presumably, step 2 is good to do branching off up-to-date master.

I usually press a button on the web interface for this ("Sync fork" on the forked page).

artkhyzha (commented):
LGTM -- this is more or less what I do too.

I added two nitpick-y comments, to check if I understood the flow correctly than anything else.

README.md
```diff
@@ -54,6 +54,39 @@ When adding a new pre-commit check, please run Pre-Commit manually first:
...
+2.  Create a git branch: `git checkout -b feature`
+3.  Implement new feature/change.
+4.  Commit with meaningful messages:
+    `git commit -m "[tag] one line descriptive title, followed by detailed description"`
+    The tag(s) should refer to the component that the feature relates to, e.g.,
+    `[herd]`, `[litmus]`, etc. Generally, the tags correlate to the sub-folders.
+    Also, `[all]` applies to changes that effect several components.
+5.  Test the feature and make sure all relevant tests pass, including
+    `make test`.
+6.  Assuming `origin` is your clone of `herdtools7`, push with
```
relokin, line 75: As Artem said this is the default behaviour, if you want you can add a comment to this extend when you instruct users to add the upstream remote, but here this is unnecessary.

HadrienRenaud (commented):
Hi @Roman-Manevich, thank you for writing this.

I agree with the comments that most of those steps need more explanation.
I would like to add that external references can (and probably should) be used to provide more detailed and up-to-date explanations. For example, the steps on rebasing are confusing, and we could point to an external tutorial, for example [git-scm](https://git-scm.com/book/en/v2/Git-Branching-Rebasing).

A fear I have is that such a text would deter occasional contributors. For example, while rebasing is good for people that do 5 PR a week (me? 🙄), it might not be the easiest if you don't have a good knowledge of the code base and you only occasionally work on it, in which case a merge commit is probably fine.

PR #780, opened 2024-02-07
[asl][fix] Use anonymous types before static analysis

HadrienRenaud:
A bug reported by @maranget.

The following test did not type-check:

```
type pagros of integer{1,2,4,8,16};

var ones = Ones(64);

func f(sz:pagros) => bits(sz)
begin
  return ones[sz-1:0];
end

func main() => integer
begin
  let x = f(8);
  DEBUG(x);
  return 0;
end
```

The cause was the use of named types inside slices, which were not correctly
handled. This has been fixed.

This PR is not a systematic fix nor a complete survey of correct handling of
named types.

The reported test has been added to the regression test set, and is now
accepted by the type-checker.

### DIFF
--- a/asllib/StaticInterpreter.ml
+++ b/asllib/StaticInterpreter.ml
@@ -508,6 +508,14 @@ module Normalize = struct
            ctnts_and (Conjunction (PMap.singleton p sign)) ctnts)
          li2)
 
+  let rec make_anonymous (env : env) (ty : ty) : ty =
+    match ty.desc with
+    | T_Named x -> (
+        match IMap.find_opt x env.global.declared_types with
+        | Some ty' -> make_anonymous env ty'
+        | None -> fatal_from ty (Error.UndefinedIdentifier x))
+    | _ -> ty
+
   let rec to_ir env (e : expr) : ir_expr =
     match e.desc with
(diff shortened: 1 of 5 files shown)


asllib/StaticInterpreter.ml
```diff
@@ -508,6 +508,14 @@ module Normalize = struct
            ctnts_and (Conjunction (PMap.singleton p sign)) ctnts)
          li2)
 
+  let rec make_anonymous (env : env) (ty : ty) : ty =
+    match ty.desc with
+    | T_Named x -> (
+        match IMap.find_opt x env.global.declared_types with
+        | Some ty' -> make_anonymous env ty'
+        | None -> Error.(fatal_from ty (UndefinedIdentifier x)))
```
maranget, line 516: Avoid a warning as follows
```suggestion
        | None -> fatal_from ty (Error.UndefinedIdentifier x))
```

maranget:
Hi @HadrienRenaud, thanks for the swift fix.

Some file ("named-types-in-slices.asl" ?) is apparently missing, `make test` fails, asking for it.

maranget:
LGTM, tests are fine, the title is cryptic but the intend is clear.
Thanks @HadrienRenaud, merging is on its way.

PR #781, opened 2024-02-09
[asl] Rework local declarations left-hand side

### LINKED ISSUE #766: [ASL] Remove LDUninitialisedTuple
This rule is unsound cc @HadrienRenaud.

HadrienRenaud:
This commit rework the way local declaration are represented to make the type indications a full constructor of the type `local_declaration_item`, and removes them of the other constructors.
For example, `let x: integer = 4;`, the local declaration item was previously `LDI_Var ("x", Some integer)`, and is now `LDI_Typed (LDI_Var "x", integer)`.
This makes both the typing and semantics rules simpler and more explicit.

This commits also brings transliteration with examples of the new abstract syntax, the semantics rules and the typing rules handling local declaration items.

While the concrete syntax stays the same, the AST can now represents strange things, for example:
```
var x: integer : boolean;
```
(note that this example cannot be constructed by the parser).

However, the conflicts that arrise from this new possibilities could also happen with the previous AST, but in more intricate ways, for example:
```
var (x: integer, -): (boolean, boolean);
```

How to solve those conflicts and how to handle them was not explicit nor straight-forward as the AST was complex with a lot of cases. With this new AST, reasonning on the various cases is simplified and solutions are often simple.

Fix #766.

### DIFF
--- a/asllib/Interpreter.ml
+++ b/asllib/Interpreter.ml
@@ -79,9 +79,7 @@ module Make (B : Backend.S) (C : Config) = struct
 
   type expr_eval_type = (B.value * env) maybe_exception m
   type stmt_eval_type = control_flow_state maybe_exception m
-
-  type func_eval_type =
-    (value_read_from list * IEnv.global) maybe_exception m
+  type func_eval_type = (value_read_from list * IEnv.global) maybe_exception m
 
   (*****************************************************************************)
   (*                                                                           *)
@@ -97,9 +95,7 @@ module Make (B : Backend.S) (C : Config) = struct
   (* ------ *)
   let return = B.return
   let return_normal v = Normal v |> return
-
-  let return_continue env : stmt_eval_type =
-    Continuing env |> return_normal
+  let return_continue env : stmt_eval_type = Continuing env |> return_normal
 
   let return_return env vs : stmt_eval_type =
     Returning (vs, env.IEnv.global) |> return_normal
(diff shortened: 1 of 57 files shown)


Roman-Manevich (changes requested):

asllib/Interpreter.ml
```diff
@@ -846,28 +810,31 @@ module Make (B : Backend.S) (C : Config) = struct
   (* Evaluation of Local Declarations *)
   (* -------------------------------- *)
   and eval_local_decl s ldi env m_init_opt : env maybe_exception m =
+    let () =
+      if false then Format.eprintf "Evaluating %a.@." PP.pp_local_decl_item ldi
+    in
     match (ldi, m_init_opt) with
     (* Begin LDDiscard *)
-    | LDI_Discard _ty, _ -> return_normal env |: SemanticsRule.LDDiscard
+    | LDI_Discard, _ -> return_normal env |: SemanticsRule.LDDiscard
```
Roman-Manevich, line 818: nit: can you please add one space between LDI_Discard and the following comma, please?

herd/AArch64ASLSem.ml
Roman-Manevich, line 1: Why is this file even affected?
It looks like the only differences are formatting.

herd/tests/instructions/ASL-pseudo-arch/LB-02.litmus.expected
Roman-Manevich, line 1: Why are the hash values in the litmus files different?

asllib/Typing.ml
```diff
@@ -1595,106 +1513,69 @@ module Annotate (C : ANNOTATE_CONFIG) = struct
...
-    (* Begin LDDiscardNone *)
-    | LDI_Discard None -> (env, ldi) |: TypingRule.LDDiscardNone
+    (* Begin LDDiscard *)
+    | LDI_Discard -> (env, ldi) |: TypingRule.LDDiscard
     (* End *)
-    (* Begin LDDiscardSome *)
-    | LDI_Discard (Some t) ->
+    (* Begin LDTyped *)
+    | LDI_Typed (ldi', t) ->
+        (* check is t is valid *)
```
Roman-Manevich, line 1521: What do you mean by valid for `t` here?

Roman-Manevich (approved):

PR #782, opened 2024-02-09
[herd] Add support for Arm FEAT_NV2

relokin:
When FEAT_NV2 is implemented, system register access that would be trapped to EL1 and EL2 via MRS/MSR instructions is transformed to perform Memory Read/Write effects. The address on the memory access is calculated via the address held in VNCR_EL2 and an offset as defined in:

Table D8-66 Memory address offset associated with transformed register access 

of the Arm ARM DDI 0487J.a


relokin:
Thanks @HadrienRenaud  for your review!

If there are no objections, I would like to merge this before Friday (ping @maranget just to bring this to your notice).

maranget:
LGTM Thanks to alll

relokin:
Thanks Luc!

PR #783, opened 2024-02-12
[asl] Escape string literals

HadrienRenaud:


Roman-Manevich (approved):
Seems thorough enough.

maranget:
Hi @HadrienRenaud. Wouldn't be possible to use `Scanf.unescaped` in place of a custom scan of the lexed string? However notice that lexing strings character by character is still necessary if one accepts `\"` sequences in strings.

maranget:
Fair enough. Adding a reference to ASL lexing convention in a comment is a good idea, A reference to the appopriate documentation may be sufficient.

PR #784, opened 2024-02-12
[asl] Make Print a full statement.

HadrienRenaud:
This simplifies the type-system, as the declaration of `print` was not supported by the current type-system: a very permissive version of it should accept any number of arguments of any type.

### DIFF
--- a/asllib/AST.mli
+++ b/asllib/AST.mli
@@ -272,7 +272,9 @@ type stmt_desc =
           represent the implicit throw, such as [throw;]. *)
   | S_Try of stmt * catcher list * stmt option
       (** The stmt option is the optional otherwise guard. *)
-  | S_Debug of expr
+  | S_Print of { args : expr list; debug : bool }
+      (** A call to print, as an explicit node as it does not require
+          type-checking. *)
 
 and stmt = stmt_desc annotated
 and case_alt = (pattern * stmt) annotated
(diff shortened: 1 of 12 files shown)


asllib/AST.mli
```diff
@@ -272,7 +272,9 @@ type stmt_desc =
           represent the implicit throw, such as [throw;]. *)
   | S_Try of stmt * catcher list * stmt option
       (** The stmt option is the optional otherwise guard. *)
-  | S_Debug of expr
+  | S_Print of { args : expr list; debug : bool }
```
Roman-Manevich, line 275: Putting the constructor arguments in a record is really nice. That way, you can name them and document their intention.
I wish we had that for other constructors like E_Cond, for example.

Roman-Manevich (approved):
Looks good.
Adjusting the Abstract Syntax Refernce will be done in a separate PR.

asllib/Parser.mly
```diff
@@ -463,9 +463,9 @@ let stmt ==
       | RETURN; ~=ioption(expr);                             < S_Return >
       | x=IDENTIFIER; args=plist(expr); ~=nargs;             < S_Call   >
       | ASSERT; e=expr;                                      < S_Assert >
-      | DEBUG; e=expr;                                       < S_Debug >
-      | le=lexpr; EQ; e=expr;
-          {  S_Assign (le,e,V1) }
+      | PRINT; args=plist(expr);                             { S_Print { args; debug = false } }
+      | DEBUG; args=clist(expr);                             { S_Print { args; debug = true } }
```
maranget, line 467: Why not use plist to parse the arguments of DEBUG? Syntax `DEBUG x,y` is rather weird.

PR #785, opened 2024-02-12
Added ocamlformat hook to pre-commit

Roman-Manevich:
I think it is a good idea to have a standard style throughout our OCaml code base.
Also, using ocamlformat would reduce the number of cosmetic code differences, making reviewing the code simpler and faster.
If testing is taking too long for you, you can commit while skipping testing like so: `SKIP=make-test git commit -m "foo"`
However, it is recommended that you do perform testing on every commit (unless you're squashing commit in which case, it is okay to test just once at the last commit in the sequence).


Roman-Manevich:
HI @HadrienRenaud. Sure, I'll review and make sure things work, but it may have to wait to next week.

.pre-commit-config.yaml
```diff
@@ -44,6 +44,11 @@ repos:
             # Expected files should be left verbatim.
             \.expected(-failure)?$
           )
+- repo: https://github.com/arenadotio/pre-commit-ocamlformat
```
relokin, line 47: Isn't this going to apply ocamlformat to the whole repo?

relokin: Ok, I missed the -ignore files because I skipped the `[[fmt] run dune fmt](https://github.com/herd/herdtools7/pull/785/commits/29eaed86c6cb0363ea6e0fb506edf4c785461154)` because I thought it was just checking in the formatting changes but it does add some configuration as well. If we want to retain the commits in this PR can we add these changes as a separate commit?

litmus/.ocamlformat
```diff
@@ -0,0 +1 @@
+disable = true
```
relokin, line 1: I really like the idea of the -ignore files because it meant that any new file will be checked, should we do this for litmus, tools and jingle and internal as well?

relokin (commented):
LGTM, but I have a high level question.

It will be a bit annoying if our editors can't be configured to follow the same style. Essentially ocamlformat will be fighting the developers. Is there a way to configure popular editors to follow this style? I've seen people using vim, emacs and VS Code.

relokin:
> > It will be a bit annoying if our editors can't be configured to follow the same style. Essentially ocamlformat will be fighting the developers. Is there a way to configure popular editors to follow this style? I've seen people using vim, emacs and VS Code.
> 
> I don't know. The way I use ocamlformat is the following:
> 
>     1. Write code, not caring about any style.
> 
>     2. Run ocamlformat before committing or when I've finished my function/whatever.
> 
>     3. Repeat
> 
>     4. Run ocamlformat (i.e. `dune fmt`) just before commiting and running `make test`
> 
> 
> That being said, there are a lot of editors which accept an external reformatter:
> 
>     1. VSCode [official ocaml extension](https://marketplace.visualstudio.com/items?itemName=ocamllabs.ocaml-platform) supports `ocamlformat` natively (once it is installed)
> 
>     2. If you're running emacs or vim, probably running `dune fmt` is just a few keystrokes away ...

Thanks @HadrienRenaud. That makes sense. I would vote for enabling it for other folders too. Not a blocker though if peolpe want to try it in a limited set of folders first.

relokin (dismissed):

maranget:
Well I am not very keen on imposing some format on OCaml code.

We each code in slightly different styles and I see no strong benefit in uniformity here. Provided of course that we refrain from changing complete files for no other reason than style.

As an example. Hadrien tends to put the `in` keyword in a line of its own. I'd rather have `in` at the end of the last line of the bound expression. I see no compelling reason for Hadrien to adopt my style or for me to adopt Hadrien's style.

relokin:
> Well I am not very keen on imposing some format on OCaml code.
> 
> We each code in slightly different styles and I see no strong benefit in uniformity here. Provided of course that we refrain from changing complete files for no other reason than style.
> 
> As an example. Hadrien tends to put the `in` keyword in a line of its own. I'd rather have `in` at the end of the last line of the bound expression. I see no compelling reason for Hadrien to adopt my style or for me to adopt Hadrien's style.

I think this PR needs to be accepted by everyone, and for this reason I'm removing my approval.

Having said that imo, I am not attached to any particular style. I find it useful when there is a uniform style though whatever that is. It doesn't matter if a block of code is indented 4, 8 spaces or a tab, but it helps a lot identifying the block when it's indented with the same spacing.

For example, in this code:
```
let foo a =
  let b = a + 1 in
  let c = b * 2 in
  d / c
```
I would find it confusing if someone wrote:
```
let foo a =
  let b = a + 1 in
let c = b * 2 in
  d / c
```
and I don't know if this is ever intentional.

But let me be clear I don't care if a style enforces this:
```
let foo a =
  let b = a + 1 in
  let c = b * 2 in
  d / c
```

over this:

```
let foo a =
     let b = a + 1 in
     let c = b * 2 in
     d / c
```

or this
```
let foo a =
  let b = a + 1 in
  let c = b * 2
  in d / c
```

maranget:
Option 2 looks reasonable.

PR #787, opened 2024-02-14
[ASL] Sort topologicallly the declarations of an AST before typing.

HadrienRenaud:
- This allows a single pass typing;
- This allows detection of badly recursive types and global storage elements;
- This will allow refined type validation in an up-coming PR.

### DIFF
--- a/asllib/Interpreter.ml
+++ b/asllib/Interpreter.ml
@@ -185,14 +185,6 @@ module Make (B : Backend.S) (C : Config) = struct
     let names =
       List.fold_left (fun k (name, _) -> ISet.add name k) ISet.empty env0
     in
-    let def d =
-      match d.desc with
-      | D_Func { name; _ }
-      | D_GlobalStorage { name; _ }
-      | D_TypeDecl (name, _, _) ->
-          name
-    in
-    let use d = use_constant_decl ISet.empty d in
     let process_one_decl d =
       match d.desc with
       | D_GlobalStorage { initial_value; name; ty; _ } ->
@@ -214,7 +206,12 @@ module Make (B : Backend.S) (C : Config) = struct
               IEnv.declare_global name v env |> return
       | _ -> Fun.id
     in
-    dag_fold def use process_one_decl
+    let fold = function
+      | TopoSort.ASTFold.Single d -> process_one_decl d
+      | TopoSort.ASTFold.Recursive ds ->
+          List.fold_right process_one_decl ds
+    in
+    fun ast -> TopoSort.ASTFold.fold fold ast
 
   (** [build_genv static_env ast primitives] is the global environment before
       the start of the evaluation of [ast]. *)
--- a/asllib/TopoSort.ml
+++ b/asllib/TopoSort.ml
@@ -71,7 +71,7 @@ module Make (O : OrderedHashedType) = struct
         else (acc, v_data)
 
   and strong_connected succs s fold acc v =
-    (* Assumption that [acc.datas] does not contain [v]. *)
+    (* Assumption that [s.data] does not contain [v]. *)
     let () =
       if false then Format.eprintf "strong_connected %s@." (O.to_string v)
     in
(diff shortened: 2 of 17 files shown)


maranget:
Hi @HadrienRenaud. This PR has conflicts, maybe introduced by PR #781.

asllib/Interpreter.ml
```diff
@@ -182,17 +182,21 @@ module Make (B : Backend.S) (C : Config) = struct
...
 
   let build_global_storage env0 eval_expr base_value =
+    let build_graph =
+      let def d =
+        match d.desc with
+        | D_Func { name; _ }
+        | D_GlobalStorage { name; _ }
+        | D_TypeDecl (name, _, _) ->
+            name
+      in
```
maranget, line 192: Isn't local `def` the same functions as `ASTUtils. identifier_of_decl`?

asllib/Typing.ml
```diff
@@ -2210,26 +2193,84 @@ module Annotate (C : ANNOTATE_CONFIG) = struct
...
+          | D_Func ({ body = SB_Primitive; _ } as f) ->
+              D_Func (rename_primitive d env f) |> add_pos_from d
+          | _ -> assert false)
+        ds
+    in
+    (List.rev_append ds acc, env)
+
+  (* Begin Specification *)
+  let type_check_ast =
+    let def d =
```
maranget, line 2251: Again the local function `def` looks the same as `ASTUtils. identifier_of_decl`.

asllib/TopoSort.mli
```diff
@@ -0,0 +1,81 @@
+(** This modules implement a topological sorting algorithm. *)
```
maranget, line 1: Hum this module implements more than simple "topological sorting". It implements computation of strongly connected components.

asllib/TopoSort.mli
```diff
@@ -0,0 +1,81 @@
...
+
+(** In the following, {i dependencies} are understood in the ASL declaration
+    level. For example, in the following ASL code:
+    {v
+       constant x: integer = 4;
+       constant y: integer = x + 1;
+    v}
+    the declaration of [y] should happen after the declaration of [x],
+    because the declaration of [y] uses the value of [x].
+*)
```
maranget, line 13: The example may lead to believe that, generally speaking ASL definition must appear in source  according to their dependencies, which does not look to be the case. Said otherwise "should happen" above is ambiguous.

Perhaps "should happen" refers to typing or evaluation order?

Besides the question of mutual dependencies is overlooked. Are mutually recursive functions possible?

asllib/TopoSort.mli
```diff
@@ -0,0 +1,81 @@
...
+  type t
+  (** Represents a step in the folding *)
+
+  val build :
+    def:(AST.decl -> AST.identifier) ->
+    extra_def:(AST.decl -> AST.identifier list) ->
+    use:(AST.decl -> AST.identifier list) ->
+    AST.t ->
+    t
+  (** [build ~def ~extra_def ~use ast] is the internal representation of the
```
maranget, line 26:
```suggestion
  (** [build ~def ~extra_def ~use ast] returns the internal representation of the
```

maranget: Moreover, if I am not wrong, all usages of this module immediately apply the `fold` function to the result of `build`. As a consequence it may not be necessary to export the intermediate type `t`. Accordingly `build` would not be exported either and the new type of fold would be something like:
```OCaml
val fold : (step -> 'acc -> 'acc) ->  AST.t -> 'acc -> 'acc
```
Assuming that the `def`, `extra_def` and `use` functions are the same for all usages (which seems to be case), they can be found from the modeule `ASTUtils`.

asllib/TopoSort.ml
```diff
@@ -0,0 +1,212 @@
...
+        if v_data.lowlink > w_data.lowlink then
+          (acc, set_lowlink s v v_data w_data.lowlink)
+        else (acc, v_data)
+    | Some w_data ->
+        if OStack.mem w s.stack && v_data.lowlink > w_data.index then
+          (acc, set_lowlink s v v_data w_data.index)
+        else (acc, v_data)
+
+  and strong_connected succs s fold acc v =
+    (* Assumption that [acc.datas] does not contain [v]. *)
```
maranget, line 74: Shouldn't [acc.datas] be [s.data]? This would justify using the `add` function of hash tables below.

maranget:
Hi @HadrienRenaud, is this PR ready for merge?

maranget:
Ok so let's go.

PR #794, opened 2024-02-21
Add github action to build herd-www

relokin:
The artifact `www` can be built by explicit action.


maranget:
Thanks @relokin, looming merge.

PR #799, opened 2024-02-27
[asl] Check type validity on user input

HadrienRenaud:
This PR aims at checking type-validity on every user-provided type.


jalglave:
thanks @HadrienRenaud please do go ahead.
Jade

PR #802, opened 2024-02-28
Introduce basic support for SVE

murzinv:
Hi All,

This PR introduces basic support for SVE instructions in `litmus7`, `diy7` and `herd7`.

Scalable Vector Extension (SVE) is Vector Length Agnostic (VLA):
- Vector Length (VL) is a hardware implementation choice from 128 up to 2048 bits.
- New programming model allows software to scale dynamically to available vector length.
- No need to define a new ISA, rewrite or recompile for new vector lengths.

Scalable vector registers:
- Z0-Z31 extending NEON’s 128-bit V0-V31
- Packed DP, SP & HP floating-point elements
- Packed 64, 32, 16 & 8-bit integer elements
    
 Scalable predicate registers:
   
 - P0-P7 governing predicates for load/store/arithmetic
 - P8-P15 additional predicates for loop management
 - FFR first fault register for software speculation
 
 I personally find tutorial at [1] really great introduction to the SVE programming model.
 
 Implementation choices and known limitations:
     - SVE memory model is not addressed
     - `herd7` implements 128 bits vector length (on top of existing Neon infrastructure)
 
[1] https://www.stonybrook.edu/commcms/ookami/support/_docs/ARM_SVE_tutorial.pdf


relokin (commented):
Some very quick comments, I'll try and do another more thorough pass tomorrow.

I understand that the current implementation re-uses the Neon registers just renames them. How hard would it be to be able to change the register size?

relokin:
> > Some very quick comments, I'll try and do another more thorough pass tomorrow.
> > I understand that the current implementation re-uses the Neon registers just renames them. How hard would it be to be able to change the register size?
> 
> At the first glance, IIUC, we'd need to support arithmetic on larger register size. Another problem to solve would relate to the way to observe bits of register above 128, now we (ab)use Neon register for that.

Thanks! If it was trivial to make it scalable then it would be nice. OTOH, I am not sure that the questions on the memory model change so maybe having one implementation where the register size is 128 is sufficient.

Makefile
```diff
@@ -137,6 +137,17 @@ test.neon::
...
 	@ echo "herd7 AArch64 NEON instructions tests: OK"
 
+test:: test.sve
+test.sve::
+	@ echo
+	$(HERD_REGRESSION_TEST) \
+		-herd-path $(HERD) \
+		-libdir-path ./herd/libdir \
+		-litmus-dir ./herd/tests/instructions/AArch64.sve \
+		-variant neon \
```
relokin, line 147: Should we introduce a synonym for -variant neon, something like "vector". I find it counter intuitive that we need to enable -variant neon for sve.

herd/AArch64Arch_herd.ml
```diff
@@ -212,13 +219,31 @@ module Make (C:Arch_herd.Config)(V:Value.AArch64) =
...
       V.Val (Constant.ConcreteVector vs)
 
+    let predicate_mask psize =
+      let mask = match psize with
+      | 1 -> "0x1"
+      | 2 -> "0x3"
+      | 4 -> "0x7"
+      | 8 -> "0xff"
+      | _ ->  assert false in
+      V.stringToV mask
```
relokin, line 229: Is there a reason for this? Another simpler implementation could be:
```suggestion
    let predicate_mask psize = V.intToV (psize * 2 - 1)
```

herd/AArch64Arch_herd.ml
```diff
@@ -212,13 +219,31 @@ module Make (C:Arch_herd.Config)(V:Value.AArch64) =
...
     | Vreg (_,(_,16)) -> MachSize.Short
     | Vreg (_,(_,32)) -> MachSize.Word
     | Vreg (_,(_,64)) -> MachSize.Quad
     | _ -> assert false (* Unsupported arrangement specifier *)
 
+    let predicated_mem_access_size rs = match List.hd rs with
+    | Zreg (_,(8)) -> MachSize.Byte
+    | Zreg (_,(16)) -> MachSize.Short
+    | Zreg (_,(32)) -> MachSize.Word
+    | Zreg (_,(64)) -> MachSize.Quad
```
relokin, line 244:
```suggestion
    | Zreg (_,8) -> MachSize.Byte
    | Zreg (_,16) -> MachSize.Short
    | Zreg (_,32) -> MachSize.Word
    | Zreg (_,64) -> MachSize.Quad
```

relokin: But also this function seems to be unused?

lib/AArch64Base.ml
```diff
@@ -56,6 +66,16 @@ let arrange_specifier =
   (0,8),".B"    ; (0,16),".H"  ; (0,32),".S" ; (0,64),".D";
 ]
 
+let sve_arrange_specifier =
+[
+  0,"";8,".B"; 16,".H"; 32,".S"; 64,".D"; 128,".Q";
```
relokin, line 71: What is the purpose of 0?

relokin (commented):
Some more comments

Makefile
maranget: Don't we need two different variants as neon and sve are different?

relokin: We could have two different variants `sve` and `neon` but the underlying functionality that is enabled is the same. So either we introduce `sve` and we make it a synonym to `neon` or we introduce a common one `vector`. The only solution I think would be confusing would be to ask someone to enable `-variant neon` to run an SVE test.

relokin (commented):
Some comments, but I haven't finished my review yet.

Makefile
relokin: But in any case, the way Vladimir has implemented this change, and I think that's the right approach, -variant neon still works.

maranget: But what about generators? Is it acceptable to  generate sve  code with `-variant neon`?

lib/AArch64Base.ml
```diff
@@ -369,9 +455,33 @@ let pp_simd_scalar_reg rl r = match r with
...
+  (try List.assoc r' zvrs with Not_found -> assert false) ^
+  (try List.assoc s sve_arrange_specifier with Not_found -> assert false)
+| _ -> assert false
+
+let pp_preg_simple r = match r with
+| Preg (r',_) ->
+  (try List.assoc r' pvrs with Not_found -> assert false)
+| PMreg (r',_) ->
+  (try List.assoc r' pvrs with Not_found -> assert false)
+| _ -> assert false
```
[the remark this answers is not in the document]
maranget: Hi @relokin, `assert false` is slightly more informative as it it leads to a message with line number included. That said, `assert false` should never be executed...

Makefile
relokin: ah sorry I thought we were referring to `herd7`. For the generator I am not sure, what do use `-variant neon` for? Or why would we need to introduce -variant sve? If a user specifies an SVE or a Neon relaxation, shouldn't diyone7 just produce the test?

maranget: Well I'd expect `-variant neon` to accept a strict subset of the annotations accepted by `-variant sve`. More precisely, with head version we have:
```
$ diyone7 -arch AArch64 -variant neon -show annotations
 X XL XA XAL I L Q A PaI PaN Pa Ne4i Ne3i Ne2i Ne4 Ne3 Ne2 Ne1 NePaN NePa NeP NeL NeQ
```
While with this PR, we have:
```
$ /opt/WIP/bin/diyone7 -arch AArch64 -variant neon -show annotations
 X XL XA XAL I L Q A PaI PaN Pa Sv4i Sv3i Sv2i Sv1 SvV Ne4i Ne3i Ne2i Ne4 Ne3 Ne2 Ne1 NePaN NePa NeP NeL NeQ
```
Obviously some annotations have been introduced: `Sv4i Sv3i Sv2i Sv1 SvV`. It would perhaps be more natural if they were available only with a specific `-variant sve` option. Notice that I have no idea whether `Sv...` annotations lead to code that is neon compatible or not. I am just considering the names themselves.

relokin: Ok, I see diyone7 will use `-variant sve|neon` to decide which annotations and relaxations it will list.

Then, it sounds like the cleanest option would be to have two different variants `sve` and `neon`. For herd7, they would be synonyms, but for diyone7, they would enable different annotations. Also, we drop `vector`. `vector` makes sense for herd7 but it doesn't make sense for `diyone7` and it's not worth having different variants for the two. Do you agree?

maranget: Not only which annotation it would list, but more significantly, which annotation it would accept.

Then, `vector` could exist for all tools as a synonymous of "the most advanced vector extension"... ?

BTW, can someone try running sve tests on actual hardware?  The appropriate gcc compile option should be something like `-march=armv8.2-a+sve`. It may be useful to define a litmus configuration file similar to `armv8.1.cfg`.

herd/tests/instructions/AArch64.sve/V01.litmus
```diff
@@ -0,0 +1,13 @@
...
+}
+P0                    ;
+ MOV Z0.S, #1         ;
+ MOV Z1.S, #2         ;
+ ADD Z2.S,Z1.S,Z0.S   ;
+ MOV Z3.D, #3         ;
+ MOV Z4.D, #4         ;
+ ADD Z5.D,Z4.D,Z3.D   ;
+
+forall 0:V2.4S={3,3,3,3} /\ 0:V5.4S={7,0,7,0}
```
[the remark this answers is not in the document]
relokin: That's a good point, any feature has to be feasible in both herd7 and litmus7

herd/tests/instructions/AArch64.sve/V19.litmus
```diff
@@ -0,0 +1,9 @@
+AArch64 V19
+{}
+P0                     ;
+MOV Z0.S,#1            ;
+MOV W0,#2              ;
+WHILELT P0.S, WZR,W0   ;
+MOVPRFX Z1.S,P0/Z,Z0.S ;
```
relokin, line 7: gcc complaints about this, but it might be ok
```Warning: SVE instruction expected after `movprfx' -- `ldr x0,[x19]'```

File: herd/tests/instructions/AArch64.sve/V20.litmus [hunk 8]
```diff
@@ -0,0 +1,10 @@
+AArch64 V20
+{}
+P0                     ;
+MOV Z0.S,#1            ;
+MOV Z1.S,#2            ;
+MOV W0,#3              ;
+WHILELT P0.S, WZR,W0   ;
+MOVPRFX Z1.S,P0/M,Z0.S ;
```
relokin, line 8: ditto

relokin (commented):
A couple of more comments but if you fixed some glitches we could merge this soon. Also, it might be worth adding something like this as it allowed me to run the sve tests with litmus7:
```
$> cat litmus/libdir/armv8+sve-qemu.cfg
size_of_test = 10k
number_of_run = 100
avail = 4
limit = true
memory = direct
stride = 1
carch = AArch64
barrier = userfence
smt = 4
smt_mode = seq
affinity = incr0
ccopts = -O2 -march=armv8-a+sve
crossrun = qemu:qemu-aarch64
```

herd/AArch64Sem.ml
```diff
@@ -2033,6 +2183,187 @@ module Make
...
+        let orig = match pg with
+        | AArch64Base.PMreg (_,AArch64Base.Zero)
+          -> mzero
+        | AArch64Base.PMreg (_,AArch64Base.Merge)
+          -> read_reg_scalable false dst ii
+        | _ -> assert false
+        in
+        orig >>|
+        read_reg_predicate false pg ii >>= fun (orig,pred) ->
+          get_predicate_any pred psize nelem >>= fun any ->
```
relokin, line 2225: That's an interesting implementation choice.

Your implementation generates one Register Read/Write effect per register. Another approach would be to generate one Register Read/Write effect per lane.

A first observation is that if the predicate is \M and all, then you don't have to read the dst register. But more importantly, the pseudocode makes it look like these register effects are per lane.

What do you think?

relokin: Ignore this, I misread (again) the pseudocode. I think your code is fine!

relokin:
I am happy with this PR, so we could merge soon! Is anyone else planning to have to review this? Otherwise, we'll aim to merge this by the end of the week.

Makefile
```diff
@@ -138,6 +138,17 @@ test.neon::
 		$(REGRESSION_TEST_MODE)
 	@ echo "herd7 AArch64 NEON instructions tests: OK"
 
+test:: test.sve
+test.sve::
+	@ echo
+	$(HERD_REGRESSION_TEST) \
```
maranget, line 144: maybe consider running test in parallel:
```suggestion
	$(HERD_REGRESSION_TEST) \
	-j $(J) \
```

maranget:
Hi @murzinv, with the following [patch](https://cambium.inria.fr/~maranget/vrac/0001-herd-Optimise-SVE-predicate-regsiters.patch), predicated loads and store usually run _much_ faster.

litmus/libdir/armv8+sve-qemu.cfg
```diff
@@ -0,0 +1,13 @@
...
+limit = true
+memory = direct
+stride = 1
+carch = AArch64
+barrier = userfence
+smt = 4
+smt_mode = seq
+affinity = incr0
+ccopts = -O2 -march=armv8-a+sve
+crossrun = qemu:qemu-aarch64
```
maranget, line 13: It may be appropriate to specify the usual name for aarch64 cross compiler and static linking:
```suggestion
crossrun = qemu:qemu-aarch64
gcc = aarch64-linux-gnu-gcc
linkopt = -static
```
static linking is useful when the aarch64 `.so` for libc is not installed, which is often the case.

litmus/AArch64Arch_litmus.ml
```diff
@@ -36,6 +36,12 @@ module Make(O:Arch_litmus.Config)(V:Constant.S) = struct
   | Internal i -> sprintf "i%i" i
   | Vreg (vr, _) | SIMDreg vr ->
       (try Misc.lowercase (List.assoc vr vvrs) with Not_found -> assert false)
+  | Preg (pr, _) ->
+      (try Misc.lowercase (List.assoc pr pvrs) with Not_found -> assert false)
+  | PMreg (pr, _) ->
+      (try Misc.lowercase (List.assoc pr pvrs) with Not_found -> assert false)
```
maranget, line 42: As actions are identical, one can use a or-pattern:
```suggestion
  | Preg (pr, _) | PMreg (pr, _) ->
      (try Misc.lowercase (List.assoc pr pvrs) with Not_found -> assert false)
```

lib/AArch64Base.ml
```diff
@@ -1248,13 +1579,38 @@ let pp_cond = function
...
+  | VL7 -> "VL7"
+  | VL8 -> "VL8"
+  | VL16 -> "VL16"
+  | VL32 -> "VL32"
+  | VL64 -> "VL64"
+  | VL128 -> "VL128"
+  | VL256 -> "VL256"
+  | MUL4 -> "MUL4"
+  | MUL3 -> "MUL3"
+  | ALL -> "ALL"
```
maranget, line 1599: One can re-use code from `lib/AArch64Base.ml`:
```suggestion
let pp_pattern p = A.pp_pattern p |> Misc.lowercase
```

litmus/AArch64Compile_litmus.ml
```diff
@@ -918,6 +939,336 @@ module Make(V:Constant.S)(C:Config) =
...
+      | VL7 -> "vl7"
+      | VL8 -> "vl8"
+      | VL16 -> "vl16"
+      | VL32 -> "vl32"
+      | VL64 -> "vl64"
+      | VL128 -> "vl128"
+      | VL256 -> "vl256"
+      | MUL4 -> "mul4"
+      | MUL3 -> "mul3"
+      | ALL -> "all"
```
maranget, line 960: One car re-use code from `lib/AArch64Base.ml`:
```suggestion
    let pp_pattern p = A.pp_pattern p |> Misc.lowercase
```

litmus/AArch64Compile_litmus.ml
```diff
@@ -918,6 +939,336 @@ module Make(V:Constant.S)(C:Config) =
...
+    | Preg (_,s) -> "^" ^ io ^ string_of_int (i+offset) ^
+       (try Misc.lowercase (List.assoc s sve_arrange_specifier) with Not_found -> assert false)
+    | PMreg (_,s) -> "^" ^ io ^ string_of_int (i+offset) ^
+       (try Misc.lowercase (List.assoc s sve_pred_modifier) with Not_found -> assert false)
+    | _ -> assert false
+
+    let print_zreg io offset i r = match r with
+    | Zreg (_,s) -> "^" ^ io ^ string_of_int (i+offset) ^
+      (try Misc.lowercase (List.assoc s sve_arrange_specifier) with Not_found -> assert false)
+    | _ -> assert false
```
maranget, line 972: It would be more elegant to define `pp_sve_arrange_specifier` and `pp_sve_pred_modifier` functions in `lib/AArch64Base.ml` and then:
```suggestion
    let print_preg io offset i r = match r with
    | Preg (_,s) -> "^" ^ io ^ string_of_int (i+offset) ^
       (pp_sve_arrange_specifier s |> Misc.lowercase)
    | PMreg (_,s) -> "^" ^ io ^ string_of_int (i+offset) ^
       (pp_sve_pred_modifier s |> Misc.lowercase)
    | _ -> assert false

    let print_zreg io offset i r = match r with
    | Zreg (_,s) -> "^" ^ io ^ string_of_int (i+offset) ^
      (pp_sve_arrange_specifier s |> Misc.lowercase)
    | _ -> assert false
```

maranget:
Hi @murzinv, nice work, LGTM. I have nevertheless suggested a patch and could not resist making a few cosmetic remarks. You probably find the [patch](https://github.com/herd/herdtools7/pull/802#issuecomment-2088721154) to be worth considering.

maranget:
> Wow! That runs at speed of light! I'm wondering if similar optimization can be applied to other instructions which set predicate like `WHILELT` and friends?

[Here is the same optimisation](https://cambium.inria.fr/~maranget/vrac/0001-herd-Optimise-SVE-predicate-registers-WHILE-OP.patch) applied to `WHILE<OP>` instructions.

I also notice that all those `WHILE<OP>` instructions could be implemented by one single constructor with an extra argument that describes the operation : that is, `WHILE(<OP>,....)` in place of `WHILELT(...)`, `WHILEEQ(...)` etc. However I would not like this to prevent merging. I am in favor of merging now, once the patch is applied.

maranget:
Fine, if you are happy with the current state, would you lease rebase on master. You may also squash all commits into one, or at least squash the last two commit into one. Once this is done, I'll merge.

maranget:
Merged, thanks @murzinv and @relokin.

PR #804, opened 2024-02-29
[all]: Rework Neon load/store and load/store pair

murzinv:
Currently Neon load/store is split in two instructions: one for handling immediate/register offset and another for handling post-index variant. The similar applies to Neon load/store pair. Note, missing support for pre-index variant for Neon load/store.

Let's refactor those instructions to take advantage of existing infrastructure for handling addressing modes. That removes need for special instruction for post-index variant and as a bonus naturally brings support for pre-index variant for Neon load/store.

### DIFF
--- a/herd/AArch64Sem.ml
+++ b/herd/AArch64Sem.ml
@@ -2511,7 +2511,6 @@ module Make
             ldrs sz (tr_variant v) rd rs e ii
         | I_LDUR(var,rd,rs,k) ->
             let sz = tr_variant var in
-            let k = match k with Some k -> k | None -> 0 in
             ldr sz rd rs (MemExt.k2idx k) ii
         | I_LDAR(var,t,rd,rs) ->
             let sz = tr_variant var in
--- a/herd/AArch64Sem.ml
+++ b/herd/AArch64Sem.ml
@@ -1903,9 +1903,9 @@ module Make
       let simd_ldr = do_simd_ldr  Annot.N
       let simd_ldar = do_simd_ldr  Annot.Q
 
-      let do_simd_str an sz rs rd kr s ii =
-        get_ea rs kr s ii >>|
-        read_reg_neon true rd ii >>= fun (addr, v) ->
+      let do_simd_str an sz ma rd ii =
+        ma >>|
+        read_reg_neon true rd ii >>= fun (addr,v) ->
         if sz == MachSize.S128 then
           do_write_mem_2_ops sz an aexp Access.VIR addr v ii >>= B.next2T
         else
@@ -1914,9 +1914,9 @@ module Make
       let simd_str = do_simd_str Annot.N
       let simd_stlr = do_simd_str Annot.L
 
-      let simd_str_p sz rs rd k ii =
-        read_reg_ord rs ii >>|
-        read_reg_neon true rd ii >>= fun (addr, v) ->
+      let simd_str_p sz ma rd rs k ii =
+        ma >>|
+        read_reg_neon true rd ii >>= fun (addr,v) ->
         if sz == MachSize.S128 then
           (* 128-bit Neon LDR/STR and friends are split into two 64-bit
            * single-copy atomic accesses. *)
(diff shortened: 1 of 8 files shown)


lib/AArch64Base.ml
```diff
@@ -1385,27 +1381,15 @@ let do_pp_instruction m =
...
-
-  let pp_smemp memo v r1 r2 ra k =
-    pp_memo memo ^ " " ^ pp_vsimdreg v r1 ^ "," ^ pp_vsimdreg v r2 ^
-    ",[" ^ pp_xreg ra ^ (if m.compat
-      then pp_kr false false (K k) else "," ^ m.pp_k k)  ^ "]" in
+  let pp_smemp memo v r1 r2 ra idx =
+    pp_memo memo ^ " " ^
+    pp_vsimdreg v r1 ^ "," ^
+    pp_vsimdreg v r2 ^ "," ^
+    do_pp_idx true ra idx in
```
relokin, line 1392: This is getting a bit complicated but if you made the change below you wouldn't have to change the hashes of tests with LDP/STP. I think this means that we need to change the declaration of `do_pp_idx` and make sure that the argument from_pair is set to true for non-neon pair instructions only.

```suggestion
    do_pp_idx false ra idx in
```

relokin (commented):
Two comments but otherwise looks good to me.

lib/AArch64Base.ml
```diff
@@ -1472,10 +1472,8 @@ let do_pp_instruction m =
       pp_mem_ext "LDRSW" V64 r1 r2 idx
   | I_LDRS ((v,bh),r1,r2,idx) ->
      pp_mem_ext (ldrs_memo bh) v r1 r2 idx
-  | I_LDUR (_,r1,r2,None) ->
-      sprintf "LDUR %s, [%s]" (pp_reg r1) (pp_reg r2)
-  | I_LDUR (_,r1,r2,Some(k)) ->
-      sprintf "LDUR %s, [%s, %s]" (pp_reg r1) (pp_reg r2) (m.pp_k k)
+  | I_LDUR (_,r1,r2,k) ->
+      sprintf "LDUR %s, [%s%s]" (pp_reg r1) (pp_reg r2) (pp_kr false false (K k))
```
relokin, line 1476: It's this space that makes the hash change? I think it's fine unless @maranget has runs with LDUR and the change of hash causes problems for his database.

herd/AArch64Sem.ml
```diff
@@ -2625,57 +2622,85 @@ module Make
...
-        | I_STR_SIMD(var,r1,rA,kr,s) ->
             let access_size = tr_simd_variant var in
-            simd_str access_size rA r1 kr s ii
-        | I_STR_P_SIMD(var,r1,rA,k) ->
+            get_ea rA (K k) S_NOEXT ii >>= fun addr ->
+            simd_ldar access_size addr r1 ii >>= B.next1T
+        | I_STR_SIMD(var,r1,rA,MemExt.Reg (v,kr,sext,s)) ->
+            let access_size = tr_simd_variant var in
+            get_ea_reg rA v kr sext s ii >>= fun addr ->
+            simd_str access_size addr r1 ii
```
relokin, line 2653: This has the unfortunate consequence of changing the intra-instruction relations. So I would suggest instead doing this:
```suggestion
            let ma = get_ea_reg rA v kr sext s ii in
            simd_str access_size ma r1 ii
```

herd/AArch64Sem.ml
```diff
@@ -1901,9 +1900,8 @@ module Make
       let simd_str = do_simd_str Annot.N
       let simd_stlr = do_simd_str Annot.L
 
-      let simd_str_p sz rs rd k ii =
-        read_reg_ord rs ii >>|
-        read_reg_neon true rd ii >>= fun (addr, v) ->
+      let simd_str_p sz addr rd rs k ii =
+        read_reg_neon true rd ii >>= fun v ->
```
relokin, line 1904: Check my comment below but to retain the iico_* relations as before, then you could do:
```suggestion
      let simd_str_p sz ma rd rs k ii =
        ma >>| 
        read_reg_neon true rd ii >>= fun (addr, v) ->
```

relokin (commented):
Two quick comments

herd/AArch64Sem.ml
relokin: And the same applies to all store variants in this PR. The loads are ok.

relokin (approved):
Great thanks for the quick fix!

If @maranget doesn't see any problems with the hash change we can merge. Let's give people some time to react. Let's aim to have this merged by the end of Thursday?

maranget:
> If @maranget doesn't see any problems with the hash change we can merge. Let's give people some time to react. Let's aim to have this merged by the end of Thursday?

I have no objection, `LDUR` does not appear in our hardware logs.

relokin:
@murzinv can you please rebase on top of master and then we can merge this.

relokin:
Thanks @murzinv!

PR #812, opened 2024-03-25
[herd]: Fix address generation for STG, LDG and friends

murzinv:
The address used for these instructions is calculated from the base register and an immediate signed offset scaled by the Tag granule.


relokin (approved):
LGTM!

maranget:
Merged, thanks @murzinv

PR #816, opened 2024-03-28
[test] Do not test carpenter when dependencies are not available

maranget:
Add appropriate (?) `if_enabled` clause in `dune` file.


maranget:
Well I did not realize either, until it failed.

PR #821, opened 2024-04-03
[ci] Add www build in github make test action

relokin:
This is to catch cases like this https://github.com/herd/herdtools7/pull/817

### DIFF
--- a/.github/workflows/make-test.yml
+++ b/.github/workflows/make-test.yml
@@ -39,3 +39,7 @@ jobs:
       - run: opam exec -- make build
 
       - run: opam exec -- make test
+
+      - run: opam install js_of_ocaml js_of_ocaml-ppx zarith_stubs_js
+
+      - run: opam exec -- make -C herd-www
--- a/herdtools7.opam
+++ b/herdtools7.opam
@@ -23,7 +23,9 @@ depends: [
   "zarith"
 ]
 depopts: [
+  "js_of_ocaml"
   "js_of_ocaml-ppx"
+  "zarith_stubs_js"
 ]
 conflicts: ["ocaml-option-bytecode-only"]
 url {


maranget:
I am a bit concerned about the overall time taken by CI. Do  we think that the benefit of testing systematic testing of www is worth the extra ressource? After all www  updates are not that frequent.

relokin:
> I am a bit concerned about the overall time taken by CI. Do we think that the benefit of testing systematic testing of www is worth the extra ressource? After all www updates are not that frequent.

Are you concerned that it might delay us or that it's wasting resources? 
- As far as I understand, the way this action works right now it doesn't block merging. One could merge before the regressions are complete.
- The concern about resource utilisation is a fair one. I don't want to waste resources although, this is an OSS project and github doesn't bill.

The rational behind this is the fix you did https://github.com/herd/herdtools7/pull/817 because of my PR https://github.com/herd/herdtools7/pull/777 . Indeed www doesn't change very often but it's difficult to know when changes to herd while affect www and one way and to be on the safe side the easy way is just to make in www and make sure nothing failed.

But this is just an idea and I am happy to drop this if the downsides outweigh any benefits.

relokin:
Actually looking at the log that provides a breakdown, the increase in runtime doesn't seem be due to a glitch in the system. The 2nd opam install takes a bit more than 1m and make in www takes a bit more than 1m as well.

There might be ways to bring this down:
- Merging the 2 opam install commands into 1
- Using some sort of cache for the environment, it's probably a bit of a waste starting from scratch and setting up opam every time.

maranget:
I also ran the action  (on a temporary branch), we can count on 2 minutes extra time at least.

Thinking again, I am not in favor of systematic testing of the www build. In practice, some PR may break this build, but this does not harm as long as we do not update the site. At that time and one easily discover that the build is broken. Fixes are usually simple. If complex they will require some expertise that the  PR author may not have. Said otherwise, we can live with a broken www build for some time.

However, we should also avoid significant drift of the www interface, Why not a periodic action on a weekly pr monthly basis?

maranget:
I am also wondering who will be warned in case of failure of a periodic action.

relokin:
Sounds good!

In the actions we already have, the person that triggers a run gets notified if the action failed. For the `www-build` action it's the person who pressed the button, for the `make-test` it's the person that open or updated a PR. For periodic actions
"Notifications for scheduled workflows are sent to the user who last modified the cron syntax in the workflow file. For more information, see "[Notifications for workflow runs](https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/notifications-for-workflow-runs)." https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule

maranget:
Hum it would be more convenient if one could define a list of recipients to be warned in case of failure.

relokin:
Closing this PR, as discussed, this is not the right solution at this point.

PR #822, opened 2024-04-04
[ASL] New Asserted Type Conversions

HadrienRenaud:
Notice that this is a PR to merge on branch `asl-valid-types`, ie #799 

This implements new logic for Asserted Type Conversions, which only focuses on structures and not on subtypes.
This new version also only allows ATCs on bitvector and integer types (if they are not equal).

TODO:
 - [x] Implementation
 - [ ] Testing


Roman-Manevich:
Let's wait a tiny bit more.

PR #824, opened 2024-04-05
[asl] Fixup missing value in asl-carpetenter

relokin:


relokin:
It appears that the latest master doesn't compile. @maranget @HadrienRenaud can you check that `experimental` has the right value for carpenter?

relokin:
> Hi @relokin, thank you for noticing However, I think the latest master (after merging #819 which removes the `experimental` flag) does not.

Ah cool I wasn't following #819. But this is indeed unnecessary now, so I am going to close it.

PR #828, opened 2024-04-11
[herd] Fixes for -variant vmsa,mte

relokin:
This PR fixes three different issues in herd7 when simulating with `-variant vmsa,mte`.
1. Tag locations in post-condition are Physical symbols when running with `-variant vmsa`.
2. Fixes the possible executions for `STZ{,2}G` when there is a fault.
3. Fixes `lift_kvm` to correctly handle cases where a checked instruction runs into a fault.

For the 3 fixes, this PR contributes unit tests that demonstrate the issue and serve as regressions after the fix is applied.

### DIFF
--- a/herd/AArch64Sem.ml
+++ b/herd/AArch64Sem.ml
@@ -1312,18 +1312,14 @@ module Make
             fun (r,_) -> M.unitT r
         else m
 
-      let lift_kvm dir updatedb is_tag mop ma an ii mphy =
+      let lift_kvm dir updatedb mop ma an ii mphy =
         let lbl_v = get_instr_label ii in
         let mfault ma a ft =
-          if is_tag then
-            insert_commit_to_fault ma (fun _ -> mzero) (Some "Tag") ii >>!
-            B.Fault [AArch64Base.elr_el1, lbl_v]
-          else
-            insert_commit_to_fault ma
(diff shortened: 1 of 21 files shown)


relokin:
@maranget from my perspective this PR is now ready and the instruction semantics have been reviewed and agreed. Would you like to have a look? If you do, there are some motivating unit tests but if there is something that doesn't make sense please let me know.

maranget:
LGTM. I agree that the PR can be merged, once rebased on master, as usual.

As far as I have checked the output and diagrams of tests A010, A011 and A012 has not changed from before the PR: a small contraction with commit  bfae4ce title "Fix fault behaviour for -variant vmsa,mte". Or am I missing something?

relokin:
> LGTM. I agree that the PR can be merged, once rebased on master, as usual.
> 
> As far as I have checked the output and diagrams of tests A010, A011 and A012 has not changed from before the PR: a small contraction with commit [bfae4ce](https://github.com/herd/herdtools7/commit/bfae4ceb8ce726b94fa5d82d1a9ce1bd3384c143) title "Fix fault behaviour for -variant vmsa,mte". Or am I missing something?

Good point, these are interesting tests illustrative of the semantics that we want to model, but they didn't change due to this commit. I've added 3 more tests that would fail before this fix.

Also rebased to the latest master.

maranget:
Merged, thanks @relokin.

relokin:
Super! Many thanks @maranget!

PR #830, opened 2024-04-12
7.57 rc3 prepare

maranget:
Prepare for release 7.57


herdtools7.opam
```diff
@@ -22,9 +22,6 @@ depends: [
   "menhir" {>= "20200123"}
   "zarith"
 ]
-depopts: [
-  "js_of_ocaml-ppx"
```
relokin, line 26: Is this incorrect? It's required to build www

maranget: This is not necessary to build the package. Standard users are not meant to build www. Moreover, not all dependencies for building www are listed . I'd rather not maintain those. Advanced used will follow the instructions, which are up to date.

As a matter of facts, I do not understand the purpose of such "optional" dependencies. I guess it may triggers some re-compilations en case of upgrades.

PR #836, opened 2024-04-25
[asl][doc] fix make asldoc on empty version

HadrienRenaud:
Fix `make asldoc` on an empty directory.


Roman-Manevich (approved):
Looks good to me.

PR #837, opened 2024-04-25
[asl] Fix: moved macros around to enable compiling all references.

Roman-Manevich:


jalglave:
`make asldoc` works for me here. thanks for the fix @Roman-Manevich.

PR #839, opened 2024-04-29
[test] Check catalogue existence before running catalogue tests

maranget:
Namely, releases of herdtools do not include the catalogue. As an untimely consequence `make test` was failing.


relokin (approved):
Looks good to me!

maranget:
Thanks @relokin.  With this last minute change, I guess we are ready for releasing 7.57.

PR #845, opened 2024-05-02
[Makefile] Add some basic tests for litmus7

relokin:
I am not sure this ready for prime time but it's high time we added some basic regression testing for litmus7. So I'm posting this just to start a conversation really. How can we improve this?

Known limitations of `make litmus-test` :
- [ ] Assumes it runs on an Arm machine without checking. Checking and exiting is rather straightforward.
- [ ] Does not run any litmus tests, so the check is limited to whether the generated C code compiles.
- [ ] Two regressions that depend on kvm-unit-tests, automatically clone, and compile kvm-unit-tests but don't clean-up the kvm-unit-tests directory.


maranget:
Thanks @relokin. I have already found a problem thanks to to PR: option `-mode presi` does not work without `-ascall true`.

Makefile
```diff
@@ -617,3 +618,65 @@ clean-asl-pseudocode:
...
+	@ echo "litmus7 in -mode presi catalogue aarch64 tests: OK"
+
+ifeq ($(OS),Darwin)
+KUT_CONFIG_PARAMS=--page-size=4k --cross-prefix=aarch64-elf-
+endif
+KUT_DIR:=$(shell mktemp -d)
+
+litmus-dep:
+	cd $(KUT_DIR); \
+	git clone -q git@gitlab.com:kvm-unit-tests/kvm-unit-tests.git; \
```
maranget, line 647: Access to the kvm-unit-test repository looks easier using http. Namely, public key access does not work for me, probably by lack of some configuration:
```suggestion
	git clone -q https://gitlab.com/kvm-unit-tests/kvm-unit-tests.git; \
```

maranget:
With the patch above `make litmus-test` works for me on a Macbook pro M1.

maranget:
Hi @relokin. What about applying the patch above and merging? There still are three task pending, but the PR already looks quite useful as it is, since you apparently have used it at least twice...

maranget:
Hi @murzinv. This looks very interesting, would you submit these adjustments here?

relokin:
I am sorry for the radio silence. I had a quick look at Vladimir's suggestions, they fix some of the issues with this patch on Linux but they break the regression on my MacBook, I'll try to find a way to make it work on both and will update this PR.

relokin:
@murzinv, I've made some improvements. I've tested on MacOS with Apple Silicon and on an Arm server. I don't have any x86 machines to try at the moment.

relokin:
With this new version, I can run litmus-test on an Arm Linux machine and the litmus-aarch64-test on a MacBook Pro M1.

maranget:
Hi @relokin ans @murzinv. Would consider running the tests optionally ? For the record, `make litmus-x86_64-test RUN_TESTS=true` additionally runs the compiled tests (meaningful on an x86_64 machine, as you can guess).

maranget:
Let us merge this PR, @relokin would you proceed?

relokin:
@maranget, I just posted a new version with RUN_TESTS. However I realised that if there is a failure then the make will not exit. I don't have the time to look at this problem now. I guess whoever is using RUN_TESTS=yes they have to proceed with caution.

maranget:
Merged, thanks @relokin and @murzinv.

PR #847, opened 2024-05-03
Fix BL/BLR semantics

relokin:
Prior to this change, for BL and BLR, there was an intrinsic data dependency from the the Register Write effect of the LR to the Intrinsic Branching effect. This incorrect as the Intrinsic Branching effect is not dependent on the data of the LR. This change, changes the semantics of BL and BLR such there is an intrinsic order dependency from Register Write effect of the LR to the Intrinsic Branching effect.

In addition, this PR adds support for labels as values of registers in the post-condition of tests, like the following:

```
AArch64 A258
{
}
 P0        ;
L0:        ;
 ADR X0,L0 ;
forall(0:X0=label:"0:L0")
```


maranget:
Hi @relokin. The following test is rejected:
```
% cat A.litmus
AArch64 A258

{
}
 P0        ;
 L0:       ;
 ADR X0,L0 ;

forall (0:X0=P0:L0)
% herd7 A.litmus
Warning: File "A.litmus": L0 is not a register (User error)
```
Namely, the syntax `P0:L0` for a label is not recognised in the final condition, while it is in the init section. That is, teh following test is accepted:
```
 AArch64 A258

{
0:X1=P0:L0;
}
 P0        ;
 L0:       ;
 ADR X0,L0 ;

locations [0:X1;]
```
Would it be possible to accept the syntax `P<i>:<label>` in final conditions? 

(For completeness, there is another, similar,  glitch, labels in the init section are accepted as `P0:L0` and prinuted as `0:L0`)

maranget:
The `A.litmus` test above was generated with `mprog7`. It looks difficult to fix the parser so as to accept for `P<i>:<label>`. Hence an easier solution is to fix the pretty-printing of final conditions, for `mprog7` (and other tests pretty-printers) to output  syntax that can be parsed. The fix is comming.

maranget:
Here is the 
[fix](https://github.com/herd/herdtools7/files/15501610/0001-all-Pretty-print-labels-in-conditions-as-parsable-sy.PATCH).

lib/stateParser.mly
```diff
@@ -179,8 +180,8 @@ maybev_list:
 
 maybev_label:
 | maybev { $1 }
-| PROC COLON NAME { mk_lab $1 $3 }
-| NUM COLON NAME { mk_lab (Misc.string_as_int $1) $3 }
+| PROC COLON NAME { mk_lab ($1, $3) }
```
maranget, line 183: It seems possible to restore one deleted line without having a conflict:
```suggestion
|  NUM COLON NAME { mk_lab (Misc.string_as_int $1) $3 }
| PROC COLON NAME { mk_lab ($1, $3) }
```

maranget: That way we'd restore previous behaviour to accept initialisations such as `0:X1=0:L0`.

maranget:
> Here is the [fix](https://github.com/herd/herdtools7/files/15501610/0001-all-Pretty-print-labels-in-conditions-as-parsable-sy.PATCH).

Hum this fix does not handle all cases of  condition pretty-printing. Some other places should undergo similar changes (_e.g._ [here](https://github.com/herd/herdtools7/files/15502265/0001-herd-Output-parsable-syntax-for-labels-in-conditions.PATCH)). Some refactoring may simplify things.

litmus/skelUtil.ml
```diff
@@ -700,6 +715,71 @@ end
...
+          List.iteri
+            (fun i (p,lbl) ->
+               let flbl = OutUtils.fmt_lbl_var p lbl in
+               O.f "#define %-25s  %d" (instr_symb_id flbl) (i + 1))
+            lbls ;
+          O.o "" ;
+          O.f "static const char *instr_symb_name[] = {" ;
+          O.oi "\"UNKNOWN\"," ;
+          (* Define names for inst symbols *)
+          List.iter (fun (_,lbl) -> O.fi "\"%s\"," lbl) lbls ;
```
maranget, line 730: **herd7** output includes the processor info, change for compatibility:
```suggestion
          List.iter (fun (p,lbl) -> O.fi "\"%d:%s\"," p lbl) lbls ;
```
```suggestion
          List.iter (fun (_,lbl) -> O.fi "\"%s\"," lbl) lbls ;
```

litmus/skel.ml
```diff
@@ -1090,8 +1094,6 @@ module Make
                            (CType.dump (register_type rloc (CType.Base t)))
                            sloc k::pp_rec (k+1) in
                      String.concat "," (pp_rec 0)
-                 | t when CType.is_ins_t t ->
-                    sprintf "pretty_opcode(o[%s_f])" sloc
```
maranget, line 1094: Removing those lines looks problematic. For instance, **litmus7** now produces C code that does not compile on test [L049](https://github.com/herd/herdtools7/blob/master/herd/tests/instructions/AArch64/L049.litmus).
Restoring the deleted lines apparently fixes the anomaly.

catalogue/aarch64-ifetch/tests/MP+rel+blr.litmus
```diff
@@ -0,0 +1,14 @@
+AArch64 MP+rel+blr
```
maranget, line 1: For this test to run on hardware, one must declare `X30` to be allocated  to itself by gcc:
```suggestion
AArch64 MP+rel+blr
Stable=X30
```
See also https://github.com/herd/herdtools7/pull/847#discussion_r1622564008, as printed `1:X6` is of instruction ttype.

relokin:
@maranget thanks for the reviews! I've added your commits to the PR but as I was looking at them, it occurred to me that we can require that labels in post-conditions always use the full format `label:"P<n>:<str>". This should ease things for the parser as it always provides its type. What do you think?

maranget:
Hi @relokin, it looks like a good idea. Please proceed.

maranget:
This PR is going beyond its initial aim to fix BL semantics. There still are minor problems (see the last two commits of this [branch](https://github.com/herd/herdtools7/tree/blr-fixes-luc)).

Nevertheless I am in favor of limiting the ambition of this PR and to aim at merging it as is, once label parsing has been changed as you have suggested. Your opinion?

litmus/skel.ml
maranget: Here is a suggestion of commit: ff8738d72443536774c96ba994463512475ca92c, as the "suggest" button is not available for a deletion :)

relokin:
hey @maranget apologies for the delay on this.

I've now had another look and applied you 2/3 of your patches from your branch (https://github.com/herd/herdtools7/tree/blr-fixes-luc). I didn't apply https://github.com/herd/herdtools7/commit/a0746aabb8ea25d6b5cdb664295d617a987c05bc because litmus7 produced extra warnings. I've forced pushed and merged my label-related commits to one, but I left your commits intact.

I've tested with make litmus-test after applying https://github.com/herd/herdtools7/pull/845

I think this is ready from my perspective, please let me know if you would like to see any changes.

maranget:
Hi @relokin. I still wish to perform a few more tests before merging. I'll keep you informed.

As regard commit https://github.com/herd/herdtools7/commit/a0746aabb8ea25d6b5cdb664295d617a987c05bc, here is an example of a test whose produced C code does not compile without this commit:
```
AArch64 Label01
{
ins_t *0:X0;
}
   P0       ;
 L0:        ;
  ADR X0,L0 ;
locations [0:X0;]
```
However the same example was already problematic before this PR, although in a different way. So let us not address this problem for now. I'll submit a specific PR later.

maranget:
Ok, nice work. Let us merge this PR. Beforehand, would you please rebase on master and squash all my commits into one, keeping all comments. Then I'll gladly merge :)

relokin:
> Hi @relokin. I still wish to perform a few more tests before merging. I'll keep you informed.
> 
> As regard commit [a0746aa](https://github.com/herd/herdtools7/commit/a0746aabb8ea25d6b5cdb664295d617a987c05bc), here is an example of a test whose produced C code does not compile without this commit:
> 
> ```
> AArch64 Label01
> {
> ins_t *0:X0;
> }
>    P0       ;
>  L0:        ;
>   ADR X0,L0 ;
> locations [0:X0;]
> ```
> 
> However the same example was already problematic before this PR, although in a different way. So let us not address this problem for now. I'll submit a specific PR later.

I agree, if this was an issue before and since this fix needs to be refined a separate PR is needed.

> Ok, nice work. Let us merge this PR. Beforehand, would you please rebase on master and squash all my commits into one, keeping all comments. Then I'll gladly merge :)

Done, thanks Luc!

maranget:
Merged! Thanks @relokin.

PR #848, opened 2024-05-03
[asl] Differentiate accessers with/without any args

HadrienRenaud:
This PR implements checks that differentiate accessers with from without `[]`.
For example:
```
getter f1[] => integer
begin return 0; end

getter f2 => integer
begin return 0; end

func main () => integer
begin
  let x = f1[];
  let x = f2;

  return 0;
end
```
Before this PR, the two syntaxes could be used without distinction.

Also fixes a bug on `make test-aarch64-asl`.

### DIFF
--- a/asllib/AST.mli
+++ b/asllib/AST.mli
@@ -313,7 +313,26 @@ and catcher = identifier option * ty * stmt
 
 (** {2 Top-level declarations} *)
 
-type subprogram_type = ST_Procedure | ST_Function | ST_Getter | ST_Setter
+type subprogram_type =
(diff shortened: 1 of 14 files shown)


asllib/AST.mli
```diff
@@ -313,7 +313,26 @@ and catcher = identifier option * ty * stmt
 
 (** {2 Top-level declarations} *)
 
-type subprogram_type = ST_Procedure | ST_Function | ST_Getter | ST_Setter
+type subprogram_type =
```
jalglave, line 316: This is nice.

PR #862, opened 2024-05-15
[herd] BPF Architecture fixed size support for herd

puranjaymohan:
This PR adds the support of BPF assembly to herd. It also adds a cat model and multiple litmus tests. A README file has been added that explains the semantics of all instructions.


catalogue/bpf/README.md
```diff
@@ -0,0 +1,239 @@
...
+lock *(u64 *)(rd + offset16) |= rs
+lock *(u64 *)(rd + offset16) ^= rs
+```
+
+2. 32 and 64 bit atomic operations with fetch:
+These instructions fetch the value from memory and then do an ALU op on the memory location (register + offset) atomically. It can work with 32-bit and 64-bit data.
+```
+rs = atomic_fetch_add ((u32 *)(rd + offset16), rs)
+rs = atomic_fetch_and ((u32 *)(rd + offset16), rs)
+rs = atomic_fetch_aor ((u32 *)(rd + offset16), rs)
```
hernanponcedeleon, line 89: Typo: it should be `or` (same below)

catalogue/bpf/tests/IRIW+fencembonceonces+OnceOnce.litmus
```diff
@@ -0,0 +1,35 @@
...
+2:r1=y;
+
+3:r0=x;
+3:r1=y;
+3:r5=tmp2;
+}
+
+P0		     | P1		                         | P2                   | P3		                              ;
+*(u32 *)(r0 + 0) = 1 | r2 = *(u32 *)(r0 + 0)                     | *(u32 *)(r1 + 0) = 1 | r2 = *(u32 *)(r1 + 0)                       ;
+                     | r6 = atomic_fetch_add((u64*)(r5 + 0), r6) |                      | r6 = atomic_fetch_add((u64*)(r5 + 0), r6)   ;
```
hernanponcedeleon, line 31: You are adding `r6`, but the value of this register is not initialized. Other tests have the same problem.

hernanponcedeleon (commented):
Two small things I noticed by quickly skimming over the changes.

catalogue/bpf/tests/IRIW+fencembonceonces+OnceOnce.litmus
puranjaymohan: Yes, `atomic_fetch_add` is just being used here to enforce ordering because BPF doesn't have a real fence instruction. r6 will be `0` by default and that is okay for this usage.

catalogue/bpf/README.md
puranjaymohan: Will fix in next version.

catalogue/bpf/tests/R+poonceonces.litmus
```diff
@@ -0,0 +1,23 @@
...
+
+1:r0 = x;
+1:r1 = y;
+}
+
+P0                      | P1                      ;
+*(u32 *)(r0 + 0) = 1    | *(u32 *)(r1 + 0) = 2    ;
+*(u32 *)(r1 + 0) = 1    | r2 = *(u32 *)(r0 + 0)   ;
+
+~exists (y=2 /\ 1:r2=0)
```
hernanponcedeleon, line 23: Assuming this test is intended to match [this one](https://github.com/torvalds/linux/blob/master/tools/memory-model/litmus-tests/R%2Bpoonceonces.litmus), it should be `exists` rather than `~exists`

puranjaymohan: Yes,
Thanks for catching it. I will fix it in the next version.

catalogue/bpf/tests/IRIW+fencembonceonces+OnceOnce.litmus
puranjaymohan: Fixed this and similar usages of r6 but initialising it to 0

maranget:
Hi here is a patch that enable using the tools on BPF tests.
[0001-tools-Enable-BPF-parsing.PATCH](https://github.com/herd/herdtools7/files/15418303/0001-tools-Enable-BPF-parsing.PATCH)

puranjaymohan:
@maranget I will implement the pretty-printing and update the pull request, I will also add your patch to the pull request.

Thanks

maranget:
For regression testing it would be nice to have a "kind.txt" file in `catalogue/bpf/tests/`. This file summarizes the result of tests as keywords: Allow or Forbid. I attach the [kinds.txt](https://github.com/herd/herdtools7/files/15418505/kinds.txt)
 file constructed from `herd7` current results. I am not sure these are  the intended kinds. Please check before adding.

There are other files to add for the regression to be performed from the main `Makefile`, please see `catalogue/aarch64/ ` and entry `cata-aarch64-test:` in Makfile, for instance. We can delay this.

puranjaymohan:
@maranget 
I have implemented the instruction pretty-printing and also added the support for regression testing.

Thanks

lib/BPFBase.ml
```diff
@@ -0,0 +1,422 @@
...
+  | DIV
+  | REM
+  | AND
+  | OR
+  | XOR
+  | LSL
+  | LSR
+  | ASR
+  | AMOXCHG
+  | AMOCMPXCHG
```
maranget, line 146: Wouldn't it be better to have two types for operators, one standard and the the other for AMO's. Even if some operators are common to the two sets (_e.g._ `ADD`) they can be represented by different constructors. One can even have the same name for different constructors by uslng local modules:
```OCaml
module Std = struct
  type op = ... | ADD | ...
end

module AMO = struct
  type op.= ... | ADD | ...
end

(* Here refer to `AMO.ADD` or `Std.ADD` *)

type instruction =
| Op of Std.op * ...
| ...
| AMO of AMO.op * ...
| ... 
```
That way, typing controls more and we avoid some `assert false`.

puranjaymohan: Sure,
I will fix this in the next version

puranjaymohan: I tried this but it is making the implementation difficult.
In the Lexer, `+=` can mean both `Std.ADD` and `AMO.ADD`. To make this work I will have to change a lot of code. So, if this is not a blocker, I would like to keep the current implementation.

maranget: Ok let it as it is.  Out of curiosity, how is disambiguation peformed?

maranget: Ok, let us leave it that way for now.

puranjaymohan: The disambiguation is performed in the Parser as the normal atomic ops look like `r1 += r2` and the atomic ops look like `lock *(u64 *)(r1 + 32) += r2`

herd/BPFSem.ml
```diff
@@ -0,0 +1,249 @@
...
+
+    let read_reg_ord = read_reg false
+    let read_reg_data = read_reg true
+
+    let do_read_mem sz ato a ii =
+      M.read_loc false (mk_read sz ato) (A.Location_global a) ii
+    ;;
+
+    let read_mem sz a ii = do_read_mem sz BPF.N a ii
+    let read_mem_atomic sz a ii = do_read_mem sz BPF.X a ii
```
maranget, line 94: This function is not called

maranget, line 94: Function `read_mem_atomic` is not called.

herd/BPFSem.ml
```diff
@@ -0,0 +1,249 @@
...
+        ii
+    ;;
+
+    let write_mem_atomic sz a v resa ii =
+      let eq = [ M.VC.Assign (a, M.VC.Atom resa) ] in
+      M.mk_singleton_es_eq
+        (Act.Access (Dir.W, A.Location_global a, v, BPF.X, (), sz, Access.VIR))
+        eq
+        ii
+    ;;
```
maranget, line 122: This function `write_mem_atomic` is not called.

herd/BPFSem.ml
```diff
@@ -0,0 +1,249 @@
...
+
+    let create_barrier b ii = M.mk_singleton_es (Act.Barrier b) ii
```
maranget, line 124: Not called either

catalogue/bpf/tests/LB+fcas-ctrlcvg-once+once-scas.litmus
```diff
@@ -0,0 +1,33 @@
+BPF LB+fcas-ctrl-once+once-scas
```
maranget, line 1: Having identical names for different tests will confuse some tools
```suggestion
BPF LB+fcas-ctrlcvg-once+once-scas
```

catalogue/bpf/tests/LB+fcas-addr-once+once-scas.litmus
```diff
@@ -0,0 +1,25 @@
+BPF LB+fcas-data-once+once-scas
```
maranget, line 1:
```suggestion
BPF LB+fcas-addr-once+once-scas
```

maranget: Notice: I found this inconstency (and the next one) with the `mdiag7` tools:
```
% mdiag7 *.litmus
Error: name LB+fcas-ctrl-once+once-scas has different hashes
 LB+fcas-ctrl-once+once-scas.litmus
 LB+fcas-ctrlcvg-once+once-scas.litmus
Error: name LB+fcas-data-once+once-scas has different hashes
 LB+fcas-data-once+once-scas.litmus
 LB+fcas-addr-once+once-scas.litmus
Input: 34
By name: 32
By hash: 34
```

catalogue/bpf/tests/kinds.txt
```diff
@@ -0,0 +1,34 @@
+ARMv8DOB5-PPC Forbidden
+CoRR+poonceonce+Once Forbidden
+CoRW+poonceonce+Once Forbidden
+CoWR+poonceonce Forbidden
+CoWW+poonceonce Forbidden
+IRIW+fencembonceonces+OnceOnce Forbidden
+IRIW+poonceonces+OnceOnce Allowed
+ISA2+poonceonces Allowed
+LB+fcas-ctrl-once+once-scas Forbidden
+LB+fcas-ctrl-once+once-scas Forbidden
```
maranget, line 10: Update following name change:
```suggestion
LB+fcas-ctrlcvg-once+once-scas Forbidden
```

catalogue/bpf/tests/kinds.txt
```diff
@@ -0,0 +1,34 @@
...
+LB+fcas-data-once+once-scas Forbidden
```
maranget, line 11:
```suggestion
LB+fcas-addr-once+once-scas Forbidden
```

maranget:
Hi @puranjaymohan , my review is over, remarks are here for documentation and are  non-blocking, _except_ for the change of the internal names  of two of the tests (see above). 

Here is the reason why: having the same internal name for different tests, will confuse some of the tools. For instance, try `herd7 *.litmus` in unmodified `catalogue/bpf/tests` and you'll see that some of the tests are not executed.

maranget:
I could not resist adding a test in the catalogue. The test is for the "lock" prefix, which, if I am not mistaken, is absent from the other tests.

herd/BPFSem.ml
puranjaymohan: Fixed in next version

herd/BPFSem.ml
puranjaymohan: Fixed in next version

herd/BPFSem.ml
puranjaymohan: Fixed in next version

catalogue/bpf/tests/LB+fcas-ctrlcvg-once+once-scas.litmus
puranjaymohan: Thanks for catching this,
fixed in next version

catalogue/bpf/tests/LB+fcas-addr-once+once-scas.litmus
puranjaymohan: Thanks, I fixed these and reran `mdiag7` to verify it is working.

catalogue/bpf/tests/kinds.txt
puranjaymohan: Fixed in next version

puranjaymohan:
Hi @maranget,
Thanks for the review. I have made all the changes that you suggested and also added the Lock test, it was not there before, thanks for catching this.

catalogue/bpf/tests/kinds.txt
puranjaymohan: fixed in next version

maranget:
Merged thanks @puranjaymohan and @hernanponcedeleon.

PR #867, opened 2024-05-30
[asl] Fix let binding of tuples

maranget:
Fix various problems related to tuple bindings.
- [X] Cyclic program order.
- [X] Impossibility to transmit non-determined value from tuple construction to binding.

Two tests that illustrate the fixed problems are also added.

- Cyclic program order:
```
ASL return-tuple

{

}

func f() => (integer, integer)
begin
  return (1,2);
end

func main() => integer
begin
  let (a,b) = f();
  let c = a+b;
  return 0;
end

forall 0:main.0.c=3
```
Before the fix, this test yielded no execution at all, because if the cyclic `po` introduced while evaluating `let (a,b) = f();`.
- Creating tuples with non-resolved arguments:
```
ASL frozen-tuple-arg

{
x=15;
0:X0=x;
}

func main() => integer
begin
  let x = read_register(0);  
  let (a,b) = (1,UInt(read_memory(x,32)));
  let c = a+b;
  return 0;
end

forall 0:main.0.c=16
```
Before the fix, evaluation of `let (a,b) = (1,UInt(read_memory(x,32)));` was failing because the `read_memory` primitive returns a non-resolved value, resulting in a pair whose second argument is a non-resolved value.

### DIFF
--- a/asllib/Interpreter.ml
+++ b/asllib/Interpreter.ml
@@ -834,7 +834,8 @@ module Make (B : Backend.S) (C : Config) = struct
     (* Begin LDTuple *)
     | LDI_Tuple ldis, Some m ->
         let n = List.length ldis in
-        let liv = List.init n (fun i -> m >>= B.get_index i) in
+        let* vm = m in
+        let liv = List.init n (fun i -> B.return vm >>= B.get_index i) in
         let folder envm ldi1 vm =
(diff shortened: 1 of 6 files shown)


maranget:
Merged, thanks @HadrienRenaud forthe review.

PR #870, opened 2024-06-03
[gen] Restore default behaviour where -relaxlist is an alias for -safe

relokin:
Prior to 17eda0c637554cd709f28107119c93628b711ee0 -relaxlist was an alias for -safe. Restore this behaviour to avoid breaking existing .conf files like gen/libdir/forbidden.conf.

### DIFF
--- a/gen/config.ml
+++ b/gen/config.ml
@@ -308,8 +308,8 @@ let speclist () =
     sprintf "<n> test relaxations considering <n> or more different relaxations (default %i). Implies -mix true." !min_relax)::
    ("-safe", Arg.String (fun s -> safes := !safes @ [s]),
     "<relax-list> specify a safe list")::
-   ("-relaxlist", Arg.String (fun s -> relaxs := !relaxs @[s]),
-    "<relax-list> specify a list of relaxations of interest")::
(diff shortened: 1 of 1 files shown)


relokin:
@maranget, I saw you were the author of 17eda0c637554cd709f28107119c93628b711ee0. Was the change to -relaxlist intentional? The commit message indicates that the intent was to make -relaxlist, -safe and -relax cumulative but didn't mention anything about the change to -relaxlist which is now an alias of -relax.

maranget:
Hi @relokin, this probably was a mistake of mine. Perhaps, the behaviour "-relaxlist is an alias for -safe" should be documented.

relokin:
> Hi @relokin, this probably was a mistake of mine. Perhaps, the behaviour "-relaxlist is an alias for -safe" should be documented.

I agree, it's not very intuitive due to the choice of names for the arguments `-relaxlist`, `-relax` and `-safe`. I've added a quick note in the help text for the argument. I couldn't find any other mention of `-relaxlist` in `doc/` or anywhere else. If you think we should document it somewhere else please let me know.

Otherwise, are you happy for me to merge this today?

maranget:
Merged, thanks @relokin. Notice: PR #874 documents `-relaxlist` as an alias of `-relax`.

PR #871, opened 2024-06-03
[herd] Neon signed128bit

maranget:
Patching printing and parsing glitches by using signed integers as a base for the neon variant.

Only a patch, because unsigned base integer types should probably be able to parse and print signed integers.


maranget:
Superseeded by PR #872

Issue #813, opened 2024-03-26
[herd] No fault on crossing memtag granularity


murzinv:
Consider litmus test
```
AArch64 T
{
int x[8]={1,2,3,4,5,6,7,8};

0:X0=x:green; 0:X1=x:red;
}
P0                               ;
STG X1,[X0,#1]                   ; (* set tag :red for x[4:7] *)
DMB SY                           ;
MOV W2,#9                        ;
MOV W3,#9                        ;
STP W2,W3,[X0,#12]               ; (* write to x[3:4] with tag :green *)

(* We expect tag fault due to access to x[4] *)
exists x={1,2,3,9,9,6,7,8} /\ ~fault(P0,x)
```
Note, that `STG` here relies on fix #812 
Output of `herd7` suggests that test is allowed
```
$ herd7 -variant memtag,sync T
Test T Allowed
States 1
x={1,2,3,9,9,6,7,8};  ~Fault(P0,x);
Ok
Witnesses
Positive: 1 Negative: 0
Condition exists (x={1,2,3,9,9,6,7,8} /\ not (fault(P0,x)))
Observation T Always 1 0
Time T 0.02
Hash=df433a02055836977667c54e7708421d
```

maranget:
Well, as apparent from diagram, the tag is check once only...
[T.pdf](https://github.com/herd/herdtools7/files/14760999/T.pdf)

PR #876, opened 2024-06-17
[all]: Add support for SVE fault tolerant instructions

murzinv:
SVE adds instructions which can suppresses memory fault, specifically:

 - Load first-fault
 - Load non-fault
 - Auxiliary instructions to inspect First Fault Register (FFR)

### DIFF
--- a/herd/AArch64Arch_herd.ml
+++ b/herd/AArch64Arch_herd.ml
@@ -111,12 +111,13 @@ module Make (C:Arch_herd.Config)(V:Value.AArch64) =
     | I_MOPL _
     | I_WHILELT _ | I_WHILELE _ | I_WHILELO _ | I_WHILELS _
     | I_UADDV _
+    | I_LDNF1 _ | I_LDFF1SP _
     | I_LD1SP _ | I_LD2SP _ | I_LD3SP _ | I_LD4SP _
     | I_ST1SP _ | I_ST2SP _ | I_ST3SP _ | I_ST4SP _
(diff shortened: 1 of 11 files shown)


relokin:
Thanks for this @murzinv, perhaps before adding support for these new instructions, or in parallel with this change, we need to add vmsa support for the existing vector instructions. Looking at the pseudocode, it looks like we have one translation/tag check per element memory read/write.

I am looking specifically at `store_scatter_predicated_elem_or_merge` what is that prevents from wrapping the call to `do_write_mem` inside a call to `lift_memop`?

maranget:
Hi @murzinv and @relokin, I have no precise idea on the next step. I agree with @relokin that combining SVE and VMSA is a prerequisite for this PR. Another, perhaps less interesting but less ambitious  direction could be to implement additional arithmetical and logical operations.

relokin:
> I assume that in such case every store of the element would end-up with either `B.next1T` or `B.Exit`, I'm not sure if it is effect we want 🤔 I also noticed that load/store pairs do not do `lift_memop` for each element of the pair (that's why I mentioned #813 ) and I suspect that there is a good reason for that... Anyway, I can explore your suggestion and report back.

I think it's B.Next or B.Fault that we want. The instruction should raise a fault, but, also, I suspect that it's architecturally allowed to see from 0 to all non-faulting, active element write happening before entering the fault handler. That might mean multiple architecturally allowed executions just because one element read/write generates a fault.

PR #886, opened 2024-06-27
[herd] Systematic delaying of failures.

maranget:
The first phases of herd should behave as follows:

When some operation fails by raising some exception, the general rule is to delay the failure, hoping that the corresponding candidate execution will be eliminate when solving equations.

However, this behaviour may complicate debugging. Hence the command line option `-debug exception` replace it with immediately raising the exception.

This PR generalises the expected behaviour.

Notice that exceptions used for controlling herd engine (Undetermined, Contradiction and Timeout) follow their own rules

### DIFF
--- a/herd/eventsMonad.ml
+++ b/herd/eventsMonad.ml
@@ -1664,15 +1664,22 @@ Monad type:
     (* Operations, no events generated, only equations *)
     (***************************************************)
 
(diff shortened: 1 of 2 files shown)


maranget:
Quick merge of this simple PR. The problem fixed has been discovered during debugging the asl-vmsa branch.

PR #890, opened 2024-07-03
[litmus]: Use 'Upl' constraint modifier for predicate register

murzinv:
Constraint modifier we currently use for predicate register is

 `Upa` - Any of the SVE predicate registers (P0 to P15)

That made a trick so far, yet with GCC 14.1.0 it seems that register allocation happens top to bottom, yet most of SVE instructions we support allow predicate register only in a range P0 to P7. So build fails.

Let's work it around by using `Upl` constraint modifier for predicate register:

  `Upl` - One of the low eight SVE predicate registers (P0 to P7)


maranget:
Hi @murzinv, I have checked this PR with an older gcc version. It is ready for merge, once rebased on master. Thanks.

maranget:
Merged, thanks @murzinv.

PR #892, opened 2024-07-03
[asl] Statically link ASL stdlib.

HadrienRenaud:
This PR statically links ASL stdlib.

This is following some troubles I had where the tests were using an installed version of the stdlib, and not the one in dev.

I'm open to more flexibility or amending this PR.


maranget:
I like this PR, do you consider it ready?

maranget:
Yes, I'll merge now. Thanks @HadrienRenaud. The new code is much simpler than the former one, as we avoid the duplication of the search patch code, which I felt uncomfortable with. It is not that nice to carry a string in code... but I guess stdlib will not grow much and that the size of this string will not increase much.

PR #894, opened 2024-07-04
[asl] Implementation of expression propagation

HadrienRenaud:
This PR implements some storage of immutable expressions accross different statements of a function. This allows some type-checking examples to work.

### DIFF
--- a/asllib/StaticEnv.ml
+++ b/asllib/StaticEnv.ml
@@ -35,6 +35,7 @@ type global = {
 type local = {
   constant_values : literal IMap.t;
   storage_types : (ty * local_decl_keyword) IMap.t;
+  expr_equiv : expr IMap.t;
   return_type : ty option;
 }
 
@@ -58,15 +59,15 @@ module PPEnv = struct
       (PP.pp_print_seq ~pp_sep pp_print_string)
       (ISet.to_seq s)
 
-  let pp_local f { constant_values; storage_types; return_type } =
+  let pp_local f { constant_values; storage_types; return_type; expr_equiv } =
     fprintf f
       "@[<v 2>Local with:@ - @[constants:@ %a@]@ - @[storage:@ %a@]@ - \
-       @[return type:@ %a@]@]"
+       @[return type:@ %a@]@ - @[expr equiv:@ %a@]@]"
       (pp_map PP.pp_literal) constant_values
       (pp_map (fun f (t, _) -> PP.pp_ty f t))
       storage_types
       (pp_print_option ~none:(fun f () -> fprintf f "none") PP.pp_ty)
-      return_type
+      return_type (pp_map PP.pp_expr) expr_equiv
 
   let pp_subprogram f func_sig =
     fprintf f "@[<hov 2>%a@ -> %a@]"
@@ -116,6 +117,7 @@ let empty_local =
     constant_values = IMap.empty;
     storage_types = IMap.empty;
     return_type = None;
+    expr_equiv = IMap.empty;
   }
 
 let empty_local_return_type return_type = { empty_local with return_type }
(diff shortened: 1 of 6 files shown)


asllib/StaticEnv.ml
```diff
@@ -218,6 +225,15 @@ let add_local x ty ldk env =
       };
   }
 
+let add_local_immutable_expr x e env =
```
Roman-Manevich, line 228: Perhaps add a comment that it is **assume** that `e` already refers to the oldest expressions?
Actually, maybe it's better to take `e` and substitute existing expressions for its variables here instead of in `annotate_local_decl_item`?

asllib/Typing.ml
```diff
@@ -571,14 +570,17 @@ module Annotate (C : ANNOTATE_CONFIG) = struct
         match IMap.find_opt s env.global.storage_types with
         | Some (_, (GDK_Constant | GDK_Config | GDK_Let)) -> true
         | Some (_, GDK_Var) -> false
-        | None -> undefined_identifier loc s)
+        | None -> (* Undefined identifier, but we will raise it later *) false)
```
Roman-Manevich, line 573: This is a bit fishy. Why not catch it here? It's safer. What if elsewhere the code changes and doesn't catch it anymore?

asllib/Typing.ml
```diff
@@ -1936,15 +1943,24 @@ module Annotate (C : ANNOTATE_CONFIG) = struct
...
         (* Rule LCFD: A local declaration shall not declare an identifier
            which is already in scope at the point of declaration. *)
         let+ () = check_var_not_in_env loc env x in
-        let new_env = add_local x ty ldk env in
+        let env2 = add_local x ty ldk env in
+        let new_env =
+          match e with
+          | None -> env2
+          | Some e ->
+              let () = assert (ldk = LDK_Let || ldk = LDK_Constant) in
```
Roman-Manevich, line 1961: I think we need to check that `e` is statically evaluable.

```
var x = 3;
let y = x;
let z = y; // z is equal to z here
x = 4;
let w = y; // w is not equal to z here
```
I guess you take care of this in `S_Decl`?
If so, isn't it better to do it here and avoid having to pass `?e` and re-check that `ldk` is `LDK_LET` or `LDK_Constant`?

asllib/Typing.ml
```diff
@@ -1961,6 +1977,7 @@ module Annotate (C : ANNOTATE_CONFIG) = struct
         let new_env, new_ldis =
           List.fold_right2
             (fun ty' ldi' (env', les) ->
+              (* TODO: propagate epxressions? *)
```
Roman-Manevich, line 1980: I think that if you perform this inside the static environment then the `annotate_local_decl_item` below would do the right thing.

asllib/Typing.ml
```diff
@@ -1983,16 +2000,15 @@ module Annotate (C : ANNOTATE_CONFIG) = struct
         in
         (new_env, LDI_Typed (new_ldi', t')) |: TypingRule.LDUninitialisedTyped
 
-  let declare_local_constant loc env t_e v ldi =
+  let declare_local_constant env v ldi =
```
Roman-Manevich, line 2003: I'm not sure what this change is doing. Why did you remove `t_e`?

Roman-Manevich (approved):

asllib/Typing.ml
Roman-Manevich: This is related to the other comment about making sure that expressions always refer to the earliest variables. I thought it would be good to have just one function where immutable expressions are added to the environment and replacement of variables by their equivalent expressions happens. This ensures that if you always use this function the invariant of referring to the oldest expressions holds.

Roman-Manevich (approved):

PR #896, opened 2024-07-05
[Asl] Explode constant intervals before multiplication

HadrienRenaud:
This PR implements a transformation of intervals into sets before doing a multiplication in order not to lose any precision on those. This can only be done for constant sets and up to a certain point, here the maximum set size for which this is done is `1 lsl 31`, or a bit more than 1 billion.


Roman-Manevich (approved):
LGTM

PR #898, opened 2024-07-09
[format] Bump ocamlformat to a version compatible with ocaml 5.2

HadrienRenaud:
Ocamlformat version 0.26.1 is not compatible with ocaml 5.2 ([source](https://github.com/ocaml/opam-repository/blob/master/packages/ocamlformat/ocamlformat.0.26.1/opam)), but ocaml version 0.26.2 is ([source](https://github.com/ocaml/opam-repository/blob/master/packages/ocamlformat/ocamlformat.0.26.2/opam)), so as the bump is trivial (no diff on any source file), we can bump it to this version.

### DIFF
--- a/.github/workflows/check-format.yml
+++ b/.github/workflows/check-format.yml
@@ -28,6 +28,6 @@ jobs:
         with:
           ocaml-compiler: 5.1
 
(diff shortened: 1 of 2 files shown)


relokin (approved):
LGTM!

PR #900, opened 2024-07-11
[asl] Fix the non-determinstic ASL primitive `somebool`

maranget:
Before, the evaluation of the `somebool` primitive by "ASLSem" was yielding no candidate, because of some unsolved equations. Now, we have two candidates, as expected.

As an example the following litmus test now yields two candidates:
```
ASL non-deterministic

{ }

// Non deterministically returns 1 or 2
func g() => integer
begin
  let b = SomeBoolean();
  if b then
    return 1;
  end
  return 2;
end

func main() => integer
begin
  let z = g();
  return 0;
end

locations [0:main.0.z;]
```

### DIFF
--- a/herd/ASLSem.ml
+++ b/herd/ASLSem.ml
@@ -431,7 +431,7 @@ module Make (C : Config) = struct
 
     (*
      * Add equation, the effect will be silent
-     * discard of executon candidate if a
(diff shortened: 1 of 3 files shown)


maranget:
Merged, as being a (small) bug fix.

PR #901, opened 2024-07-11
[asl] Fix definition of Lowest Common Ancestor on tuples.

Roman-Manevich:
Removed the requirement that elements of types must type-satisfy each other.
This means that the type of `if cond then (0, 63) else (0, 31)` is `(integer{0}, integer{63,31})`.


jalglave (approved):
LGTM

PR #904, opened 2024-07-19
[all]: Introduce basic support for SME

murzinv:
Arm Scalable Matrix Extension (SME) is an architecture extension that provides enhanced support for matrix operations. SME builds on the Scalable Vector Extensions (SVE and SVE2), adding new capabilities to efficiently process matrices. Key features include:
 - Outer product between two vectors
 - Matrix tile storage
 - Load, store, insert, and extract of tile vectors, including on-the-fly transposition
 - Streaming SVE mode

Like SVE, SME is a scalable vector length extension which enables Vector Length Agnostic (VLA).

SME comes with two major changes:
 - Streaming SVE mode - implements a subset of the SVE/SVE2 instruction set and adds new SME-specific instructions.
 - ZA array - 2D square byte array of variable length

Most SME instructions are only available in Streaming SVE mode. The streaming vector length (SVL) in Streaming SVE mode might be different from the Non-streaming vector length (NSVL).

The ZA array can be accessed as:
 - ZA array vectors
 - ZA tiles
 - ZA tile slices

Implementation choices and known limitations:

 - SME memory model is not addressed
 - `herd7` doesn't support different vector lengths for NSVL and SVL
 - `herd7` doesn't check if instruction is prmitted in Streaming Mode
 - `litmus7` requires GCC 14+ or Clang 18+ to build SME instructions

 - Following SVE instructions are implemented:
   + LD1B (scalar plus scalar, tile slice)
   + LD1H (scalar plus scalar, tile slice)
   + LD1W (scalar plus scalar, tile slice)
   + LD1D (scalar plus scalar, tile slice)
   + LD1Q (scalar plus scalar, tile slice)
   + ST1B (scalar plus scalar, tile slice)
   + ST1H (scalar plus scalar, tile slice)
   + ST1W (scalar plus scalar, tile slice)
   + ST1D (scalar plus scalar, tile slice)
   + ST1Q (scalar plus scalar, tile slice)
   + MOVA (tile to vector, single)
   + MOVA (tile to vector, single)
   + ADDHA
   + ADDVA
   + SMSTART
   + SMSTOP


gen/AArch64Arch_gen.ml
```diff
@@ -884,8 +902,11 @@ include
       let free_registers = allowed_for_symb
 
       type special = reg
+      type special2 = reg
```
hugookeeffe, line 905: why do we need the type special2 here? As it is the same as special, can we not just use special

hugookeeffe (commented):
Hi @murzinv thanks for the PR! I have one question, but the generation code looks good!

maranget:
Hi @murzinv. I have rebased this PR on top of master and resolved the conflicts. May I force-push onto your branch?

maranget:
Hi again @murzinv, I have installed gcc 14.1.0. I am using qemu for testing, having no idea that this setting is adequate.

Anyway there are significant differences between **herd7** and **litmus7** + **qemu**. More precisely, output differ on tests V06, V07, V08, V09, V12, V15, V16. I here is the **litmus7** + **qemu** output for these tests: [run.txt](https://github.com/user-attachments/files/16442759/run.txt), Of course the difference may originate from different size parameters. However as I do note quite understand this SME extension I draw your attention on these differences.

maranget:
Hum there may be a problem with installed **qemu**. Consider the following test:
```
AArch64 W03
{
  uint32_t x[4];
  0:X0=x;
}
P0 ;
SMSTART                       ;
PTRUE P0.S,VL4                ;
PTRUE P1.S                    ;
MOV W12,#0                    ;
INDEX Z0.S,#1,#1              ;
ADDHA ZA0.S,P1/M,P1/M,Z0.S    ;
ST1W {ZA0H.S[W12,0]},P0,[X0]  ;
SMSTOP                        ;

forall x={1,2,3,4}
```
As far as I understand , the final value of `x` is `{1,2,3,4}` and **herd7** agrees. However, **litmus7**+**qemu** give a final value of `{0,0,0,0}` for `x`.

maranget:
Hi @murzin, I have installed **qemu** from git sources and nothing changed: I have the same results on the tests as before. However, I have better results on some other tests. Stay tuned.

maranget:
I have a remark on lengths. As the implementation requires the SVE end SME to be equal, wouldn't be more convenient for `-variant sme:<n>` to also set the SVE vector lenght in addition to the SME vector length?

maranget:
I have doubts on the calculation of offsets into the bit vector that implements the `ZA` storage. Consider this test:
```
AArch64 W02
{
  uint32_t x[4];
  uint32_t y[4];
  0:X0=x;   0:X1=y;
}
P0 ;
SMSTART                      ; (* ZA is full of zero's *)
INDEX Z0.S,#1,#1             ;
PTRUE P0.S,VL4               ;
PTRUE P1.S                   ;
MOV W12,WZR                  ;
MOVA ZA0H.S[W12,0],P1/M,Z0.S ; (* Store {1,2,3,4,...} into first row *)
MOVA ZA0H.S[W12,1],P1/M,Z0.S ; (* Store {1,2,3,4,...} into second row *)
ST1W {ZA0H.S[W12,0]},P0,[X0] ; (* Save first four elts of first row *)
ST1W {ZA0V.S[W12,0]},P0,[X1] ; (* Save first four elts of first column *)
SMSTOP                       ;
locations [x;y;]

forall x={1,2,3,4} /\ y={1,1,0,0}
```
Result is fine with default sizes of 128 bits, but wrong with other sizes:. For instance:
```
% herd7 -variant sme:512,sve:512 W02.litmus
Test W02 Required
States 1
x={1,2,3,4}; y={1,1,5,9};
No
Witnesses
Positive: 0 Negative: 1
Flag Scalable-Matrix-Extensions-is-work-in-progress
Flag Scalable-Vector-Extensions-is-work-in-progress
Condition forall (x={1,2,3,4} /\ y={1,1,0,0})
Observation W02 Never 0 1
Time W02 0.12
Hash=46977a41525c2afb3f673a1d63254dac
```
I may have fixed this and may submit a patch soon.

herd/AArch64Sem.ml
```diff
@@ -391,6 +409,65 @@ module Make
...
       let write_reg_scalable = write_reg_scalable_sz MachSize.S128
 
+      let za_getoffset tile slice idx esize =
+        let dim = scalable_nbits / esize in
+        let tile = V.op Op.Mul (V.intToV tile) (V.intToV (dim*dim)) in
+        let tiles = V.intToV (esize/8) in
+         M.op Op.Mul slice tiles >>= fun v ->
+          M.op Op.Add tile v  >>= fun v ->
+            M.op Op.Add v idx >>= fun v ->
+              M.op Op.Mul v (V.intToV esize)
```
maranget, line 419: Here is my patch, however beware,  the output of tests V12, V15 and V16 have changed.
```suggestion
      let za_getoffset tile slice idx esize =
        (* ZA offset, in bits, see ARM ARM B.1.4.10 "ZA tile access" *)
      let za_getoffset tile slice idx esize =
        let esize_to_shift = function
          | 8 -> 0
          | 16 -> 1
          | 32 -> 2
          | 64 -> 3
          | 128 -> 4
          | _ -> assert false in
        let shift = esize_to_shift esize in
        let mk_shift k = M.op1 (Op.LeftShift k) in
          begin
            begin
              mk_shift shift slice >>=
              M.add (V.intToV tile) >>=
              M.op Op.Mul (V.intToV scalable_nbits)
            end >>|
            mk_shift (shift+3) idx
          end >>= fun (base,idx) -> M.add base idx

```

maranget: Nipicking: one can use `M.op1 (Op.AddK tile)` in place of `M.add (V.intToV tile)`.

maranget:
As we do not handle different vector length values, I'd be in favor of sme:<n> also to set sve vector length. However I heve not looked a lot at differences between streaming and non-streaming vectors. Maybe there is catch there?

maranget:
I have worked yesterday night and would like to submit a few commits. First for efficiency. Namely at the moment the values of `PSTATE.ZA` and `PSTATE.SM` are non determined at startup.As a consequence, executimg  the first `SMSTART` instruction yields several candidate executions. The next commit will add  bindings of those "registers" to zero in the initial semantics environemnt.

maranget:
The next commit is not necessary. However it improves efficiency and facilitates debugging by allowing more operations to be performed at semantics time in place of later, at equation resolution time,

maranget:
What is your opinion about [patch](https://github.com/herd/herdtools7/pull/904#discussion_r1700279308). I have coded it from the ARM ARM (section B1.4.10 "Tile Access"). I am not that confident.

herd/AArch64Sem.ml
maranget: I agree. Both my **qemu** give these results.

herd/AArch64Sem.ml
```diff
@@ -3400,6 +3668,105 @@ module Make
...
+           mova_vt za ri k p z ii
+           >>= nextSet za
+        | I_MOVA_TV (z,p,za,ri,k) ->
+           check_sme inst;
+           mova_tv z p za ri k ii
+           >>= nextSet z
+        | I_ADDA (dir,za,pslice,pelem,z) ->
+           check_sme inst;
+           adda dir za pslice pelem z ii
+           >>= nextSet z
```
maranget, line 3749: Fix typo
```suggestion
           >>= nextSet za
```

maranget: Indeed, the `ADD<H|A>` modify the `ZA` storage.

maranget:
Hi @murzinv, for your convenience I have packed three modifications as commits:
  1. Change tile access this is this [patch](https://github.com/herd/herdtools7/pull/904#discussion_r1700279308) as commit c8e777cd.
  2. Compute `ADD<V|H>` differently. Because a given `S` tile is only a subpart of `ZA` storage, the computation of the new `ZA` must start from its initial value and not from zero. Commit 4bc0b827.
  3. Change tests. Commit 319cebd2.

With 1. and 2. **herd7** output matches hardware (**qemu**) output. Hence the change 3. More precisely tests `V12`, `V15` and `V16` are changed, exactly as you suggest above. Other discrepancies were due to a litmus bug fixed by PR #923.

maranget:
> I'm really eager to learn your approach to debug (both performance and functionality), so I'm wondering if you can do some write up (may be worth diy.inria.fr) using this PR as use case?

Nothing very fancy and I am not remembering every step.

 I have installed **gcc** 14 from sources and **qemu**, of which binary installation proved sufficient in the end, although I have also installed the head version from sources. Then, I compared **herd7** and hardware (_i.e._**qemu**) output on your test suite. Once I have found that some of the discrepancies were due to **litmus7** bug (see PR #923), differences were present in tests `V12`, `V15` and `V16`.

 Those tests have several instructions and I made them smaller, first  focusing first on tile access. I wrote a lot of print debug code, in order to understand which cells of the `ZA` storage were accessed and what values were read and written. It is to be noticed that initially there were a lot of non-determined variables in the output, which I found strange as the small tests I was using were meant to be deterministic. Hence the "efficiency" patches whose benefits  are clearer debug output more than improvements in efficiency. In the end, I read documentation and followed the formulas given there and coded them.

For `ADD<H|V>A` the modifications of the final condition helped a lot, convincing me that there were too many zeros in  **herd7** output, as I already saw thanks to the debug printing code. I found the origin of all those zeros: (the  initialisation to zero) by looking at the code. More precisely while rewriting it to my taste.

As a summary, I tested, used small tests and attempted to understand code logic by debug printing and rewriting part of it. Notice that I did not started from scratch and that your code is well structured, which helps a lot. Finally A very good read on debugging good practices is "The Practice of Programming" by Kernigham and Pike, Addison Wesley.

gen/AArch64Compile_gen.ml
```diff
@@ -80,6 +93,19 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
...
+    | _ -> assert false
+
+    let next_zaslice n st =
+      let dir = match n with
+                | SIMD.SmV -> Vertical
+                | SIMD.SmH -> Horizontal
+                | _ -> assert false
+      in
+      let r,st = A64.alloc_special3 st in
+    with_directon dir (snd r),(fst r),st
```
maranget, line 107: Suggestion, more idiomatic:
```suggestion
      let (i,r),st = A64.alloc_special3 st in
    with_directon dir r,i,st
```

maranget: There may also be a typo `with_directon` -> `with_direction` ?

lib/AArch64Base.ml
```diff
@@ -2059,6 +2177,25 @@ let do_pp_instruction m =
       sprintf "ADDVL %s,%s,%s" (pp_xreg r1) (pp_xreg r2) (m.pp_k k1)
   | I_CNT_INC_SVE (op,r,pat,k) ->
       pp_cnt_inc op r pat k
+(* Scalable Matrix Extention *)
+  | I_LD1SPT (v,r,ri,k,p,ra,idx) ->
+      sprintf "LD1%s {%s[%s,%s]},%s,%s" (pp_simd_variant v) (pp_zareg r) (pp_wreg ri) (m.pp_k k) (pp_preg p) (pp_addr ra idx)
```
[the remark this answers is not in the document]
maranget: Hi @murzinv. The function `do_pp_instruction`  has a polymorphic type,  more precisely `'a basic_pp -> 'a kinstruction -> string`, where `'a` is the type of constant `k`.

 This permits using the same function for different tasks. In particular `'a` can be `int` as in `dump_instruction` below or `MetaConst.t` as in `dump_parsedIntruction`. One possible option to avoid the leading `#` for some constant arguments such as `offs` would be to add a `pp_k_nohash` member to the `basic_pp` record type and to to replace some calls to `m.pp_k` by calls to `m.pp_k_no_hash`. Of course `pp_k_n_hash` would not add the leading `#`.

maranget:
Hi @murzinv . This PR is ready for merge. I'll merge it as soon as you have rebased it on top of master. Thanks a lot for the good work.

PR #906, opened 2024-07-22
[herd] Asl bool condition

maranget:
The choice operator of the "concurrent" ASL interpreted now assigns unresolved condition  to TRUE or FALSE depending on the selected branch.

As a result the "SomeBoolean" primitive can now reduce to a simple non-resolved variable and could have been replaced by the construct `UNKNOWN : boolean`. Typically the following running **herd7** on the following test will result in two outcomes with values `0` and `1` for `x`. Moreover the final values of `b` are `FALSE` and `TRUE`, respectively.
```
ASL U
{}

func main() => integer
begin
  var x = 0;
  let b = UNKNOWN : boolean;
  if b then
    x = 1;
  end
  return 0;
end

locations [0:main.0.x;]
```
Previously, the same result could be achieved by adding option `-variant ASL` or by replacing the `UNKNOWN` expression by a primitive call `SomeBoolean ()`.

Unfortunately this "complete" resolution does not succeed in general. See the example `herd/tests/instructions/ASL/unknown-integer.litmus` for instance.

### DIFF
--- a/herd/ASLSem.ml
+++ b/herd/ASLSem.ml
@@ -262,13 +262,33 @@ module Make (C : Config) = struct
     let commit (ii,poi) msg =
       M.mk_singleton_es (Act.Branching msg) (use_ii_with_poi ii poi)
 
+    let vbool b = V.Val (Constant.Concrete (ASLScalar.S_Bool b))
+
+(* Handle choice from boolean condition *)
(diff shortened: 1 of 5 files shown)


maranget:
Hi @HadrienRenaud, if you could write it with `SomeBoolean()` before, it would work...

PR #907, opened 2024-07-22
[gen] Generate C atomic exchange.

maranget:
The `Rmw` or `Exch` edge produces a primitive call `atomic_exchange_explicit(...,<mo>)`, while `Fetch.Add` will produce a prinitive call to `atomic_fetch_add_explicit(...,<mo>)`.

The memory_order `<mo>` depends on the combination of anotations `<ar>` and `<aw>` from the sequence
`<ar> Rmw <aw>` with:
  + `Sc Rmw Sc` yields `memory_order_seq_cst`
  + `Rlx Rmw Rlx` yields `memory_order_relaxed`
  + `Acq Rmw Rlx` yields `memory_order_acquire`
  + `Rlx Rmw Rel` yields `memory_order_release`
  + `Acq Rmw Rel` yields `memory_order_acq_rel`

Other sequences yield a fatal error.

Try for instance: `diyone7 -arch C Acq Rmw Rel PodWW Rfe PodRW Rlx Rfe`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -641,6 +641,27 @@ diy-test-mte::
 		$(REGRESSION_TEST_MODE)
 	@ echo "herd7 AArch64.MTE diycross7 tests: OK"
 
+diy-test::  diy-test-C
+diy-test-C:
+	@ echo
+	$(HERD_DIYCROSS_REGRESSION_TEST) \
+		-j $(J) \
+		-herd-path $(HERD) \
+		-diycross-path $(DIYCROSS) \
+		-libdir-path ./herd/libdir \
+		-expected-dir ./herd/tests/diycross/C \
+		-conf ./herd/tests/diycross/C/C.cfg \
+		-diycross-arg -arch \
+                -diycross-arg C \
+		-diycross-arg [Rlx,Coe,Rlx],[Rlx,Rfe,Rlx],[Rlx,Fre,Rlx],[Rlx,Hat,Rlx] \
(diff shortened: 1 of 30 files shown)


db7:
This is very nice. I've tried locally and it seems to be working. Is there a plan to do the same for CmpExchange?

maranget:
> This is very nice. I've tried locally and it seems to be working. Is there a plan to do the same for CmpExchange?

Hi @db7, `compare_exchange_...` for C is more involved.

In particular, the **herd7** implementation is not functional.

However, you can write tests by hand and **litmus7** can handle them. For instance consider the test:
```
% cat IncrCmpExch.litmus
C IncrCmpExch

{}

P0 (atomic_int *x) {
  int old;
  int new;
  do {
    old = atomic_load_explicit(x,memory_order_relaxed);
    new = old+1;
  } while
    (!atomic_compare_exchange_weak_explicit
        (x,&old,new,memory_order_relaxed,memory_order_relaxed));  
}

P1 (atomic_int *x) {
  int old;
  int new;
  do {
    old = atomic_load_explicit(x,memory_order_relaxed);
    new = old+1;
  } while
    (!atomic_compare_exchange_strong_explicit
        (x,&old,new,memory_order_relaxed,memory_order_relaxed));  
}

forall [x] = 2
```
One can comple and run this tests as follows:
```
% mkdir -p R
% litmus7 -c11 true  -a 4  -o R IncrCmpExch.litmus
% cd R
% make
% sh ./run.sh
```

PR #908, opened 2024-07-24
[asl] Added StaticInterpreter.mli

Roman-Manevich:
This is in order to avoid shadowing definitions from Instrumentation and to be able to mark code in StaticInterpreter.ml for typing rules. 
Marked static_eval with the new TypingRule.StaticEval


jalglave (approved):
LGTM

PR #914, opened 2024-07-26
[asl] Rework inlining of getter/setters

HadrienRenaud:
This PR improves the handling of getters and setters to support all wanted behaviours.

The only strange case is the following:
```
getter F => bits(3)
begin return '000'; end

func main () => integer
begin
  let -: bits(0) = F[];

  return 0;
end
```
Which now passes, as it's permitted to have 0-length bitvectors.

### DIFF
--- a/asllib/ASTUtils.ml
+++ b/asllib/ASTUtils.ml
@@ -534,6 +534,8 @@ let case_to_conds : stmt -> stmt =
       S_Seq (assign, cases_to_cond x cases)
   | _ -> raise (Invalid_argument "case_to_conds")
 
+let slice_is_single = function Slice_Single _ -> true | _ -> false
+
 let slice_as_single = function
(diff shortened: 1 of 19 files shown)


maranget:
Hi @HadrienRenaud, the following tests yields a typing error:
```
var _z : bits(8) = '11111111';

getter F[] => bits(8)
begin
  return _z;
end

getter F => bits(8)
begin
  return z;
end


func main () => integer
begin
  __DEBUG__(F);
  __DEBUG__(F[]);
  return 0;
end
```
The error spells:
```
File T.asl, line 3, character 0 to line 6, character 3:
ASL Typing error: cannot declare already declared element "F".
```

Is this normal?

maranget:
Fine, I was just curious. I think this PR is ready.  Merging on its way. Thanks @HadrienRenaud.

PR #917, opened 2024-07-29
[gen] bug in check_fetch causing unwanted output

hugookeeffe:
This PR solves a bug where a print statement in check fetch is triggered when using **diy7**, generating unwanted output.

### DIFF
--- a/gen/cycle.ml
+++ b/gen/cycle.ml
@@ -720,8 +720,7 @@ let remove_store n0 =
       begin match m.evt.loc, m.evt.dir with
         | Code.Code _, Some R when not (E.is_ifetch m.edge.E.a1) ->
             if is_read_same_nonfetch m then begin
-              Printf.printf "%s" (str_node m);
-              Warn.user_error "Multiple ifetch reads to same code location"
+              Warn.user_error "Multiple ifetch reads to same code location [%s]" (str_node m)
               end;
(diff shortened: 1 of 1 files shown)


gen/cycle.ml
```diff
@@ -720,8 +720,7 @@ let remove_store n0 =
       begin match m.evt.loc, m.evt.dir with
         | Code.Code _, Some R when not (E.is_ifetch m.edge.E.a1) ->
             if is_read_same_nonfetch m then begin
-              Printf.printf "%s" (str_node m);
```
relokin, line 723: Would this be useful if it was gated by a debug flag?

hugookeeffe: Hi @relokin thanks for comment! I'm not sure that a debug flag is necessary as using Warn.user_error will give the same result (as it will only be printed if debug mode is on, or if using diyone7, in which case it should indeed fail).

relokin: ah I sorry I spoke too fast and didn't realise that you added the call to str_node in the user_error.

relokin (approved):
LGTM. Planning to merge this tomorrow, unless I hear any objections :)

maranget:
Fine by me. Thanks @hugookeeffe and @relokin. FYI. master has just been updated. Would you please rebase on master before merging?

maranget:
Merged, thanks @hugookeeffe and @relokin.

PR #920, opened 2024-07-30
[asl] Change type of index in for loop

HadrienRenaud:
This PR changes the type of the index in a for loop. Before it was a complex operation to determine the maximum and the minimum of the bounds of the loop. Now the type of `i` in `for i = a to b` is simply:
 - `integer {a..b}` if `a` and `b` are well-constrained;
 - the unconstrained integer type otherwise.

`TPositive8` is now split in two:
- `TPositive8` keeps the bulk of the test. ASLRef does not report any error on this part after #926. For now, it still report a type error.
- ASLRef reports a type error on `TPositive8-1`.


Roman-Manevich (approved):
LGTM

PR #937, opened 2024-08-13
[herd]: Demote result of Neon register read for `FMOV`

murzinv:
Consider litmus test

```
AArch64 V70
{
 int x=1;
 0:X0=x; 0:X3=y;
}
P0                 ;
LDR S1,[X3]        ;
FMOV W1,S1         ;
LDR W2,[X0,W1,SXTW];

forall 0:X2=1
```

Currently `herd7` would refuse to run the test with

```
Warning: File "V70.litmus": operation sxt non-existent for wide integers
```

The reason is that `W1` holds `wide` integer value from `S1` which blocks `sxt` operation. General purpose registers don't require `wide` integer and work pretty well with `narrow` one, so `demote` result of Neon register read for `FMOV`

### DIFF
--- a/herd/AArch64Sem.ml
+++ b/herd/AArch64Sem.ml
@@ -3367,8 +3367,8 @@ module Make
               fun v -> write_reg_neon_rep (neon_sz r1) r1 v ii)
         | I_FMOV_TG(_,r1,_,r2) ->
             check_neon inst;
-            !(read_reg_neon false r2 ii >>=
-              fun v -> write_reg r1 v ii)
+            read_reg_neon false r2 ii >>= demote
+            >>= fun v -> write_reg_dest r1 v ii >>= nextSet r1
         | I_MOV_VE(r1,i1,r2,i2) ->
             check_neon inst;
             !(read_reg_neon_elem false r2 i2 ii >>=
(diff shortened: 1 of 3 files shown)


herd/AArch64Sem.ml
```diff
@@ -3363,7 +3363,7 @@ module Make
               fun v -> write_reg_neon_rep (neon_sz r1) r1 v ii)
         | I_FMOV_TG(_,r1,_,r2) ->
             check_neon inst;
-            !(read_reg_neon false r2 ii >>=
+            !(read_reg_neon false r2 ii >>= demote >>=
               fun v -> write_reg r1 v ii)
```
maranget, line 3367: Hi @murzin. We may seize the opportunity to perform the store to register now, resulting in one equation less:
```suggestion
            read_reg_neon false r2 ii >>= demote
             >>= fun v -> write_reg_dest r1 v ii >>= nextSet r1
```
Notice that other Neon instruction could undergo a similar "optimisation". This in particular applies to other kinds of move instructions. Nothing urgent, though: just seizing the opportunity...

maranget:
Hi @murzinv. Nice catch. Please notice the suggestion above. The PR can be merged as soon as rebased on master, if needed.

PR #922, opened 2024-08-01
[aarch64,cat] a fix to tagcheck/translation orderings to restore intent

artkhyzha:
The AArch64 cat model is being fixed to restore architectural intent for the hardware requirement arising from address translations and tag checks.

1) `[Imp & TTD & R]; tr-ib; po; [CSE]; po` is added into `CSE-ob` so that the following litmus test changes its verdict to "Forbidden".

```
AArch64 MP+dmb.st+trib-po-isb-po
Variant=vmsa,fatal
{
TTD(y)=(oa:PA(y),valid:0);
0:X1=x; 0:X4=(oa:PA(y),valid:1); 0:X5=PTE(y);
1:X1=x; 1:X3=y;
}
P0          |P1          ;
 MOV W0,#1  | LDR W2,[X3];
 STR W0,[X1]|L0:         ;
 DMB ST     | ISB        ;
 STR X4,[X5]| LDR W0,[X1];
exists ~Fault(P1,y) /\ 1:X0=0
```

2) `[Imp & Tag & R]; tc-ib; [CSE]; po` is added into `CSE-ob` so that the following litmus test changes its verdict to "Forbidden".

```
AArch64 MP+dmb.stpt+tcib-tagfault-po
Variant=memtag,sync
{
0:X1=x; 0:X3=y:green; 0:X2=y:red;
1:X1=x; 1:X3=y:green;
}
P0          | P1          | P1.F        ;
MOV W0,#1   | LDR W2,[X3] | LDR W0,[X1] ;
STR W0,[X1] |             |             ;
DMB ST      |             |             ;
STG X2,[X3] |             |             ;
exists Fault(P1,y) /\ 1:X0=0
```

3) `[Imp & Tag & R]; iico_data; [B]; iico_ctrl; [((Exp & M) \ Tag) | TagCheck & FAULT]; po; [CSE]; po` is added into `CSE-ob` so that the following litmus test changes its verdict to "Forbidden".

```
AArch64 MP+dmb.stpt+tcib-po-isb-po
Variant=memtag,sync
{
0:X1=x; 0:X3=y:green; 0:X2=y:red;
1:X1=x; 1:X3=y:red; 1:X2=1;
}
P0          | P1          ;
MOV W0,#1   | LDR W2,[X3] ;
STR W0,[X1] |L0:          ;
DMB ST      | ISB         ;
STG X2,[X3] | LDR W0,[X1] ;
exists ~Fault(P1,y) /\ 1:X0=0
```

[...]

PR #946, opened 2024-08-23
[catalogue,aarch64] correcting two VMSA tests to have a fault handler

artkhyzha:
This PR corrects two litmus tests in the catalogue that have instructions after a faulting instruction. Such tests benefit from an explicit fault handler.

Note that kinds for `CASAL+FAIL+FH` are different from those for `CASAL+FAIL`: the test should be forbidden in line with #922.


relokin:
@artkhyzha this has been superseded by #952, hasn't it? Can we close it?

artkhyzha:
Closing as suggested: indeed this became a part of another PR (#952).

PR #949, opened 2024-09-03
[herd] Option `-debug` accepts several, comma separated, arguments.

maranget:
In place of `-debug solver -debug mem`, one can now write `-debug solver,mem`.


maranget:
> Look good to me!

Thanks @HadrienRenaud.
 
> If you fancy a bit of documentation, the `.mli` is quite obscure.

Nice suggestion. I tried.

PR #950, opened 2024-09-03
[asl] Add missing EOF in cram tests

yuxiliu-arm:
Otherwise, `make test` may generate a diff ``+  warning: here-document at line N delimited by end-of-file (wanted `EOF')`` for some shells.

### DIFF
--- a/asllib/tests/atcs.t
+++ b/asllib/tests/atcs.t
@@ -4,6 +4,7 @@ Deferred to execution ATCs
   >   let x = (3 as integer {42});
   >   return 0;
   > end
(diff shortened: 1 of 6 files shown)


yuxiliu-arm:
Hi, I'm assuming `dune` cram test uses the encompassing shell, which for me is:

```
$ bash --version
GNU bash, version 5.2.32(1)-release (aarch64-apple-darwin23.6.0)
```

I wasn't getting the error on a Linux machine before. I don't have the exact shell version of that at the moment.

PR #951, opened 2024-09-04
[all] Add support for MADD, MUL, MSUB, MNEG

yuxiliu-arm:
Multiply add and sub.


herd/tests/instructions/AArch64/Y001.litmus
```diff
@@ -0,0 +1,18 @@
+AArch64 Y001
+(* MADD *)
+
+{
+    0: X1 = 2;
+    0: X2 = 3;
+    0: X3 = 5;
```
maranget, line 7: Hi @yuxiliu-arm. Nice PR.

One slight remark: when you are using the `X` form of register for scalars, it is better to declare them as 64bit quantities:
```suggestion
    int64_t 0: X1 = 2;
    int64_t 0: X2 = 3;
    int64_t 0: X3 = 5;
    int64_t 0:X4;
```
The reason is as follows: by default **herd7** and **litmus7** assume that printed scalar quantities are 32bits (type int) by default. As test Y001 prints the four registers from `0:X1` to `0:X4`, it will assume type `int`, while the code uses them as 64bits quantities. As a result higher order bits will not appear in output while using `X` registers suggests not to ignore them. For instance consider this test
```
AArch64 N
{
  0:X0=1;
}
  P0           ;
 LSL X0,X0,#32 ;

locations [0:X0;]
```
Ouput of will be `0:X0=0;` while on would probably expect `0:X0=4294967296;`

Notice that **herd7** does not warn you about this. By contrast **litmus7** does:
```
% litmus7 -mach aarch64 -o /tmp/N.tar  N.litmus 
Warning: File "N.litmus" Register x0 has different types: <int> and <int64_t>
```
It does so because your code correctly assigns the right type to the register arguments of the new instructions --- see `reg_env = add_v v (rd@rs@ra);` in function `mop` in litmus/AArch64Compile.ml` below.

herd/tests/instructions/AArch64/Y002.litmus
```diff
@@ -0,0 +1,16 @@
+AArch64 Y002
+(* MUL *)
+
+{
+    0: X1 = 2;
+    0: X2 = 3;
```
maranget, line 6: As above:
```suggestion
    int64_t 0: X1 = 2;
    int64_t 0: X2 = 3;
    int64_t 0:X3;
```

herd/tests/instructions/AArch64/Y001.litmus
yuxiliu-arm: Thank you for the detailed explanation. It's fixed and rebased.

maranget:
Merged, thanks @yuxiliu-arm.

PR #952, opened 2024-09-04
[aarch64,cat] Add FEAT_ETS3 and reorganise relations as in Arm ARM issue L.a

artkhyzha:
This release presents a number of changes reflected in the Arm ARM issue L.a. For more details please contact memory-model@arm.com.

This version of the model:
* reorganises the ordering rules to clarify the architectural intent and enhance readability
    - CSE Effects are renamed into Instruction Fetch Barrier (IFB) Effects to distinguish a relevant subset of the semantics of Context Synchronisation Events
    - endpoint constraints of `local-hw-reqs` are propagated into the relations defining it (most notably, into the `lob`)
* further elaborates the architectural intent for ordering implications of exception entries (in addition to previously established rules in `CSE-ob`)
* introduces new ordering requirements of the ETS3 feature, to provide order to Implicit TTD Memory Read Effects due to address translations and Implicit Tag Memory Read Effects due to tag checks of memory accesses resulting in a synchronous fault.


catalogue/aarch64-VMSA/tests/CASAL+FAIL+FH.litmus
```diff
@@ -0,0 +1,18 @@
+AArch64 CASAL+FAIL+FH
+Variant=handled
```
relokin, line 2: Can we remove the handled variant? Also add vmsa.

catalogue/aarch64-VMSA/tests/CASAL+FAIL2+FH.litmus
```diff
@@ -0,0 +1,24 @@
+AArch64 CASAL+FAIL2+FH
+Variant=handled
```
relokin, line 2: ditto

relokin (commented):
For all tests, my suggestion would be to pass them through mprog. For example, for CASAL+FAIL+FH:
$> mprog7 -mode text CASAL+FAIL+FH
will give you a canonical form.

herd/libdir/aarch64hwreqs.cat
```diff
@@ -56,104 +56,124 @@ let tc-ib = if "MTE" then [Imp & Tag & R]; iico_data; [B]; iico_ctrl; [((Exp & W
...
+ | [Exp & R]; addr; [Exp & M]; po; [CSE]; po
+ | [Exp & R]; pick-addr-dep; [Exp & M]; po; [CSE]; po
  | DSB-ob; [CSE]; po
+ | [Imp & Instr & R]; f-ib; [CSE]; po
  | [Imp & TTD & R]; tr-ib; [CSE]; po
  | [Imp & TTD & R]; tr-ib; po; [CSE]; po
  | [Imp & Tag & R]; tc-ib; [CSE]; po
- | [Imp & Tag & R]; iico_data; [B]; iico_ctrl; [((Exp & M) \ Tag) | TagCheck & FAULT]; po; [CSE]; po
- | [Imp & Instr & R]; f-ib; [CSE]; po
+ | [Imp & Tag & R]; iico_data; [B]; iico_ctrl; [((Exp & M) \ Tag) | TagCheck & FAULT & Sync-EXC]; po; [CSE]; po
```
relokin, line 85: I think you don't need the `\ Tag`, LDG, STG, etc are unchecked instructions. Also I think, this part of the clause:
```
 | [Imp & Tag & R]; iico_data; [B]; iico_ctrl; [TagCheck & FAULT & Sync-EXC]; po; [CSE]; po
```
is already included in
```
| [Imp & Tag & R]; tc-ib; [CSE]; po
```
so I think you can simplify this to
```
[Imp & Tag & R]; iico_data; [B]; iico_ctrl; [Exp & M]; po; [CSE]; po
```
or even
```
[Imp & Tag & R]; po; [CSE]; po
```
Can't you?

herd/libdir/aarch64hwreqs.cat
```diff
@@ -56,104 +56,124 @@ let tc-ib = if "MTE" then [Imp & Tag & R]; iico_data; [B]; iico_ctrl; [((Exp & W
...
 (* Dependency-ordered-before *)
-let dob = addr 
-        | data
-        | ctrl; [W]
-        | addr; [Exp]; po; [W]
-        | addr; [Exp]; po; [CSE]; po; [R]
-        | addr; [Exp]; lrs
-        | data; [Exp]; lrs
+let dob =
+ addr; [Exp & M | Imp & Tag & R | Imp & TTD & R | HU | TLBI | DC.CVAU | IC]
```
relokin, line 89: Isn't this all of addr? Is anything excluded?

herd/libdir/aarch64hwreqs.cat
```diff
@@ -56,104 +56,124 @@ let tc-ib = if "MTE" then [Imp & Tag & R]; iico_data; [B]; iico_ctrl; [((Exp & W
...
+ | data; [Exp & W]
+ | ctrl; [Exp & W | HU | TLBI | DC.CVAU | IC]
+ | addr; [Exp & M]; po; [Exp & W | HU]
+ | addr; [Exp & M]; lrs; [Exp & R | Imp & Tag & R]
+ | data; [Exp & M]; lrs; [Exp & R | Imp & Tag & R]
+ | [Imp & TTD & R]; tr-ib; po; [Exp & W]
+ | [Imp & Tag & R]; iico_data; [B]; iico_ctrl; [((Exp & M) \ Tag) | TagCheck & FAULT & Sync-EXC]; po; [Exp & W]
```
relokin, line 96: same comment as above.

herd/libdir/aarch64hwreqs.cat
```diff
@@ -56,104 +56,124 @@ let tc-ib = if "MTE" then [Imp & Tag & R]; iico_data; [B]; iico_ctrl; [((Exp & W
...
+
+let Sync-EXC = EXC-ENTRY | EXC-RET
+
 let CSE-ob =
  [Exp & R]; ctrl; [CSE]; po
+ | [Exp & R]; pick-ctrl-dep; [CSE]; po
+ | [Exp & R]; addr; [Exp & M]; po; [CSE]; po
+ | [Exp & R]; pick-addr-dep; [Exp & M]; po; [CSE]; po
  | DSB-ob; [CSE]; po
+ | [Imp & Instr & R]; f-ib; [CSE]; po
```
relokin, line 81: This is included in po-fetch-ob, isn't it?

herd/libdir/aarch64deps.cat
```diff
@@ -62,30 +65,29 @@ let rec dtrm =
 let ADDR = Rreg \ DATA
 let basic-dep =
    [Exp & R | Rreg]; dtrm?
-let data = (basic-dep; [DATA]; iico_data+; [W]) \ same-instance
-let addr = [Exp & M]; (basic-dep; [ADDR]; iico_data+; [M | TLBI | IC.IVAU | DC.CVAU]) \ same-instance
-let ctrl = (basic-dep; [BCC]; po) \ same-instance
+let data = [Exp & R]; (basic-dep; [DATA]; iico_data+; [W]) & ~same-instance
+let addr = [Exp & R]; (basic-dep; [ADDR]; iico_data+; [M | TLBI | IC.IVAU | DC.CVAU]) & ~same-instance
+let ctrl = [Exp & R]; basic-dep; [BCC]; po
```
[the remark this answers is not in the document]
artkhyzha: I dropped `~same-instance` from the control dependency because the definition includes a `po` edge. There is no way for `ctrl` to connect to effects internal to one instruction instance. That said, it can be returned for presentation reasons.

herd/libdir/aarch64deps.cat
```diff
@@ -62,30 +65,29 @@ let rec dtrm =
...
-   pick-basic-dep; [ADDR]; iico_data+; [M]
+   [Exp & R]; (pick-basic-dep; [ADDR]; iico_data+; [M | TLBI | IC.IVAU | DC.CVAU]) & ~same-instance
 let pick-data-dep =
-   pick-basic-dep; [DATA]; iico_data+; [W]
-let pick-addr-dep =
-   pick-basic-dep; [ADDR]; iico_data+; [M]
+   [Exp & R]; (pick-basic-dep; [DATA]; (iico_data|iico_ctrl)+; [W]) & ~same-instance
 let pick-ctrl-dep =
-   pick-basic-dep; [BCC]; po
+   [Exp & R]; pick-basic-dep; [BCC]; po
```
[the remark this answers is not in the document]
artkhyzha: Indeed, same as with `ctrl`, presence of `po` should make the use of `~same-instance` redundant.

herd/libdir/aarch64deps.cat
```diff
@@ -46,7 +46,10 @@ catdep (* This option tells that the cat file computes dependencies *)
 include "aarch64util.cat"
 
 (* Local write successor *)
-let lws = [M]; (po & same-loc); [W | MMU & FAULT]
+let lws =
```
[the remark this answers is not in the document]
artkhyzha: It is intentional in the sense that the clause `[Exp & M]; po & same-loc; [HU]` was not in the current model in the first place, so it is not being excluded.

----

In the kinds file for `aarch64-VMSA`, this following test is marked as Allowed. Adding the clause you propose would change the verdict.

```
AArch64 coWWExpNExp
TTHM=HD
{
int x; int y;
0:X1=pte_x;
pte_x=(oa:phy_x,db:0,dbm:1,valid:1);
0:X2=(oa:phy_y,db:0,dbm:1,valid:1);
0:X4=x;
}
P0         ;
STR X2,[X1];
MOV W3,#1  ;
STR W3,[X4];
exists(x=1 /\ pte_x=(oa:phy_y,db:0,dbm:1))
```

herd/libdir/aarch64deps.cat
```diff
@@ -46,7 +46,10 @@ catdep (* This option tells that the cat file computes dependencies *)
...
+ [Exp & M | Imp & Tag & R]; (po & same-loc); [Exp & W]
```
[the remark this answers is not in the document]
artkhyzha: I am not aware of arguments either way. I conjecture that the switch from `po-loc` to `po & same-loc` happened before the release of the K.a, because the latter phrasing was found to be easier to transliterate automatically. Perhaps, the latter should be preferred then?

herd/libdir/aarch64hwreqs.cat
artkhyzha: Yes, this is all of `addr` as currently defined.

artkhyzha: I suppose, the list of endpoint constraints could be moved into the definition of `addr` itself, if that helps.

herd/libdir/aarch64deps.cat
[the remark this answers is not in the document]
relokin: yes that makes sense, thanks!

herd/libdir/aarch64deps.cat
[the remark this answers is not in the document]
relokin: Great, thanks for checking!

herd/libdir/aarch64deps.cat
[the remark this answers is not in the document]
relokin: Good point, I wonder if we could require a lrs to avoid the ~same-instance in data and addr, but let's leave that for now.

herd/libdir/aarch64hwreqs.cat
relokin: Maybe that would improve the readability of the graphs too, what do you think? Whichever option we choose here should apply for data and ctrl.

herd/libdir/aarch64hwreqs.cat
artkhyzha: I like the second-to-last simplification. It is a bit sneaky, in that it relies on the instruction semantics in async/asym mode not to draw the `iico_ctrl` edge to the memory effect, and on `CSE` effects being synchronous exceptions. However, those are the matters of definition and strike me as completely adequate. I'd like to use this phrasing.

----

I like the very last suggestion too, as it looks very clean, but it does not allow for discriminating between sync/async/asym fault handling mode, does it? Not every Implicit Tag Read enjoys this order, but only the one that leads to a synchronous Tag Check fault exception. I suppose, precisely that can be defined as a filter on `Imp & Tag & R` rather than on Fault effects. Namely, we could do:
```
let MayRaiseSyncExc = domain([Imp & Tag & R]; iico_data; [B]; iico_ctrl; [Exp & M | TagCheck & EXC-ENTRY])
```
and adjust your last suggestion to
```
[Imp & Tag & R & MayRaiseSyncExc]; po; [CSE]; po
```

To attach this point to a litmus test, consider this:
```
AArch64 MP+dmb.stpt+tcib-po-isb-po--async
Variant=memtag,async
{
0:X1=x; 0:X3=y:green; 0:X2=y:red;
1:X1=x; 1:X3=y:red; 1:X2=1;
}
P0          | P1          ;
MOV W0,#1   | LDR W2,[X3] ;
STR W0,[X1] |L0:          ;
DMB ST      | ISB         ;
STG X2,[X3] | LDR W0,[X1] ;
exists ~Fault(P1,y) /\ 1:X0=0
```
The last suggestion would forbid this, although it only needs to be forbidden in sync mode. (And async mode, I guess.)

herd/libdir/aarch64hwreqs.cat
artkhyzha: Good point. Yes, I agree -- it is included by construction.

herd/libdir/aarch64hwreqs.cat
relokin: Makes sense, thanks!

In that case, I think I vote for:
```
[Imp & Tag & R]; iico_data; [B]; iico_ctrl; [Exp & M]; po; [CSE]; po
```
and I think it might be easier to transliterate but it's up to you. And you can also add a comment why this works (ie., in async there is no iico_ctrl).

herd/libdir/aarch64hwreqs.cat
artkhyzha: An advantage of moving the endpoint constraints for `addr`, `data` and `ctrl` closer to their definitions is that the next time the definition extends to new events, there is one place to edit it, not both the definition and the `dob` like now -- which is a readability/maintainability win for the transliteration of the rules.

The disadvantage is that we do need to define dependencies more generally, and use them in more specific ways:
* for `ctrl`, the `dob` does not provide order to`CSE` effects, but we do use `ctrl;[CSE];po` elsewhere.
* for `pick-addr-dep`, the `pob` does not provide order to `Exp & R`, but `pick-addr-dep; [Exp & M]; po; [Exp & W]` makes use of it.

Hence, to preserve the same behaviour we'd need to:
* Move the endpoint constraints for `ctrl` into its definition, extend them with `CSE`, but let `ctrl; [~CSE]` in `dob`. Admittedly, there are no tests that tell the difference between having or not having `CSE` there.
* Move the endpoint constraints for `pick-addr-dep` into its definition, extend them with `Exp & R`, but let `pick-addr-dep; [~(Exp & R)]` in `pob`. The need for this is illustrated by `catalogue/aarch64-pick/tests/T3-bis.litmus`, which is forbidden.

catalogue/aarch64-VMSA/tests/VMSA-kinds.txt
```diff
@@ -773,3 +773,214 @@ NT-33-tlbi+dsb                                 Allowed
...
+MP+dmb.st+addrisb-Imp_TTD_R            Forbid
+RWRR+HU-dmb.sy+acq                     Allow
+RWRR+HU-dsb.sy+acq                     Forbid
+RWRR+HU-rel+acq                        Allow
+S-I2V+dmb.st+ctrl_Imp_TTD_R            Allow
+S-I2V+dmb.st+data_Imp_TTD_R            Allow
+S+dmb.st+addr-HUaf                     Allow #
+S+dmb.st+addr-HUdb                     Forbid #
+S+dmb.st+ctrl-HUdb                     Forbid
+S+dmb.st+data-HUaf                     Allow # Observed on d06
```
relokin, line 824: ditto

catalogue/aarch64-VMSA/tests/order-via-faults/aob/@all
relokin, line 1: is this intentionally empty? Maybe we could remove it?

catalogue/aarch64/tests/kinds.txt
relokin, line 1: This hasn't changed, has it? Could maybe revert the changes?

catalogue/aarch64-VMSA/tests/CASAL+FAIL+FH.litmus
relokin: Just marking this unresolved to bring it back to your attention. I don't mind as much the test not having the variant vmsa but now handed is going to apply only for faults in P0 which I think is not the intention is it?

tools/miaou.ml
```diff
@@ -440,8 +440,14 @@ module Make
...
          Some
            (Item
-              (sprintf "\\expandafter\\MakeUppercase\\notthecase{%s}" txt))
-      | List _|DiffPair _|IfCond _ ->
+              (sprintf "\\expandafter{\\MakeUppercase\\notthecase{%s}}" txt))
+      | List (op,intro_txt,sep_txt,es) ->
+         (* eprintf "~~~~~~~~~~~~~~~~~~~~\n"; *)
+         (* eprintf "%s\n" (sprintf "\\expandafter{\\MakeUppercase\\notthecase{%s}}" intro_txt); *)
+         (* eprintf "%s\n" (List.fold_left (fun acc e -> acc ^ "," ^ (pp_txt e)) " " es) ; *)
+         (* eprintf "~~~~~~~~~~~~~~~~~~~~\n"; *)
```
relokin, line 448: It looks like this was used for debugging can we remove it now? Or otherwise add that as code behind some debug flag?

tools/miaou.ml
```diff
@@ -590,9 +590,11 @@ module Make
     and pp_txt indent s = function
       | Item txt
       | List (_,_,_,[Item txt]) ->
+         (* eprintf "%s\\item %s%s:\n" indent txt s ; *)
```
relokin, line 593: ditto

tools/miaou.ml
```diff
@@ -590,9 +590,11 @@ module Make
...
          printf "%s\\item %s%s\n" indent txt s
       | List (_,txt,(s1,s2),txts) ->
          printf "%s\\item %s:\n" indent (Misc.capitalize txt) ;
+         (* eprintf "123%s\\item %s:\n" indent txt ; *)
```
relokin, line 597: ditto

catalogue/aarch64-MTE/shelf.py
relokin, line 1: Remind was the problem that the regressions complain if the tests are missing from the shelf? Is adding them as comment helpful?

relokin (commented):
Thanks Artem. The PR looks good to me. I have a added a couple of minor comments and I want to ask if we could move or replicate the comment you have in PR also in the commit that changes the cat file? Should also in the title of the commit that these changes are for the Arm ARM issue L.a?

catalogue/aarch64-VMSA/tests/VMSA-ETS2-kinds.txt
```diff
@@ -773,3 +773,211 @@ NT-33-tlbi+dsb                                 Allowed
...
+MP+dmb.st+addrisb-Imp_TTD_R            Forbid
+RWRR+HU-dmb.sy+acq                     Allow
+RWRR+HU-dsb.sy+acq                     Forbid
+RWRR+HU-rel+acq                        Allow
+S-I2V+dmb.st+ctrl_Imp_TTD_R            Allow
+S-I2V+dmb.st+data_Imp_TTD_R            Allow
+S+dmb.st+addr-HUaf                     Allow #
+S+dmb.st+addr-HUdb                     Forbid #
+S+dmb.st+ctrl-HUdb                     Forbid
+S+dmb.st+data-HUaf                     Allow # Observed on d06
```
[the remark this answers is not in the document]
artkhyzha: Fixed now, marking resolved.

catalogue/aarch64-VMSA/tests/order-via-faults/aob/@all
artkhyzha: Fixed now, marking resolved.

tools/miaou.ml
artkhyzha: Fixed, thanks.

tools/miaou.ml
artkhyzha: Fixed, thanks.

tools/miaou.ml
artkhyzha: Not sure a debug flag is useful in this case (or rather, more concentrated effort is needed to understand which sort of debug info is helpful), these remnants were an oversight. Should be fixed now.

catalogue/aarch64/tests/kinds.txt
artkhyzha: Thanks, reverted.

catalogue/aarch64-VMSA/tests/VMSA-kinds.txt
artkhyzha: Fixed, marking resolved.

catalogue/aarch64-MTE/shelf.py
artkhyzha: The issue is:
* regressions do complain when tests are not in the shelf
* regressions also complain when tests in the shelf feature control dependencies and print warnings into stderr, making it harder to satisfy the complaint

I have updated the shelf file to uncomment tests that should not cause the aforementioned problems.

artkhyzha: (By the way, I realise I did not edit the shelf file for VMSA, but also I am not sure why/if the regression tests require editing such files. I wonder if the actual action item is orthogonal to this specific PR...)

catalogue/aarch64-VMSA/tests/CASAL+FAIL+FH.litmus
artkhyzha: Ah, I previously updated the wrong commit. Updated tests are now a part of the first commit too.

catalogue/aarch64-MTE/shelf.py
```diff
@@ -1150,4 +1150,61 @@
...
+    "tests/from-tagcheck-async/LB-tagfault+rel+dmb.ldpt--async.litmus",
+    "tests/from-tagcheck-async/MP+dmb.stpt+tcib-po-isb-po--async.litmus",
+    "tests/from-tagcheck-async/MP+dmb.stpt+tcib-tagfault-po--async.litmus",
+    "tests/from-tagcheck-async/S+dmb.stpt+tcib-po--async.litmus",
+    "tests/from-tagcheck-async/S-tagfault+dmb.st+dmb.ldpt--async.litmus",
+    "tests/from-tagcheck-sync/LB-notagfault+po+dmb.ldpt--sync.litmus",
+    "tests/from-tagcheck-sync/LB-tagfault+po+dmb.ldpt--sync.litmus",
+    "tests/from-tagcheck-sync/MP+dmb.stpt+tcib-po-isb-po.litmus",
+    "tests/from-tagcheck-sync/MP+dmb.stpt+tcib-tagfault-po.litmus",
+    "tests/from-tagcheck-sync/S+dmb.stpt+tcib-po.litmus",
```
relokin, line 1170: These 3 tests are missing from the repository, I think you might have missed adding them to the relevant commit.

catalogue/aarch64-MTE/shelf.py
relokin: Ah see it's the warning about loop unrolling. Sounds like a sensible workaround for now. Going forward I think we might need a separate shelf for regressions and for the subset of tests that are presented through web, and we should make the warning about missing tests optional.

relokin (commented):
I just added one minor comment and I have a last request; can you reword the title of the 2nd commit from
 [aarch64] Changes to the model and new tests 
something along the line of
 [aarch64] Add FEAT_ETS3 and reorganise relations as in Arm ARM issue M.a
Just to be a bit more specific about the type of changes in this commit and also make it obvious from the title that this the cat file that corresponds to the specific version of the Arm ARM.

catalogue/aarch64-MTE/shelf.py
artkhyzha: Oops! Sorry about this, should be fixed now.

artkhyzha:
I have come to realise that `herd7` generates faults without marking them as exception entries. In some cases this causes buggy verdicts for MTE and ETS3. A possibly radical small solution is proposed in 8584a29a42e27377fdf952c5590faf0527ba1ae2: it simply marks almost all faults as exception entry effects, so that the cat phrasing applies as is.

Would you suggest an alternative approach, @relokin?

relokin (approved):
Thank you for all the hard work on this @artkhyzha. Looks good to me and I would like to merge. Can you please rebase on top of master?

@jalglave, @maranget are you happy for us to merge this?

artkhyzha:
Thank you so much for all the reviewing, @relokin .
This is now rebased on top of master.

artkhyzha:
Just to double check with @relokin, @jalglave and @maranget: as there are no known outstanding todo items, should we merge this PR or wait for additional reviews? Let me know if I can help in any way.

jalglave:
@artkhyzha thank you so much for doing this work. Please go ahead @relokin and @artkhyzha.

relokin:
Thanks @jalglave!

@artkhyzha this PR looks good to me and ready to be merged!

PR #959, opened 2024-09-09
[herd,asl] Do not output ASL typing warnings for herd

maranget:
The solution to silence _all_ warnings may be a bit too radical. Notice that there are two warnings only for now.


maranget:
Thanks, @HadrienRenaud. Merge on its way.

PR #970, opened 2024-09-14
[herd,asl] Fix unsound XOR optimization

yuxiliu-arm:
For XOR, the optimization when the two values are equal (`compare` returns 0) is to return `V.zero`. However, the actual result might be different for different scalar types. Thus `V.zero` should not be used in all the cases.

An example is the `ASLConstant` module, when XOR-ing two bitvectors, the result should be a zeroed bitvector of the same length, rather than `V.zero`, i.e. `S_Int Z.zero`.

This PR implements a fix by removing the optimization. An alternative fix would be to make `zero` generic over the scalar's type (and similarly `one` if it's used for generic optimization).

### DIFF
--- a/lib/symbValue.ml
+++ b/lib/symbValue.ml
@@ -341,8 +341,12 @@ module
       | _ -> binop_cs_cs Op.Or Cst.Scalar.logor v1 v2
 
   and xor v1 v2 =
-    if compare v1 v2 = 0 then zero else
-    binop Op.Xor (Cst.Scalar.logxor) v1 v2
+    if equal v1 v2 && Cst.Scalar.unique_zero then zero else
+    match v1,v2 with
+    | (Val (Symbolic id1),Val (Symbolic id2))
+      when Constant.symbol_eq id1 id2
+        -> zero
+    | _ -> binop Op.Xor Cst.Scalar.logxor v1 v2
 
   and maskop op sz v = match v,sz with
   | Val (Tag _),_ -> v (* tags are small enough for any mask be idempotent *)
--- a/Makefile
+++ b/Makefile
@@ -438,6 +438,8 @@ cata-test:: mte-test
 mte-test:
 	@ echo
 	$(HERD_CATALOGUE_REGRESSION_TEST) \
+		-herd-timeout $(TIMEOUT) \
+		-j $(J) \
 		-herd-path $(HERD) \
 		-libdir-path ./herd/libdir \
 		-kinds-path catalogue/aarch64-MTE/tests/kinds.txt \
--- a/herd/eventsMonad.ml
+++ b/herd/eventsMonad.ml
@@ -1689,10 +1689,7 @@ Monad type:
     and op op v1 v2 =
       any_op
         (fun () -> V.op op v1 v2)
-        (fun () ->
-          match op with
-          | Op.Xor when V.compare v1 v2 = 0 -> VC.Atom V.zero
-          | _ -> VC.Binop (op,v1,v2))
+        (fun () -> VC.Binop (op,v1,v2))
 
     and op3 op v1 v2 v3 =
       any_op
--- a/herd/mem.ml
+++ b/herd/mem.ml
@@ -2008,6 +2008,8 @@ let match_reg_events es =
             Constant.is_symbol cst ||
             Constant.is_label cst
             -> ()
+        | Some (A.Location_global (V.Var _))
+            -> ()
         | Some loc ->
             Warn.user_error "Non-symbolic memory access found on '%s'"
               (A.pp_location loc)
--- a/herd/tests/instructions/AArch64/L103.litmus
+++ b/herd/tests/instructions/AArch64/L103.litmus
@@ -1,0 +1,16 @@
+AArch64 L103
+(* Test xor'ing symbolic addresses *)
+{
+int z = 1;
+int *p = z;
+int *q = z;
+int x = 2;
+int y = 2;
+0:X0=p; 0:X1=q; 0:X2=x;
+1:X0=p; 1:X1=q; 1:X2=y;
+}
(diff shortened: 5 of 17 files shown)


yuxiliu-arm:
Amended to add comments in code. One thing to be noticed that in ec19d70, if a similar fixed is implemented to [`eventsMonad`'s `op`](https://github.com/herd/herdtools7/blob/4112e1ea161e0af405e9cf068bbcdce673d2afd5/herd/eventsMonad.ml#L1694), the test will fail at

```
_build/default/internal/herd_catalogue_regression_test.exe \
	-herd-path _build/install/default/bin/herd7 \
	-libdir-path ./herd/libdir \
	-kinds-path catalogue/aarch64-MTE/tests/kinds.txt \
	-shelf-path catalogue/aarch64-MTE/shelf.py \
	test
Fatal error: exception Failure("Herd returned stderr: Warning: File \"catalogue/aarch64-MTE/tests/LB+dmb.ld+addrtp.litmus\": Non-symbolic memory access found on '[S17]' (User error)")
make: *** [Makefile:440: mte-test] Error 2
```

So this behavior is indeed depended upon in this test.

maranget:
> @maranget, as you implemented this optimisation: what was the reason behind it? Do you think that this is still necessary?

Frankly I do not remember. If tests are ok, everything should be fine.

maranget:
Well, I see I need to have a look at commit  ec19d70 before definitive approval. Stay tuned.

maranget:
Hi @yuxiliu-arm and @HadrienRenaud. Tests in the MTE catalogue perform a lot xor operations on (identical) symbolic addresses. By giving up on "polymorphic" xor on identical values, commit ec19d70 leads to runtime errors as a result of xor'ing symbolic addresses.

Commit ba837cb9df214572b5608b75cc40d945574ab224 allows explicit xor on (identical) symbolic addresses. This fixes the MTE catalogue failure.

Unfortunately, the "polymorphic" xor on identical values was performed very early by herd and its removal yields prohibitive running times. The commit mitigates this slowdown by concurrent evaluation and timeouts.

yuxiliu-arm:
In light of the performance issue, shall we look at an alternative fix mentioned in my original comment: add a function e.g. `zero_of : v -> v option` to `Value.S`, which returns the zero value of the correct type if possible. And Xor is only optimized to `zero` if the return value is `Some zero`.

In ASL, it could match on the concrete sum type variant to fix the bitvector issue. In AArch64, it can be `let zero_of _ = Some zero`.

maranget:
Performance was better because we had `V?? ^ V??` to be integer zero, where `V??` is a yet unsolved variable (constructor `Var` from `lib/value.mli`. Such variables do not have a determined type.

maranget:
By the way here us another commit with a test of xor'ing symbolic addresses that is simpler than the tests in the MTE catalogue: 098a12e63c83a2cb6ed1377e3fd2acee5bd421ec.

maranget:
> In light of the performance issue, shall we look at an alternative fix mentioned in my original comment: add a function e.g. `zero_of : v -> v option` to `Value.S`, which returns the zero value of the correct type if possible. And Xor is only optimized to `zero` if the return value is `Some zero`.
> 
> In ASL, it could match on the concrete sum type variant to fix the bitvector issue. In AArch64, it can be `let zero_of _ = Some zero`.



> In light of the performance issue, shall we look at an alternative fix mentioned in my original comment: add a function e.g. `zero_of : v -> v option` to `Value.S`, which returns the zero value of the correct type if possible. And Xor is only optimized to `zero` if the return value is `Some zero`.
> 
> In ASL, it could match on the concrete sum type variant to fix the bitvector issue. In AArch64, it can be `let zero_of _ = Some zero`.

Hi @yuxiliu-arm. Your idea looks worth a try.

yuxiliu-arm:
Hi, @maranget, this PR is rebased to use the `zero_of` idea, together with your commits.

lib/symbValue.ml
```diff
@@ -341,8 +341,18 @@ module
...
+    | (Val (Symbolic id1),Val (Symbolic id2))
+      when Constant.symbol_eq id1 id2
+        -> zero
+    | (Val cv1, _ )
+      when compare v1 v2 = 0
+      ->
+        (match Cst.zero_of cv1 with
+        | Some zero -> Val zero
+        | _ -> do_xor ())
+    | _ -> do_xor ()
```
maranget, line 355: Hi @yuxiliu-arm I do not think we gain much here. Namely, the efficiency gain is huge when `v1` and `v2` are teh same variable (`Var _`). Here we do not gain a lot if anything, because `do_xor` will perform teh reduction to zero anyway. A more efficient change would be to apply the optimisation `if compare v1 v2 = 0 then zero else ...` when the architecture is not `ASL`. Doing so, we assume that integers are the only meaningful scalars, which I guess is the case for all architectures except `ASL`  where scalars also include bitvectors.

yuxiliu-arm: I see.

> Doing so, we assume that integers are the only meaningful scalars, which I guess is the case for all architectures except ASL where scalars also include bitvectors.

I didn't know that was the case, so I only implemented the conservative change for concrete values. `zero_of` is indeed the way you described (always returns `Some zero` for al backends except ASL). I guess a simpler solution is to just extend `Constant.S` to have `is_xor_result_always_numeric : bool`?

maranget:
Hi @yuxiliu-arm, I was not clear in my previous message, even to myself. I have just pushed some change that should be more clear as to what I meant by restoring the xor optimisation in some cases. Your comments are welcome.

maranget:
Nice, would you please at the first occasion squash the various commits into one or two and rebase on master, aiming at a clean merge.

maranget:
Merged, thanks @yuxiliu-arm.

PR #972, opened 2024-09-16
[herd] Some optimisation aiming at faster Cat file interpretation

maranget:
With ASL the cat interpreter now faces inputs of thousands events. This PR regroups attempts to avoid some quadratic behaviours in the computation of pre-defined relations such as the `loc` relation.

More precisely, we compute the `loc` relation by first computing a mapping from locations to event sets. Then we union all the subrelation made by applying the cartesian product to each  event set. Relations `same-low-order-bits`, `int` and `ext` are computed by a similar technique.


maranget:
Converted to draft as a similar optimisation applies to  the 'int' relation and maybe to others.

maranget:

PR #975, opened 2024-09-21
[asl] Fix a few parsing issues.

HadrienRenaud:
This PR implements a few issues with parsing:
- [x] Bitvectors as masks
- [x] duplicated otherwises in case statements

### Bitvectors as masks

All lines except the first in the following test in `lexer.t` were failing because of a parsing error. This PR fixes this by raising bitvectors as masks when needed.

```
func main() => integer
begin
    var b = ''; // works
    let expr_a = '' IN '1'; // fails
    let expr_b = '1' IN ''; // fails
    let expr_c = '' IN '0'; // fails
    let expr_d = '0' IN '0';
    return 0;
end
```

Thanks @Roman-Manevich for raising this issue.

### DIFF
--- a/asllib/Parser.mly
+++ b/asllib/Parser.mly
@@ -228,6 +228,13 @@ let plist2(x) == pared(
   ~=x; COMMA; li=separated_nonempty_list(COMMA, x); { x :: li }
 )
 
+(* Produces a list of [x]s, optionally followed by a single [y] that would go
+   in the tail of the list *)
+let nlist_opt_terminated(x, y) :=
+  | ~=x; { [ x ] }
+  | ~=x; ~=y; { [ x; y ] }
+  | ~=x; l=nlist_opt_terminated(x, y); { x :: l }
+
 (* ------------------------------------------------------------------------
 
                              First parsing rules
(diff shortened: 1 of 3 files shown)


asllib/Lexer.mll
```diff
@@ -202,7 +202,7 @@ and token = parse
     | real_lit as lxm          { REAL_LIT(Q.of_string lxm)        }
     | '"'                      { string_lit (Buffer.create 16) lexbuf }
     | '\'' (bits as lxm) '\''  { bitvector_lit lxm                }
-    | '\'' (mask as lxm) '\''  { mask_lit lxm                     }
+    | '\'' (mask as lxm) '\''  { mask_lit lxm                     }  (* Warning: masks with no unknown 'x' characters will be lexed as bitvectors. *)
```
Roman-Manevich, line 205: I think it would be cleaner to have just the regular expression for bitmasks and lex it into some kind of token like `mask_or_bits` that will later be converted by the parser to either a bitmask or a bitvector. The parser has the right context to decide which one.

Roman-Manevich (changes requested):

asllib/Lexer.mll
Roman-Manevich: Right. I tried it and saw the issues for myself.

Roman-Manevich (approved):
Let's go ahead with this.

jalglave:
@HadrienRenaud please could you rebase and merge when you can? This now has conflicts which need resolving.
Thanks in advance,
Jade

asllib/Parser.mly
```diff
@@ -228,6 +228,13 @@ let plist2(x) == pared(
   ~=x; COMMA; li=separated_nonempty_list(COMMA, x); { x :: li }
 )
 
+(* Produces a list of [x]s, optionnaly followed by a single [y] that would go
```
Roman-Manevich, line 231: typo optionnaly -> optionally

PR #980, opened 2024-09-24
Streamlining the static model

hrutvik:
This PR attempts to streamline static normalisation in a few ways:
- Avoid using both static evaluation and normalisation if one should suffice
- Weaken normalisation based on its usage
  - Remove "affectations"
  - Reduce the kinds of constraints it can track to just equality/inequality with zero
- Fix a couple of potential bugs
- Avoid reducing polynomials by maintaining a stronger invariant about the absence of 0-coefficients
- Limit catching of `ConjunctionBottomInterrupt` exception to within a single function, to simplify control flow
- Along the way, do various refactors - including a final significant refactor attempting to modularise the various moving parts in normalisation (monomials/polynomials/internal representation)

### DIFF
--- a/asllib/StaticModel.ml
+++ b/asllib/StaticModel.ml
@@ -57,14 +57,14 @@ end
 module PMap = Map.Make (PolynomialOrdered)
 (** Map from polynomials. *)
 
-(** A constraint on a numerical value. *)
-type sign = Null | NotNull
+(** A (in)equation for a numerical value. *)
+type eq = Zero | NonZero
 
 (** A conjunctive logical formulae with polynomials.
 
-      For example, {m X^2 \leq 0 } is represented with {m X^2 \to \leq 0 }.
+      For example, {m X^2 = 0 } is represented with {m X^2 \to Null }.
   *)
-type ctnts = Conjunction of sign PMap.t | Bottom
+type ctnts = Conjunction of eq PMap.t | Bottom
 
 (** Case disjunctions. *)
 type ir_expr =
@@ -112,13 +112,13 @@ let pp_poly f (Sum poly) =
     PP.pp_print_seq ~pp_sep pp_mono f (MMap.to_seq poly);
     pp_close_box f ())
 
-let pp_sign f s =
-  let s = match s with Null -> "= 0" | NotNull -> "!= 0" in
+let pp_eq f s =
+  let s = match s with Zero -> "= 0" | NonZero -> "!= 0" in
   Format.pp_print_string f s
 
 let pp_ctnt f (p, s) =
   let open Format in
-  fprintf f "@[<h>%a@ %a@]" pp_poly p pp_sign s
+  fprintf f "@[<h>%a@ %a@]" pp_poly p pp_eq s
 
 let pp_ctnts f =
   let open Format in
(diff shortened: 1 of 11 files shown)


hrutvik:
NB there were a few things I could not update in the Typing Reference document because they are not yet formalised:
- The PR's change to `StaticInterpreter.ml` strengthens static evaluation by recursing into let-bound expressions, but I believe the static environments in the Typing Reference do not store these expressions.
- Most of the PR's changes to `StaticModel.ml` remove things that are not yet formalised. For example: `equal_mod_branches` (and the associated "conjunction" machinery, including its effect on reduction) and polynomial division. This PR should reduce the number of things that are yet to be formalised, particularly the "affectation" machinery.

asllib/StaticModel.ml
```diff
@@ -2,366 +2,428 @@ open AST
 open ASTUtils
 open Error
 
-exception NotYetImplemented
+exception NotSupported
 
-type atom = identifier
-(** Our basic variables. *)
+(*---------- Expression helpers - consider moving to AST.mli ----------*)
```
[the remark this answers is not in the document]
Roman-Manevich: How about `truth` -> `lit_true`?

asllib/StaticModel.ml
```diff
@@ -2,366 +2,428 @@ open AST
...
-      {m X^2 \to 1, X \to -1, 1 \to 4 } *)
-type polynomial = Sum of Q.t MMap.t  (** Maps each monomial to its factor. *)
+  val empty : t
+  val is_bottom : t -> bool
+  val is_empty : t -> bool
+  val of_bool : bool -> t
+  val single_conjunct : Polynomial.t -> eq -> t
+  val conj : t -> t -> t
+  val to_expr : t -> expr option
+  val is_trivial : t -> bool option
```
Roman-Manevich, line 224: I don't understand what this does. Why does it return a `bool option` and not just `bool`?

asllib/StaticModel.ml
```diff
@@ -2,366 +2,428 @@ open AST
...
-  | StrictPositive
-  | Positive
-  | Negative
-  | StrictNegative
-  | NotNull
+  module PolynomialMap = Map.Make (Polynomial)
+  (** Map from polynomials. *)
 
-(** A conjunctive logical formulae with polynomials.
+  type t = eq PolynomialMap.t option
```
Roman-Manevich, line 243: Why do you need an option here?

Roman-Manevich (changes requested):

asllib/StaticModel.ml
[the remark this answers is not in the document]
hrutvik: I've moved them to `ASTUtils` and renamed:
- `truth` --> `lit_true`
- `from_rational` --> `expr_of_rational`
- `from_int` --> `expr_of_z`

asllib/StaticModel.ml
hrutvik: I've added an `ocamldoc` comment to explain: if the conjunction is trivially true or false, this returns `Some true` or `Some false`. If not, then it returns `None`.

asllib/StaticModel.ml
hrutvik: I've added more to  @the comment to explain: `None` is a trivially false conjunction

asllib/StaticModel.ml
Roman-Manevich: It feels like instead of a comment, this needs to be a type. perhaps something like:
`TriviallyTrue | TriviallyFalse | Nontrivial`.

hrutvik: I've introduced this type as you suggest!

hrutvik:
> On the code part, I only have one comment, and the more general remark that your use of `=` frightens me (I rather use `==` for optimisations or `Module.equal` for precise typed equality).

Also I have addressed this by using `expr_equal` instead of polymorphic equality.

Roman-Manevich (approved):
You're good to go!
Thanks for the good work!

PR #987, opened 2024-09-26
[doc] Fix spelling mistake

knatten:
I assume these should say "TSO" and not "TSP"


maranget:
Merged via command line, thanks @knatten.

PR #993, opened 2024-10-01
[lib] Simplify  the `mySet` module

maranget:
Function `of_list` is now part of OCaml standard library. Also delete some unused functions.

### DIFF
--- a/lib/mySet.ml
+++ b/lib/mySet.ml
@@ -27,9 +27,6 @@ module type S = sig
   val exists2 : (elt -> elt -> bool) -> t -> t -> bool
   (* Like exists, but returns an elt that satisfy the predicate,
      raises Not_found, if no such elt exists *)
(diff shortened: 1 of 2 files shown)


maranget:
Are sure? `Stdlib.Set.S.exists` type is `(elt -> bool) -> t -> bool`, while MySet.S.find type is `(elt -> bool) -> t -> elt`.

Don't you mean `Stdlib.Set.S.find`?

maranget:
I do not really understand what you mean. Ii quite likely, without checking it very hard, that `find_first` or `find_last` would be adequate replacements for out `MySet.S.find`, which is used once only in the obsolete `slrc11.ml` file.

However I'd rather keep things as they are.

maranget:
Merge on its way, thanks @HadrienRenaud.

PR #995, opened 2024-10-02
[herd] Implement system registers read from multiple previous writes

TiberiuBucur:
This commit implements a semantic for system registers that is similar to memory locations, in that any read can read from multiple previous write operations, generating different execution paths.
Take, for example, the following test
```
AArch64 LDR-TagCheck-MRS-async
Variant=mte,async
{
 0:X0=x:red;
}
 P0              ;
 LDR WZR,[X0]    ;
 MRS X1,TFSR_ELx ;
exists(0:X1=0)
```

The load operation will update the `TFSR_ELx` register, but the update is not required to be seen by the MRS instruction, which could instead observe the initial write of the default value 0 to the system register. As a result, there are 2 execution paths being generated:
```
Test LDR-TagCheck-MRS-async Allowed
States 2
0:X1=0;
0:X1=1;
Ok
Witnesses
Positive: 1 Negative: 1
Condition exists (0:X1=0)
Observation LDR-TagCheck-MRS-async Sometimes 1 1
Time LDR-TagCheck-MRS-async 0.01
Hash=08f8931dde2fbfcf9dd273a836baab86
```

This change, naturally, gets in the way of already existing tests that make use of system registers, assuming the old behaviour. Take, for example, `herd/tests/instructions/AArch64.MTE/B004.litmus`:
```
AArch64 B004
Variant=async
{
 0:X1=x:red
}
 P0          ;
 MOV W0,#1   ;
 STR W0,[X1] ;
forall([x]=1 /\ 0:TFSR_ELx=1)
```

The `STR` instruction sets the `TFSR_ELx` register to 1, but, under the semantics assumed in this PR, the postcondition is not required to observe this last write. It could also read the initial value of 0, producing, just like in the first example we showed, 2 executions. This behaviour **is wrong** and needs to be forbidden by adding extra rules to the cat file to forbid the execution where the postcondition reads the initial value of the register.

Another example causing a different problem is `herd/tests/instructions/AArch64.MTE/AK003.litmus`:
```
AArch64 AK003
{
    tag(z)=:green;
    0:X1=z:red;
}
P0           | P0.F           ;
 STR W0,[X1] | ADR X9,L0      ;
L0:          | MSR ELR_EL1,X9 ;
 NOP         | ERET           ;
```

In this test, the `ADR` instruction updates the `X9` register to point to the L0 label. Then, the value of register `X9` is copied into the `ELR_EL1` system register. Upon running this test, herd will crash with a fatal error, saying it cannot determine the target of the `ERET` instruction. This is due to the fact that reading system registers now always return a fresh symbolic variable instead of the value recorded in the register map, in order to make them behave more like memory locations. Having the register point to an unresolved symbolic variable instead of a concrete label makes herd panic because it does not know where to fetch from after the ERET. We do not yet have a solution for this problem, and it is to be investigated further.


relokin:
Thanks @TiberiuBucur!

I have two quick comments:
* The 1st commit adds a lot of boiler code; can you please reduce this to the functions you need? If I understand correctly your goal is to implement the set `("SysReg", is_sysreg_any)`. The other filtering functions are unnecessary, in cat we can get the intersection and unions of sets and with these two operators we would be able to get all possible sets anyway. In other words, we won't need a `SysRegRead` set if we just had `SysReg` because we would do `Rreg & SysReg`.
* The 2nd commit adds a new parameter to `Access`. It would a lot simpler to add a new annotation, much like we have `Exp`, `NExp` for `M`, you could add `Direct` for `Indirect` for Register `Accesses`. This will simplify things a lot.

herd/semExtra.ml
relokin, line 1: The changes in this files introduce functions that are not used anywhere, can you please remove them?

herd/AArch64Annot.ml
```diff
@@ -18,6 +18,7 @@
 type t =
   A | XA | L | XL | X | N | Q | XQ | NoRet | S
   | NTA (* Non-Temporal, avoid clash with NT in AArch64Base *)
+  | SysDirect (* Direct accesses to system registers *)
```
relokin, line 21: Rather than adding a new annotation can we please use
`Exp | NExp`
That is already there for memory accesses? Then in the cat file we can use Exp to mean Direct and NExp to mean indirect.

That way you won't have to another type here and you will only need to change `herd/AArch64Sem.ml`

herd/ASLAction.ml
relokin, line 1: Was this introduced by a mistake in rebasing? This should go to the 1st patch in the series.

herd/mem.ml
```diff
@@ -1151,8 +1180,11 @@ let match_reg_events es =
           E.EventSet.filter code_store es.E.events in
         let kont es rfm cns res =
           (* We get here once code accesses are solved *)
-          let loads =  E.EventSet.filter E.is_mem_load es.E.events
-          and stores = E.EventSet.filter E.is_mem_store es.E.events in
+          (* System registers behave like memory locations *)
```
relokin, line 1183: Please remove this comment, it's not correct as a blanket statement.

herd/mem.ml
```diff
@@ -786,6 +791,7 @@ let match_reg_events es =
             S.RFMap.add (S.Load er) rf k)
           loads k)
       loc_loads S.RFMap.empty in
+
```
relokin, line 794: please remove

herd/mem.ml
```diff
@@ -991,6 +1005,21 @@ let match_reg_events es =
                 else c)
               map_load)
           stores (map_load_init loads) in
+
+      (* Add init write for system registers when one does not
```
relokin, line 1009: Can you please remind why you need to add this?

lib/AArch64Base.ml
```diff
@@ -595,6 +599,16 @@ let type_reg r =
...
 
+let is_sysreg = function
+  | SysReg _ -> true
+  | _ -> false
+
+let is_spsysreg = function
+  | SysReg r -> is_special_purpose r
+  | _ -> false
+
+let is_non_sp_sysreg r = is_sysreg r && not (is_spsysreg r)
```
relokin, line 610: This is not necessary and since this is not supported by other architectures, let derived where it is needed.

herd/memUtils.mli
```diff
@@ -76,6 +76,8 @@ module Make : functor (S: SemExtra.S) -> sig
 (* Collect various events, indexed by location *)
   val collect_reg_loads : S.event_structure -> S.event list LocEnv.t
   val collect_reg_stores : S.event_structure -> S.event list LocEnv.t
+  val collect_gpreg_and_spsysreg_loads : S.event_structure -> S.event list LocEnv.t
+  val collect_gpreg_and_spsysreg_stores : S.event_structure -> S.event list LocEnv.t
```
relokin, line 80: instead of load/store can you please use read/write?

herd/memUtils.mli
```diff
@@ -76,6 +76,8 @@ module Make : functor (S: SemExtra.S) -> sig
 (* Collect various events, indexed by location *)
   val collect_reg_loads : S.event_structure -> S.event list LocEnv.t
   val collect_reg_stores : S.event_structure -> S.event list LocEnv.t
```
relokin, line 78: After you PR these two are unused.

herd/memUtils.ml
```diff
@@ -366,6 +366,10 @@ let lift_proc_info i evts =
   let not_speculated es e = not (E.EventSet.mem e es.E.speculated)
   let collect_reg_loads es = collect_by_loc es E.is_reg_load_any
   and collect_reg_stores es = collect_by_loc es E.is_reg_store_any
+  and collect_gpreg_and_spsysreg_loads es = collect_by_loc es (fun e ->
+    E.is_non_sysreg_load_any e || E.is_spsysreg_load_any e)
```
relokin, line 370: I think it would be more clear to define this as
```
and collect_gpreg_and_spsysreg_loads es = collect_by_loc es (fun e ->
    (E.is_load_any e) && (E.is_spsysreg e || (not (E.is_sysreg e)))
```
This should inform what functions you need from E and which ones you should remove.

herd/memUtils.ml
```diff
@@ -366,6 +366,10 @@ let lift_proc_info i evts =
...
+  and collect_gpreg_and_spsysreg_stores es = collect_by_loc es (fun e ->
+    E.is_non_sysreg_store_any e || E.is_spsysreg_store_any e)
```
relokin, line 372: ditto

herd/mem.ml
TiberiuBucur: Tests that make use of memory locations have to mention the location in the initial state (perhaps assign it to a register, since otherwise it would not be accessible by the test). As soon as that happens, the location becomes part of the "observed_locs" set. If the `-initwrites` flag is set to true, this creates a write event to that location and an `S.Store ew` is added to the reads-from map (where `ew` is the generated write event). If `-initwrites` is false, an `S.Init` is added to the reads-from map, as per the `map_load_init` function above.

The same thing happens with registers when they are part of the observed locations (mentioned in the init state or the postcondition). However, we also want to be able to use registers with default initialisation values (0) _without_ having to mention them in the init state. Since there is no way for us to know when generating the init writes that a register would be used in the code (since the event structure, with all the loads and stores of a program, is not generated at that point), a read operation of the uninitialised register would encounter no mapping in the reads-from map. This code adds an `S.Init` read-from, so that the default value can be used for uninitialised registers.

PR #996, opened 2024-10-03
[herd] Check the liveness of read-from relations

maranget:
Intuitively, this "liveness" property is that a read cannot indefinitely read from a non-final write. Our formulation involves 'pruned' executions and requires that, for all locations x the last read of x in a pruned po is from the final write to x.


maranget:
Superseded by PR #997.

PR #997, opened 2024-10-04
[herd]  Check the liveness of some read-from relations

maranget:
Intuitively, this "liveness" property is that a read cannot indefinitely read from a non-final write. Our formulation involves 'pruned' executions and requires that, for all locations x the last read of x in a pruned po is from the final write to x.

Our implementation builds on the "CutOff" (ex "TooFar") feature. Namely, for the sake of avoiding infinite or excessive running times, some executions are pruned, resulting in execution prefixes. Those are normally discarded, unless `-variant CutOff` is specified. In those executions, pruned `po` relation end with a specific `CutOff` effect.

The liveness condition applies to those execution prefixes. The condition will reject executions for which there exists
some `fr` source in a pruned po that is not followed by a `rf` target:

That is, given
```
      +-fr->W
      |
      |
      R ----po---> CutOff
```
there does not exists a rf s.t.
 ```
      +-fr->W-rf-+
      |          |
      |         \ /
      R ----po---R-----> CutOff
```
Notice that this "liveness" condition applies to writes that are followed by a DSB effect (_i.e._ stores followed by a DSB fence instruction).

### DIFF
--- a/herd/AArch64ASLSem.ml
+++ b/herd/AArch64ASLSem.ml
@@ -936,7 +936,7 @@ module Make (TopConf : AArch64Sig.Config) (V : Value.AArch64ASL) :
         | ASLS.Act.NoAction ->
            (* As long as aarch64.cat ignores "NoAction" effects *)
            None
-        | ASLS.Act.TooFar msg -> Some (Act.TooFar msg)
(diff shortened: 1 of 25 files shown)


maranget:
Delayed by PR #1004

PR #1000, opened 2024-10-09
[LKMM] Fix regression of atomic_add_unless 

hernanponcedeleon:
@akiyks noticed I introduced a regression in c0ce66d75f482f5386036783e9a44dc588fed515 (PR #865). In the version previous to that commit, `atomic_add_unless` could be used even if not properly defined in the *.bell file.

This PR restores the previous behavior, but it also allows to "redefine" the `atomic_add_unless` macro in the *.bell file.


akiyks:
Quick tests show that this restores `atomic_add_unless()` for legacy LKMM models and works well with the model for lkmmv2 variant as well. 
The fix looks much smaller than what I had anticipated!

@maranget, this should be good to pull.

maranget:
Merged, thanks @hernanponcedeleon and @akiyks.

PR #1001, opened 2024-10-10
[ASL+AArch64] Allow access to UDF under variant

HadrienRenaud:
### How to run an instruction not implemented in herd?

For example, for `FADD`, a how-to.
```
cd herd/libdir/asl-pseudocode/aarch64/instrs
mv udf/UDF_only_perm_undef.opn udf/UDF_only_perm_undef.opn.bak
cp float/arithmetic/add-sub/FADD_D_floatdp2.opn udf/UDF_only_perm_undef.opn
```

Then edit the file to add the correct variables as the decode part of the instruction would usually do.
Then run `make install` or run with the correct `-libdir`, and a test like:
```
AArch64 UDF
variant=ASL+AArch64+UDF
{}
P0     ;
UDF 1  ;
```

### DIFF
--- a/herd/AArch64ASLSem.ml
+++ b/herd/AArch64ASLSem.ml
@@ -730,6 +730,8 @@ module Make (TopConf : AArch64Sig.Config) (V : Value.AArch64ASL) :
                   "domain" ^= var (barrier_domain dom);
                   "types" ^= var (barrier_typ btyp);
                 ] )
+      | I_UDF k when C.variant Variant.ASL_AArch64_UDF ->
+          Some ("udf/UDF_only_perm_undef.opn", stmt [ "imm16" ^= litbv 16 k ])
       | i ->
           let () =
             if _dbg then
--- a/herd/variant.ml
+++ b/herd/variant.ml
@@ -96,6 +96,8 @@ type t =
   | ASLType of [`Warn|`Silence|`TypeCheck]
 (* Activate ASL experimental mode *)
   | ASLExperimental
+(* UDF control in AArch64 mode *)
+  | ASL_AArch64_UDF
 (* Signed Int128 types *)
   | S128
 (* Strict interpretation of variant, e.g. -variant asl,strict *)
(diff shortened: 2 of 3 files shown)


relokin:
Thanks for this Hadrien!

Just to check that my understanding is correct.

I did the following:
```
cp integer/arithmetic/add-sub/extendedreg/ADD_32_addsub_ext.opn udf/UDF_only_perm_undef.opn
```
and then edited it to look like this:
```
// Execute
// =======

bits(datasize) result;
bits(datasize) operand1 = X[0, 32];
bits(datasize) operand2 = ExtendReg(1, ExtendType_UXTH, 0, 32);

(result, -) = AddWithCarry(operand1, operand2, '0');

X[2, 32] = result;
```

Then I run the following test:

```
AArch64 UDF
Variant ASL+AArch64+UDF
{
0:X0=21;
0:X1=42;
}
P0     ;
UDF 1  ;
locations[0:X2;]
```
But I get 
```
Test UDF Required
States 1
0:X2=0;
Ok
Witnesses
Positive: 1 Negative: 0
Condition forall (true)
Observation UDF Always 1 0
Time UDF 0.01
Hash=53bb635e83af9884274b948b0c2a6276
```

relokin:
ofc! I saw that space in Variant and I thought ah I didn't know this was allowed but never checked :)

Thanks Hadrien!

herd/AArch64ASLSem.ml
```diff
@@ -730,6 +730,8 @@ module Make (TopConf : AArch64Sig.Config) (V : Value.AArch64ASL) :
                   "domain" ^= var (barrier_domain dom);
                   "types" ^= var (barrier_typ btyp);
                 ] )
+      | I_UDF k when C.variant Variant.ASL_AArch64_UDF ->
```
relokin, line 733: Out of curiosity, can't unconditionally handle UDF with that ASL code? Because if that's the case you don't really need the variant either, do you?

relokin: ah yes, I think this something that happens in decode.

relokin (approved):
LGTM, thanks @HadrienRenaud!

PR #1002, opened 2024-10-13
Distinguish some overloading usages of [] in ASL

yuxiliu-arm:
1. Replace [] around nested bitvector's or record's field name list with () parentheses.
2. Replace [] around bitvectors to be concatenated with :: bit concatenation operator
3. Remove unadorned [lsb] for single-bit slice and add
     a. [lsb:] as syntactic sugar for [lsb+:1] (single-bit slice)
     b. [:wid] as syntactic sugar for [0+:wid] (least sig wid bits)

TODO:
- [x] Decide how to change `LE_Concat` to best match the new bitvector concat syntax. (Decision: leave it as is.)
- [x] Single concat is no longer representable. Decide whether to drop its usage, or inventing a new syntax for it. (Decision: either drop the usage, or use `:: ''` if desired.)
- [ ] Update the spec.

Note on the implementation:
1. `Slice_Single` is removed and `Slice_Arg of expr` is introduced. For ASLv1, originally `bitvector[x]` parses to `Slice_Single`, now `bitvector[x:]` parses to `Slice_Length (x, 1)` (i.e. desugared during parsing), while `getter[x]` parses to `Slice_Arg`. Similarly for `array[index]`. Note the type error for `bitvector[x]` is only triggered only for ASLv1.
2. `E_Concat` is removed, the left-associative `BV_Concat` binop is introduced.


asllib/error.ml
```diff
@@ -155,6 +156,11 @@ module PPrint = struct
           (error_handling_time_to_string t)
           length pp_slice_list slices
     | BadSlice slice -> fprintf f "ASL error: invalid slice %a." pp_slice slice
+    | UnexpectedSliceArg e ->
+        fprintf f
+          "ASL error:@ Unexpected slice argument %a, but a slice is expected.@ \
```
Roman-Manevich, line 161: "...but a slice is expected." => "...[where] a slice is expected."

asllib/AST.mli
```diff
@@ -79,6 +79,7 @@ type binop =
   | RDIV  (** Division for reals *)
   | SHL  (** Shift left for ints *)
   | SHR  (** Shift right for ints *)
+  | COLON_COLON  (** Bitvector concatenation *)
```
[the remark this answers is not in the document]
yuxiliu-arm: There's a string CONCAT token (`++`). How about changing this to `BV_CONCAT` and the string one to `STR_CONCAT`?

yuxiliu-arm:
> What is the point of the AST construct `E_Concat` now that it's a binary operation? I'm aware that it's constructed with `expr_of_lexpr`, but that could surely be changed to used `COLON_COLON`?

Yes and I explored this choice a bit before. So we want to also desugar ASLv0's `E_Concat` construct into `COLON_COLON` (or `BV_CONCAT`) as well, right?

asllib/Typing.ml
```diff
@@ -102,7 +102,7 @@ let slices_width env =
   let minus = binop MINUS in
   let one = !$1 in
   let slice_width = function
-    | Slice_Single _ -> one
+    | Slice_Arg _ -> one
```
Roman-Manevich, line 105: How come a `Slice_Arg` has width if it's not a bitvector slice? Unless this is only for ASLv0?

asllib/Typing.ml
```diff
@@ -1196,11 +1210,12 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
           Format.eprintf "Annotating slice %a@." PP.pp_slice_list [ s ]
       in
       match s with
-      | Slice_Single i ->
-          (* LRM R_GXKG:
-             The notation b[i] is syntactic sugar for b[i +: 1].
-          *)
-          annotate_slice (Slice_Length (i, !$1)) |: TypingRule.Slice
+      | Slice_Arg i ->
```
Roman-Manevich, line 1213: I don't understand why a getter argument is annotated as a slice of length 1.

asllib/AST.mli
```diff
@@ -143,10 +143,10 @@ and pattern =
   | Pattern_Single of expr
   | Pattern_Tuple of pattern list
 
-(** Indexes an array, a bitvector. *)
+(** Indexes an array, a bitvector, an argument to a getter/setter *)
 and slice =
-  | Slice_Single of expr
-      (** [Slice_Single i] is the slice of length [1] at position [i]. *)
+  | Slice_Arg of expr
```
Roman-Manevich, line 148: Since this is always a getter argument perhaps we should not even consider it a slice but rather a separate type of AST node? Something like
`getter_arg of expr`?
This would avoid having to fail in matching slices.

asllib/Parser.mly
```diff
@@ -179,6 +183,10 @@ let     pared(x) == delimited(    LPAR, x, RPAR    )
 let    braced(x) == delimited(  LBRACE, x, RBRACE  )
 let bracketed(x) == delimited(LBRACKET, x, RBRACKET)
 
+let bracketed_or_pared(x) :=
```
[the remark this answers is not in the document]
yuxiliu-arm: This should be removed.

asllib/Typing.ml
```diff
@@ -410,7 +410,11 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
           else DI.Interval.make x y
         in
         match slice with
-        | Slice_Single e ->
+        | Slice_Arg e ->
+            let+ () =
+              check_true false @@ fun () ->
+              fatal_from e (Error.UnexpectedSliceArg e)
+            in
```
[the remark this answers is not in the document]
yuxiliu-arm: The most apt name would be `fail_from_for_v1`, but if we want to be more principled, maybe `fail_from_if_strict`?

asllib/Typing.ml
yuxiliu-arm: This is for compatibility with ASLv0. Since `Slice_Single` is desugared at parsing in ASLv1, there has to be something emitted for v0 parsing.

asllib/AST.mli
yuxiliu-arm: Because the AST is shared between ASLv0 and ASLv1, the same node has different meanings for the different parsing process. In ASLv1, `Slice_Arg` is used for getter/setter arguments and array indexing. In ASLv0, *in addition to* those, it's also used for slicing a single bit. The aim of this change is to disambiguate some bracket usages in ASLv1 and therefore this distinction is expected.

As for the name, I can't think of other names that can capture the above meaning. I do want to avoid use `Slice_Single` again, because the reader might be confused of the complete different usage in outdated documents.

asllib/Typing.ml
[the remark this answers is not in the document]
yuxiliu-arm: I used `fatal_from_if_not_silence` because `` `Silence `` appears in the code.

PR #1008, opened 2024-10-15
[ASL] Rework binary operation effects on symbolic constraints

HadrienRenaud:
- [x] Reorganise the code for binary-operation checking: it is now more explicit and shorter.
- [x] Improve DIV results by considering it decreasing on its second argument (see bug report later)
- [x] Add test for this
- [x] Merge static intervals after divisions using a diet
- [x] Implement `constraint_binop` for `MOD` and remove all usages of this strange default to `UnConstrained`
- [x] Fix DIV implementation: DIV is not decreasing for negative numerators.
- [x] Fix MUL, SHR, SHL, POW implementation: they are not increasing if one operand is negative.

---

### Bug report

With the following test, we can see that the results of divisions on constraints are defaulting to the unconstrained integer. This is unwanted behaviour.

```
func main() =>integer
begin
  let a : integer {10} = 10;
  let b : integer {10..20} = 10;
  var c = b DIV a;
  var d = a DIV b;
  c = 1.0; // will error so that the type is printed
  d = 1.0; // will error so that the type is printed
  return 0;
end
```

Previous output for this test:
```
❯ aslref tests-asl/1008.asl --type-check-warn
File tests-asl/1008.asl, line 7, characters 2 to 3:
ASL Typing error: a subtype of integer {1..2} was expected, provided real.
File tests-asl/1008.asl, line 8, characters 2 to 3:
ASL Typing error: a subtype of integer was expected, provided real.
```

Current output for this test:
```
❯ dune exec aslref -- tests-asl/1008.asl --type-check-warn
File tests-asl/1008.asl, line 7, characters 2 to 3:
ASL Typing error: a subtype of integer {1..2} was expected, provided real.
File tests-asl/1008.asl, line 8, characters 2 to 3:
ASL Typing error: a subtype of integer {1} was expected, provided real.
```

### DIFF
--- a/asllib/StaticOperations.mli
+++ b/asllib/StaticOperations.mli
@@ -26,3 +26,7 @@ module Make : functor (C : CONFIG) -> sig
     int_constraint list ->
     int_constraint list
 end
+
+(* Used by asllib/tests/ConstraintBinops.ml *)
+
+val filter_reduce_constraint_div : int_constraint -> int_constraint option
--- a/asllib/tests/ConstraintBinops.ml
+++ b/asllib/tests/ConstraintBinops.ml
@@ -1,0 +1,166 @@
+open Asllib
+open AST
+
+let eval_binop x op y =
+  let pos = ASTUtils.dummy_annotated and t = Error.Static in
+  match Operations.binop_values pos t op (L_Int x) (L_Int y) with
+  | L_Int z -> z
+  | _ -> assert false
+
+let eval_expr e =
+  match StaticInterpreter.static_eval StaticEnv.empty e with
+  | L_Int z -> z
+  | _ -> assert false
+
+let try_eval_expr e = try eval_expr e |> ASTUtils.expr_of_z with _ -> e
+
+let z_in_constraint z c =
+  match
+    StaticOperations.filter_reduce_constraint_div c |> Option.value ~default:c
+  with
+  | Constraint_Exact e -> ( try Z.equal z (eval_expr e) with _ -> false)
+  | Constraint_Range (e1, e2) -> (
+      try
+        let z1 = eval_expr e1 and z2 = eval_expr e2 in
+        Z.leq z1 z && Z.leq z z2
+      with _ -> false)
+
+let z_in_constraints z cs = List.exists (z_in_constraint z) cs
+
+let property op (x, y, cs1, cs2) =
+  assert (z_in_constraints x cs1);
+  assert (z_in_constraints y cs2);
+  (match op with
+  | SHR | SHL | POW -> assert (Z.sign y >= 0)
+  | DIV | DIVRM | MOD -> assert (Z.sign y > 0)
+  | _ -> ());
+  z_in_constraints (eval_binop x op y)
+    (StaticOperations.constraint_binop op cs1 cs2)
+
+let gen_xy op =
+  let open QCheck2.Gen in
+  let base_nat = small_nat in
+  let strict_positive_nat = base_nat >|= ( + ) 1 in
+  let signed_nat = oneof [ base_nat; base_nat >|= ( ~- ) ] in
+  match op with
+  | SHR | SHL | POW -> pair signed_nat base_nat
+  | DIV ->
+      let* y = strict_positive_nat in
+      let+ x = signed_nat >|= ( * ) y in
+      (x, y)
+  | DIVRM | MOD -> pair signed_nat strict_positive_nat
+  | _ -> pair signed_nat signed_nat
+
+let gen_cs x =
+  let open QCheck2.Gen in
+  let+ a = small_nat >|= ( - ) x and+ b = small_nat >|= ( + ) x in
+  assert (a <= x);
+  assert (x <= b);
+  (ASTUtils.expr_of_int a, ASTUtils.expr_of_int b)
+
+let gen_test_abcd op =
+  let open QCheck2.Gen in
+  let* x, y = gen_xy op in
+  let+ a, b = gen_cs x and+ c, d = gen_cs y in
+  ( Z.of_int x,
+    Z.of_int y,
+    [ Constraint_Range (a, b) ],
+    [ Constraint_Range (c, d) ] )
+
+let gen_test_abc op =
+  let open QCheck2.Gen in
+  let* x, y = gen_xy op in
+  let+ a, b = gen_cs x in
+  let c = ASTUtils.expr_of_int y in
+  (Z.of_int x, Z.of_int y, [ Constraint_Range (a, b) ], [ Constraint_Exact c ])
+
+let gen_test_acd op =
+  let open QCheck2.Gen in
+  let* x, y = gen_xy op in
+  let+ c, d = gen_cs y in
+  let a = ASTUtils.expr_of_int x in
+  (Z.of_int x, Z.of_int y, [ Constraint_Exact a ], [ Constraint_Range (c, d) ])
+
+let print_test op (x, y, cs1, cs2) =
+  try
+    let res = eval_binop x op y in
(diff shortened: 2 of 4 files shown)


maranget:
Hi @HadrienRenaud, I am not sure that the warnings are always useful : "trop de warning tue le warning". Maybe we should consider various levels of verbosity controlled by some command-line option?

asllib/ASTUtils.ml
```diff
@@ -593,60 +594,83 @@ let list_cross f li1 li2 =
     [] li1
   |> List.rev
 
+let list_flat_cross f li1 li2 =
+  List.fold_left
+    (fun xys x ->
+      List.fold_left (fun xys' y -> List.rev_append (f x y) xys') xys li2)
+    [] li1
+  |> List.rev
+
```
Roman-Manevich, line 603: (**
Tests whether the value of binary operation [op] increases along with its left-hand-side operand when all operands are non-negative. That is, if [is_left_increasing op] is true then for every two expressions [x1 op y] and [x2 op y] such that [x1 >= 0 /\ x2 >= 0 /\ y >=0] and [x1 <= x2] then [x1 op y <= x2 op y].
*)

asllib/ASTUtils.ml
```diff
@@ -593,60 +594,83 @@ let list_cross f li1 li2 =
...
 let is_left_increasing = function
-  | MUL | DIV | DIVRM | SHL | SHR | POW | PLUS | MINUS -> true
-  | AND | BAND | BEQ | MOD | BOR | EOR | EQ_OP | GT | GEQ | IMPL | LT | LEQ
-  | NEQ | OR | RDIV ->
+  | MUL | SHL | SHR | POW | PLUS | MINUS -> true
```
Roman-Manevich, line 605: If you agree with the definition given by my comment above, shouldn't DIV/DIVRM be true?

asllib/ASTUtils.ml
```diff
@@ -593,60 +594,83 @@ let list_cross f li1 li2 =
...
+  | DIV | DIVRM | AND | BAND | BEQ | MOD | BOR | EOR | EQ_OP | GT | GEQ | IMPL
```
Roman-Manevich, line 606: Shouldn't the result for MOD be false, rather than a failed assertion?
It looks like the function either returns true or fails.

asllib/ASTUtils.ml
```diff
@@ -593,60 +594,83 @@ let list_cross f li1 li2 =
...
+  | LT | LEQ | NEQ | OR | RDIV ->
       assert false
 
 let is_right_increasing = function
```
Roman-Manevich, line 610: (**
Tests whether the value of binary operation [op] increases along with its right-hand-side operand when all operands are non-negative. That is, if [is_right_increasing op] is true then for every two expressions [x op y1] and [x op y2] such that [y1 >= 0 /\ y2 >= 0 /\ x >=0]  and [y1 <= y2] then [x op y1 <= x op y2].
*)

asllib/ASTUtils.ml
```diff
@@ -593,60 +594,83 @@ let list_cross f li1 li2 =
...
   | MUL | SHL | SHR | POW | PLUS -> true
-  | DIV | DIVRM | MINUS -> false
-  | AND | BAND | MOD | BEQ | BOR | EOR | EQ_OP | GT | GEQ | IMPL | LT | LEQ
-  | NEQ | OR | RDIV ->
+  | MINUS -> false
```
Roman-Manevich, line 612: shouldn't SHR be false? 8 >> 1 is greater than 8 >> 2

asllib/Typing.ml
```diff
@@ -618,7 +618,8 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
 
   let binop_is_exploding = function
     | MUL | SHL | POW -> true
-    | PLUS | DIV | MINUS | MOD | SHR | DIVRM -> false
+    | DIV | DIVRM -> (* general case loses too much precision *) true
```
Roman-Manevich, line 621: I don't understand the comment. Perhaps add a comment explaining this function?

asllib/ASTUtils.ml
```diff
@@ -593,60 +594,83 @@ let list_cross f li1 li2 =
...
+  | DIV | DIVRM | AND | BAND | MOD | BEQ | BOR | EOR | EQ_OP | GT | GEQ | IMPL
+  | LT | LEQ | NEQ | OR | RDIV ->
       assert false
 
 let is_right_decreasing = function
```
Roman-Manevich, line 617: (**
Tests whether the value of binary operation [op] decreases along with its right-hand-side operand when all operands are non-negative. That is, if [is_right_decreasing op] is true then for every two expressions [x op y1] and [x op y2] such that [y1 >= 0 /\ y2 >= 0 /\ x >=0] and [y1 <= y2] then [x op y1 <= x op y2].
*)

Roman-Manevich, line 617: Do we even need this function? It's the boolean negation of is_right_incvreasing.

asllib/ASTUtils.ml
```diff
@@ -593,60 +594,83 @@ let list_cross f li1 li2 =
...
-    | Constraint_Exact e | Constraint_Range (_, e) ->
-        Constraint_Range (zero_expr, e)
+    | Constraint_Exact e | Constraint_Range (_, e) -> range zero_expr e
+  in
+  let constraint_divisions op c1 c2 =
+    match (c1, c2) with
+    | Constraint_Exact e1, Constraint_Exact e2 -> [ exact (binop op e1 e2) ]
+    | Constraint_Range (e11, e12), Constraint_Exact e2 ->
+        [ range (binop op e11 e2) (binop op e12 e2) ]
+    | Constraint_Exact e1, Constraint_Range (_e21, e22) ->
```
Roman-Manevich, line 658: Is this correct?
For example `10 DIV 1..10` = `1, 2, 5, 10`, not `[10..1, 1..10]`

asllib/ASTUtils.ml
```diff
@@ -593,60 +594,83 @@ let list_cross f li1 li2 =
...
+        (* {a} DIV {b..c} == {a .. (a DIV c), (a DIV c) .. a} *)
+        [ range e1 (binop op e1 e22); range (binop op e1 e22) e1 ]
+    | Constraint_Range (e11, e12), Constraint_Range (_e21, _e22) ->
```
Roman-Manevich, line 661: Likewise, I'm not sure the result is a fixed number of ranges.

asllib/ASTUtils.ml
```diff
@@ -108,6 +108,16 @@ let list_concat_map f l =
   in
   aux f [] l
 
+let list_take =
```
Roman-Manevich, line 111: How about renaming to list_prefix?

Roman-Manevich: Okay, but how about a documentation comment saying that it returns a prefix?

asllib/StaticOperations.ml
```diff
@@ -0,0 +1,440 @@
...
+  | SHR -> list_flat_cross constraint_shr cs1 cs2
+  | MOD -> List.map constraint_mod cs2
+  | SHL -> list_flat_cross constraint_shl cs1 cs2
+  | POW -> list_flat_cross constraint_pow cs1 cs2
+  | AND | BAND | BEQ | BOR | EOR | EQ_OP | GT | GEQ | IMPL | LT | LEQ | NEQ | OR
+  | RDIV ->
+      assert false
+(* End *)
+
+(* Begin FilterReduceConstraintDiv *)
```
Roman-Manevich, line 148: It's getting hard to follow the logic here. I think some comments could help.

Roman-Manevich (changes requested):

PR #1016, opened 2024-10-22
[asl] Prevent use of empty slices on integers and bitvectors.

HadrienRenaud:
This makes it impossible to call a bracket-less getter with brackets.

See the now forbidden tests:
```
getter f1 => bits(4)
begin
  return '0000';
end

setter f1 = v: bits(4)
begin
  pass;
end

func main () => integer
begin
  // both of the following are forbidden
  let - = f1[];
  f1[] = '';
  
  return 0;
end
```

### DIFF
--- a/asllib/StaticEnv.ml
+++ b/asllib/StaticEnv.ml
@@ -272,7 +272,15 @@ let add_subtype s t env =
     global = { env.global with subtypes = IMap.add s t env.global.subtypes };
   }
 
-let is_undefined name env =
+let is_global_undefined x (genv : global) =
   not
-    (IMap.mem name env.local.storage_types
-    || IMap.mem name env.global.storage_types)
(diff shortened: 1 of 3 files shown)


asllib/error.ml
```diff
@@ -149,6 +150,9 @@ module PPrint = struct
           pp_ty ty
           (pp_print_list ~pp_sep:pp_print_space pp_print_string)
           fields
+    | EmptySlice ->
+        fprintf f
+          "ASL Static Error:@ cannot@ slice@ with@ empty@ slicing@ operator"
```
Roman-Manevich, line 155: Do you mind adding that this may be due to incorrect invocation of a getter or a setter? Something like "expression is either a slice with an empty list of slices or an incorrect getter/setter invocation"

Roman-Manevich (approved):

PR #1018, opened 2024-10-24
[asl] Fix ATCs on types with same structure

HadrienRenaud:
ATCs on types with same structures are allowed by the type checker, but not supported by the interpreter. We change that by removing at type-checking time atcs on types whose structures type-satisfies one-another.


Roman-Manevich (approved):
Looks good to me.

Roman-Manevich (approved):
Looks good to me!

PR #1021, opened 2024-10-25
[asl][ci] Check that `make asldoc` runs

HadrienRenaud:
There has been a few times where the specification was broken, sometimes because innocuous-looking changes to the code could crash bento.

This create a new check that check that `make asldoc` runs on `ubuntu-latest` with `texlive-full` (and herdtools' dependencies) installed.

### DIFF
--- a/.github/workflows/build-asl-reference.yml
+++ b/.github/workflows/build-asl-reference.yml
@@ -1,0 +1,34 @@
+name: Build ASL reference
+
+on:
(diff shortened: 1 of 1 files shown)


Roman-Manevich (approved):
I only undestand the last line, but I guess it's okay.

Roman-Manevich (approved):
Looks good to me!

PR #1022, opened 2024-10-28
[Asl] Fix pattern for string types and improve error message

HadrienRenaud:
This PR comes from the following test:
```
func myfunction( s : string)
begin
   case s of
       when "hello" => print("helloworld\n");
       otherwise => return;
    end
end
```
This test should be accepted, which is done in the first commit of this PR.
Secondly, the error message on this test was quite confusing, as it was talking about the `==` binary operation. We change this back to an error message specific to patterns in the second commit of this PR.

### DIFF
--- a/asllib/error.ml
+++ b/asllib/error.ml
@@ -63,6 +63,7 @@ type error_desc =
   | BadRecursiveDecls of identifier list
   | UnrespectedParserInvariant
   | BadATC of ty * ty  (** asserting, asserted *)
+  | BadPattern of pattern * ty
   | ConstrainedIntegerExpected of ty
   | ParameterWithoutDecl of identifier
   | BaseValueEmptyType of ty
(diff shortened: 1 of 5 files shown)


asllib/error.ml
```diff
@@ -269,6 +270,10 @@ module PPrint = struct
            corresponding@ getter@ of@ signature@ @[@[%a@]@ ->@ %a@]"
           func.name (pp_comma_list pp_ty) args pp_ty ret
     | UnexpectedATC -> pp_print_text f "ASL Typing error: unexpected ATC."
+    | BadPattern (p, t) ->
+        fprintf f
+          "ASL Typing error:@ Bad@ pattern@ %a@ for@ expression@ of@ type@ %a."
```
Roman-Manevich, line 275: Nit: Perhaps change "Bad" to something like "Erroneous"?

Roman-Manevich (approved):
Looks good to me.

PR #1023, opened 2024-10-29
[Asl] Static base values

HadrienRenaud:
In ASL, you can write:
```
var x: integer;
```
And here, an ASL interpreter will declare `x` as having the "base value" of its type `integer`.

Before this PR, this is done dynamically in ASLRef. However, this is a bit complicated, so to simplify this, we want to make it a static transformation.

Now base values of types constrained by symbolic variable are forbidden:
```
func foo {N} (bv: bits(N))
begin
  var x: integer {0..N}; // Forbidden
end
```

We are also implementing base values as a static transformation of code in ASLv0, but in order not to fail on symbolic values, we plug-in arbitrary expressions, e.g. `Zeros (N)` or the first expression in a range for a constrained integer.

### DIFF
--- a/asllib/Typing.ml
+++ b/asllib/Typing.ml
@@ -2018,6 +2018,10 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
         let t = Types.make_anonymous env t in
         base_value_v0 ~loc env t
 
+  (** [base_value ~loc env e] is [base_value_v1 ~loc env e] if running for ASLv1.
+
+      Otherwise, it tries a more accepting algorithm with [base_value_v0 ~loc env e].
+  *)
   let base_value ~loc env e =
     try base_value_v1 ~loc env e
     with Error.ASLException _ as error ->
--- a/asllib/error.ml
+++ b/asllib/error.ml
@@ -167,12 +167,12 @@ module PPrint = struct
     | UndefinedIdentifier s ->
         fprintf f "ASL Error: Undefined identifier:@ '%s'" s
     | MismatchedReturnValue s ->
-        fprintf f "ASL Error: Mismatched use of return value from call to '%s'"
+        fprintf f "ASL Error: Mismatched use of return value from call to '%s'."
           s
     | BadArity (name, expected, provided) ->
         fprintf f
           "ASL Error: Arity error while calling '%s':@ %d arguments expected \
-           and %d provided"
+           and %d provided."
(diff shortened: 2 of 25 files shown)


asllib/Typing.ml
```diff
@@ -2986,24 +3097,28 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
       | Some t1, None -> t1
       | None, None -> Error.fatal_from loc UnrespectedParserInvariant
     in
+    let initial_value'' =
```
Roman-Manevich, line 3100: I think omputing `initial_value'` and then `initial_value''` is very fussy. Why not first compute just `initial_value_type`, then `declared_t` and then `initial_value'`?

asllib/error.ml
```diff
@@ -227,36 +228,41 @@ module PPrint = struct
...
           "ASL Typing error:@ explicit@ parameter@ %S@ does@ not@ have@ a@ \
-           corresponding@ defining@ argument"
+           corresponding@ defining@ argument."
           s
     | BaseValueEmptyType t ->
-        fprintf f "ASL Execution error: base value of empty type %a" pp_ty t
+        fprintf f "ASL Typing error: base value of empty type %a." pp_ty t
+    | BaseValueNonStatic (t, e) ->
+        fprintf f
+          "ASL Typing error:@ base@ value@ of@ type@ %a@ containing@ \
```
Roman-Manevich, line 256: I'm not sure the term "non-static value" makes sense.
How about something like "base value of type %a cannot be statically determined since it consists of %a"?

Roman-Manevich (commented):
Other than some minor comments, this look very good.

asllib/Typing.ml
```diff
@@ -1904,6 +1904,116 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
...
+          match length with
+          | ArrayLength_Enum (_, i) -> expr_of_int i
+          | ArrayLength_Expr e -> e
+        in
+        E_Array { length; value } |> add_pos
+    | T_Named _id ->
+        let t = Types.make_anonymous env t in
+        base_value_v1 loc env t
+
+  let base_value env loc e =
```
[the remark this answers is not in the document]
Roman-Manevich: Right. maybe just add that as a comment?

Roman-Manevich (approved):
Looks good to me!

PR #1027, opened 2024-10-31
Store version information in AST annotations

hrutvik:


asllib/ASTUtils.ml
```diff
@@ -49,9 +49,13 @@ module IMap = struct
...
 let dummy_pos = Lexing.dummy_pos
+let dummy_version = V1
 let desc v = v.desc
-let annotated desc pos_start pos_end = { desc; pos_start; pos_end }
-let add_dummy_pos desc = annotated desc dummy_pos dummy_pos
+
+let annotated desc pos_start pos_end version =
+  { desc; pos_start; pos_end; version }
+
+let add_dummy_pos desc = annotated desc dummy_pos dummy_pos dummy_version
```
Roman-Manevich, line 58: rename `annotate_dummy_pos` to something like `dummy_annotation`, since annotations are not just about source position now.

asllib/ASTUtils.mli
```diff
@@ -50,7 +50,10 @@ end
 val dummy_pos : Lexing.position
 (** A dummy position. *)
 
-val annotated : 'a -> position -> position -> 'a annotated
+val dummy_version : version
```
Roman-Manevich, line 53: perhaps rename to `default_version`?

asllib/ASTUtils.mli
```diff
@@ -50,7 +50,10 @@ end
...
+(** A dummy version. *)
+
+val annotated : 'a -> position -> position -> version -> 'a annotated
 (** [annotated v start end] is [v] with location specified as from [start] to
```
Roman-Manevich, line 57: document version as well.

asllib/Typing.ml
```diff
@@ -2019,15 +2019,13 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
...
-      Otherwise, it tries a more accepting algorithm with [base_value_v0 ~loc env e].
+  (** [base_value ~loc env e] is [base_value_v1 ~loc env e] if running for ASLv1,
+      or [base_value_v0 ~loc env e] if running for ASLv0.
   *)
   let base_value ~loc env e =
-    try base_value_v1 ~loc env e
-    with Error.ASLException _ as error ->
-      let+ () = fun () -> raise error in
-      base_value_v0 ~loc env e
+    match loc.version with
```
Roman-Manevich, line 2026: Nice

Roman-Manevich (changes requested):
This is actually looking quite good. I only have a few minor comments.

Roman-Manevich (approved):
Looks good to me.

asllib/ASTUtils.ml
```diff
@@ -49,17 +49,23 @@ module IMap = struct
...
+
+let dummy_annotated = add_dummy_annotation ()
 let to_pos pos = { pos with desc = () }
 
 let add_pos_from_st pos desc =
   if pos.desc == desc then pos else { pos with desc }
 
 let add_pos_from pos desc = { pos with desc }
-let with_pos_from pos { desc; _ } = add_pos_from pos desc
+let with_pos_from pos { desc; version } = add_pos_from { pos with version } desc
```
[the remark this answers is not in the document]
hrutvik: My thinking was that if given an AST generated by the V0 parser, we don't want to accidentally consider this as V1 AST, as it may become incompatible with V1 checking logic. Looking more closely, I realise that `with_pos_from` is used only once (in `StaticModel.ml`) - perhaps it would be better if I inlined its previous definition there?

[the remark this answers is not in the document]
hrutvik: Also forgot to say - for the compiler not giving warnings, from [what I can see](https://ocaml.org/manual/5.2/patterns.html#sss:pat-record) it doesn't mind for records:
> The record value can define more fields than field1 … fieldn; the values associated to these extra fields are not taken into account for matching.

PR #1028, opened 2024-10-31
[asl] Check that functions return something

HadrienRenaud:
The main goal of this PR is to forbid the type checking of a function that is meant to return something, but do not actually return anything.

The analysis is for the moment very simple, we can improve it if needs be.

This PR also makes case statements syntactic sugar for the purpose of the type checker (they were already syntactic sugar for the interpreter).

We also move `Unreachable ()` from the stdlib to a full AST node, as it has a special behaviour for this control-flow analysis, acting as an assertion that this control-flow path is not reached.

Two other implementation improvements are shipped: position annotation are added to pattern expressions, and CSV output of errors now displays a label with the error text.

### DIFF
--- a/asllib/aslref.ml
+++ b/asllib/aslref.ml
@@ -199,7 +199,7 @@ let () =
     match args.output_format with
     | Error.CSV ->
         Printf.eprintf
-          {|"File","Start line","Start col","End line","End col","Exception"
+          {|"File","Start line","Start col","End line","End col","Exception label","Exception"
 |}
     | Error.(HumanReadable | Silence) -> ()
   in
--- a/asllib/error.ml
+++ b/asllib/error.ml
@@ -103,6 +103,58 @@ type warning_desc =
 
 type warning = warning_desc annotated
 
+let error_label = function
+  | ReservedIdentifier _ -> "ReservedIdentifier"
+  | BadField _ -> "BadField"
+  | BadPattern _ -> "BadPattern"
+  | MissingField _ -> "MissingField"
+  | BadSlices _ -> "BadSlices"
+  | BadSlice _ -> "BadSlice"
+  | EmptySlice -> "EmptySlice"
+  | TypeInferenceNeeded -> "TypeInferenceNeeded"
+  | UndefinedIdentifier _ -> "UndefinedIdentifier"
+  | MismatchedReturnValue _ -> "MismatchedReturnValue"
+  | BadArity _ -> "BadArity"
+  | UnsupportedBinop _ -> "UnsupportedBinop"
+  | UnsupportedUnop _ -> "UnsupportedUnop"
+  | UnsupportedExpr _ -> "UnsupportedExpr"
+  | InvalidExpr _ -> "InvalidExpr"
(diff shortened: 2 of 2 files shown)


asllib/Typing.ml
```diff
@@ -2314,7 +2316,29 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
...
+              t (* Assertion that the condition always holds *)
+          | Interrupt, Interrupt -> Interrupt)
+
+    let check_interrupts ~loc name t () =
+      match t with
+      | Unreachable | Interrupt -> ()
+      | Nothing -> fatal_from loc (Error.NonReturningFunction name)
+  end
+
+  let rec annotate_stmt env s : stmt * env * ControlFlow.t =
```
Roman-Manevich, line 2341: Do you have to perform the check while annotating statements?
Can't this check be done in a separate phase, after annotating the body of a subprogram?

Roman-Manevich: It's not necessarily easier to document, but I think it's cleaner in terms of separating concerns and making it easier to understand the static semantics.

Roman-Manevich: I think [2a3bb70](https://github.com/herd/herdtools7/commit/2a3bb709c941d89cd7fe7e1d0684303ad3f6c589) is really good. Please let's go with it. I left some comments.

asllib/Typing.ml
```diff
@@ -2917,6 +2903,77 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
...
 
+  module ControlFlow : sig
+    val check_stmt_interrupts : identifier -> env -> stmt_desc annotated -> prop
+    (** [check_stmt_interrupts name env body] checks that the function named
+        [name] with the statement body [body] returns a value or throws an
+        exception. *)
+  end = struct
+    (** Possible Control-Flow actions of a statement. *)
+    type t =
+      | Nothing  (** Continuing simply to the next statement. *)
```
Roman-Manevich, line 2914: [Naming] In terms of a dataflow analysis, I think this is a may analysis with `Nothing` as the top element of the lattice. Essentially, I think it means "considering all the control flow paths of a given statement, there is at least one path that does not return a value or throws an exception."
`Nothing` doesn't convey any information and the comment I think does not reflect the property above.
How about something like `Top` or `MaybeErroneous`?

Roman-Manevich (approved):
Looks good to me.

asllib/Typing.ml
Roman-Manevich: I think this is okay.
I the lattice I am thinking of is the powerset of the kinds of runtime configurations that the semantics of a statement may result with: Returning/Continuing/Normal/Throwing, with union as the join operator.
This is slightly precise than needed, so I think you're using a further abstraction of that lattice where `Nothing` corresponds to the entire set of configurations (top), Interrupt corresponds to any non-empty subset of Returning/Throwing, and `Unreachable` corresponds to the empty set (bottom).

Roman-Manevich (approved):
Looks good to me.

PR #1031, opened 2024-11-04
[gen] allow address dependencies landing on code location

hugookeeffe:
This PR allows address dependencies to land on code locations when DC's or IC's are present only. This is necessary to allow tests such as the following to be generated:

```
AArch64 MP.RF+dmb.st+addr-dc.cvau-dsb.ish-ic.ivau-dsb.ish-isb
Variant=ifetch
{
  0:X0=instr:"NOP";
  0:X1=P1:L0; 1:X1=P1:L0;
  0:X3=y; 1:X3=y;
  1:X9=0;
}
 STR W0,[X1] | LDR W2,[X3]       ;
 DMB ST      | EOR W6,W2,W2      ;
 MOV W2,#1   | ADD X1,X1,W6,SXTW ;
 STR W2,[X3] | DC CVAU,X1        ;
             | DSB ISH           ;
             | IC IVAU,X1        ;
             | DSB ISH           ;
             | ISB               ;
             |L0:                ;
             | B over            ;
             | MOV W9,#1         ;
             |over:              ;
exists 1:X2=1 /\ 1:X9=0
```

This PR allows a very similar test to be generated, with the following command and output:
`diyone7 -arch AArch64 -variant self -moreedges true "DMB.ISHdWW Rfe DpAddrdRPI DC.CVAUn DSB.ISH IC.IVAUn DSB.ISH ISB FreIP"`
```
AArch64 A
"DMB.ISHdWW Rfe DpAddrdRPI DC.CVAUn DSB.ISH IC.IVAUn DSB.ISH ISB FreIP"
Generator=diyone7 (version 7.57+1)
Com=Rf Fr
Orig=DMB.ISHdWW Rfe DpAddrdRPI DC.CVAUn DSB.ISH IC.IVAUn DSB.ISH ISB FreIP
{
0:X0=instr:"NOP"; 0:X1=P1:Lself00; 0:X3=x;
1:X0=x; 1:X3=P1:Lself00;
}
 STR W0,[X1] | LDR W1,[X0]       ;
 DMB ISH     | EOR W2,W1,W1      ;
 MOV W2,#1   | ADD X3,X3,W2,SXTW ;
 STR W2,[X3] | DC CVAU,X3        ;
             | DSB ISH           ;
             | IC IVAU,X3        ;
             | DSB ISH           ;
             | ISB               ;
             | Lself00:          ;
             | B .+12            ;
             | MOV W4,#2         ;
             | B .+8             ;
             | MOV W4,#1         ;
exists (1:X1=1 /\ 1:X4=1)
```

an important note here is the change in the code to make the first two registers of the Add instruction point to the code location, this is necessary to create forbidden tests.

### DIFF
--- a/gen/AArch64Compile_gen.ml
+++ b/gen/AArch64Compile_gen.ml
@@ -2762,11 +2762,22 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
         let n2 = C.find_non_pseudo n in
         let loc = match n2.C.evt.C.loc with
         | Data loc -> loc
-        | Code _ -> Warn.fatal "Can't create address dependency for a code location" in
+        | Code lab -> match f with
+          | CMO (_,Next) -> lab
+          | _ -> Warn.fatal "Address dependency to code location must be directly followed by DC.CVAUn or IC.IVAUn" in
+        (try ignore (C.find_node
+          (fun m -> match m.C.edge.E.edge with
+            | E.Insert (CMO (_,Prev)) when (loc_eq m.C.evt.C.loc n.C.evt.C.loc) ->
+              Warn.fatal "Address dependency to code location cannot have IC.IVAUp or DC.CVAUp"
+            | _ -> false )
+          n)
+        with Not_found -> ());
         let r2,st = next_reg st in
         let cs0,st =  calc0_gen csel st vdep r2 r1 in
         let rB,init,st = U.next_init st p init loc in
-        let r,st = tempo1 st in
+        let r,st = match f with
+          | CMO _ -> rB,st
+          | _ -> tempo1 st in
         let cs2 = (if do_morello then
           [do_addcapa rB r r2]
         else


hugookeeffe:
this PR has been tested by running diy7 with forbidden_ifetch.conf in gen/libdir/ which succesfully generated desired tests such as the one mentioned above. Furthermore the tests were ran through herd7, with the following PR cherrypicked https://github.com/herd/herdtools7/pull/952, showing all tests as forbidden

artkhyzha:
Thank you for writing up the explanations, Hugo. If I understand correctly, this increases the number of forbidden tests the previous PR generates. Looks good to me!

hugookeeffe:
@maranget do you have any thoughts on this PR?  From the generated tests I looked at, I couldn't see anything wrong but perhaps I missed something

hugookeeffe:
thank you @relokin for pointing that out. Indeed this wouldn't produce an error when it should. I have just pushed a new commit that makes sure the error is thrown when DC.CVAU is pointing to the previous register

relokin (approved):
Thanks Hugo, LGTM. Let's give this PR some time if @maranget wants to review it as well. Otherwise I'll ping you on Friday to rebase on top of master and merge.

maranget:
Hi @hugookeeffe and @relokin. Code is fine, with minimal intervention :).

Out of curiosity I had the iidea to have one `p` and one `n`...
```
% diyone7 -arch AArch64 -variant self -moreedges true "DMB.ISHdWW Rfe DpAddrdRPI DC.CVAUn DSB.ISH IC.IVAUp DSB.ISH ISB FreIP"
AArch64 A
"DMB.ISHdWW Rfe DpAddrdRPI DC.CVAUn DSB.ISH IC.IVAUp DSB.ISH ISB FreIP"
Generator=diyone7 (version 7.57+1)
Com=Rf Fr
Orig=DMB.ISHdWW Rfe DpAddrdRPI DC.CVAUn DSB.ISH IC.IVAUp DSB.ISH ISB FreIP
{
0:X0=instr:"NOP"; 0:X1=P1:Lself00; 0:X3=x;
1:X0=x; 1:X3=P1:Lself00;
}
 STR W0,[X1] | LDR W1,[X0]       ;
 DMB ISH     | EOR W2,W1,W1      ;
 MOV W2,#1   | ADD X3,X3,W2,SXTW ;
 STR W2,[X3] | DC CVAU,X3        ;
             | DSB ISH           ;
             | IC IVAU,X0        ;
             | DSB ISH           ;
             | ISB               ;
             | Lself00:          ;
             | B .+12            ;
             | MOV W4,#2         ;
             | B .+8             ;
             | MOV W4,#1         ;
exists (1:X1=1 /\ 1:X4=1)

```
That is, success, while swapping `p` and `n` yields a failure:
```
% diyone7 -arch AArch64 -variant self -moreedges true "DMB.ISHdWW Rfe DpAddrdRPI DC.CVAUp DSB.ISH IC.IVAUn DSB.ISH ISB FreIP"
/opt/WIP/bin/diyone7: Fatal error: Test A [DMB.ISHdWW Rfe DpAddrdRPI DC.CVAUp DSB.ISH IC.IVAUn DSB.ISH ISB FreIP] failed:
can't create address dependency for a code location without DC.CVAU or IC.IVAU
```
(I notice that the error message does not help to fix the problem)

Is this the intended behaviour? Maybe there is nothing to worry about anyway.

hugookeeffe:
thank you @maranget for having a look! thank you for spotting that. Indeed that is not the intended behaviour and after discussing with @artkhyzha it seems we don't want to allow DC/IC to the previous register at all (atleast for now). I have just pushed a commit that addresses this problem. It does unfortunately require traversing the nodes, so the code is not very pretty but I believe it should work. Let me know if you have any suggestions on how to better write this.

gen/AArch64Compile_gen.ml
```diff
@@ -2764,7 +2764,14 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
         | Data loc -> loc
         | Code lab -> match f with
           | CMO (_,Next) -> lab
-          | _ -> Warn.fatal "can't create address dependency for a code location without DC.CVAU or IC.IVAU" in
+          | _ -> Warn.fatal "Address dependency to code location must be directly followed by DC.CVAUn or IC.IVAUn" in
+        (try ignore (C.find_node
+          (fun m -> match m.C.edge.E.edge with
+            | E.Insert (CMO (_,Prev)) when m.C.evt.C.loc == n.C.evt.C.loc ->
```
maranget, line 2770: Are you sure about using physical equality?

hugookeeffe: Hi @maranget no, thank you for spotting that. I have just changed it to use loc_compare instead.

gen/AArch64Compile_gen.ml
```diff
@@ -2762,11 +2762,22 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
         let n2 = C.find_non_pseudo n in
         let loc = match n2.C.evt.C.loc with
         | Data loc -> loc
-        | Code _ -> Warn.fatal "Can't create address dependency for a code location" in
+        | Code lab -> match f with
+          | CMO (_,Next) -> lab
+          | _ -> Warn.fatal "Address dependency to code location must be directly followed by DC.CVAUn or IC.IVAUn" in
+        (try ignore (C.find_node
+          (fun m -> match m.C.edge.E.edge with
```
relokin, line 2769: Could you do?
```suggestion
        (try ignore (C.find_edge
          (function
```
Another suggestion would be to add an `iter_node/iter_edge` function in cycle.ml which would simplify the code here as well.

gen/AArch64Compile_gen.ml
```diff
@@ -2762,11 +2762,22 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
...
+            | E.Insert (CMO (_,Prev)) when (loc_compare m.C.evt.C.loc n.C.evt.C.loc) = 0 ->
```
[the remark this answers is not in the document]
hugookeeffe: thank you for the suggestion!

gen/AArch64Compile_gen.ml
hugookeeffe: i dont think we can use find_edge, because we need to get the edge type from the node, which contains the actual edge and two atom options, and then edge. find_edge would only work for the actual edge. We could change the input to the function to be n.C.edge, but i'm not sure that would make the code much cleaner.

i agree an iter_node/iter_edge function would be very useful. Would it be ok to do it in a different PR? just because there are other places it could also be used, and i think it could use its own PR because of that

relokin:
Thanks @hugookeeffe! Thanks @artkhyzha and @maranget for the reviews!

PR #1041, opened 2024-11-08
[herd]: Fix address generation for STG, LDG and friends (again)

murzinv:
Partially revert c00eaf38b63 ("[herd]: Fix address generation for STG, LDG and friends").

Altough it is correct that address generation using immediate signed offset scaled by the Tag granule, it only applies to encoding of the instruction but immediate for assembler has different requirement:

```
<simm> Is the optional signed immediate offset, a multiple of 16
in the range -4096 to 4080, defaulting to 0 and encoded in the "imm9"
field.
```
and it is what assembler implements

```
Error: immediate value must be a multiple of 16 at operand 2 -- `stg X1,[X0,#1]'
Error: immediate value must be a multiple of 16 at operand 2 -- `ldg X3,[X0,#1]'
```

So roll back herd part and update MTE tests, also add extra check for immediate and dedicated litmus test to exercise the check.

### DIFF
--- a/herd/AArch64Sem.ml
+++ b/herd/AArch64Sem.ml
@@ -3090,7 +3090,7 @@ module Make
         end
 
       let ldg rt rn k ii =
-        let ma = get_ea rn (AArch64.K k) (AArch64.S_LSL MachSize.granule_log2bytes) ii in
+        let ma = get_ea rn (AArch64.K k) AArch64.S_NOEXT ii in
         let do_ldg a_virt ac ma =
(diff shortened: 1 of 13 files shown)


relokin (approved):
Looks good to me!

relokin:
@maranget I would like to merge this, unless you want to have a look or need more time.

maranget:
Hi @relokin and @murzinv. Please proceed.

relokin:
Merged thanks @murzinv and @maranget

PR #1042, opened 2024-11-08
Basic MTE support in litmus

murzinv:
Introduce basic support for MTE.  Simple litmus test
```
AArch64 T
{
int y=2;
int x=1;
tag(x)=:red;
0:X0=x:green; 
0:X2=y:blue;
0:X3=y:green;
}
 P0          ;
 STG X2,[X3] ;
 LDR W1,[X0] ;
forall [tag(y)]=:blue /\ [tag(x)]=:red /\ x=1 /\ y=2 /\ fault(P0,x)
```
can be build with
```
$ litmus7 -variant memtag  -mach kvm-armv8.5-a+memtag -a 1 -s 100 -r 10 T.litmus -o /kvm-unit-tests/t/
```
and ran
```
$ QEMU=qemu-system-aarch64 ACCEL=tcg sh t/run.sh
Tue Jul 29 04:18:54 PM BST 2025
%%%%%%%%%%%%%%%%%%%%%%%%%
% Results for T.litmus %
%%%%%%%%%%%%%%%%%%%%%%%%%
AArch64 T

{
 int y=2;
 int x=1;
 [tag(x)]=:red;
 0:X0=x:green; 0:X2=y:blue; 0:X3=y:green;
}
 P0          ;
 STG X2,[X3] ;
 LDR W1,[X0] ;

forall ([tag(y)]=:blue /\ [tag(x)]=:red /\ [x]=1 /\ [y]=2 /\ fault(P0,x))
Generated assembler
#START _litmus_P0
	nop
	stg x2,[x1]
	ldr w3,[x0]
	nop
qemu-system-aarch64 -nodefaults -machine virt -accel tcg -cpu max -device virtio-serial-device -device virtconsole,chardev=ctd -chardev testdev,id=ctd -device pci-testdev -display none -serial stdio -kernel t/./T.flat -smp 1 -append -q -machine mte=on # -initrd /tmp/tmp.nombuyASov
Test T Required
Histogram (1 states)
1000  :>[x]=1; [y]=2; [tag(x)]=:red; [tag(y)]=:blue; fault(P0,x,TagCheck);
Ok

Witnesses
Positive: 1000, Negative: 0
Condition forall ([tag(y)]=:blue /\ [tag(x)]=:red /\ [x]=1 /\ [y]=2 /\ fault(P0,x)) is validated
Hash=9bb639ee811dc6a38b59edcb2eccfe3d
Observation T Always 1000 0
Faults T 1000 P0:1000
Time T 0.02

EXIT: STATUS=1
Revision d3d0a61746bb864deff8a69bc73310c5b44c3921, version 7.58+1
Command line: litmus7 -mach kvm-armv8.5-a+memtag -a 1 -s 100 -r 10 T.litmus -variant memtag -o /kvm-unit-tests/t
Parameters
#define SIZE_OF_TEST 100
#define NUMBER_OF_RUN 10
#define AVAIL 1
/* gcc options: -Wall -std=gnu99 -std=gnu99 -ffreestanding -I $(SRCDIR)/lib -I $(SRCDIR)/libfdt -Wall -Werror  -fomit-frame-pointer -Wno-frame-address   -fno-pic  -no-pie -Wmissing-parameter-type -Wold-style-declaration -Woverride-init -O2 $(call optional-ccopt, -mno-outline-atomics) -march=armv8.5-a+memtag */
/* barrier: userfence */
/* alloc: static */
/* proc used: 1 */
LITMUSOPTS=
Tue Jul 29 04:18:54 PM BST 2025
```

P.S.
All required support from `kvm-unit-test` available upstream, so make sure you use latest and greatest version :wink:


maranget:
> * `AK002.litmus` - is that related to MTE?

Hi @murzinv, this is unrelated to MTE and could probably be moved into `herd/tests//instructions/AArch64`  --- beware a different test with the same name exists there.

In any case, the test now compiles with master.

litmus/variant_litmus.ml
```diff
@@ -40,24 +42,30 @@ let parse s = match Misc.lowercase s with
 | "telechat" -> Some Telechat
 | "sve" -> Some SVE
 | "sme" -> Some SME
+| "memtag" | "mte" -> Some MemTag
 | tag ->
   match
    Misc.app_opt (fun p -> FaultHandling p) (Fault.Handling.parse tag)
   with
   | Some _ as r ->  r
```
relokin: This is just a suggestion but would it better to flatten the pattern matching? Something along the lines of

```suggestion
| tag when Option.is_some (Fault.Handling.parse tag) ->
  let fh = Option.as_some (Fault.Handling.parse tag) in
  FaultHandling fh
| tag when  Option.is_some (Precision.parse tag) ->
  let p = Option.as_some (Precision.parse tag) in
  MTEPrecision p
 | tag ->
  ...
```

lib/misc.ml
```diff
@@ -742,6 +742,33 @@ let pp_physical = sprintf "PA(%s)"
...
+let tag_of_int v = match v with
+  | 0 -> "green"
+  | 1 -> "red"
+  | 2 -> "blue"
+  | 3 -> "black"
+  | 4 -> "white"
+  | 5 -> "cyan"
+  | 6 -> "yellow"
+  | 7 -> "magenta"
+  | _ -> raise (Fatal (sprintf "cannot convert %d of int" v))
```
relokin: minor nit but it might slightly better to have an associative list [("green", 0) ... ] and the implement int_of_tag with List.assoc and tag_of_int with a custom iterator that way the two functions cannot get out of sync.

litmus/libdir/_aarch64/intrinsics.h
```diff
@@ -13,4 +13,61 @@
...
-#endif
\ No newline at end of file
+#endif
+
+#ifdef __ARM_FEATURE_MEMORY_TAGGING
+#define MTE_GRANULE_SIZE        16UL
+#define MTE_GRANULE_MASK        (MTE_GRANULE_SIZE - 1)
+#define MTE_TAG_SHIFT           56
+
+#include <arm_acle.h>
```
relokin: It's a little unclear, do you need this?

litmus/libdir/_aarch64/_memtag.c
relokin, line 1: Please feel free to ignore this, but ideally these constants would be #defines

litmus/libdir/_aarch64/kvm-headers.h
```diff
@@ -182,6 +184,15 @@ static inline void litmus_set_pte_attribute(pteval_t *p,pte_attr_key k) {
   case attr_Device_GRE:
     *p = litmus_set_memattr(*p, MT_DEVICE_GRE);
     break;
+  case attr_TaggedNormal:
+#ifdef __ARM_FEATURE_MEMORY_TAGGING
```
relokin, line 188: Is this ifdef needed?

litmus/libdir/_aarch64/intrinsics.h
murzinv: That's leftover from my early thoughts on useing [MTE intrinsics](https://arm-software.github.io/acle/main/acle.html#mte-intrinsics). I did not go that path, so include not required

litmus/libdir/_aarch64/kvm-headers.h
murzinv: IIRC logic is as follows: if it is truly MTE test we get know about it from compiler via `__ARM_FEATURE_MEMORY_TAGGING` so we then want to mandate `kvm-unit-test` version which comes with `MT_NORMAL_TAGGED`.
In case it is the test which just happen to set `TaggedNormal` attribute but has not requested compiler to support MTE (so we have no `__ARM_FEATURE_MEMORY_TAGGING`) because it doesn't do anything MTE specific then `TaggedNormal` behaves the same as `Normal, iWB, iWB` which is default attribute and we can:
 - skip call to `litmus_set_memattr`
 - work with old(is) `kvm-unit-test` (with no `MT_NORMAL_TAGGED` being defined)

PR #1045, opened 2024-11-10
[asl] Fully support constant functions

HadrienRenaud:
In this PR, we use another instance of the dynamic interpreter to implement the static interpreter.

In this way, complex constants such as `double(32)` or `Zeros(2)` are replaced by their evaluations, here respectively `64` and `'00'`.

### DIFF
--- a/asllib/StaticInterpreter.ml
+++ b/asllib/StaticInterpreter.ml
@@ -45,12 +45,7 @@ let static_eval (senv : SEnv.env) (e : expr) : literal =
   let env =
     let open SI.IEnv in
     {
-      global =
-        {
-          static = senv.global;
-          funcs = senv.global.subprograms;
-          storage = global_storage;
-        };
+      global = { static = senv.global; storage = global_storage };
       local = empty_scoped ~storage:local_storage (SB.Scope.global ~init:true);
     }
   in
--- a/asllib/Interpreter.ml
+++ b/asllib/Interpreter.ml
@@ -206,17 +206,6 @@ module Make (B : Backend.S) (C : Config) = struct
   (*                                                                           *)
   (*****************************************************************************)
 
-  (* Functions *)
-  (* --------- *)
-
-  (** [build_funcs] initialize the unique calling reference for each function
-      and builds the subprogram sub-env. *)
-  let build_funcs ast (funcs : func IMap.t) =
-    List.to_seq ast
-    |> Seq.filter_map (fun d ->
-           match d.desc with D_Func func -> Some (func.name, func) | _ -> None)
-    |> Fun.flip IMap.add_seq funcs
-
   (* Global env *)
   (* ---------- *)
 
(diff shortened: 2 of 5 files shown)


asllib/Native.ml
```diff
@@ -118,7 +122,7 @@ module NativeBackend = struct
   let binop op v1 v2 =
```
Roman-Manevich, line 122: It would be nice to have a source location for the error reports here.

asllib/StaticInterpreter.ml
```diff
@@ -25,92 +25,44 @@ open ASTUtils
...
-          let pfactor = eval_to_int efactor and plength = eval_to_int elength in
-          let ptop = (pfactor * plength) + plength - 1 in
-          (ptop, plength, efactor)
-    in
-    let bot = top - len + 1 in
-    if top >= bot && bot >= 0 then interval top len
-    else fatal_from pos @@ Error.BadSlice slice
+let static_eval (senv : SEnv.env) (e : expr) : literal =
+  let funcs =
+    IMap.map
```
Roman-Manevich, line 41: I think it's better to do it once after building the global environment and store it inside the global static environment.

asllib/StaticInterpreter.ml
```diff
@@ -25,92 +25,44 @@ open ASTUtils
...
+      (fun (func : AST.func) -> (ref 0, func))
+      senv.SEnv.global.subprograms
+  and global_storage =
+    IMap.map SB.v_of_literal senv.SEnv.global.constant_values
+    |> Storage.of_v_map
```
Roman-Manevich, line 46: Not sure about it, but perhaps you should also include local/global `let` variables that can be normalized to a constant.

asllib/StaticInterpreter.ml
Roman-Manevich: Sorry for being vague. I meant computing and caching `funcs`.

Roman-Manevich (approved):
Looks good to me!

asllib/StaticEnv.mli
```diff
@@ -25,15 +25,18 @@ open ASTUtils
 
 (** Static Environment used for type-checking (cf {!Typing}. *)
 
+type func = int ref * AST.func
```
Roman-Manevich, line 28: So this allows referring to each subprogram uniquely?
Is this instead of the name mangling for overloaded functions (that is, naming them foo1, foo2, etc.)?
Why do we need a reference though?

asllib/StaticInterpreter.ml
```diff
@@ -37,19 +37,21 @@ module SB = Native.StaticBackend
...
+  in
   let env =
     let open SI.IEnv in
     {
-      global = { static = senv.global; funcs; storage = global_storage };
-      local = local_of_v_map local_v_map;
+      global =
+        {
+          static = senv.global;
+          funcs = senv.global.subprograms;
```
Roman-Manevich, line 51: Do you actually need funcs here if it's just an alias for `senv.global.subprograms`?

Roman-Manevich (approved):
Looks good to me.
There's only one comment.

PR #1046, opened 2024-11-11
[asl] Reintroduce equality with masks as sugar for patterns in v1.

HadrienRenaud:
- [x] Remove possibility to write `'1010' IN '10x0'`: braces are mandatory and this should be written `'1010' IN {'10x0'}`
- [x] Add possibility to write `'1010' == '10x0'`: this is a sugar for `'1010' IN {'10x0'}`

### DIFF
--- a/asllib/Parser.mly
+++ b/asllib/Parser.mly
@@ -344,10 +344,14 @@ let make_expr(sub_expr) ==
     | e=sub_expr; DOT; x=IDENTIFIER;                              < E_GetField           >
     | e=sub_expr; DOT; fs=bracketed(nclist(IDENTIFIER));          < E_GetFields          >
     | ~=bracketed(nclist(expr));                                  < E_Concat             >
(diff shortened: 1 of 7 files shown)


asllib/Parser.mly
```diff
@@ -372,33 +376,28 @@ let int_constraint ==
...
   annotated (
     | ~=expr_pattern; < Pattern_Single >
     | e1=expr_pattern; SLICING; e2=expr; < Pattern_Range >
     | MINUS; { Pattern_All }
     | LEQ; ~=expr; < Pattern_Leq >
     | GEQ; ~=expr; < Pattern_Geq >
-    | ~=MASK_LIT; < Pattern_Mask >
     | ~=plist2(pattern); < Pattern_Tuple >
   )
+  | pattern_mask
```
Roman-Manevich, line 393: Just curious: why did you move `pattern_mask` outside of this `annotated` block and into its own. It seems equivalent, no?

Roman-Manevich (approved):
Looks good to me!

PR #1047, opened 2024-11-11
[asl] Use parentheses for accessors

hrutvik:
This PR implements the following for accessors (getters/setters):
- forbid accessors without arguments
- use parentheses to declare and invoke accessors, not square brackets
- permit only a few forms on assignment to setters: `Setter() = rhs`, `Setter.fld = rhs`, and `Setter.[fld1,fld2] = rhs`.

All forms are still permitted for ASL0, to maintain backwards compatibility - but the relevant behaviour is gated using an AST versioning check (https://github.com/herd/herdtools7/pull/1027). Note that if calling a no-argument ASL0 accessor from an ASL1 specification, an empty argument list `()` must still be provided.

`herd` relies on an empty accessor to intercept reads/writes to the `_PC` variable seamlessly - this had to be moved to an ASL0 implementation to preserve its behaviour.

### DIFF
--- a/asllib/doc/ASLmacros.tex
+++ b/asllib/doc/ASLmacros.tex
@@ -680,6 +680,7 @@
 \newcommand\builddecl[0]{\hyperlink{build-decl}{\textfunc{build\_decl}}}
 \newcommand\builddeclitem[0]{\hyperlink{build-declitem}{\textfunc{build\_decl\_item}}}
 \newcommand\builduntypeddeclitem[0]{\hyperlink{build-untypeddeclitem}{\textfunc{build\_untyped\_decl\_item}}}
+\newcommand\desugarsetter[0]{\hyperlink{def-desugarsetter}{\textfunc{desugar\_setter}}}
 \newcommand\buildstmt[0]{\hyperlink{build-stmt}{\textfunc{build\_stmt}}}
 \newcommand\buildexpr[0]{\hyperlink{build-expr}{\textfunc{build\_expr}}}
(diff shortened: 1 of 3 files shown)


asllib/ASTUtils.ml
```diff
@@ -804,3 +804,33 @@ let rename_locals map_name ast =
...
+  let () = assert (loc.version = V1) in
+  let here desc = add_pos_from loc desc in
+  match fields with
+  | [] -> S_Call (name, rhs :: args, []) (* Setter(rhs, ...); *)
+  | _ ->
+      let temp = fresh_var "__setter_v1_temporary" in
+      (* temp = Getter(...); *)
+      let read =
+        let getter_call = E_Call (name, args, []) |> here in
+        S_Decl (LDK_Var, LDI_Var temp, Some getter_call) |> here
```
Roman-Manevich, line 819: Maybe use `LDK_Let` instead of `LDK_Var`?

hrutvik: My understanding was that because we then then modify `temp`, it needs to be `LDK_Var`?

Roman-Manevich: Oh, sorry. I thought it wasn't being modified.

hrutvik:
I'm a little wary about simplifying the V0 side of things as part of this PR - in particular, removing the various checks for `ST_Empty{Getter,Setter}` when reducing a variable to an accessor call, and removing checks that e.g. the calls of V0 getters using `[...]` don't accidentally get conflated with syntactically distinct function calls using `(...)`. 

Can I suggest that we leave this simplification out of scope for this PR, and look into it separately?

PR #1049, opened 2024-11-13
[asl] Stdlib updates

HadrienRenaud:
- [x] Add new `Align{Up,Down}{Size,P2}` functions for integers and bitvectors
- [x] Add new implementation of `SQRTRounded`
- [x] Add new implementation of `ILog2`, an equivalent of `log2` for reals.
- [x] Switch loop limit implementation to use `Z.t` as those functions have virtually unbounded loop limit, e.g. `2^128`, which does not fit in a standard ocaml integer.

### DIFF
--- a/asllib/libdir/stdlib.asl
+++ b/asllib/libdir/stdlib.asl
@@ -85,6 +85,49 @@ begin
     return FloorPow2(x) == CeilPow2(x);
 end;
 
+// AlignDownSize()
+// ===============
+// For a non-negative integer x and positive integer size, returns the greatest
+// multiple of size that is less than or equal to x.
+
+func AlignDownSize(x: integer, size: integer) => integer
+begin
+    assert size > 0;
+    return (x DIVRM size) * size;
+end;
+
+// AlignUpSize()
+// =============
+// For a non-negative integer x and positive integer size, returns the smallest
+// multiple of size that is greater than or equal to x.
+
+func AlignUpSize(x: integer, size: integer) => integer
+begin
+    assert size > 0;
+    return AlignDownSize(x + (size - 1), size);
+end;
+
+// AlignDownP2()
+// =============
+// For non-negative integers x and p2, returns the greatest multiple of 2^p2
+// that is less than or equal to x.
+
+func AlignDownP2(x: integer, p2: integer) => integer
+begin
+    assert p2 >= 0;
+    return AlignDownSize(x, 2^p2);
+end;
+
+// AlignUpP2()
+// ===========
+// For non-negative integers x and p2, returns the smallest multiple of 2^p2
+// that is greater than or equal to x.
+
+func AlignUpP2(x: integer, p2: integer) => integer
+begin
+    assert p2 >= 0;
+    return AlignUpSize(x, 2^p2);
+end;
 
 //------------------------------------------------------------------------------
 // Functions on reals
@@ -210,7 +253,7 @@ begin
   return if isZero then Zeros{N} else Ones{N};
 end;
 
-// Returns a bit vector of width N, containing (N DIV M) copies of input bit
+// Returns a bitvector of width N, containing (N DIV M) copies of input bit
 // vector x of width M. N must be exactly divisible by M.
 func Replicate{N,M}(x: bits(M)) => bits(N)
 begin
@@ -326,6 +369,50 @@ begin
   else
     return x[N-1:y]+1 :: Zeros{y};
   end;
+end;
+
+// Bitvector alignment functions
+// =============================
+
+// AlignDownSize()
+// ===============
+// A variant of AlignDownSize where the bitvector x is viewed as an unsigned
+// integer and the resulting integer is represented by its first N bits.
+
+func AlignDownSize{N}(x: bits(N), size: integer {1..2^N}) => bits(N)
+begin
+    return AlignDownSize(UInt(x), size)[:N];
+end;
+
+// AlignUpSize()
+// =============
+// A variant of AlignUpSize where the bitvector x is viewed as an unsigned
+// integer and the resulting integer is represented by its first N bits.
+
+func AlignUpSize{N}(x: bits(N), size: integer {1..2^N}) => bits(N)
+begin
+    return AlignUpSize(UInt(x), size)[:N];
+end;
+
+// AlignDownP2()
+// =============
+// A variant of AlignDownP2 where the bitvector x is viewed as an unsigned
+// integer and the resulting integer is represented by its first N bits.
+
+func AlignDownP2{N}(x: bits(N), p2: integer {0..N}) => bits(N)
+begin
+    if N == 0 then return x; end;
+    return x[N-1:p2] :: Zeros{p2};
+end;
+
+// AlignUpP2()
+// ===========
+// A variant of AlignUpP2 where the bitvector x is viewed as an unsigned
+// integer and the resulting integer is represented by its first N bits.
+
+func AlignUpP2{N}(x: bits(N), p2: integer {0..N}) => bits(N)
+begin
+    return AlignDownP2{N}(x + (2^p2 - 1), p2);
 end;
 
 // The shift functions LSL, LSR, ASR and ROR accept a non-negative shift amount.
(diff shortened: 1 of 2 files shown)


asllib/Typing.ml
```diff
@@ -2670,55 +2669,42 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
     (** Sequencial combination of two control flows. *)
     let seq t1 t2 = if t1 = MayNotInterrupt then t2 else t1
 
-    (** [guard env e t1 t2] correspond to the conditional combination of [t1]
+    (** [or_ t1 t2] correspond to the conditional combination of [t1]
```
Roman-Manevich, line 2672: How about `or_` -> `join_control_flow_states`?

asllib/libdir/stdlib.asl
```diff
@@ -34,6 +34,11 @@ end;
 
 // Log2
 
+func IsPowerOfTwo(x: integer) => boolean
+begin
+  return x > 0 && x == 2 ^ Log2 (x);
```
Roman-Manevich, line 39: The previous documentation of `Log2` says "// Calculate the logarithm base 2 of the input. Input must be a power of 2." so you probably shouldn't count on `Log2` to round down.

asllib/libdir/stdlib.asl
```diff
@@ -46,6 +51,30 @@ begin
     return (a MOD 2) == 1;
 end;
 
+func AlignUpP2(x: integer, y: integer) => integer
```
Roman-Manevich, line 54: We probably need a line of comment per function. Ideally with the precondition/assertion on the arguments (e.g., non-negative).

asllib/tests/ASLSemanticsReference.t/SemanticsRule.SReturnOne.asl
```diff
@@ -6,6 +6,9 @@ begin
     assert x == 1; // Only the first loop is executed
```
Roman-Manevich, line 6: `Only the first loop [iteation] is [ever] executed`

asllib/tests/ASLSemanticsReference.t/SemanticsRule.SReturnSome.asl
```diff
@@ -6,6 +6,9 @@ begin
     assert x == 1; // Only the first loop is executed
```
Roman-Manevich, line 6: same comment as above.

asllib/libdir/stdlib.asl
Roman-Manevich: Why do you need the special cases below 2047?

Roman-Manevich: I understand, but why do we need to optimize? If there's a benefit for herd then it's a good reason, but in terms of providing an example implementation, it's not necessary. I guess if we had an optimized version for ASLRef/herd and one for external consumption, we won't have to compromise.

asllib/libdir/stdlib.asl
```diff
@@ -46,6 +46,69 @@ begin
...
 end;
 
+// FloorPow2()
+// ===========
+// For a positive integer X, return the largest power of 2 <= X
+
+func FloorPow2(x : integer) => integer
+begin
+    assert x > 0;
+    var p2 : integer = 2;
```
Roman-Manevich, line 56: I understand the implementation, but I think the following is slighly clearer
```
var p2 : integer = 1;
while x >= p2 * 2 looplimit 2^128 do // i.e. unbounded
    p2 = p2 * 2;
end;
return p2;
```

asllib/libdir/stdlib.asl
```diff
@@ -46,6 +46,69 @@ begin
...
+    while x >= p2 looplimit 2^128 do // i.e. unbounded
+        p2 = p2 * 2;
+    end;
+    return p2 DIV 2;
+end;
+
+// CeilPow2()
+// ==========
+// For a positive integer X, return the smallest power of 2 >= X
```
Roman-Manevich, line 65: `For a positive integer x, returns the smallest power of 2 that is greater or equal to x`

asllib/libdir/stdlib.asl
```diff
@@ -46,6 +46,69 @@ begin
...
+// Return TRUE if integer X is positive and a power of 2. Otherwise,
+// return FALSE.
+
+func IsPow2(x : integer) => boolean
+begin
+    if x <= 0 then return FALSE; end;
+    return FloorPow2(x) == CeilPow2(x);
+end;
+
+// Integer alignment to multiple of size
```
Roman-Manevich, line 84: For a non-negative integer x and positive integer size, returns the greatest multiple of size that is less than or equal to x.

asllib/libdir/stdlib.asl
```diff
@@ -46,6 +46,69 @@ begin
...
+
+func AlignDownSize(x: integer, size: integer) => integer
+begin
+    assert size > 0;
+    return (x DIVRM size) * size;
+end;
+
+func AlignUpSize(x: integer, size: integer) => integer
```
Roman-Manevich, line 92: For a non-negative integer x and positive integer size, returns the smallest multiple of size that is greater than or equal to x.

asllib/libdir/stdlib.asl
```diff
@@ -46,6 +46,69 @@ begin
...
+begin
+    assert size > 0;
+    return AlignDownSize(x + (size - 1), size);
+end;
+
+// Integer alignment to power of two
+
+func AlignDownP2(x: integer, p2: integer) => integer
```
Roman-Manevich, line 100: For non-negative integers x and p2, returns the greatest multiple of 2^p2 that is less than or equal to x.

asllib/libdir/stdlib.asl
```diff
@@ -46,6 +46,69 @@ begin
...
+begin
+    assert p2 >= 0;
+    return AlignDownSize(x, 2^p2);
+end;
+
+func AlignUpP2(x: integer, p2: integer) => integer
```
Roman-Manevich, line 106: For non-negative integers x and p2, returns the smallest multiple of 2^p2 that is greater than or equal to x.

asllib/libdir/stdlib.asl
```diff
@@ -79,29 +142,82 @@ begin
...
+    end;
 
-  // Initial guess
-  let x0 = x;
+    return high;
+end;
 
-  let precision = 1.0 / (2.0 ^ sf);
+// SqrtRounded()
+// Compute square root of VALUE with FRACBITS of precision, rounding inexact values to Odd
```
Roman-Manevich, line 184: What is `Odd` is in this context?

asllib/libdir/stdlib.asl
```diff
@@ -235,6 +351,31 @@ begin
   end;
 end;
 
+// Bit vector alignment to multiple of size
```
Roman-Manevich, line 354: Let's consistently use the term "bitvector" rather than "bit vector".

asllib/libdir/stdlib.asl
```diff
@@ -235,6 +351,31 @@ begin
...
+
+func AlignDownSize{N}(x: bits(N), size: integer {1..2^N}) => bits(N)
```
Roman-Manevich, line 356: Ho about "A variant of AlignDownSize where the bitvector x is viewed as an unsigned integer and the resulting integer is represented by its first N bits."?

asllib/Interpreter.ml
```diff
@@ -544,10 +544,11 @@ module Make (B : Backend.S) (C : Config) = struct
         let*^ m_array, env1 = eval_expr env e_array in
         let*^ m_index, new_env = eval_expr env1 e_index in
         let* v_array = m_array and* v_index = m_index in
-        match B.v_to_int v_index with
+        match B.v_to_z v_index with
         | None -> unsupported_expr e
-        | Some i ->
-            let* v = B.get_index i v_array in
+        | Some z when not (Z.fits_int z) -> unsupported_expr e
```
Roman-Manevich, line 549: Why do we need this case?

asllib/libdir/stdlib.asl
Roman-Manevich: If you want to keep the implementation as is, perhaps rename `p2` to something like `result_times_two` or add a comment? In any case, please decide how you like an resolve this issue.

Roman-Manevich (approved):
Looks good to me.
Please resolve one remaining comment any way you choose to.

PR #1050, opened 2024-11-13
Bpf: Add Load acquire and Store release instructions and multiple litmus tests.

puranjaymohan:
This series adds the support for load-acquire and store release instructions to BPF and adds multiple litmus tests.


herd/BPFArch_herd.ml
```diff
@@ -54,6 +64,8 @@ struct
     | X -> "*"
     | SC -> "SC"
     | N -> ""
+    | A -> "ACQ"
+    | R -> "REL"
```
maranget, line 68: Hi @puranjaymohan. This is not very important, but there is some inconsistencies in names here:  the cat name for the acquire (resp. release) effect set being `AQ` (resp. `RL`), while the annotation is pretty-printed as `ACQ` (resp. `REL`). Is it ok to keep names as they are?

maranget:
Hi @puranjaymohan, I have a small question. Saved from that, the PR is ready for merging.

We favor simplicity in history, which we most often achieve by (1) squashing all commits in a PR into one and (2) rebasing on top of master just before merging. In any case we prefer that tests succeed after all commits. Would you please proceed?

herd/BPFArch_herd.ml
puranjaymohan: Hi @maranget 
I will fix it in the next update.

puranjaymohan:
Hi @maranget 

I have squashed all commits into one and rebased on the master branch, I have also changes `ACQ` to `AQ` and `REL` to `RL` as you suggested.

herd/BPFArch_herd.ml
```diff
@@ -54,6 +64,8 @@ struct
     | X -> "*"
     | SC -> "SC"
     | N -> ""
+    | A -> "AC"
```
maranget, line 67: You may have meant `AQ` here?

puranjaymohan: Yes, sorry for typo. Fixed and pushed again

maranget:
Merged, thanks @puranjaymohan.

PR #1053, opened 2024-11-14
[asl] Explicit parameters

hrutvik:
Implement explicit, positional parameters for ASL1.

In particular:

- All parameters must be explicitly declared at function declaration. They must also be in a specific order: textually left-to-right of the return type and then the argument types.
- Functions must be called with explicit parameter instantiations, except for the following special cases.
  - Standard library functions can omit their input parameter `N` if they take a `bits(N)` argument:
    ```
      let - = UInt('1111');
      let - = ZeroExtend{64}('1111');
    ```
  - Functions immediately on the right-hand sides of assignments can inherit return parameters from explicit type annotations on the left-hand sides:
    ```
      let - : bits(4) = Foo{}(args);    --desugars-->  Foo{4}(args);
      let - : bits(4) = Bar{,M}(args);  --desugars-->  Bar{4,M}(args);
    ```
- There is a shorthand for calling functions which accept parameters but no arguments. Note that this cannot be used with the "right-hand sides" above.
  ```
    let - = Zeros{64};
    let - : bits(64) = Zeros{64};   // OK
    let - : bits(64) = Zeros{}(); // OK
    let - : bits(64) = Zeros{};   // INVALID - parsed as empty record Zeros{}
  ```

## Implementation details

- Checking of ASL0 and ASL1 function declarations and call sites are mostly partitioned off (using AST versioning, https://github.com/herd/herdtools7/pull/1027), so the ASL1 implementations are new.
- But by factoring out ASL1 checking of function declarations, the ASL0 checking is also streamlined and refactored.
- ASL0 function calls also rely on positional parameters too:
  - checking of function declarations infers a list of parameter declarations in the correct order (as if they had been written in ASL1); and
  - checking of call sites inserts inferred parameters into the correct positions.
- The "right-hand side" and "empty argument" shorthands above are desugared during parsing.
- Detecting standard library functions reliably in the presence of overloading requires augmenting the `func` type with a `builtin` flag.

## Backwards compatibility

Some care is taken to preserve backwards compatiblity:
- Various function signatures change in `stdlib.asl` as they are now illegal in ASL1, so an ASL0-compatible `stdlib0.asl` is introduced.
- The `write_memory{,gen}` ASL1 primitive stubs internal to `herd` were no longer valid, so their signatures were changed for equivalent legal ones.
- Some ASL1 patched functions used by `herd` were no longer valid. A couple of these could be made valid, and the others were ported to ASL0 (as with `_PC` in https://github.com/herd/herdtools7/pull/1047).

## Other changes

Along the way, some refactoring could be done:
- ASL0 function declaration checking (see above).
- Using labelled arguments to pass around parameters/arguments in type checking and interpretation. This is because parameters and arguments now have the same type, which could easily be confused/reordered.

## Documentation

This PR also includes synchronisation for https://github.com/herd/herdtools7/pull/1047


asllib/Interpreter.ml
```diff
@@ -485,8 +485,8 @@ module Make (B : Backend.S) (C : Config) = struct
         return_normal (v, new_env) |: SemanticsRule.ESlice
     (* End *)
     (* Begin EvalECall *)
-    | E_Call { name; args; params } ->
-        let**| ms, new_env = eval_call (to_pos e) name env args params in
+    | E_Call { name; params; args } ->
+        let**| ms, new_env = eval_call (to_pos e) name env ~params ~args in
```
jalglave, line 489: the `~` might be worth a comment (throughout)

asllib/Interpreter.ml
```diff
@@ -1240,17 +1240,15 @@ module Make (B : Backend.S) (C : Config) = struct
...
+  and eval_call pos name env ~params ~args =
     let*^ vargs, env1 = eval_expr_list_m env args in
-    let*^ nargs2, env2 = eval_expr_list_m env1 nargs1 in
-    let* vargs = vargs and* nargs2 = nargs2 in
-    let nargs3 = List.combine names nargs2 in
+    let*^ vparams, env2 = eval_expr_list_m env1 params in
+    let* vargs = vargs and* vparams = vparams in
     let genv = IEnv.incr_stack_size name env2.global in
-    let res = eval_subprogram genv name pos vargs nargs3 in
+    let res = eval_subprogram genv name pos ~params:vparams ~args:vargs in
```
jalglave, line 1251: have you used the same names in the transliteration?

asllib/Parser0.mly
```diff
@@ -453,7 +455,16 @@ let getter_decl ==
...
-        D_Func { name; args; return_type; body; parameters; subprogram_type; recurse_limit }
+        D_Func {
+          name;
+          args;
+          return_type;
+          body;
+          parameters;
+          subprogram_type;
+          recurse_limit;
+          builtin = false;
```
jalglave, line 466: is this meant to identify whether something is part of the blessed functions or not?

jalglave (commented):
a very quick first pass - I've not managed to go through it all but thought I would share early

asllib/error.ml
```diff
@@ -317,6 +334,13 @@ module PPrint = struct
           "ASL Typing error:@ explicit@ parameter@ %S@ does@ not@ have@ a@ \
            corresponding@ defining@ argument."
           s
+    | BadParameterDecl (name, expected, actual) ->
+        fprintf f
+          "ASL Typing error: incorrect parameter declaration for %S, expected \
+           %S but %S provided"
+          name
+          ("{" ^ String.concat ", " expected ^ "}")
+          ("{" ^ String.concat ", " actual ^ "}")
```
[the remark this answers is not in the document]
hrutvik: My intention here was to make the "correct" declaration copy-pastable (braces and all) - so I'll incorporate your suggestion without the `pp_quote_string`, as (if I understand correctly) this will introduce double-quotes around each identifier

asllib/desugar.ml
```diff
@@ -53,8 +53,17 @@ let desugar_setter call fields rhs =
...
+        S_Call { name; args = temp_e :: args; params; call_type = ST_Setter }
         |> here
       in
       S_Seq (s_then read modify, write)
+
+let desugar_elided_parameter ldk lhs (call : call annotated) =
+  let bits_e =
+    match lhs with
+    | LDI_Typed (_, { desc = T_Bits (bits_e, []) }) -> bits_e
+    | _ -> Error.fatal_from (to_pos call) CannotParse
```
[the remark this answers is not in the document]
hrutvik: I think this is more of a parse error than obsolete syntax, in that it should be illegal to write e.g. `let x = foo{,M}(args);` because there is no way to desugar this to insert the elided parameter. However the old syntax omitting the parameters entirely is still legal syntactically (but not for type checking). I could add a desugaring error if you'd like?

asllib/Typing.ml
```diff
@@ -2496,191 +2577,192 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
...
-          | Some t1 -> annotate_type ~loc env1 t1
-          (* Type should be valid in the env with no param declared. *)
+          | Some ty ->
+              (* valid in environment which has no parameters declared *)
+              annotate_type ~loc env ty
         in
-        let+ () = check_constrained_integer ~loc env1 t in
-        (add_local x t LDK_Let env1', IMap.add x t acc)
-        |: TypingRule.AnnotateOneParam
+        let+ () = check_constrained_integer ~loc new_env ty in
```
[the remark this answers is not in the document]
hrutvik: Good spot, will fix to `env`. As for where, this was mostly for uniformity of checks and it seemed cleaner than
```
let ty' = annotate_type ... in
let+ () = check ... ty' in
ty'
```
If you feel strongly, I can change it!

asllib/tests/explicit-parameters.t/run.t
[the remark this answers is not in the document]
hrutvik: I will put each non-passing test into a separate file

asllib/Parser0.mly
hrutvik: Yes exactly! The flag is set to `true` for stdlib/primitives in `builder.ml`

hrutvik:
> I fear the deduplication of code in the annotate_func_sig code, do you think it would be possible to refacto that to use some common sub-blocks?

My suggestion is not to - I think of the four major blocks (check parameters, annotate parameters, annotate arguments, annotate return type), only one of them is actually shared (the annotate parameters looks close, but is different due to checking if the parameter is actually declared as an argument in V0).

I could potentially factor out the annotating of loop limit and return type - but I wonder if this brings enough benefits, particularly because the separated version maps more directly to the reference documents?

asllib/doc/AbstractSyntax.tex
```diff
@@ -30,6 +30,7 @@ \chapter{Abstract Syntax\label{chap:AbstractSyntax}}
   \item Expressions (\secref{Expressions})
   \item Patterns (\secref{Patterns})
   \item Slices (\secref{Slices})
+  \item Subprogram calls (\secref{sec:SubprogramCalls})
```
Roman-Manevich, line 33: `\secref{sec:SubprogramCalls}` -> `\secref{SubprogramCalls}`

asllib/ASTUtils.mli
```diff
@@ -336,5 +336,8 @@ val list_fold_left_map :
 (** [fold_left_map] is a combination of [fold_left] and [map] that threads an
     accumulator through calls to [f]. Taken from stdlib 4.11. *)
 
+val list_nub : 'a list -> 'a list
```
Roman-Manevich, line 339: How about `remove_duplicates` or `uniq` (similar to the standard `sort_uniq`)?

asllib/Typing.ml
```diff
@@ -2780,7 +2886,7 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
           | (_, ret_type) :: args -> (ret_type, List.map snd args)
         in
         let _, _, func_sig' =
-          try Fn.subprogram_for_name loc env func_sig.name arg_types
+          try Fn.subprogram_for_name loc env V0 func_sig.name arg_types
```
Roman-Manevich, line 2914: Why specifically the `V0` version? If we pass an annotated `func` we can apply the correct version, no?

asllib/doc/ASLFormal.tex
```diff
@@ -303,6 +303,27 @@ \subsection{Lists}
 \]
 \end{definition}
 
+\hypertarget{def-uniquelist}{}
+\begin{definition}[Finding unique elements of a list]
+The parametric function
+\[
+\uniquelist : \overname{T^*}{l} \rightarrow T^*
+\]
+retains the only the first occurence of each element of the list $l$.
```
Roman-Manevich, line 312: "the only the" -> "only the"

asllib/ASTUtils.mli
hrutvik: I like `uniq`! Will update

asllib/doc/ASLFormal.tex
```diff
@@ -303,6 +303,27 @@ \subsection{Lists}
...
+It relies on the helper function \texttt{unique'}:
```
Roman-Manevich, line 313: please make unique' a macro.

asllib/doc/ASLmacros.tex
```diff
@@ -215,6 +215,7 @@
 \newcommand\cardinality[1]{\hyperlink{def-cardinality}{|}#1\hyperlink{def-cardinality}{|}}
 \newcommand\filterlist[0]{\hyperlink{def-filterlist}{\texttt{filter\_list}}}
 \newcommand\unziplist[0]{\hyperlink{def-unziplist}{\texttt{unzip}}}
+\newcommand\uniquelist[0]{\hyperlink{def-uniquelist}{\texttt{unique}}}
```
Roman-Manevich, line 218: please use \textfunc instead of \texttt. Actually, same for the functions above.

asllib/doc/ErrorCodes.tex
```diff
@@ -69,10 +69,10 @@ \section{Static Error Codes}
...
 See \secref{TypingRule.SubprogramForName} for an example.
 
-\hypertarget{def-parameterwithoutdecl}{}
-\item[$\ParameterWithoutDecl$]
-A subprogram includes a parameter that is not associated with any variable appearing in one of the arguments.
-See \secref{TypingRule.AnnotateParams} for an example.
+\hypertarget{def-badparameterdecl}{}
+\item[$\BadParameterDecl$]
+A subprogram has an invalid declaration of its parameters.
+See \secref{TypingRule.CheckParamDecls} for an example.
```
Roman-Manevich, line 75: I think `\nameref{sec:TypingRule.CheckParamDecls}` is more readable.

asllib/doc/Statements.tex
```diff
@@ -675,6 +696,133 @@ \subsubsection{Formally}
...
+}
+\end{mathpar}
+
+
+% \Nstmt \derivesinline\
+%    & \Nlocaldeclkeyword \parsesep \Ndeclitem \parsesep \Teq \parsesep \Nelidedparamcall \parsesep \Tsemicolon &\\
+% |\ & \Tvar \parsesep \Ndeclitem \parsesep \Teq \parsesep \Nelidedparamcall \parsesep \Tsemicolon &\\
+
+\subsection{Typing and semantics}
+As given by the applying the relevant rules to the desugared AST (see \secref{DeclarationStatements}).
```
Roman-Manevich, line 824: typo: "As given by [the] applying..."

asllib/doc/SubprogramCalls.tex
```diff
@@ -112,150 +91,198 @@ \subsubsection{Formally}
...
-  }\\
-  \checktrans{\callee.\subprogramtype = \calltype}{\MismatchedReturnValue} \checktransarrow \True\OrTypeError\\\\
-  \equallength(\callee.\funcargs, \vargsone) \typearrow \aritymatch\\
-  \checktrans{\aritymatch}{\CallBadArity} \checktransarrow \True \OrTypeError\\\\
+  \unziplist(\texttt{typed\_args}) = (\argtypes, \args)\\
+    \subprogramforname(\tenv, \name, \argtypes) \typearrow
+  (\nameone, \funcsig) \OrTypeError \\
   {
-  \annotateparameterdefining\left(
+  \hyperlink{def-checktrans}{\textfunc{check}}\left(
```
Roman-Manevich, line 146: I'm not sure what's happending here. You probably want to use the `checktrans{}{}` macro, which allows you to include the error code.

asllib/doc/SubprogramCalls.tex
```diff
@@ -112,150 +91,198 @@ \subsubsection{Formally}
...
-        to check that the actual arguments have correct types with respect to $\callerargtypes$
+  \item applying $\unziplist$ to $\texttt{typed\_args}$ yields the corresponding lists of types
+        and expressions $\argtypes$ and $\args$;
+  \item applying $\subprogramforname$ to match $\name$ and $\argtypes$ in $\tenv$
+        yields \\ $(\namep, \funcsig)$\ProseOrTypeError;
+  \item either the $\subprogramtype$ of $\funcsig$ equals $\calltype$,
+        or the $\subprogramtype$ of $\funcsig$ is $\STGetter$ and $\calltype$ is $\STFunction$\ProseOrTypeError;
+  \item applying $\insertstdlibparam$ to $\funcsig$, $\params$, and $\argtypes$ yields new parameters $\paramsone$;
+  \item checking that the lengths of $\funcsig.\funcparameters$ and $\paramsone$ are the same yields $\True$\ProseOrTypeError;
+  \item checking that the lengths of $\funcsig.\funcargs$ and $\args$ are the same yields $\True$\ProseOrTypeError;
```
Roman-Manevich, line 126: You can specify the error code by using the `\ProseTerminateAs{}` macro. Also in the cases above.

asllib/doc/SubprogramCalls.tex
```diff
@@ -112,150 +91,198 @@ \subsubsection{Formally}
...
     \right)
-    \typearrow \eqsthreep \OrTypeError
-  }\\
-  \eqsthree \eqdef \eqsthreep \concat \eqsone\\
-  \checkargstypesat(\tenv, \callee.\funcargs, \callerargtypes, \eqsthree) \typearrow \True \OrTypeError\\\\
-  \checkcalleeparams(\tenv, \callee.\funcparameters, \eqsthree) \typearrow \True \OrTypeError\\\\
-  \annotateretty(\tenv, \calltype, \callee.\funcreturntype) \typearrow \rettyopt \OrTypeError
+  } \checktransarrow \True\OrTypeError\\\\
+  \insertstdlibparam(\funcsig, \params, \argtypes) \typearrow \paramsone\\\\
+  \equallength(\funcsig.\funcparameters, \paramsone) \typearrow \texttt{param\_arity\_match} \\\\
```
Roman-Manevich, line 155: it's better to define a macro for `param_arity_match`. Otherwise, error tend to creep in.

asllib/doc/SubprogramCalls.tex
```diff
@@ -112,150 +91,198 @@ \subsubsection{Formally}
...
-  \item define $\eqsthree$ is the concatenation of $\eqsthreep$ and $\eqsone$;
-  \item applying $\checkargstypesat$ to $\calleeargtypes$
-        to check that the actual arguments have correct types with respect to $\callerargtypes$
+  \item applying $\unziplist$ to $\texttt{typed\_args}$ yields the corresponding lists of types
+        and expressions $\argtypes$ and $\args$;
+  \item applying $\subprogramforname$ to match $\name$ and $\argtypes$ in $\tenv$
+        yields \\ $(\namep, \funcsig)$\ProseOrTypeError;
+  \item either the $\subprogramtype$ of $\funcsig$ equals $\calltype$,
+        or the $\subprogramtype$ of $\funcsig$ is $\STGetter$ and $\calltype$ is $\STFunction$\ProseOrTypeError;
+  \item applying $\insertstdlibparam$ to $\funcsig$, $\params$, and $\argtypes$ yields new parameters $\paramsone$;
```
Roman-Manevich, line 124: But how do you know whether this is a builtin function?
This is a new concept and it's not defined in the reference and it's not part of the abstract syntax. Perhaps there should be an extra field in the abstract syntax and we can add it at the top level to the standard library?

asllib/doc/SubprogramCalls.tex
```diff
@@ -112,150 +91,198 @@ \subsubsection{Formally}
...
-  {
-    \begin{array}{r}
-    \subprogramforname(\tenv, \name, \callerargtypes) \typearrow \\
-  (\eqsone, \nameone, \callee) \OrTypeError
-    \end{array}
-  }\\
-  \checktrans{\callee.\subprogramtype = \calltype}{\MismatchedReturnValue} \checktransarrow \True\OrTypeError\\\\
-  \equallength(\callee.\funcargs, \vargsone) \typearrow \aritymatch\\
-  \checktrans{\aritymatch}{\CallBadArity} \checktransarrow \True \OrTypeError\\\\
+  \unziplist(\texttt{typed\_args}) = (\argtypes, \args)\\
```
Roman-Manevich, line 142: please define a macro for \texttt{typed\_args}.

asllib/doc/SubprogramCalls.tex
```diff
@@ -1481,16 +1266,16 @@ \section{Semantics}
 \[
   \begin{array}{c}
     \evalcall{\overname{\envs}{\env} \aslsep
-    \overbracket{\Identifiers}^{\texttt{name}} \aslsep
-    \overbracket{\expr^*}^{\texttt{args}} \aslsep
-    \overbracket{(\overname{\Identifiers}{\id_i}\times\overname{\expr}{\ve_i})^*}^{\texttt{named\_args}}} \;\aslrel\; \\
+    \overbracket{\Identifiers}^{\name} \aslsep
```
Roman-Manevich, line 1269: \overbracket -> \overname. here and in the next two cases.

asllib/doc/SubprogramCalls.tex
hrutvik: In the "Formally" subsubsection for this `\insertstdlibparam`, I attempted to replicate the way evaluation of primitives are handled - in other words, by stating that this transition parametrises the typing/semantics of ASL. Does this seem sufficient, or do you think it needs reworking?

Roman-Manevich: I guess it's okay for now.

Roman-Manevich (approved):
Looks good overall.

PR #1065, opened 2024-11-19
Get rid of asserts in AArch64 instruction parser

murzinv:
Currently sanity checks for vector instructions done during parsing and asserts if instruction is malformed - that definitely is not user-friendly and we could do better. Let's convert all asserts for vector instructions to checks in dedicated validation (`is_valid`) pass.


maranget:
This PR is ready, I'll merge it once rebased. I have a remark:  is it correct that  EOR apparently is our only OP3 operation in Neon and SVE?

maranget:
Waiting for tests to complete...

maranget:
Merged, thanks @murzinv.

PR #1068, opened 2024-11-19
[asl][reference] Update progress appendices

Roman-Manevich:


jalglave (commented):
thanks for updating this @Roman-Manevich

PR #1071, opened 2024-11-21
[doc] Documentation on how to introduce new languages

diaolo01:


jalglave:
hi @maranget @relokin @artkhyzha @yuxiliu-arm please could you take a look at Diana's documentation?
Thanks in advance,
Jade

yuxiliu-arm (commented):
Looks great. Just a few comments. Regarding the monad section, I wonder if it's worth to touch a bit about `event_structure`, which would provide some motivation for using a monad in the first place.

Another subtle point is that because `event_structure` distinguishes input and output, and they might be combined differently, so there are some variants. Furthermore, not all operators satisfies monad laws like `(f <*> g) <*> h = f <*> (g <*> h)`. But these might be not as useful to add here. I'm happy either way.

One comment on the format, shall the document be written in [GitHub flavored markdown](https://github.github.com/gfm/)? Basically adding back-ticks for inline code, using GFM's table syntax, and linking to folders with path.

diaolo01:
@relokin, Jade and I have been working on documenting how to insert new languages in herd. She suggested to further add a couple of paragraphs about litmus7 and kvm-unit-tests, and that you could contribute. Could you please add the write-up for this?

diaolo01:
@maranget, Jade suggested you might be able to integrate this content into a new chapter of the http://diy.inria.fr/ documentation. Could you please help with the integration?

maranget:
Hi @diaolo01, I have integrated your work as a new part in the documentation.  The html and PDF version of this enhanced documentation are available [here](https://diy.inria.fr/tst7/doc/developer.html) .

doc/languages.tex
```diff
@@ -0,0 +1,248 @@
+\cutname{developer.html}
+\section{Introducing new languages in \texttt{herdtools7}}.
```
diaolo01, line 2: Is the full stop at the end required? It makes the paragraph begin with ".".

diaolo01 (commented):
Thanks @maranget! One minor comment, otherwise rendering looks good to me!

maranget:
Thanks @diaolo01. Following the usual procedure I have rebased this PR on top of master (I have also squashed my contributions together). Merge is on its way. Thanks again.

PR #1072, opened 2024-11-21
[asl][reference] Add pragma AST, typing and semantics

IGGeorgiev:
This PR adds:
 * The `S_Pragma` and `D_Pragma` AST nodes
 * Type checking of pragma arguments
 * Interpretation semantics of pragmas (semantically a `pass;`)

### DIFF
--- a/asllib/AST.mli
+++ b/asllib/AST.mli
@@ -347,6 +347,9 @@ type stmt_desc =
   | S_Unreachable
       (** The unreachable statement, as an explicit node as it has a specific
           control-flow behaviour. *)
+  | S_Pragma of identifier * expr list
+      (** A pragma statement, as an explicit node to be used by tools which need
+          AST level hints. *)
 
 and stmt = stmt_desc annotated
 and case_alt_desc = { pattern : pattern; where : expr option; stmt : stmt }
@@ -397,6 +400,9 @@ type decl_desc =
   | D_Func of func
   | D_GlobalStorage of global_decl
   | D_TypeDecl of identifier * ty * (identifier * field list) option
+  | D_Pragma of identifier * expr list
+      (** A global pragma, as an explicit node to be used by tools which need
+          AST level hints. *)
 
 type decl = decl_desc annotated
 
--- a/asllib/ASTUtils.ml
+++ b/asllib/ASTUtils.ml
@@ -278,6 +278,7 @@ let rec use_s s =
   | S_Try (s, catchers, s') ->
       use_s s $ use_option use_s s' $ use_catchers catchers
   | S_Print { args; debug = _ } -> use_es args
+  | S_Pragma (name, args) -> ISet.add name $ use_es args
   | S_Unreachable -> Fun.id
 
 and use_ldi = function
@@ -314,6 +315,7 @@ and use_decl d =
       $ use_option use_ty return_type
       $ use_named_list (use_option use_ty) parameters
       $ match body with SB_ASL s -> use_s s | SB_Primitive -> Fun.id)
+  | D_Pragma (name, args) -> ISet.add name $ use_es args
 
 and use_subtypes (x, subfields) = ISet.add x $ use_named_list use_ty subfields
 
@@ -607,8 +609,10 @@ let default_array_ty =
 
 let identifier_of_decl d =
   match d.desc with
-  | D_Func { name; _ } | D_GlobalStorage { name; _ } | D_TypeDecl (name, _, _)
-    ->
+  | D_Func { name; _ }
+  | D_GlobalStorage { name; _ }
+  | D_TypeDecl (name, _, _)
+  | D_Pragma (name, _) ->
       name
 
 let patch ~src ~patches =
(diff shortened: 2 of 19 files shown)


asllib/TopoSort.ml
```diff
@@ -186,13 +186,7 @@ module ASTFold = struct
...
     | Some prev -> Tbl.replace tbl key (List.rev_append values prev)
 
-  let def d =
-    let open AST in
-    match d.desc with
-    | D_Func { name; _ } | D_GlobalStorage { name; _ } | D_TypeDecl (name, _, _)
-      ->
-        name
-
+  let def d = ASTUtils.identifier_of_decl d
```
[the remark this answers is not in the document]
IGGeorgiev: Does this want to hold true more generally for `ASTUtils.identifier_of_decl`?

I think it makes sense that the pragma name is effectively non-consequential and can be ignored by all aslref processing.

asllib/ASTUtils.ml
```diff
@@ -278,6 +278,7 @@ let rec use_s s =
   | S_Try (s, catchers, s') ->
       use_s s $ use_option use_s s' $ use_catchers catchers
   | S_Print { args; debug = _ } -> use_es args
+  | S_Pragma (_, args) -> use_es args
```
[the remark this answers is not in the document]
IGGeorgiev: I was a bit worried about unintended side effects here. Could a pragma's name clash with the name of a local/global variable? My naive assumption would be "yes", which makes me think that a pragma's name should really not have any impact on any reference counting (and as you mentioned earlier renaming)

asllib/doc/Statements.tex
```diff
@@ -3098,5 +3098,61 @@ \subsection{Syntax}
...
+\begin{mathpar}
+\inferrule{
+  \annotateexprlist(\tenv, \vargs) \typearrow \vargsp \OrTypeError
+}{
+  \annotatestmt(\tenv, \overname{\SPragma(\id, \vargs)}{\vs}) \typearrow (\overname{\SPragma(\id, \vargsp)}{\news}, \tenv)
+}
+\end{mathpar}
+\CodeSubsection{\SPragmaBegin}{\SPragmaEnd}{../Typing.ml}
+
 \subsection{Semantics}
```
Roman-Manevich, line 3137: I think you can remove `SemanticsRUle.SPramga` and write `Pragmas are not associated with a semantics.`

asllib/TopoSort.ml
```diff
@@ -192,6 +192,7 @@ module ASTFold = struct
     | D_Func { name; _ } | D_GlobalStorage { name; _ } | D_TypeDecl (name, _, _)
       ->
         name
+    | D_Pragma _ -> "__pragmas"
```
Roman-Manevich, line 195: Since you're separating global pragmas from the rest of the global declarations, they shouldn't appear here, so perhaps `assert false`?

asllib/Typing.ml
```diff
@@ -3201,7 +3220,12 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
     in
     let fold_topo ast acc = TopoSort.ASTFold.fold fold ast acc in
     fun env ast ->
-      let ast_rev, env = fold_topo ast ([], env) in
+      let is_pragma d = match d.desc with D_Pragma _ -> true | _ -> false in
+      (* Type check D_Pragma declarations separately from the main AST
```
Roman-Manevich, line 3224: This block is not documented.

Roman-Manevich (changes requested):

asllib/AST.mli
```diff
@@ -397,6 +400,9 @@ type decl_desc =
   | D_Func of func
   | D_GlobalStorage of global_decl
   | D_TypeDecl of identifier * ty * (identifier * field list) option
+  | D_Pragma of identifier * expr list
+      (** A global pragma, as an explicit node to be used by tools which need
+          AST level hints about adjacent code. *)
```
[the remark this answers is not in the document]
IGGeorgiev: We do have line number annotation, so we do retain some idea of proximity.

Pragmas' only function is purely to be a code hint as far as I understand them. Them being reordered for type checking in aslref should be okay so long as a tool can parse them and perform some action using them in the untyped ast context (directly out of a parse). Once we get to typing and evaluation - they lose their relevance. Type checking their expressions is just a sensible thing to do for a sanity check, I think.

If you want I can remove the `about adjacent code` part of the comment to avoid confusion?

asllib/doc/Specifications.tex
```diff
@@ -275,6 +290,58 @@ \subsubsection{Formally}
 \end{mathpar}
 \CodeSubsection{\TypeCheckMutuallyRecBegin}{\TypeCheckMutuallyRecEnd}{../Typing.ml}
 
+\subsubsection{TypingRule.TypeCheckGlobalPragmas\label{sec:TypingRule.TypeCheckGlobalPragmas}}
```
Roman-Manevich, line 293: This is correct, but I think it would be simpler to have a function to check a single global pragma, e.g., `check_global_pragma` (TypingRUle.CheckGlobalPragma) and then apply that to the entire list as a premise in checking a specifications, as opposed to explicitly processing the list here. Also, the term `annotate` should be `checked`, as annotate is used for "returns a typed AST node".

Roman-Manevich (approved):

Roman-Manevich (approved):
Looks good to me!

PR #1074, opened 2024-11-22
[herd,aarch64,asl] Rewrite physical memory access  functions in V1

maranget:
This PR moves the `PhysMemRead` and `PhysMemWrite` functions from the `implementation0.asl` file to the `implementation.asl` file. Of course, the functions are now in V1 style. Additionnaly, the seemingly now useless function `AltDecodeBitMasks` is deleted from the `implementation.asl` file.

I have two reasons for this moves:
  + The move facilitatees the periodic rebasing of the experimental `asl-vmsa` branch that maintains two versions of the physical memory access functions.
  + If feasible,  translating functions from V1 to V0 does, as in PR #1053 should probably be avoided as one can anticipate the reverse translation in the future.

Would you please @HadrienRenaud and @hrutvik check and comment this PR?


hrutvik:
I originally moved these functions to V0 in https://github.com/herd/herdtools7/pull/1053 because they became illegal in V1 - both `PhysMem{Read,Write}` do not explicitly declare their `size` parameters, and do not call `{read,write}_memory_gen` with an explicit parameter. I couldn't see a way of making these signatures compatible with both usage in V0 XML (using the parameter-defining argument for `size`) and V1 (using an explicit `size` parameter which could not be duplicated as an argument) - in particular, the trick I used for `write_memory_gen` in https://github.com/herd/herdtools7/pull/1053 didn't seem applicable (converting an integer parameter-defining argument to an exact constrained integer, relying on a second bit vector parameter-defining argument).

However, it seems that the "best effort" type checking/interpretation can cope with these issues enough to pass all the tests. I'm probably not best placed to say whether the reliance on this "best effort" approach is OK in this instance, but I thought I would provide my reasoning for the move to V0 so that you have all the information.

For reference, below is a paste of some of the errors caused by these functions in V1 (obtained by setting `TypeCheck.check = Warn` in `ASLSem.ml`).
```
File ./herd/libdir/asl-pseudocode/implementations.asl, line 449, character 0
  to line 463, character 4:
ASL Typing error: incorrect parameter declaration for "PhysMemWrite",
  expected {size} but {} provided
File ./herd/libdir/asl-pseudocode/implementations.asl, line 453,
  characters 16 to 20:
ASL Error: Undefined identifier: 'size'
File ./herd/libdir/asl-pseudocode/implementations.asl, line 456,
  characters 2 to 58:
ASL Static Error: Arity error while calling 'write_memory_gen':
  1 parameters expected and 0 provided
File ./herd/libdir/asl-pseudocode/implementations.asl, line 467, character 0
  to line 482, character 4:
ASL Typing error: incorrect parameter declaration for "PhysMemRead", expected
  {size} but {} provided
File ./herd/libdir/asl-pseudocode/implementations.asl, line 471,
  characters 31 to 35:
ASL Error: Undefined identifier: 'size'
ASL Typing error: constrained integer expected, provided integer.
```

maranget:
Thanks @hrutvik for your answer. I understand that it is not possible at the moment to both declare the the function PhysMemWrite in V1 style and to call it from V0 code. I am wondering how you would write this function to be type correct in a completely V1 environment?

hrutvik:
I believe they would be declared like this:
```
func PhysMemWrite{size}(desc: AddressDescriptor, accdesc: AccessDescriptor, value: bits(8*size)) => PhysMemRetStatus
begin
  write_memory_gen{size*8}(desc.vaddress, size*8, value, accdesc);
  ...
end;

func PhysMemRead{size}(desc: AddressDescriptor, accdesc: AccessDescriptor) => (PhysMemRetStatus, bits(8*size))
begin
  let value = read_memory_gen(desc.vaddress,size*8,accdesc)[8*size-1:0];
  ...
end;
```
And then called as:
```
PhysMemWrite{sz}(d, ad, v);
PhysMemRead{sz}(d, ad);
```
The duplication of `size*8` in the call to `write_memory_gen` is the "trick" I mention above - perhaps in a fully V1 setting, the definitions of `{read,write}_memory_gen` could be updated to better suit the use of explicit parameters.

maranget:
Hi @hrutvik.  I have just written `PhysMemWrite` and `PhysMemRead` as you suggested. Additionally I have slightly enhanced V0 syntax, so as to allow wrappers from V0 to V1. That way, one can continue to develop subprograms in V1 style. Later, when V1 will be everywhere, getting rid of the wrappers and a few renamings will be the only necessary operations. Would you agree with this solution? Also poke @HadrienRenaud.

As a side note, if, as I am suspecting now, the primitive `write_memory_gen` is called by `PhysMemWrite` only, may I suggest simplifying it by removing the size argument, keeping the size parameter.

hrutvik (approved):
This sounds good to me!

In terms of simplifying the primitive `write_memory_gen` - I would be in favour of this. Off the top of my head I can think of two things worth noting:
- We must also update the "pseudo-arch" tests, as I believe these invoke `write_memory_gen` directly.
- I think the handling of parameters for primitives in `ASLSem.ml` needs to change slightly - I believe parameters don't currently get passed to the underlying OCaml function (see e.g. the builders `p1`, `p2`, ...), but we would need access to the `size` parameter to produce the correct effects.

maranget:
Thanks @hrutvik, we'll see for `write_memory_gen` in a forthcoming PR. I'll merge this one soon.

PR #1080, opened 2024-11-28
Small fixes for ASL Reference

yuxiliu-arm:
These are flagged by `pandoc` and may also make sense for the upstream LaTeX tool.


Roman-Manevich (approved):
Looks good to me.

asllib/doc/LexicalStructure.tex
yuxiliu-arm, line 1: Actually, this seems like a MathJax error, it reports `\textbackslash` is only available in math mode, while it should be available in text mode. Will revert this change.

Roman-Manevich (approved):
Looks good to me!

PR #1084, opened 2024-12-02
[asl] Simplify V0 standard library

hrutvik:
Using V0 grammar changes in https://github.com/herd/herdtools7/pull/1074, the V0 standard library functions used for backwards compatibility can simply call their V1 versions with appropriate parameter instantiations.

### DIFF
--- a/asllib/libdir/stdlib0.asl
+++ b/asllib/libdir/stdlib0.asl
@@ -2,30 +2,22 @@
 // Backwards compatibility for ASLv0
 
 bits(N) ReplicateBit(boolean isZero, integer N)
-  return if isZero then Zeros(N) else Ones(N);
+  return ReplicateBit{N}(isZero);
 
 bits(M*N) Replicate(bits(M) x, integer N)
-  if M == 1 then
-    return ReplicateBit(IsZero(x), M*N);
-  else
-    r = Zeros(M*N);
-    for i=0 to N-1
-        r<M*(i+1) - 1 : M*i> = x;
-    return r;
+  return Replicate{M*N,M}(x);
 
 bits(N) Zeros(integer N)
-  return 0<N-1:0>;
+  return Zeros{N}();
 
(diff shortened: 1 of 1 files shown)


maranget:
Hi @hrutvik. I did not notice that V0 and V1 functions can have the same name, as they have different signatures. Nice trick.
LGTM

asllib/libdir/stdlib0.asl
```diff
@@ -2,30 +2,22 @@
...
 
 bits(M*N) Replicate(bits(M) x, integer N)
-  if M == 1 then
-    return ReplicateBit(IsZero(x), M*N);
-  else
-    r = Zeros(M*N);
-    for i=0 to N-1
-        r<M*(i+1) - 1 : M*i> = x;
-    return r;
+  return Replicate{M*N}(x);
```
maranget, line 8: Hi @hrutvik I do not understand here, wouldn't the signature de V1 Replicate be `func Replicate{N,M}(x: bits(M)) => bits(N)`...

hrutvik: The signature of V1 `Replicate` has changed as you describe:
   ```
   bits(M*N) Replicate(bits(M) x, integer N)    // V0
   func Replicate{N,M}(x: bits(M)) => bits(N)   // V1
   ```
And then I've elided the "input" parameter of the standard library function. Does this seem right to you, or have I missed something somewhere?

maranget: Well,  my question was not very clear.  Although I have not a very strong opinion on this, I think that giving all parameters in the call would be preferable: `return Replicate{M*N,M}(x);`

maranget:
@hrutvik LGTM.

asllib/libdir/stdlib0.asl
hrutvik: That makes sense - I've updated to the fully-explicit form as you suggest

PR #1086, opened 2024-12-02
[asl] change behaviour of print + println

HadrienRenaud:
This PR introduces a sibling of `print`: `println`, that adds a new line at the end of the printed section. Now, `print` does not do that anymore.

It also remove the separators from between the different arguments to `print`/`println`.

For some examples, see `asllib/tests/print.t`.


maranget:
Hi @HadrienRenaud. Test failures probably are easy to fix. Why not add one word or two to motivate the PR?

maranget:
Also there are two `asl-630-behaviour-of-print` branches, this one et the other on origin. This is a bit confusing.

asllib/tests/print.t
```diff
@@ -27,3 +25,36 @@
...
+  1234
+  0x03
+  TRUE
+  1/2
+  0
+  20
+
+  $ cat >print4.asl <<EOF
+  > func main () => integer begin
+  >   println ((1, 2));
```
Roman-Manevich, line 55: Attempting to print tuples should result in a type error.

asllib/tests/print.t
```diff
@@ -8,9 +8,7 @@
   > EOF
 
   $ aslref printer1.asl
-  Wow 2 3.14 some other string
-  no type-checking
-  32
+  Wow 2 157/50 some other stringno type-checking32
```
Roman-Manevich, line 11: There should be no spaces between the individual sub-strings.

asllib/Interpreter.ml
```diff
@@ -1036,9 +1037,10 @@ module Make (B : Backend.S) (C : Config) = struct
               vs
           else (
             List.map B.debug_value vs |> String.concat " " |> print_string;
```
Roman-Manevich, line 1039: Remove the `" "` from `String.concat` - there should be no spaces between the individual strings.

asllib/error.ml
```diff
@@ -168,6 +169,7 @@ let error_label = function
   | RecursionLimitReached -> "RecursionLimitReached"
   | EmptyConstraints -> "EmptyConstraints"
   | UnexpectedPendingConstrained -> "UnexpectedPendingConstrained"
+  | BadPrintType _ -> "BadPrintType"
```
Roman-Manevich, line 172: Consider adding a type for a clearer type message. If there is a list of expressions, it's going to be hard to tell which one caused the problem.

asllib/tests/print.t
Roman-Manevich: Consider printing `0 as real`, `-1 as real`, `-1.0 as real`.

Roman-Manevich (approved):

maranget:
Never mind. Having two branches with the same name lead to minor trouble with the `git switch` command. Nothing serious.

asllib/bitvector.mli
```diff
@@ -81,6 +81,10 @@ val to_string : t -> string
 (** Returns a string representing the bitvector, indexed from the right and
     delimited by apostrophes. *)
 
+val to_string_hexa : t -> string
```
hrutvik, line 84: I feel like some information is lost here regarding the bit vector length. In particular:
- if the length is not a multiple of 4 you can't tell
- in the tests below it feels like there is an extra hex character, e.g. `'0111'` should be `0x7` but is `0x07`

I wonder if a solution is to print the new pseudo-syntax for bit vector literals of specific lengths, e.g. `0x07[:4]` or similar?

hrutvik (commented):
Just a minor suggestion regarding printing of bit vectors, as I think it shouldn't lose length information

hrutvik (approved):

PR #1087, opened 2024-12-03
[litmus,test] Fix a few glitches

maranget:
Restore `make litmus-aarch64-test` functionality.

### DIFF
--- a/Makefile.aarch64
+++ b/Makefile.aarch64
@@ -56,10 +56,10 @@ KUT_CONFIG_PARAMS=--arch=arm64 --page-size=4k --cross-prefix=$(KUT_AARCH64_PREFF
 KUT_DIR_AARCH64:=$(shell mktemp -d)
 
 litmus-aarch64-dep:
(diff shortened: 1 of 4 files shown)


maranget:
Hi @relokin, @murzinv. We should probably install some action to run `make litmus-x64_64-test` and `make litmus-aarch64-test` periodically, as we do for building herd-www.

maranget:
We could at least compile the tests....

maranget:
... and we thus need either to install the right cross-compiler and tools, whatever the platform, or be able to select the appropriate platform to use the local compiler.

maranget:
I had a try for a periodic action that tests litmus, x86_64 only.

maranget:
Hi @relokin, @murzinv, would your agree that I merge this PR? It is quite simple.

maranget:
I attempted a less demanding requirement: armv8.1 with FEAT_LRCPC. Apparently it works.

PR #1092, opened 2024-12-04
[asl] Update stdlib definition of `Replicate`

hrutvik:


asllib/libdir/stdlib.asl
```diff
@@ -119,15 +119,15 @@ end;
...
-  assert (N MOD M == 0);
-  if N == 1 then
+  if M == 1 then
     return ReplicateBit{N}(IsZero(x));
   else
-    var r: bits(N) = Zeros{}();
-    for i=0 to (N DIV M)-1 do
-      r[i*:M] = x;
+    let items = N DIV M; // must be exact
+    var result : bits(N) = Zeros{}();
```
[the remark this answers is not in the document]
hrutvik: @nigelito do you have a preference?

asllib/libdir/stdlib.asl
```diff
@@ -119,15 +119,15 @@ end;
 
 func Replicate{N,M}(x: bits(M)) => bits(N)
 begin
-  assert (N MOD M == 0);
-  if N == 1 then
+  if M == 1 then
```
Roman-Manevich, line 122: What is the purpose of this? If the purpose of the code is to serve as an example, there's no need for optimizations.

maranget: This code is called in mode `-variant asl` and the optimisation had some impact. I do not know whether this is still the  case or not.

asllib/libdir/stdlib.asl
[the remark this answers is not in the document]
nigelito: The proposed change looks goof to me.

asllib/libdir/stdlib.asl
nigelito: My preference would be not to include this optimisation in example/reference code, but I can live with it.

Roman-Manevich: We can opt to include this optimization in an internally-used version of the standard library.

maranget:
FYI, I have just checked that restoring the optimisation (it was in fact disabled) yields a speedup of about four on test `herd/tests/instructions/AArch64.ASL/SXTW01.litmus`.

PR #1070, opened 2024-11-19
[litmus] Fix glitch

maranget:
Previous printing of faults as _e.g._ `fault(P0:0:L0,...)` was not accepted by compare tools. Fix this.

PR #1094, opened 2024-12-05
[tools] More flexible log lexer

maranget:
Some logs exist that have been produced before PR #1070 is merged. This more flexible lexer accepts them.


relokin:
Thanks for this Luc!

Is the bug before PR #1070 affecting only my logs? If that's the case I am happy to patch them by hand and avoid complicating the log lexer. Otherwise, this PR looks good to me!

maranget:
Hi @relokin. As far as I remember, this bug affected your logs only, and only in a few places. Hence, this PR is marginally useful.

maranget:
From discussion, this PR is not useful. Closing.

PR #1098, opened 2024-12-09
[herd, aarch64]: fix nextSet for MRS instruction

murzinv:
We want to update environment with pair `destination register=value` rather than `system register=value`


maranget:
Hi @murzinv. This  PR is a welcome bugfix. Would you please rebase on master?

maranget:
Merged thanks @murzinv.

PR #1100, opened 2024-12-10
[asl] Add warning for no recursion annotation

HadrienRenaud:
This PR implements a static check for recursive functions to have a recursion limit annotation.

Todo:
- [x] **Implementation** Statically warn when a recursive function does not have a recursion limit
- [x] **Reference** No reference work needed as warnings are not yet formalised.


hrutvik (approved):

Roman-Manevich (approved):
Looks good to me!

PR #1106, opened 2024-12-14
[asl] Fix dynamic domain after enum labels

HadrienRenaud:
Dynamic domain was still using the old definition of enumeration labels: enumeration labels were compiled to integers. After PR #1091, enumeration labels are first class citizens, and so we can update the definition of dynamic domains to use those. This results in a simpler and more straight forward definition of dynamic domains.

### Before

- Prose:
Screenshot 2024-12-14 at 13 31 06

- Formally:
Screenshot 2024-12-14 at 13 31 39

### After

- Prose:
Screenshot 2024-12-14 at 13 29 05

- Formally:
Screenshot 2024-12-14 at 13 28 41


Roman-Manevich (approved):
Looks good to me!

PR #1107, opened 2024-12-16
[ci] Bump to checkout ocaml v3

HadrienRenaud:
This PR:
- [x] Fixes the CI build by bumping to checkout-ocaml@v3
- [x] Make CI faster (still by using checkout-ocaml@v3)
- [x] Move check-format into make test workflow, reducing number of workflow run (more efficient)
- [x] Bump OCaml compiler tested high version to OCaml 5.2

All in all, runtime of CI went from 11min before (see [here](https://github.com/herd/herdtools7/actions/runs/12317610957)) to 7min after (see [here](https://github.com/herd/herdtools7/actions/runs/12352369396/job/34469261343)).


maranget:
Hi @HadrienRenaud. Thanks a lot for this. As this PR fixes CI I suggest merging it as soon as possible. Would you please proceed after squashing and rebasing as usual?

PR #916, opened 2024-07-29
[herd,asl] Implement basic ASL+VMSA

maranget:
Basic ASL+VMSA, tested on simple loads and stores. The current  implementation handle successful accesses and three errors: invalid page table entries, writing on a non-writable page and access flag.

PR #1116, opened 2024-12-26
[asl,herd] Improvements to the ASL engine in herd

HadrienRenaud:
This PR is partly an extraction from @maranget #916 that I think would benefit the non-VMSA ASL executions.

This PR:

- [x] Introduce a way of profiling a cat file with command line argument `-debug profile_cat`
- [x] Introduce a way of profiling ASL executions in AArch64 with command line argument `-debug profile_asl`
- [x] Introduce CutOff events for pruned loops executions in the ASL interpreter (for modes where loops can be pruned)
- [x] Optimise ASL to AArch64 event translation
- [x] Transmit exact proc id from AArch64 to ASL
- [x] Add an optimisation in `asllib/Interpreter.ml` in the case where expressions are very simple.
- [x] Add an optimisation in `herd/ASLSem.ml` for boolean operations when one of them is a scalar boolean. This and the previous point allow an early catch of some rejected executions, and thus reduce the number of non-deterministic choices.
- [x] Remove some unused constructs in `AArch64ASLSem`

This gets `make test-aarch64-asl` from 20s to 13s and gets herd7 from 3.76s to 2.46s on a successful CAS (on my macbook).

### DIFF
--- a/herd/AArch64ASLSem.ml
+++ b/herd/AArch64ASLSem.ml
@@ -26,6 +26,13 @@ let return_0 =
   let open Asllib.ASTUtils in
   add_dummy_annotation (S_Return (Some (expr_of_int 0)))
 
+let end_profile t0 msg : unit =
+  let t1 = Sys.time () in
+  if t1 -. t0 > 1. (* We log only executions that took more than 1 second *)
+  then
+    Printf.eprintf "AArch64+ASL sem took %fs to evaluate %s.\n%!" (t1 -. t0)
+      msg
+
 module Make (TopConf : AArch64Sig.Config) (V : Value.AArch64ASL) :
   AArch64Sig.Semantics with module A.V = V = struct
   module AArch64S = AArch64Sem.Make (TopConf) (V)
@@ -35,6 +42,18 @@ module Make (TopConf : AArch64Sig.Config) (V : Value.AArch64ASL) :
   let ( let* ) = M.( >>= )
   let return = M.unitT
   let _dbg = TopConf.C.debug.Debug_herd.monad
+  let _profile = TopConf.C.debug.Debug_herd.profile_asl
+
+  let start_profile = if _profile then Sys.time else Fun.const 0.
+  let end_profile = if _profile then end_profile else fun _ _ -> ()
+
+  let profile msg f =
+    if _profile then
+      let t0 = start_profile () in
+      let res = f () in
+      let () = end_profile t0 msg in
+      res
+    else f ()
 
   module Mixed (SZ : ByteSize.S) : sig
     val build_semantics : test -> A.inst_instance_id -> (proc * branch) M.t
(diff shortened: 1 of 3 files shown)


herd/tests/instructions/ASL-pseudo-arch/non-deterministic-optimised.litmus
```diff
@@ -0,0 +1,26 @@
+ASL non-deterministic
+
+{
+  x = 0;
+  0: X1= x;
+}
+
+// Non deterministically returns 1 or 2
```
maranget, line 8: As far as I understand function `g()` always returns `2`. Am I missing something?

asllib/Interpreter.ml
```diff
@@ -442,6 +447,11 @@ module Make (B : Backend.S) (C : Config) = struct
         | NotFound -> fatal_from e @@ Error.UndefinedIdentifier x)
         |: SemanticsRule.EVar
     (* End *)
+    | E_Binop (((BAND | BOR | IMPL) as op), e1, e2)
+      when is_simple_expr e1 && is_simple_expr e2 ->
+        let* v1 = eval_expr_sef env e1 and* v2 = eval_expr_sef env e2 in
+        let* v = B.binop op v1 v2 in
+        return_normal (v, env)
```
maranget, line 454: I am not so sure about this "optimisation". Namely in some situation it may lead to the additional evaluation of the second expression. For instance in `e1 && e2`, when both expressions are simple and that `e1` reduces to `FALSE`. Then, without the above code, `e2` would not be evaluated at all...

maranget:
Hi @HadrienRenaud. Nice work. Before merging, please consider the two questions above.

herd/tests/instructions/ASL-pseudo-arch/non-deterministic-optimised.litmus
```diff
@@ -0,0 +1,28 @@
+ASL non-deterministic
+
+{
+  x = 0;
+  0: X1= x;
+}
+
+// Always return 2, but this could create 2 executions that will turn out to
+// follow the same branch. To know if this creates two different executions
+// with the same end state, run herd with -debug solver
```
maranget, line 10: Hi @HadrienRenaud. In case two executions are created, output would be different (with `Positive: 2`). Or do I miss something...

maranget: OK I have checked. The difference show up with `herd7 -v` . With the optimisation we have a message on stderr: `Warning: 1 abstract event structures`, while without it, we have: `Warning: 2 abstract event structures`.

maranget: As a test it looks easier to interpret than using `-debug solver`.

maranget:
Hi @HadrienRenaud. Up to the simpler test technique above, LGTM. Please merge.

PR #1119, opened 2025-01-03
[doc] Capturing convention about 'fault' predicate

diaolo01:
When introducing the SVC instruction, as part of the following commit: https://github.com/herd/herdtools7/commit/62b1036808b0e0ce4d6ad306ba55b1b635d1fa6c, it became apparent that there is a significant difference between checking the address or label of the instruction that caused the fault and the return address of the fault. This distinction emerged since SVC handler returns to the instruction after the SVC, unlike most instructions which return to the instruction that caused the fault.

### DIFF
--- a/catalogue/aarch64-faults/README.md
+++ b/catalogue/aarch64-faults/README.md
@@ -31,3 +31,31 @@ in the folder `${KUT_DIR}`:
     % sh ./litmus/run.sh
 
 > **_NOTE:_** litmus7 does not support MTE and, consequently, cannot run the test LDRredF.
+
+Support for fault atoms in herd7 and litmus7
+--------------------------------------------
+
+The syntax for the fault atom is `fault(<Pn>:<label>, <location>, <fault_type>)`.
+The components of the syntax are defined as follows:
+
+| Component      | Mandatory | Description                                                                             |
+|----------------|-----------|-----------------------------------------------------------------------------------------|
+| `<Pn>`         | Yes       | The processor or thread identifier. Represents the context in which the fault occurred. |
+| `<label>`      | No        | A label associated with the fault, indicating where the fault has occurred.             |
+| `<location>`   | No        | The memory address or resource where the fault occurred.                                |
+| `<fault_type>` | No        | The type of fault encountered as defined below.                                         |
+
+Currently, herdtools7 supports the following types of faults:
+
+    - `MMU:Translation` fault, illustrated by 3faults.litmus test.
+    - `MMU:AccessFlag` fault, illustrated by 3faults.litmus and LDRaf0F.litmus tests.
+    - `MMU:Permission` fault, illustrated by STRdb0F.litmus.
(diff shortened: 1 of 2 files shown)


diaolo01:
@relokin, following up from discussion on https://github.com/herd/herdtools7/pull/1079, does this match your expectation?

diaolo01:
Hello,

Reading on some of the blogs written about using kvm mode for litmus7, I found the following statement that validates the direction taken in this patch: 

"The post-condition 1:X7=1 /\ fault(P1:L0,x), asks if there are any executions where the register X7 of P1 had the value 1 and **the execution of the instruction with the label L0 in P1 raised a fault** due to a memory access to location x."

https://community.arm.com/arm-community-blogs/b/architectures-and-processors-blog/posts/expanding-memory-model-tools-system-level-architecture

catalogue/aarch64-faults/README.md
```diff
@@ -31,3 +31,21 @@ in the folder `${KUT_DIR}`:
     % sh ./litmus/run.sh
 
 > **_NOTE:_** litmus7 does not support MTE and, consequently, cannot run the test LDRredF.
+
+Herd7 support for faults
```
relokin, line 35: Can we indicate here that fault atoms are also supported in litmus7 (if not also the tools)?

catalogue/aarch64-faults/README.md
```diff
@@ -31,3 +31,21 @@ in the folder `${KUT_DIR}`:
...
+------------------------
+
```
relokin, line 37: Is it worth adding the syntax of the fault atom as well? Something like

fault(<Pn>:<label>,<location>,<fault_type>)

And then we explain what <Pn>, <label> ... is for and which fields are optional?

relokin (commented):
Thanks for this @diaolo01 just two suggestions to improve on this and we can merge this.

catalogue/aarch64-faults/README.md
```diff
@@ -31,3 +31,31 @@ in the folder `${KUT_DIR}`:
...
+--------------------------------------------
+
+Currently, herdtools7 supports the following types of faults:
+
+    - `MMU:Translation` fault, illustrated by 3faults.litmus test.
+    - `MMU:AccessFlag` fault, illustrated by 3faults.litmus and LDRaf0F.litmus tests.
+    - `MMU:Permission` fault, illustrated by STRdb0F.litmus.
+    - `TagCheck` fault, illustrated by LDRredF.litmus test. Not applicable to litmus7.
+    - `UndefinedInstruction`, illustrated by UDF.litmus and noUDF.litmus tests.
+    - `SupervisorCall`, illustrated by SVC.litmus test.
```
relokin, line 45: Should we move this after we've introduced the syntax for the fault atom?

relokin (commented):
Thanks Diana, one more comment and we can merge this PR.

relokin (approved):
Thanks Diana, I will merge this now.

PR #1125, opened 2025-01-09
Pointer authentication code

RemyCiterin:
Here is some support for the Pointer Authentication Code extension of AArch64, it support basic instructions for `FEAT_Pauth2` in herd (`xpad*`, `pac*` and `aut*`).  It support optionally `FEAT_FPAC` to generate faults in case of a `aut*` failure and `FEAT_CONSTPACFIELD` to enable `pac*` on virtual addresses with a non canonical pac field (using the xor semantic)

## Syntax

Pac fields can be added to virtual addresses using the `pac*` instructions or into the initial state using some notations:
- `0:x0=pacda(x, 42)` or `0:x0=pac(x,da,42)` will return the virtual address `x` in `x0` with a pac field `ComputePAC(x,da,42)`: with the key `da`, and the modifier `42` (modifier must be an integer in initial and final state)
- `0:x0=pacda(x,42,10)` or `0:x0=pac(x,da,42,10)` will return the virtual address `x` with the pac field `ComputePAC(x+10,da,42)`, like the other features in herd, we assume their is no aliasing here

One can write this litmus test to use herd with pac:

```
AArch64 pac example
Variant=pac,fpac,const-pac-field
{ 0:x0=pacdb(x,0); int x = pacda(y,42) }
P0;
  autdzb x0;
  ldr x1,[x0];
  xpacd x1;
  pacdza x0;
forall
  ( 0:x1=y /\ 0:x0=pac(x,da,0) /\ ~Fault(P0) )
```

some examples of usage of PAC are also provided with `make test.pac`. Some instructions are still not supported like the combined instructions or the `pacga` instruction. Also this PR completely ignores hash collisions between pac fields (two virtual addresses with syntactically different pac fields are different). Also executing pac related instructions without the `pac` variant is not supported, maybe some checks must be added for it: I don't really know the semantic of `pac` instruction in term of dependencies if it is not activated with `SCTLR_EL1`, so it will be a future work.

Here is some ressources I made to explain the basics of PAC (you can also found some very low level informations in the Arm ARM):
[notes3.md](https://github.com/user-attachments/files/18361549/notes3.md)
autda_success
autdb_failure
ldr_failure
ldr_success
pacda
str_failure
str_success
xpacd

### DIFF
--- a/herd/AArch64ASLSem.ml
+++ b/herd/AArch64ASLSem.ml
@@ -484,7 +484,7 @@ module Make (TopConf : AArch64Sig.Config) (V : Value.AArch64ASL) :
             | ADD | ADDS | SUB | SUBS ->
                 "dpreg/addsub_shift/"
               | AND | ANDS | BIC | BICS | EOR | EON | ORN | ORR ->
-                "dpreg/log_shift/"                
+                "dpreg/log_shift/"
             | _ -> assert false
           and fname =
             match op with
@@ -780,7 +780,7 @@ module Make (TopConf : AArch64Sig.Config) (V : Value.AArch64ASL) :
             | V.Var n -> n
             | _ -> assert false in
           let eq =
-            let op = AArch64Op.Extra (ASLOp.ToBV 64) in
+            let op = AArch64Op.Extra1 (ASLOp.ToBV 64) in
             M.VC.Assign (nv,M.VC.Unop (Op.ArchOp1 op,v))  in
           ASLS.A.V.Var nid,[eq]
       | V.Val cst ->
@@ -878,7 +878,7 @@ module Make (TopConf : AArch64Sig.Config) (V : Value.AArch64ASL) :
           match A.look_reg regq ii.A.env.A.regs with
           | Some v -> Some (aarch64_to_asl v)
           | None -> None
-        in    
+        in
         (match v with Some v -> ASLS.A.state_add st resaddr v | None -> st),eqs
       in
       let test = { test with Test_herd.init_state = init } in
@@ -918,16 +918,71 @@ module Make (TopConf : AArch64Sig.Config) (V : Value.AArch64ASL) :
         in
         nloc
 
-      let tr_op op acc v1 v2 = (M.VC.Binop (op, tr_v v1, tr_v v2), acc)
+      let tr_arch_op op acc v1 v2 =
+        (M.VC.Binop ((Op.ArchOp (AArch64Op.Extra op)), tr_v v1, tr_v v2), acc)
+
+      let tr_op =
+        let open Op in
+        function
+        | ArchOp op -> tr_arch_op op
+        | op ->
+            (* convert the type of `op` from `'a Op.op` to `'b Op.op` *)
+            let new_op =
+              match op with
+              | Add -> Add
+              | Sub -> Sub
+              | Mul -> Mul
+              | Div -> Div
+              | Rem -> Rem
+              | And -> And
+              | Or  -> Or
+              | Xor -> Xor
+              | Nor -> Nor
+              | AndNot2 -> AndNot2
+              | ASR -> ASR
+              | CapaAdd -> CapaAdd
+              | Alignd -> Alignd
+              | Alignu -> Alignu
+              | Build -> Build
+              | ClrPerm -> ClrPerm
+              | CpyType -> CpyType
+              | CSeal -> CSeal
+              | Cthi -> Cthi
+              | Seal -> Seal
+              | SetValue -> SetValue
+              | CapaSub -> CapaSub
+              | CapaSubs -> CapaSubs
+              | CapaSetTag -> CapaSetTag
+              | Unseal -> Unseal
+              | ShiftLeft -> ShiftLeft
+              | ShiftRight -> ShiftRight
+              | Lsr -> Lsr
+              | Lt -> Lt
+              | Gt -> Gt
+              | Eq -> Eq
+              | Ne -> Ne
+              | Le -> Le
+              | Ge -> Ge
+              | Max -> Max
+              | Min -> Min
+              | UMax -> UMax
+              | UMin -> UMin
+              | SetTag -> SetTag
+              | SquashMutable -> SquashMutable
+              | CheckPerms s -> CheckPerms s
+              | ToInteger -> ToInteger
+              | ArchOp _ -> assert false
+            in fun acc v1 v2 -> (M.VC.Binop (new_op, tr_v v1, tr_v v2), acc)
 
       let tr_arch_op1 op acc v =
-        (M.VC.Unop (Op.ArchOp1 (AArch64Op.Extra op), tr_v v), acc)
+        (M.VC.Unop (Op.ArchOp1 (AArch64Op.Extra1 op), tr_v v), acc)
 
       let tr_op1 =
         let open Op in
         function
         | ArchOp1 op -> tr_arch_op1 op
         | op ->
+            (* convert the type of `op` from `'a Op.op1` to `'b Op.op1` *)
             let new_op =
               match op with
               | Not -> Not
--- a/herd/AArch64Sem.ml
+++ b/herd/AArch64Sem.ml
@@ -41,6 +41,9 @@ module Make
     let pte2 = kvm && C.variant Variant.PTE2
     let do_cu = C.variant Variant.ConstrainedUnpredictable
     let self = C.variant Variant.Ifetch
+    let pac = C.variant Variant.Pac
+    let const_pac_field = C.variant Variant.ConstPacField
+    let fpac = C.variant Variant.FPac
 
     let check_mixed ins =
       if not mixed then
@@ -642,6 +645,7 @@ module Make
         M.op Op.Ne x V.zero >>= fun cond ->
         M.choiceT cond (mfault ma mzero) (mok ma mv)
 
+
  (* Semantics has changed, no ctrl-dep on mv *)
       let check_morello_perms a ma mv perms mok mfault =
         M.delay_kont "morello_perms"
@@ -1096,6 +1100,7 @@ module Make
 (* Page tables and TLBs *)
       let do_inv op a ii = inv_loc op (A.Location_global a) ii
 
+
 (************************)
 (* Conditions and flags *)
 (************************)
@@ -1415,6 +1420,34 @@ module Make
                (lift_fault_memtag
                   (mk_fault (Some a_virt) dir an ii ft None) mm dir ii))
 
+(* Check that the Pointer Authentication Code of a virtual address is canonical
+ * before a memory operation. Generate a Branching(pred) effect and an iico_ctrl
+ * +iico_data dependency between the event `ma` and `mop` or `mfault`, and an
+ * iico_data dependency between `mv` and `mop` in case of a success.
+ *)
+      let lift_pac_virt mop ma dir an ii =
+        let mok ma = mop Access.VIR ma >>= M.ignore >>= B.next1T in
+        let lbl_v = get_instr_label ii in
+        let ft = Some (FaultType.AArch64.MMU FaultType.AArch64.Translation) in
+        let mfault ma a =
+          do_insert_commit_to_fault ma
+            (fun _ -> set_elr_el1 lbl_v ii >>| mk_fault (Some a) dir an ii ft None)
+            None ii
+          >>! B.fault [AArch64Base.elr_el1, lbl_v]
+        in
+        let ma_with_commit ma =
+          do_append_commit ma (Some "pac") ii
+        in
+        let mcheck ma =
+          M.delay_kont "pac check" ma (fun a ma ->
+            M.op1 (Op.ArchOp1 AArch64Op.CheckCanonical) a >>= fun c ->
+            M.choiceT c (mok (ma_with_commit ma)) (mfault ma a))
+        in
+        M.delay_kont "pac check virtual" ma (fun a ma ->
+          M.op1 Op.IsVirtual a >>= fun virt ->
+          M.choiceT virt (mcheck ma) (mok ma)
+        )
+
       let lift_morello mop perms ma mv dir an ii =
         let mfault msg ma mv =
           let ft = None in (* FIXME *)
@@ -1476,6 +1509,8 @@ module Make
             let m = lift_kvm dir updatedb mop ma an ii mphy in
             (* M.short will add an iico_data only if memtag is enabled *)
             M.short (is_this_reg rA) (E.is_pred_txt (Some "color")) m
+          else if pac then
+            lift_pac_virt mop ma dir an ii
           else if checked then
             lift_memtag_virt mop ma dir an ii
           else
@@ -1489,7 +1524,7 @@ module Make
             let memtag_sync =
               memtag && (C.mte_precision = Synchronous ||
                          C.mte_precision = Asymmetric) in
-            if memtag_sync || Access.is_physical ac then
+            if memtag_sync || Access.is_physical ac || pac then
               M.bind_ctrldata ma (mop ac)
             else
               ma >>= mop ac)
@@ -1502,7 +1537,7 @@ module Make
           (fun ac ma mv ->
             let open Precision in
             let memtag_sync = memtag && C.mte_precision = Synchronous in
-            if memtag_sync || (is_branching && Access.is_physical ac) then begin
+            if pac || memtag_sync || (is_branching && Access.is_physical ac) then begin
               (* additional ctrl dep on address *)
               M.bind_ctrldata_data ma mv
                 (fun a v -> mop ac a v ii)
@@ -3350,6 +3385,90 @@ module Make
         | Some l -> ii.A.addr2v l
         | None ->  V.intToV (ii.A.addr + 4)
 
+(*******************************)
+(* Pointer Authentication Code *)
+(*******************************)
+
+      let do_pac key rd rn ii =
+        if pac then begin
+          read_reg_ord rd ii >>|
+          read_reg_ord rn ii >>= fun (addr, modifier) ->
+          M.op
+            (Op.ArchOp (AArch64Op.AddPAC (not const_pac_field, key)))
+            addr modifier >>= fun v ->
+          write_reg_dest rd v ii >>= fun v ->
+          B.nextSetT rd v
+        end else
+          read_reg_ord rd ii >>= fun v ->
+          write_reg_dest rd v ii >>= fun v ->
+          B.nextSetT rd v
+
+      let authenticate pointer modifier key ii mop mfault =
+        let mfail md mn = if fpac then
+            (md >>| mn >>= fun _ -> commit_pred ii) >>*= fun _ -> mfault
+          else
+            md >>| mn >>= fun (xd,xn) ->
+            mop (M.op (Op.ArchOp (AArch64Op.AddPAC (false,key))) xd xn)
+        in
+
+        let original_pointer md =
+          md >>= M.op1 (Op.ArchOp1 AArch64Op.MakeCanonical) in
+
+        (* data_input_next allow to not have a control depedency between the
+         * commit event and the register-read event in Xd *)
+        let mop md mn =
+          (md >>| mn >>= fun _ -> commit_pred ii) >>*= fun _ ->
+          (M.data_input_next (original_pointer md) (fun ptr -> mop (M.unitT ptr)))
+        in
+
+        let auth_check xd xn =
+          M.op (Op.ArchOp (AArch64Op.AddPAC (false,key))) xd xn >>=
+          M.op1 (Op.ArchOp1 AArch64Op.CheckCanonical)
+        in
+
+        M.delay_kont "read pointer" pointer (fun xd md ->
+          M.delay_kont "read modifier" modifier (fun xn mn ->
+            auth_check xd xn >>= fun c ->
+            M.choiceT c
+              (mop md mn)
+              (mfail md mn)
+          )
+        )
+
+      let do_aut key rd rn ii =
+        if pac then begin
+          let (>>!) = M.(>>!) in
+
+          let lbl_v = get_instr_label ii in
+          let mfault =
+              set_elr_el1 lbl_v ii >>|
+              mk_fault None Dir.R Annot.N ii
+                (Some (FaultType.AArch64.PacCheck key))
+                None
+              >>! B.fault [AArch64Base.elr_el1, lbl_v]
+          in
+
+          let mop ma =
+            ma >>= fun v ->
+            write_reg_dest rd v ii >>= fun v ->
+            B.nextSetT rd v
+          in
+
+          authenticate
+            (read_reg_ord rd ii)
+            (read_reg_ord rn ii)
+            key ii mop mfault
+        end else
+          read_reg_ord rd ii >>= fun v ->
+          write_reg_dest rd v ii >>= fun v ->
+          B.nextSetT rd v
+
+      let do_xpac r ii =
+        read_reg_ord r ii >>= fun v ->
+        M.op1 (Op.ArchOp1 AArch64Op.MakeCanonical) v >>= fun v ->
+        write_reg_dest r v ii >>= fun v ->
+        B.nextSetT r v
+
 (********************)
 (* Main entry point *)
 (********************)
@@ -3424,7 +3543,7 @@ module Make
           let (>>!) = M.(>>!) in
           let ft = Some FaultType.AArch64.SupervisorCall in
           let m_fault = mk_fault None Dir.R Annot.N ii ft None in
-          let lbl_ret = get_link_addr test ii in          
+          let lbl_ret = get_link_addr test ii in
           m_fault >>| set_elr_el1 lbl_ret ii
           >>! B.syscall [AArch64Base.elr_el1, lbl_ret]
 
@@ -4361,6 +4480,13 @@ module Make
            let lbl_v = get_instr_label ii in
            m_fault >>| set_elr_el1 lbl_v ii
            >>! B.fault [AArch64Base.elr_el1, lbl_v]
+(* Pointer Anthentication Code `FEAT_Pauth2` *)
+        | I_PAC (key, rd, rn) ->
+            do_pac key rd rn ii
+        | I_AUT (key, rd, rn) ->
+            do_aut key rd rn ii
+        | I_XPACI r | I_XPACD r ->
+            do_xpac r ii
 (*  Cannot handle *)
         (* | I_BL _|I_BLR _|I_BR _|I_RET _ *)
         | (I_STG _|I_STZG _|I_STZ2G _
--- a/herd/event.ml
+++ b/herd/event.ml
@@ -523,6 +523,7 @@ module Make  (C:Config) (AI:Arch_herd.S) (Act:Action.S with module A = AI) :
     let memtag = C.variant Variant.MemTag
     let kvm = C.variant Variant.VMSA
     let is_branching = kvm && not (C.variant Variant.NoPteBranch)
+    let pac = C.variant Variant.Pac
     let is_po_partial = A.arch = `ASL
     type eiid = int
     type subid = int
@@ -1834,7 +1835,7 @@ module Make  (C:Config) (AI:Arch_herd.S) (Act:Action.S with module A = AI) :
              [rloc.intra_causality_control;
               rmem.intra_causality_control;rreg.intra_causality_control;
               wmem.intra_causality_control;wreg.intra_causality_control;];
-           if memtag || (physical && is_branching) then
+           if memtag || pac || (physical && is_branching) then
  (* Notice similarity with data composition.  *)
              EventRel.cartesian
                (get_ctrl_output_commits rloc)
@@ -2393,7 +2394,7 @@ module Make  (C:Config) (AI:Arch_herd.S) (Act:Action.S with module A = AI) :
               wm.intra_causality_control])
           (let output_br = get_ctrl_output br in
            EventRel.union4
-             (if is_branching && is_phy then
+             (if pac || (is_branching && is_phy) then
                 EventRel.cartesian (get_ctrl_output_commits rn)
                   (EventSet.union input_rm input_wm)
               else EventRel.empty)
--- a/herd/tests/instructions/AArch64.PAC/A03.litmus
+++ b/herd/tests/instructions/AArch64.PAC/A03.litmus
@@ -1,0 +1,10 @@
+AArch64 A03
+Variant=pac,fpac,const-pac-field
+
+{ 0:x0=x; int64_t 0:x1=0 }
+
+P0            ;
+  pacdza x0   ;
+  swp x1, x1,[x0] ;
+forall
+  ( Fault(P0, MMU:Translation) )
--- a/herd/tests/instructions/AArch64.PAC/A04.litmus
+++ b/herd/tests/instructions/AArch64.PAC/A04.litmus
@@ -1,0 +1,10 @@
+AArch64 A04
+Variant=pac,fpac,const-pac-field
+
+{ 0:x0=x; int64_t 0:x1=0; int64_t x=0; int64_t 0:x2=42 }
+
+P0                 ;
+  pacdza x0        ;
+  cas x1, x2, [x0] ;
+forall
+  ( Fault(P0, MMU:Translation) )
--- a/lib/AArch64Base.ml
+++ b/lib/AArch64Base.ml
@@ -1709,6 +1709,64 @@ type 'k kinstruction =
   | I_STZ2G of reg * reg * 'k idx
   | I_LDG of reg * reg * 'k
   | I_UDF of 'k
+(* Pointer Authentication Code: Basic instruction only
+   See ARM Arm Issue K.a, C3.1.10 for the complete list of instructions
+    - First part: Add PAC to a register or SP, and and return the value in a
+    general purpose register
+*)
+  (* see ARM Arm Issue K.a, C6.2.266
+   * PACIA <Xd>, <Xn|Sp>
+   * PACIZA <Xd>, <X31>
+   * PACIASP <X30>, <Sp>
+   * PACIAZ <X30>, <X31>
+   * PACIA1716 <X17>, <X16>
+   *
+   * see ARM Arm Issue K.a, C6.2.267
+   * PACIB <Xd>, <Xn|Sp>
+   * PACIZA <Xd>, <X31>
+   * PACIBSP <X30>, <Sp>
+   * PACIBZ <X30>, <X31>
+   * PACIB1716 <X17>, <X16>
+   *
+   * see ARM Arm Issue K.a, C6.2.263
+   * PACDA <Xd>, <Xn|Sp>
+   * PACDZA <Xd>, <X31>
+   *
+   * see ARM Arm Issue K.a, C6.2.264
+   * PACDB <Xd>, <Xn|Sp>
+   * PACDZB <Xd>, <X31>
+   *)
+  | I_PAC of PAC.key * reg * reg     (* destination <- AddPACIA(source) *)
+
+(*  - Second part: check the PAC of a register *)
+  (* see ARM Arm Issue K.a, C6.2.23
+   * AUTIA <Xd>, <Xn|Sp>
+   * AUTIZA <Xd>, <X31>
+   * AUTIASP <X30>, <Sp>
+   * AUTIAZ <X30>, <X31>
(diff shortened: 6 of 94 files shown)


maranget:
Hi @RemyCiterin. I am starting a review. I see that there is some code in `litmus/`.

Is the litmus implementation functional? And if so, on which system has it been tested?

RemyCiterin:
Hi @maranget ! The implementation in litmus is not fully functional, at the moment it does not support PAC fields in pre and post conditions but supports the same instructions as herd7 (`pac*`, `aut*` and `xpac*` with the keys `da`, `db`, `ia` and `ib`) and I tested it a little with kvm_unit_tests on my Apple M3 Pro that implement `FEAT_PAUTH2` with `FEAT_FPAC` (but not `FEAT_CONSTPACFIELD`).

maranget:
I could not compile `herd/tests/instructions/AArch64.PAC/A01.litmus`, with errors:
```
Ex.c:587:70: error: implicit declaration of function ‘DATA_SYMB_ID_PACDA’; did \
you mean ‘DATA_SYMB_ID_PTE_X’? [-Werror=implicit-function-declaration]
  587 | ((p->out_0_x1_idx == DATA_SYMB_ID_Y) && (p->out_0_x0_idx == DATA_SYMB_I\
D_PACDA(X, 0X0, 0))) && (!(exists_fault(&p->th_faults[0], INSTR_SYMB_ID_UNKNOWN\
, DATA_SYMB_ID_UNKNOWN, FaultUnknown)));
      |                                                             ^~~~~~~~~~~\
~~~~~~~
      |                                                             DATA_SYMB_I\
D_PTE_X
Ex.c:587:89: error: ‘X’ undeclared (first use in this function)
  587 | = DATA_SYMB_ID_Y) && (p->out_0_x0_idx == DATA_SYMB_ID_PACDA(X, 0X0, 0))\
) && (!(exists_fault(&p->th_faults[0], INSTR_SYMB_ID_UNKNOWN, DATA_SYMB_ID_UNKN\
OWN, FaultUnknown)));
      |                                                             ^

...
```
I gues this is related to PAC fields not being supported. Would it be possible to have litmus7 to fail instead of gcc? That is to have earlier failure. That would much simplify testing the PR on hardware.

maranget:
At the moment I am checking gcc compilation of teh C files produced by litmus7. Could litmus7 also fail when there are PAC fields in the pre-condition?

litmus/preSi.ml
```diff
@@ -2002,20 +2005,77 @@ module Make
         O.o "}" ;
         O.o ""
 
+      let dump_init_pauth_def env test =
+        dump_choose_def env test ;
+        O.o "// read the SCTLR_EL1 status register" ;
```
maranget, line 2010: I have two high level remarks
  1. This code is "constant" (_.i.e._ does not change with litmus input of command-line), It would be better (1) for it to be in its own file, for instance `litmus/libdir/_aarch64/kvm_auth.c` and (2) to include the file with `Insert.insert O.o "kvm-auth.c"`.
  2. Inclusion of this code should depend on some variant such as `pac`, in order not to include this code systematically.

litmus/preSi.ml
```diff
@@ -2002,20 +2005,77 @@ module Make
...
+        O.o "" ;
+        O.o "void init_pauth_key_ia() {" ;
+        O.oi "unsigned long long x = 0xaaaaaaaaaaaaaaaa;" ;
+        O.oi "unsigned long long y = 0xaaaaaaaaaaaaaaaa;" ;
+        O.oi "asm __volatile__(\"msr APIAKeyHi_EL1, %[x]\":: [x] \"r\" (x));" ;
+        O.oi "asm __volatile__(\"msr APIAKeyLo_EL1, %[y]\":: [y] \"r\" (y));" ;
+        O.o "}" ;
+        O.o "" ;
+        O.o "void init_pauth_key_ib() {" ;
+        O.oi "unsigned long long x = 0x5555555555555555;" ;
```
maranget, line 2025: We favor type from `stdint.h`, here `uint64_t` in place of unsigned long long`

herd/tests/instructions/AArch64.PAC/A03.litmus
```diff
@@ -0,0 +1,9 @@
+AArch64 A03
+Variant=pac,fpac,const-pac-field
+
+{ 0:x0=pacda(x,0); 0:x1=0 }
+
+P0            ;
+  swp x1, x1,[x0] ;
```
maranget, line 7: litmus issues a warning on this test:
```
Warning: File "herd/tests/instructions/AArch64.PAC/A03.litmus" Register x1 has different types: <int> and <int64_t>
```
The reason is as follows: by declaring `0:X1=0; ` one implicitly declare register `0:X1` to be of type `int` (the default type in K&R C), while by using the `X` form of register R1 in instruction `swp x1, x1,[x0] `, one declare `0:X1` to hold 64bit quantities. Here it won't harm, because 0:X1 is not printed. However it is advisable to always use 32 or 64 bits quantities consistently. The easiest way to do so is to favor the `W` form of regsisters:
```
{ 0:x0=pacda(x,0); 0:x1=0 }

P0
swp w1, w1,[x0] ;
```
Alternatively, one can declare `0:X1` as follows:
``
{ 0:x0=pacda(x,0); int64_t 0:x1=0 }

P0
swp x1, x1,[x0] ;
```

**Comment by RemyCiterin** [comment]
Yes litmus7 could fail when there are PAC fields in the pre-condition, I have to look for the place initial state is pretty printed to fix this

File: litmus/variant_litmus.ml [hunk 4]
```diff
@@ -24,6 +24,7 @@ type t =
   | SVE (* Do nothing *)
   | SME (* Do nothing *)
   | NoInit (* Do not initialise variables *)
+  | Pac (* Pointer authentication instructions *)
```
maranget, line 27: Hi @RemyCiterim. I am testing your addition to litmus by by issuing the followng command:
```
litmus7 -mach kvm-pac -ascall true -o /tmp/A.tar herd/tests/instructions/AArch64.PAC/*.litmus
```
Where `kvm-pac.cfg` is the following file:
```
#ODROID-C2
size_of_test = 5k
number_of_run = 200
avail = 4
limit = true
memory = direct
stride = 1
carch = AArch64
barrier = userfence
smt = 1
smt_mode = seq
ascall = true
mode = kvm
delay = 32
makevar = AUXFLAGS=0x0
makevar = SRCDIR=$(PWD)/..
makevar = -include $(SRCDIR)/config.mak
makevar = libcflat = $(SRCDIR)/lib/libcflat.a
makevar = libeabi = $(SRCDIR)/lib/arm/libeabi.a
makevar = LIBFDT_archive = $(SRCDIR)/lib/libfdt/libfdt.a
makevar = cstart.o = $(SRCDIR)/arm/cstart64.o
makevar = FLATLIBS = $(libcflat) $(LIBFDT_archive) $(libeabi)
makevar = optional-ccopt = $(shell if $(CC) -Werror $(1) -S -o /dev/null -xc /dev/null > /dev/null 2>&1; then echo "$(1)"; fi)
ccopts = -DNOSWP -std=gnu99 -ffreestanding -I $(SRCDIR)/lib -I $(SRCDIR)/libfdt -Wall -Werror  -fomit-frame-pointer -Wno-frame-address   -fno-pic  -no-pie -Wmissing-parameter-type -Wold-style-declaration -Woverride-init -O2 $(call optional-ccopt, -mno-outline-atomics) -march=armv8.6-a
```

My idea is then to updload `/tmp/A.tar` to some AArch64 machine with kvm-unit-tests installed and to see what happens, mostly if compilation of C files proceeds smoothly.

I cannot do that because the variant listed in the tests are not accepted by litmus. Would you fix this problem?

maranget:
FYI, you'll find a tentative runtime feature check for PAC here 8a805666db554a71f4475a4f2e78725a1bfd82e0.

The one advantage of such code is a specific error message when one runs a PAC test on a system that does not implement PAC.

The code is too simple. For instance it does not really check FPAC (this would require more variants).

litmus/variant_litmus.ml
RemyCiterin: I think I can't fix the problem because the variants in my files doesn't depend of the compilation of the program but of the implementation of PAC in the machine we execute the tests, in particular by default I support only `FEAT_Pauth2` but it's possible to add some variants to support `FEAT_FPAC` with `-variant fpac` and `FEAT_CONSTPACFIELD` using `-variant const-pac-field` and these features doesn't add new instructions but instead rewrite the way `pac*` and `aut*` are executed

RemyCiterin:
Hi @maranget! I added the new variants and their checks based on your suggestion, now `feature_check` must return `0` if one of the expected feature is not implemented, or if `fpac` is not present but `FEAT_FPAC` is implemented (because `FEAT_FPAC` rewrite the way `aut*` works)

lib/AArch64Base.ml
```diff
@@ -1708,6 +1708,81 @@ type 'k kinstruction =
...
+    general purpose register
+*)
+  (* see C6.2.266
+   * PACIA <Xd>, <Xn|Sp>
+   * PACIZA <Xd>, <X31>
+   * PACIASP <X30>, <Sp>
+   * PACIAZ <X30>, <X31>
+   * PACIA1716 <X17>, <X16>
+   *)
+  | I_PAC_IA of reg * reg     (* detination <- AddPACIA(source) *)
```
maranget, line 1723: Hi @RemyCiterin. It would be more rational do define a type `pac_key` (for instance):
```OCaml
type pac_key_t = IA | IB | DA | DB
```
And then to define a single "abstract" instruction:
```OCaml
..,
| I_PAC of pac_key_t * reg * reg
...

This technique yield shorter code in many situations. For instance printing
```OCaml

let pp_pack_key = function
| IA -> "IA" | IB -> "IB" | DA -> "DA" | DB -> "DB"

let do_pp+instruction = ....
...
| | I_PA (key,r1, r2) ->
            sprintf "PAC%S %s, %s"  (pp_pack_key key) (pp_reg r1) (pp_reg r2)
...
```
Similarily for other PAC instruction: better have a "key" argument than four instructions.

lib/AArch64Parser.mly
```diff
@@ -1886,6 +1893,38 @@ instr:
 | TLBI TLBI_OP COMMA xreg
   { I_TLBI ($2, $4) }
 
+/* Pointer Authentication Code Extension */
+| PACIA1716 { I_PAC_IA (Ireg R17, Ireg R16) }
+| PACIASP { I_PAC_IA (Ireg R30, SP) }
+| PACIAZ { I_PAC_IA (Ireg R30, ZR) }
+| PACIA xreg COMMA xreg { I_PAC_IA ($2, $4) }
```
maranget, line 1900: The four rules for `PACIA`, `PACIB`, PACDA` and `PACDB` can be replaced by one:
```
| PAC xreg COMMA xreg { I_PAC ($1,$2, $4) }
```

maranget: Of coutse, the lexer has to me modified to. This amounts to change the right hand sides of rules. Typically
```OCaml
| "PACIA" | "pacia" -> PACIA
```
is replaced by
```OCaml
| "PACIA" | "pacia" -> PAC IA
```

lib/constant.ml
```diff
@@ -28,15 +28,188 @@ type offset = int
 
 let compare_tag = Misc.opt_compare String.compare
 
+module PAC = struct
```
maranget, line 31: The file `constant.ml` is already quite large. Having the `PAC` module in its own file is an option to consider.

lib/constant.ml
```diff
@@ -28,15 +28,188 @@ type offset = int
...
+
+  (* A vritual address must have a PAC field represented in the most significant
+     bits of the pointer, this PAC field may contani multiple PAC signatures using
+     the XOR off all the signatures *)
+  type signature =
+    {
+      (* "ia", "da", "ib" or "db" *)
+      key : string ;
```
maranget, line 39: Why not use a dedicated type with four elements for this field? Maybe  `AArch64Base.pac_key_t`.

lib/constant.ml
```diff
@@ -28,15 +28,188 @@ type offset = int
...
+      | x :: xs -> pp_signature x (aux xs)
+      | [] -> s
+    in
+    aux (PacSet.to_list pac)
+
+  (* A simplex like solver for linear constraints over PAC fields (linear
+     because of the XOR), it use a bi-partition of the variables in two set:
+     the basic variables and the non-basic varaibles (a variable is a PAC
+     field), and canonical is 0 by construction *)
+  type solver_state =
```
maranget, line 103: As far as I could check, the code below is not called. What is its purpose?

lib/op.ml
```diff
@@ -136,8 +143,8 @@ type 'aop op1 =
   | Promote (* Promote to higher precision *)
   | Demote  (* Demote to lower precision *)
   | ArchOp1 of 'aop
-
-
+  | CheckCanonical (* Check that the PAC field of a virtual address is canonical *)
```
maranget, line 146: Perhaps we could consider having those operations in `AArch64Op.ml` as they are arch specific. Notice that this is going halfway, as, for instance, there are no arch-specific faults for instance.

Nevertheless, as the mechanism for arch-specific operations is here, let us use it :)

lib/op.mli
```diff
@@ -45,6 +44,11 @@ type 'aop op =
  *)
   | ToInteger
   | ArchOp of 'aop
+(* PAC specific operations *)
+(* Apply the exclusive OR between the current PAC field of a virtual address
+ * and the PAC signature of this address, and a modifier for a given key.
+ *)
+  | AddPAC of bool * string
```
maranget, line 51: Perhaps we can consider the `AddPac` operation to be defined in `AArch64Op.ml`

lib/constant.ml
RemyCiterin: This is a part of the code for hash collision testing that I forgot to remove before making the PR, I will remove it in my next commit

lib/constant.ml
RemyCiterin: This is a part of the code for hash collision testing that I forgot to remove before making the PR, I will remove it in my next commit

lib/constant.ml
RemyCiterin: Yes it's a good idea, except that I will do the opposite to not have a dependency cycle

litmus/preSi.ml
```diff
@@ -2175,6 +2192,14 @@ module Make
         O.o "static int feature_check(void) {" ;
         if do_self then
           O.oi "cache_line_size = getcachelinesize();" ;
+        if Cfg.variant Variant_litmus.Pac then begin
+          O.fi "if (!check_pac_variant(%S)) return 0;" doc.Name.name;
+          let expect_fpac =
+            if Cfg.variant Variant_litmus.FPac then "true" else "false" in
```
maranget, line 2198: Hi @RemyCiterin, during my experiments I found that **stdbool.h** may not be included. May I suggest this change:
```suggestion
            if Cfg.variant Variant_litmus.FPac then "1" else "0" in
```

litmus/libdir/_aarch64/kvm_auth.c
```diff
@@ -0,0 +1,104 @@
+// read the SCTLR_EL1 status register
```
maranget, line 1: Hi @RemyCiterin. Having non-static functions in a file that is included in the C source of tests is not possible. Namely, with option `-driver C` all tests are bundled together in a unique C executable and the various instances of the functions below will clash.

I guess thet the best solution is to link the test executable with `kvm_auth.o`. I will suggest something soon.

maranget: I had a different solution in mind : do not insert the `kvm_auth.c` code in test code, but link with it instead. Doing so I discovered that kvm-unit-tests is not strictly necessary for the simplest authentification modes. And indeed one can run simple PAC tests with user qemu. I have implemented both enhancements here: 1a8bc7ba1913c70b1acd551443db46ee7f1e4491. Would you consider cherry-picking this commit here?

maranget:
For your motivation, with commit 1a8bc7ba1913c70b1acd551443db46ee7f1e4491. I am able to run this funny (?) test: on my x86_64 desktop machine:
```
lim% cat PAC01.litmus
AArch64 PAC01

{
int x=0;
0:X0=x;
0:X1=x;
0:X2=1;
0:X3=1024;
}

  P0            ;
L0:             ;
  PACDZB X0     ;
  PACDZA X1     ;
  CMP X0,X1     ;
  B.EQ L1       ;
  XPACD X0      ;
  XPACD X1      ;
  ADD X0,X0,#1  ;
  ADD X1,X1,#1  ;
  SUBS W3,W3,#1 ;
  B.NE L0       ;
L1:             ;
locations [0:X3]
exists 0:X3 != 0l
```
With this experimental configuration file. exp.cfg:
```
mode = presi
size_of_test = 10
number_of_run = 10
avail = 4
limit = true
memory = direct
stride = 1
carch = AArch64
barrier = userfence
smt = 4
smt_mode = seq
affinity = incr0
ccopts = -O2 -march=armv8.6-a
crossrun = qemu:qemu-aarch64
gcc = aarch64-linux-gnu-gcc
linkopt = -static
```
Compile (with cross compiler for aarch64 installed)
```
% mkdir -p U
% litmus7 -mach ./exp.cfg -o U PAC01.litmus
& make -C U
```
Run (with qemu-aarch64 installed)
```
% % U/PAC01.exe 
Test PAC01 Allowed
Histogram (2 states)
100   *>0:X3=1023;
100   *>0:X3=905;
Ok
...
```
The technique can be modified easily for running tests on an Apple/ARM machine.

maranget:
This next commit 2c9adfbb0bb2f537dc9a7ac4cc3592b840844a4d implements PAC for litmus standard mode,

maranget:
Hi @RemyCiterin. Have a look at the last two commits of [this branch]( https://github.com/herd/herdtools7/tree/no_collision-luc).

RemyCiterin:
> Hi @RemyCiterin. Have a look at the last two commits of [this branch](https://github.com/herd/herdtools7/tree/no_collision-luc).

I did the change of your commit [[litmus] Run (some) Pauth tests](https://github.com/herd/herdtools7/commit/660c744d60614bb4125621763382a5af21f0b428) but the variant in `objUtils.ml` is only detected if the variant is given to litmus in the arguments, but if the variant is given in the litmus test itself using `Variant=pac` then function `O.variant Variant_litmus.Pac` return false, even if `preSi.ml` see this variant as set so it will generate the `#include "auth.h"`...

RemyCiterin:
@maranget I suppose this is because the `DumpRun.Make` module is call at the end of `top_litmus.ml` with `OT` as argument, but the specific variants to each files using `Variant=...` are determined using the call to `TestVariant` inside the `from_chan` function and `OT` doesn't depend of it, so the variants seen by `DumpRun` (and `ObjUtil` as a consequence) are not the same that the variants seen by the other modules

maranget:
Hi @RemyCiterin. You are right and the problem is in some sense normal. As a matter of fact, `ObjUtil`  performs operations takin gthe complete series of tests compiled into account (here do we add `auth.c` in the list of sources files), while the settings modified by a test metadata `Variant=...` regards this test (for instance exploit a given level of authentification)...

However, adding `auth.c` to the sources does not fall into this model. As a matter of facts, we can say that adding command option `-variant pac` here means « some tests from the list of compiled tests require the extension ». Ideally, we could extract this information automatically from the result of calling `CT.from_file` (which thus has to be modified) in `run_tests` (file `dumpRun.ml`). The information could then be passed as an argument to the call to the function `Obj.dump` or to the functor application `ObjUtil.Make`.

This would be a better solution than to rely on the user to add the proper command line option.

maranget:
For a tentative implementation : commit 578c9e9a576c5de48fd25e9c31070fd7cd320b7c.

RemyCiterin:
Hi, it's better with your patch, but now the generation of the makefile is broken because the `makefile_vars` function still use `Cfg.variant` to know if `Pac` is present and doesn't add `auth.o` to the Makefile.

RemyCiterin:
This fix [[fix] add auth.o to the makefile if PAC is present](https://github.com/herd/herdtools7/pull/1125/commits/69678d722cf7870bfe54daf23e9199f39aba1ad1) seem to solve the problem

maranget:
Nice. glad you fixed this, thanks.

if your are satisfied with the current behaviour, may I suggest that you rebase the PR on top of master?

RemyCiterin:
Yes, that's sound like a good idea, but first I need to ensure that herd fail if we try to pretty print a value with a pac field (except if a given parameter or variant is set), to ensure that herd and litmus will always pretty print the same values (because adding the pretty printing of values with a pac field in litmus seem to be quite complexe...)

RemyCiterin:
Hi @maranget , after a discussion with @artkhyzha and @relokin we decided that at least in a first time, we will strip the PAC fields before the pretty printing and checking of properties in litmus and herd, so if we write the program:
```
AArch64 ldr_fault
Variant=pac
{0:x0=x}
P0;
  pacdza x0;
  ldr x1,[x0];
locations [0:x0;]
forall (Fault(P0));
```

will return 
```
0:X0=x; fault(P0,x,MMU:Translation);
```
even if the address of the fault is in reality `pacda(x,0)` because inverting the hash is not possible in general. 
Something we can do at least in theory is to pretty print that the variable `x` is non canonical. But in a lot of case we can finish the program by an `aut*` instruction to have the same information and this require extra work. And as long as we don't have some concrete examples to show that this strategy is better than just ignoring the pac field I just strip the pac fields in the final state.

I will also add a `keep-pac-field` variant in herd to keep the pac field for a debugging usage but the default parameter will be the one compatible with litmus.

herd/tests/instructions/AArch64.PAC/A03.litmus
```diff
@@ -0,0 +1,10 @@
+AArch64 A03
```
relokin, line 1: not sure this adds much more coverage than A01, A02, but happy to keep it.

herd/tests/instructions/AArch64.PAC/A04.litmus
```diff
@@ -0,0 +1,10 @@
+AArch64 A04
```
relokin, line 1: not sure this adds much more coverage than A01, A02, but happy to keep it.

lib/AArch64Base.ml
```diff
@@ -2412,6 +2470,15 @@ let do_pp_instruction m =
...
       sprintf "UDF %s" (m.pp_k k)
+  (* Pointer Authentication Code *)
+  | I_PAC (key, r1, r2) ->
+      sprintf "PAC%s %s, %s" (PAC.pp_upper_key key) (pp_reg r1) (pp_reg r2)
+  | I_AUT (key, r1, r2) ->
+      sprintf "AUT%s %s, %s" (PAC.pp_upper_key key) (pp_reg r1) (pp_reg r2)
+  | I_XPACI r ->
+      sprintf "XPACI %s" (pp_reg r)
+  | I_XPACD r ->
+      sprintf "XPACD %s" (pp_reg r)
```
relokin, line 2481: Is there a way you had to split this into XPACD and XPACI rather than just XPAC much like PAC.

lib/op.ml
relokin: I agree with @maranget here

lib/op.ml
```diff
@@ -136,8 +143,8 @@ type 'aop op1 =
...
+  | SetCanonical (* Remove the PAC field of a virtual address *)
```
relokin, line 147: I am nitpicking but is MakeCanonical a better name for this?

herd/AArch64Sem.ml
```diff
@@ -4358,6 +4405,53 @@ module Make
            let lbl_v = get_instr_label ii in
            m_fault >>| set_elr_el1 lbl_v ii
            >>! B.fault [AArch64Base.elr_el1, lbl_v]
+(* Pointer Anthentication Code `FEAT_Pauth2` with `FEAT_FPAC` *)
+        | I_PAC (key, rd, rn) ->
+            begin
```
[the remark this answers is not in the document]
relokin: and as discussed it would be nice to have implementation both when pac is enabled and when it is not.

relokin (commented):
This is really good thank you very much Remy!

A few comments, I don't think of them is a blocker.

herd/AArch64Sem.ml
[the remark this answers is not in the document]
RemyCiterin: Yes, but the comment about `FEAT_FPAC` is out of date so I just need to check for PAC here :sweat_smile:, in a first time I will add an user error and a TODO comment and came back here when I would have understand the memory model of PAC if it is deactivated by the system registers...

lib/AArch64Base.ml
```diff
@@ -1709,6 +1709,64 @@ type 'k kinstruction =
   | I_STZ2G of reg * reg * 'k idx
   | I_LDG of reg * reg * 'k
   | I_UDF of 'k
+(* Pointer Authentication Code: Basic instruction only
+   See C3.1.10 for the complete list of instructions
```
artkhyzha, line 1713: Not sure section numbers in the Arm ARM are stable. For instance, in L.a. "C6.2.266" is something else, so it must have changed since you wrote this comment.

Perhaps, mentioning the issue could help. For instance, "Arm ARM Issue K.a, C3.1.10".

lib/AArch64Base.ml
```diff
@@ -1709,6 +1709,64 @@ type 'k kinstruction =
...
+   *
+   * see C6.2.263
+   * PACDA <Xd>, <Xn|Sp>
+   * PACDZA <Xd>, <X31>
+   *
+   * see C6.2.264
+   * PACDB <Xd>, <Xn|Sp>
+   * PACDZB <Xd>, <X31>
+   *)
+  | I_PAC of PAC.key * reg * reg     (* detination <- AddPACIA(source) *)
```
artkhyzha, line 1739: Typo: "detination"

lib/constant.ml
```diff
@@ -277,6 +297,14 @@ let rec eq scalar_eq pteval_eq instr_eq c1 c2 = match c1,c2 with
   | (Tag _,(Concrete _|Symbolic _|Label _|ConcreteRecord _|ConcreteVector _|PteVal _|Instruction _|Frozen _))
     -> false
 
+(* Return if two constants are syntactically differents and can be semantically
+ equals if their is a hash collision between two pac fields *)
```
artkhyzha, line 301: differents -> different
equals -> equal
their -> there

lib/constant.mli
```diff
@@ -104,6 +106,13 @@ val eq :
         ('instr -> 'instr -> bool) ->
           ('scalar,'pte,'instr) t -> ('scalar,'pte,'instr) t -> bool
 
+(* Return if the collision of two PAC fields can implies equality of the two
```
artkhyzha, line 109: implies -> imply
differents -> different

lib/constant.mli
```diff
@@ -152,6 +161,9 @@ val of_symbolic_data : symbolic_data -> ('scalar,'pte,'instr) t
 val as_pte : ('scalar,'pte,'instr) t -> ('scalar,'pte,'instr) t option
 val is_pt : ('scalar,'pte,'instr)  t -> bool
 
+(* Remove the Pac field of a virtual address *)
+val strip_pauth : ('scalar,'pte,'instr) t -> ('scalar,'pte,'instr) t
```
artkhyzha, line 165: Really optional, I just wondered if there is any opportunity to reuse code between this and the logic for "SetCanonical" in symbValue.ml. Also, maybe this could be named "strip_pac" (or something in sync with the setcanonical)? It's just that "pauth" sounds like the name of the feature.

herd/event.ml
```diff
@@ -1839,7 +1840,7 @@ module Make  (C:Config) (AI:Arch_herd.S) (Act:Action.S with module A = AI) :
              [rloc.intra_causality_control;
               rmem.intra_causality_control;rreg.intra_causality_control;
               wmem.intra_causality_control;wreg.intra_causality_control;];
-           if memtag || (physical && is_branching) then
+           if memtag || pac || (physical && is_branching) then
```
artkhyzha, line 1843: This may be not for this PR, but it could be good to have tests sensitive to the changes to the logic for SWP and CAS here.

(I conjecture that the task where these changes will show up is checking if dependencies and pick dependencies via PACD* and AUTD* are created correctly.)

artkhyzha (changes requested):
Thank you, @RemyCiterin. Sharing a few comments, but these are just typos mostly.

One thing that surprised me at first is the reuse of "checked" argument in lift_memop: in many functions you pass an argument as `(memtag || pac)`. Can't see a problem as long as there is no such thing as turning on "-variant mte" and "-variant pac" at the same time. Perhaps, a user error could be shown when that's the case?

Plus, a curious point: isn't `DC` the only non-MTE instruction that uses lift_memop and does not care about PAC? Perhaps, we could have a test with PACD* and DC CVAU -- to record that there is no fault due to non-canonical address?

herd/AArch64ASLSem.ml
```diff
@@ -918,10 +918,65 @@ module Make (TopConf : AArch64Sig.Config) (V : Value.AArch64ASL) :
...
+      let tr_arch_op op acc v1 v2 =
+        (M.VC.Binop ((Op.ArchOp (AArch64Op.Extra op)), tr_v v1, tr_v v2), acc)
+
+      let tr_op =
+        let open Op in
+        function
+        | ArchOp op -> tr_arch_op op
+        | op ->
+            let new_op =
+              match op with
```
Roman-Manevich, line 932: Why not something like:
```
match op with
              | Add 
              | Sub 
              | Mul 
              | Div 
              | Rem 
              | And 
              | Or  
              | Xor 
              | Nor 
              | AndNot2 
              | ASR 
              | CapaAdd 
              | Alignd 
              | Alignu 
              | Build 
              | ClrPerm 
              | CpyType 
              | CSeal 
              | Cthi 
              | Seal 
              | SetValue 
              | CapaSub 
              | CapaSubs 
              | CapaSetTag 
              | Unseal 
              | ShiftLeft 
              | ShiftRight 
              | Lsr 
              | Lt 
              | Gt 
              | Eq 
              | Ne 
              | Le 
              | Ge 
              | Max 
              | Min 
              | UMax 
              | UMin 
              | SetTag 
              | SquashMutable 
              | ToInteger 
              | CheckPerms _
              -> op
              | ArchOp _ -> assert false
```

RemyCiterin: Because the goal of this function is to convert something of type `'a Op.op` to something of type  `'b Op.op` and in your proposition `op` still have the type `'a Op.op`, we may replace it by a `map_op (fun _ -> assert false) op` but their is no `map_op` operator so I used the same method that the `tr_op1` function

Roman-Manevich: I see. Perhaps add a comment that you're changing the types?

herd/AArch64Sem.ml
```diff
@@ -1415,6 +1432,33 @@ module Make
...
+        let ma_with_commit ma =
+          do_append_commit ma (Some "pac") ii
+        in
+        let mcheck ma =
+          M.delay_kont "pac check" ma (fun a ma ->
+            M.op1 Op.CheckCanonical a >>= fun c ->
+            M.choiceT c (mok (ma_with_commit ma)) (mfault ma))
+        in
+        M.delay_kont "pac check virtual" ma (fun a ma ->
+          M.op1 Op.IsVirtual a >>= fun virt ->
```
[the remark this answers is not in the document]
relokin: thanks!

herd/AArch64Sem.ml
```diff
@@ -74,6 +77,12 @@ module Make
           "SME instruction %s requires -variant sme"
           (AArch64.dump_instruction inst)
 
+    let check_pac inst =
+      if not pac then
+        Warn.user_error
+          "Pauth instruction %s requires -variant pac"
+          (AArch64.dump_instruction inst)
+
```
relokin, line 85: this is not used.

herd/AArch64Sem.ml
```diff
@@ -4358,6 +4468,65 @@ module Make
            let lbl_v = get_instr_label ii in
            m_fault >>| set_elr_el1 lbl_v ii
            >>! B.fault [AArch64Base.elr_el1, lbl_v]
+(* Pointer Anthentication Code `FEAT_Pauth2` *)
+        | I_PAC (key, rd, rn) ->
+            if pac then begin
```
relokin, line 4473: This is really optional, but I think you could also move the implementation of the PAC instructions that is inlined here into separate functions.

herd/AArch64Sem.ml
```diff
@@ -1075,6 +1085,74 @@ module Make
             (fun loc v -> Act.tag_access quad Dir.W loc v) ii
         else m a
 
+(* Perform the authentication operation of the `AUT*` instruction, assume that
+ * the CPU can "predict" if the authentication will succede, such that in case
+ * of success, their is no basic-dep (only a pick-basic-dep) between the
+ * register-read event of the morifier, and the register-write event of the
+ * destination *)
```
relokin, line 1092: I don't think this is invariably correct, I would suggest removing it.

herd/AArch64Sem.ml
```diff
@@ -1075,6 +1085,74 @@ module Make
...
+              (mfault md mn)
+          )
+        )
+
+(* Perform the authentication operation of the `AUT*` instruction, assume that
+ * the CPU can "predict" if the authentication will succede, such that in case
+ * of success, their is no basic-dep (only a pick-basic-dep) between the
+ * register-read event of the morifier, and the register-write event of the
+ * destination *)
+      let pauth2_without_fpac pointer modifier key ii mop =
```
relokin, line 1126: This is very similar as `pauth2_with_fpac` with the only difference being the `mfault` vs `mfail`, could you fact the common code in a common function?

herd/AArch64Sem.ml
```diff
@@ -1477,19 +1584,22 @@ module Make
             (* M.short will add an iico_data only if memtag is enabled *)
             M.short (is_this_reg rA) (E.is_pred_txt (Some "color")) m
           else if checked then
-            lift_memtag_virt mop ma dir an ii
+            if pac then
```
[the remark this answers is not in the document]
RemyCiterin: Yes this is what I did at the beginning of the implementation without knowing why some operations were not checked with MTE, thank you to remind me to take care of this point. And yes `lift_pac` correspond to this function so it is call at each address translation

lib/AArch64Base.ml
relokin, line 2481: Sorry just realised that my last comment didn't make much sense so let me try once more. I think you should be able to do:
```suggestion
  | I_XPAC r ->
      sprintf "XPAC%s %s" (PAC.pp_upper_key key) (pp_reg r)
```
is there a reason why you can't do that?

relokin (commented):
I've got one more comment but I think this is ready to be merged. Remy, can you please rebase on top of master?

@maranget @artkhyzha are you happy with this?

lib/AArch64Base.ml
RemyCiterin: Because `XPACD/XPACI` make a distinction between data and instruction but not between the keys `A` or `B`... and pretty printing a key return `DA` or `DB` instead of `D` (respectively for `IA` and `IB`). But for the moment (in absence of address tagging` or `logical address tagging`) `XPACD` and `XPACI` have the same semantic...

herd/event.ml
artkhyzha: Upon a consideration of the graph for SWP and CAS, I believe there is no todo item for this PR.

artkhyzha (approved):
Thank you, @RemyCiterin. LGTM.

relokin:
Thanks @artkhyzha! @maranget we're planning to merge this after rebasing on mast tomorrow. Let us know if you have any concerns or need to more time to review.

maranget:
> Thanks @artkhyzha! @maranget we're planning to merge this after rebasing on mast tomorrow. Let us know if you have any concerns or need to more time to review.


Hi @relokin, I may have some time this afternoon. If I have not, please consider that I approve this PR.

lib/PAC.ml
```diff
@@ -0,0 +1,88 @@
...
+let compare = PacSet.compare
+
+let equal = PacSet.equal
+
+let pp pac s =
+  let rec aux = function
+    | x :: xs -> pp_signature x (aux xs)
+    | [] -> s
+  in
+  aux (PacSet.to_list pac)
```
maranget, line 87: We claim compatibility from OCaml 4.08. `Set.S.to_list` was introduced later. Quick fix:
```suggestion
  aux (PacSet.elements pac)
```

relokin:
Merged. Thanks @RemyCiterin! Many thanks @maranget and @artkhyzha for the help reviewing this.

PR #1129, opened 2025-01-15
[herd] Add boolean in value types

HadrienRenaud:
### Motivation

ASL values have an explicit variant for booleans. This is usually not the case of other languages supported by herd who use 0 for false and anything else for true. As the value types have been constructed for those other languages, the engine in herd relies on this property. This has made the development of ASL in herd quite inconvenient and special conversion operations from integers to booleans and vice-versa had to be introduced. The goal of this PR is to remove those operations.

### Implementation

1. We add to the signatures of scalar, constants, and values:
   1. `true`
   2. `false`
   3. a conversion function to boolean, `as_bool : t -> bool option`

2. Implement those changes for all scalars and values implementations:
   1. `true` is `1`
   2. `false` is `0`
   3. `as_bool` is `fun v -> not (v != 0)`

3. We change the engine to use those new constructs, particularly `lib/symbvalue` and `herd/eventsMonad`

4. We change the ASL semantics to drop the unused conversion to `bool` and back

### Extra

We implement a shortcut in ASLSem for the concatenation operation when one of the operands is the empty bitvector.

### DIFF
--- a/herd/ASLSem.ml
+++ b/herd/ASLSem.ml
@@ -283,16 +283,13 @@ module Make (C : Config) = struct
     (* ASL-Backend implementation                                             *)
     (**************************************************************************)
 
-    let commit (ii,poi) msg =
+    let commit (ii, poi) msg =
       M.mk_singleton_es (Act.Branching msg) (use_ii_with_poi ii poi)
 
     let choice (m1 : V.v M.t) (m2 : 'b M.t) (m3 : 'b M.t) : 'b M.t =
-      M.asl_data
-        m1
-        (function
-         | V.Val (Constant.Concrete (ASLScalar.S_Bool b)) ->
-            if b then m2 else m3
-         | b -> M.asl_data (to_int_signed b) (fun v -> M.choiceT v m2 m3))
+      M.asl_data m1 @@ function
+      | V.Val (Constant.Concrete (ASLScalar.S_Bool b)) -> if b then m2 else m3
+      | b -> M.choiceT b m2 m3
 
     let logor v1 v2 =
       match (v1, v2) with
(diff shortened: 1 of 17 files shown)


herd/ASLSem.ml
```diff
@@ -311,33 +308,42 @@ module Make (C : Config) = struct
...
 
+    let concat v1 v2 =
+      match (v1, v2) with
+      | V.Val (Constant.Concrete (ASLScalar.S_BitVector bv)), v
+        when Asllib.Bitvector.length bv = 0 ->
+          return v
+      | v, V.Val (Constant.Concrete (ASLScalar.S_BitVector bv))
+        when Asllib.Bitvector.length bv = 0 ->
+          return v
+      | _ -> M.op (Op.ArchOp ASLOp.Concat) v1 v2
```
maranget, line 319: Hi @HadrienRenaud. Why this change that looks unrelated to this PR purpose?

lib/wideScalar.mli
```diff
@@ -58,4 +58,5 @@ module type S = sig
   val le : t -> t -> bool
   val mask : MachSize.sz -> t -> t
 
+  val as_bool : t -> bool option
```
maranget, line 61: According to `extendScalar.ml` above change.
```suggestion
```

lib/SVEScalar.ml
```diff
@@ -48,6 +48,8 @@ module BV = struct
   let mask32 = Z.of_int64_unsigned 0xffffffffL
   let mask64 = Z.of_int64_unsigned 0xffffffffffffffffL
 
+  let as_bool z = Some (Z.sign z <> 0)
+
```
maranget, line 52: Probably useless, see `extendScalar.ml`  below
```suggestion
```

lib/extendScalar.ml
```diff
@@ -58,6 +61,8 @@ module
     let of_int64 = to_narrow Narrow.of_int64
     and to_int64 = choose Narrow.to_int64 Wide.to_int64
 
+    let as_bool = choose Narrow.as_bool Wide.as_bool
+
```
maranget, line 65: Hi again, here is the missing part. I do not think there are things such as wide booleans. See the definitions of `s_true` and `s_false` above. If so, we can write :
```suggestion
    l
    let as_bool = function
      | Narrow i -> Narrow.as_bool i
      | Wide i ->
          Warn.fatal "as_bool on wide scalar '%s'" @@ Wide.pp i
```

herd/ASLSem.ml
maranget: Why not! But why so? Maybe this is worth a comment?

maranget: Perfectly so. In fact I was wondering why the shortcut is here and  not in ASLOp. In ant case, this is nitpicking.

maranget:
Thanks @HadrienRenaud. LGTM, please rebase and merge your earliest convenience.

herd/ASLSem.ml
maranget: C'est trop juste.

maranget: Seriously, such optimisations cannot be performed in `lib/SymbValue.ml` because, at this stage, we have no idea of the semantics of `ArchOp` operations. Not sure this is worth fixing...

PR #1130, opened 2025-01-16
[asl][fix] Raise error on arbitrary of an empty type

HadrienRenaud:
The following test was passing, and printing 1. This is not the expected behavior. This PR makes this forbidden.

```
func main () => integer
begin
   let b: integer {1..0} = ARBITRARY: integer {1..0};
   println(b);
   return 0;
end;
```


Roman-Manevich (approved):
Looks good to me!

PR #1136, opened 2025-01-17
[asl][reference] Rename "statically evaluable" -> "symbolically evaluable"

hrutvik:
Note that we deliberately do not rename "static interpretation" (implementation in `StaticInterpreter.ml`) as this is actually doing static interpretation during type-checking.

We only rename the concept of "statically evaluable" to "symbolically evaluable" - i.e. anything which uses side-effect analysis to decide whether an expression can be evaluated deterministically without side-effects.

### DIFF
--- a/asllib/SideEffect.ml
+++ b/asllib/SideEffect.ml
@@ -163,8 +163,8 @@ let is_pure = function
   | WritesLocal _ | WritesGlobal _ | CallsRecursive _ | ThrowsException _ ->
       false
 
-(* Begin IsStaticallyEvaluable *)
-let is_statically_evaluable = function
+(* Begin IsSymbolicallyEvaluable *)
+let is_symbolically_evaluable = function
   | ReadsLocal { immutable } | ReadsGlobal { immutable } -> immutable
   | WritesLocal _ | WritesGlobal _ | NonDeterministic | CallsRecursive _
   | ThrowsException _ | PerformsAssertions ->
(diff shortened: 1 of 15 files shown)


asllib/StaticInterpreter.ml
```diff
@@ -44,8 +44,8 @@ let eval_from ~loc env e =
   ->
     Error.fatal_from loc desc
 
-(* Begin StaticEval *)
-let static_eval (senv : SEnv.env) (e : expr) : literal =
+(* Begin SymbolicEval *)
```
Roman-Manevich, line 47: I don't think this needs renaming. Here we are actually statically evaluating. If the result is not a literal, it is a type error.

asllib/StaticInterpreter.mli
```diff
@@ -24,15 +24,15 @@
 
 module SB = Native.StaticBackend
 
-val static_eval : StaticEnv.env -> AST.expr -> AST.literal
-(** [static_eval env e] statically evaluates [e] in [env] into a literal.
+val symbolic_eval : StaticEnv.env -> AST.expr -> AST.literal
```
Roman-Manevich, line 27: Same comment as for `StaticInterpreter.ml`

Roman-Manevich (commented):
I think the term static evaluation should be kept for everything related to the static interpreter.

Roman-Manevich (approved):
Looks good to me!
Only two comments.

PR #1139, opened 2025-01-21
[asl] limit constraint size after explosion

HadrienRenaud:
Before, the explosion of constraint sets (e.g. turn `{1..4}` into `{1, 2, 3, 4}` before an operation such as `*`) was only done on a interval per interval basis, without taking into account the other side of the operations.
We now perform a very simple estimate of how big the constraint set after the operation will be, based on this decide if we want to explode intervals in a constraint set or not.
Currently the limit is fixed at `2^17`.

### DIFF
--- a/asllib/PP.ml
+++ b/asllib/PP.ml
@@ -34,6 +34,21 @@ let pp_print_seq ?(pp_sep = pp_print_cut) pp_v ppf v =
   in
   Seq.iter pp_v v
 
+(* Just to display better error messages ... *)
+let nonstandard_constraint_compare c1 c2 =
+  match (c1, c2) with
+  | Constraint_Exact _, Constraint_Exact _
+  | Constraint_Range _, Constraint_Range _ ->
+      compare c1 c2
+  | Constraint_Exact e1, Constraint_Range (e3, e4) -> (
+      match compare e1 e3 with
+      | 0 -> ( match compare e1 e4 with 0 -> 1 | n -> n)
+      | n -> n)
+  | Constraint_Range (e1, e2), Constraint_Exact e3 -> (
+      match compare e1 e3 with
+      | 0 -> ( match compare e2 e3 with 0 -> -1 | n -> n)
+      | n -> n)
+
 let pp_comma f () = fprintf f ",@ "
 let pp_comma_list pp_elt f = pp_print_list ~pp_sep:pp_comma pp_elt f
 
@@ -224,6 +239,7 @@ and pp_int_constraint f = function
   | Constraint_Range (x, y) -> fprintf f "@[<h>%a..%a@]" pp_expr x pp_expr y
 
 and pp_int_constraints f li =
+  let li = List.sort nonstandard_constraint_compare li in
   let pp_max_int_constraint_list = 10 in
   if List.length li < pp_max_int_constraint_list then
     fprintf f "@[%a@]" (pp_comma_list pp_int_constraint) li
--- a/asllib/tests/lca.t
+++ b/asllib/tests/lca.t
@@ -39,7 +39,7 @@
   $ aslref lca3.asl
   File lca3.asl, line 5, characters 2 to 25:
   ASL Typing error: a subtype of integer {N} was expected,
-    provided integer {N, 3}.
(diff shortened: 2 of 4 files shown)


asllib/StaticOperations.ml
```diff
@@ -310,32 +310,45 @@ module Make (C : CONFIG) = struct
...
       else Constraint_Range (expr_of_z x, expr_of_z y)
     in
     fun constraints ->
-      let diet, non_static =
-        List.fold_left acc_diet (DZ.empty, []) constraints
-      in
-      DZ.fold
-        (fun interval acc -> constraint_of_interval interval :: acc)
-        diet non_static
+      if List.length constraints > 1000 then constraints
```
hrutvik, line 345: Not sure I follow what this check aims to achieve - would this limit also need documenting?

asllib/error.ml
```diff
@@ -486,6 +492,13 @@ module PPrint = struct
     | NoLoopLimit ->
         fprintf f "@[%a@]" pp_print_text
           "ASL Warning: Loop does not have a limit."
+    | ConstraintSetPairToBigToBeExploded { op; left; right } ->
+        fprintf f
+          "@[Exploding@ sets@ in@ for@ the@ binary@ operation %s@ would@ \
+           result@ in@ too@ big@ a@ constraint@ set@ with@ constraints@ %a@ \
+           and@ %a. Continuing with those.@]"
```
hrutvik, line 499: nit: grammar "in for", and perhaps "those" could more clearly state the incoming constraint sets will be used without exploding

asllib/StaticOperations.ml
```diff
@@ -400,7 +415,22 @@ module Make (C : CONFIG) = struct
...
+            if Z.leq expected_constraint_length max_constraint_size then
+              (ex_cs1, ex_cs2)
+            else
+              let () =
+                C.warn_from ~loc
+                  Error.(
+                    ConstraintSetPairToBigToBeExploded
+                      { op; left = cs1; right = cs2_f })
+              in
+              (cs1, cs2_f)
```
hrutvik, line 433: There's something slightly unsatisfying about this in that we calculate an imprecise _expected_ size of the _outgoing_ constraint, and then refuse to explode even if the resulting size might actually have been fine (which I think can happen?) - so the method of computing the expected size has to be consistently documented and so on.

I wonder if instead we could limit the sizes of the _incoming_ constraints, and refuse to explode any operand constraints that would exceed this size. So in effect, add a second limit to `explode_intervals`:
- the first (existing) one is the limit for explosion for any single constraint
- the second (new) one is the limit for explosion of the whole constraint set

Though the second limit is maybe a little conservative for `MOD`, I wonder if it's a more understandable strategy.

What do you think?

asllib/StaticOperations.ml
```diff
@@ -310,32 +310,45 @@ module Make (C : CONFIG) = struct
     in
     function [] -> [] | x :: li -> aux x [ x ] li
 
+  let dz_unions =
+    let module DZ = Diet.Z in
+    let rec main = function
```
Roman-Manevich, line 315: Can you either rename `main` to something that explains the intention or add some comment please?
I don't understand what this module is doing.

asllib/error.ml
```diff
@@ -486,6 +492,13 @@ module PPrint = struct
     | NoLoopLimit ->
         fprintf f "@[%a@]" pp_print_text
           "ASL Warning: Loop does not have a limit."
+    | ConstraintSetPairToBigToBeExploded { op; left; right } ->
+        fprintf f
+          "@[Exploding@ sets@ in@ for@ the@ binary@ operation %s@ would@ \
+           result@ in@ too@ big@ a@ constraint@ set@ with@ constraints@ %a@ \
```
Roman-Manevich, line 498: how about "...would result in constraint sets larger than {max_constraint_size}... Continuing type-checking with original constraint sets"

Roman-Manevich (approved):
Looks good to me, modulo some minor comments.

asllib/error.ml
Roman-Manevich: I think `2^14` makes sense

PR #1140, opened 2025-01-22
[asl] Fix unsound controlflow analysis for try .. catch

HadrienRenaud:
- [x] Code implementation
- [x] Tests showing before/after difference
- [x] ~Documentation~ -> defered to another PR by @Roman-Manevich

### DIFF
--- a/asllib/tests/control-flow.t/run.t
+++ b/asllib/tests/control-flow.t/run.t
@@ -1,6 +1,7 @@
   $ aslref no-return.asl
   File no-return.asl, line 2, character 5:
-  ASL Typing error: function "main" does not return anything.
+  ASL Typing error: the function "main" may not terminate by returning a value
+    or raising an exception..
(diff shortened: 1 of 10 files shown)


asllib/tests/control-flow.t/run.t
```diff
@@ -26,3 +26,29 @@
...
+  ASL Typing error: function "test0" does not return anything.
+  [1]
+
+  $ aslref try-02.asl
+
+  $ aslref try-03.asl
+
+  $ aslref try-04.asl
+  File try-04.asl, line 5, character 2 to line 9, character 6:
+  ASL Typing error: function "test0" does not return anything.
```
Roman-Manevich, line 43: We may need an error message that conveys the conservative nature of the analysis. 'does not return' is too definite.
How about `may not terminate by returning a value or raising an exception`?

PR #1141, opened 2025-01-23
[asl] Small standard library improvements

HadrienRenaud:
- [x] Fix some erroneous results of `SQRTRounded` for some trivial values
- [x] Add some documentation on Odd Rounding for `SQRTRounded`
- [x] Improve ILog2's implementation

### DIFF
--- a/asllib/libdir/stdlib.asl
+++ b/asllib/libdir/stdlib.asl
@@ -188,16 +188,16 @@ begin
     end;
 
     // Binary search between low and high
-    while low <= high looplimit 2^128 do
+    while low + 1 < high looplimit 2^128 do
         var mid = (low + high) DIVRM 2;
         if 2.0 ^ mid > val then
-            high = mid - 1;
+            high = mid;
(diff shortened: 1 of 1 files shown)


asllib/tests/regressions.t/stdlib.asl
```diff
@@ -150,8 +150,7 @@ begin
     let expected_res = SqrtRounded(x, 1000);
     for p = 1 to 10 do
       let res = SqrtRounded(x, p);
-      // +2 because of the rounding to odd?
-      assert Abs(res - expected_res) / expected_res <= 2.0 ^ (-p+2);
+      assert Abs(res - expected_res) / expected_res <= 2.0 ^ (1-p);
```
solswords, line 153: I think this bound could be 2.0^(-p) and be strict. E.g., with p=1 the biggest relative error should be when the expected result is very close to 1 and the computed result is 1.5 so relative error is nearly but less than 0.5.

PR #1143, opened 2025-01-24
[asl] Fix 0.0 ^ 0 == 1.0

HadrienRenaud:
- [x] Fix bug and now `0.0 ^ 1 == 1.0`
- [x] include a very small gain for `{a..n} POW c` which now does not contain the extra value `1`.
- [x] the base count for `{a..b} OP {c..d}` got put in par with the other type of interval operations in the generated tests with QCheck
- [x] the long factor for QCheck tests of static operations on intervals has been multiplied by 10, which means they take a few minutes to complete on my computer. This is not really a problem as those tests can only be triggered manually.

### DIFF
--- a/asllib/tests/ConstraintBinops.ml
+++ b/asllib/tests/ConstraintBinops.ml
@@ -103,11 +103,11 @@ let print_test op (x, y, cs1, cs2) =
       x Z.pp_print y PP.pp_int_constraints cs1 PP.pp_int_constraints cs2
       (PP.binop_to_string op)
 
-let long_factor = 100
+let long_factor = 1000
 let base_count = 10000
 
 let test_abcd op =
-  let count = base_count * 10
+  let count = base_count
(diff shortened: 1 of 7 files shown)


asllib/tests/ConstraintBinops.ml
```diff
@@ -103,11 +103,11 @@ let print_test op (x, y, cs1, cs2) =
       x Z.pp_print y PP.pp_int_constraints cs1 PP.pp_int_constraints cs2
       (PP.binop_to_string op)
 
-let long_factor = 100
+let long_factor = 1000
 let base_count = 10000
 
 let test_abcd op =
-  let count = base_count * 10
+  let count = base_count
```
hrutvik, line 110: Should this have modified `base_count`, not `long_factor`?

hrutvik (approved):

PR #1144, opened 2025-01-24
[asl][reference] added a lexical rule for double underscores

Roman-Manevich:
* Added a lexical rule to check against double underscores in identifiers.
* Added a macro for defining conventions.
* Changed macro for high-level requirements.


asllib/doc/ASLmacros.tex
```diff
@@ -227,19 +225,25 @@
 % Typesetting macros
 \newtheorem{definition}{Definition}
 \newtheorem{example}{Example}
+\newcommand\ConventionDef[1]{\paragraph{Convention.#1:\label{sec:Convention.#1}}}
+\newcommand\RequirementDef[1]{\paragraph{Requirement.#1\label{Requirement.#1}}}
```
hrutvik, line 229: Just to check:
- is the colon after "Convention" but not "Requirement" intentional?
- as a consistency point, should the `label` for "Requirement" have a `sec:` prefix?

hrutvik (approved):

asllib/doc/ASLmacros.tex
Roman-Manevich: The colon is intenational, but I don't feel strongly about it. Let me know if you think it looks better without it.
I'm trying to avoid `sec:` and go with more specific label structures to allow referring them with the `XXXRef` macros.

PR #1153, opened 2025-01-31
[cat] Ensure that ETS3 includes ordering rules that apply for ETS2

relokin:


relokin:
@artkhyzha if you have moment I would like your thoughts on this.

I contemplated a fix for ETS3 in herd7 but I thought having the fix in the cat file will make it more accessible.

artkhyzha (approved):
Thank you, @relokin, LGTM.

I understand the alternative solution would be to automatically enable ets2 when ets3 is enabled. Not sure why the alternative would be preferable. In principle, one may assume that ets3 implies ets2, but the transliteration of the rules in the Arm ARM does not presently indicate that, so not sure if the assumption is really intuitive.

relokin:
Thanks the review @artkhyzha !

PR #1157, opened 2025-02-04
[litmus,aarch64] Fix SVC execution from EL0

maranget:
When EL0 mode is specified, SVC is executed normally to transit back from EL0 to EL1, do not override default behaviour.

Without the fix EL0 tests will not terminate. This applies for instance for this simple test:

```
AArch64 TST-EL0
Variant=precise
EL0=P0
{
int x=1;
pte_x=(el0:0);
0:X2=x;
1:X2=x;
}
  P0        |  P1         ;
LDR W0,[X2] | LDR W0,[X2] ;
forall Fault(P0,x) /\ 0:X0=0 /\ ~Fault(P1,x) /\ 1:X0=1
```

Problem introduced by PR #1079. As a result of this fix, one can no longer run tests that execute `svc` as a test instruction in EL0 mode. Is this fixable ? @HadrienRenaud, @diaolo01.

### DIFF
--- a/litmus/libdir/_aarch64/kvm_fault_handler.c
+++ b/litmus/libdir/_aarch64/kvm_fault_handler.c
@@ -63,6 +63,9 @@ static void install_fault_handler(int cpu) {
   struct thread_info *ti = thread_info_sp(user_stack[cpu]);
   ti->exception_handlers[EL0_SYNC_64][ESR_EL1_EC_DABT_EL0] = fault_handler;
   ti->exception_handlers[EL0_SYNC_64][ESR_EL1_EC_UNKNOWN] = fault_handler;
-  ti->exception_handlers[EL0_SYNC_64][ESR_EL1_EC_SVC64] = fault_handler;
+  /* SVC is executed normally to transit back from EL0 to EL1,
+     do not override default behaviour
+    ti->exception_handlers[EL0_SYNC_64][ESR_EL1_EC_SVC64] = fault_handler;
+  */
 #endif
 }


relokin:
Good point @maranget.

To fix this we will have to distinguish between `svc` calls from litmus7 and `svc` calls from the actual litmus test. We could kick the can down the road and switch to `hvc` for litmus7 or otherwise, we could keep using `svc` but reserve one immediate `svc #65535` for use by litmus7.  It's unlikely that a litmus test would use this specific immediate but we could also check in the parser. IIRC, the immediate of the call to `svc` is reported in `ESR_EL1.ISS`.

If you prefer we can merge this PR and @diaolo01 can have a closer look at the fix.

maranget:
Hi @relokin and @diaolo01. Thanks for your, if you don't mind I'd rather merge this bugfix PR now than later. Not having svc working as it does for herd7 only in EL0 mode does not look very harmful to me (I found no such test). By contrast having some tests to loop is disturbing.

diaolo01:
I am happy to fix this in a separate PR as @relokin suggested above. Feel free to progress with merging this PR for now.

The only SVC related test we introduced is [herd/tests/instructions/AArch64/A259.litmus](https://github.com/herd/herdtools7/pull/1079/files#diff-30dbe65869574e10fa70869892adb9d7cc50039f83a7edb10ed03bc3fcee249a).

relokin (approved):

maranget:
Thanks @relokin and @diaolo0. I'll merge very soon. One idea for a general fix would be to change the vector table (as we already do for explict handlers + EL0) in the EL0 case .

maranget:
Hi @relokin and @diaolo01. If you do not mind, I'd like to have a try at restoring a complete EL0 support, including explicit handlers and svc.

maranget:
Problem  persists with explicit handlers, setting draft status.
```
AArch64 F004
{
0:X2=0;
}
  P0          | P0.F           ;
L0:           | MOV W2,#1      ;
  SVC #0      | ERET           ;
L1:           |                ;
  ADD W2,W2,#2|                ;

forall 0:X2=3;lim% 
```
Try:
```
% /opt/WIP/bin/herd7 F004.litmus
Contradiction on reg 0:ELR_EL1: loaded label:"P0:L0" vs. stored label:"P0:L1"

Fatal: File "F004.litmus" Adios
Fatal error: exception File "herd/mem.ml", line 852, characters 16-22: Assertion failed
```

PR #1162, opened 2025-02-05
[herd, aarch64]: Introduce support for FEAT_MTE_STORE_ONLY

murzinv:
FEAT_MTE_STORE_ONLY introduces control if Load instructions are Tag Unchecked. Let's add support for that feature, yet (as usual) we need to take extra care for CAS and friends:

  RMNHCZ If FEAT_MTE_STORE_ONLY is implemented, for a Tag Checked
         Compare and swap instruction, if the compare fails, it is
         CONSTRAINED UNPREDICTABLE whether the Tag Check operation is
         performed.

### DIFF
--- a/herd/AArch64Sem.ml
+++ b/herd/AArch64Sem.ml
@@ -1590,11 +1590,17 @@ Arguments:
 
       let do_ldr rA sz an mop ma ii =
 (* Generic load *)
-        lift_memop rA Dir.R false memtag
+        let checked = memtag && not C.mte_store_only in
+        let ma =
+          (* Extract location without a tag from an address *)
+          if memtag && C.mte_store_only then
+            ma >>= fun a -> loc_extract a
+          else ma in
+        lift_memop rA Dir.R false checked
           (fun ac ma _mv -> (* value fake here *)
             let open Precision in
             let memtag_sync =
-              memtag && (C.mte_precision = Synchronous ||
+              checked && (C.mte_precision = Synchronous ||
                          C.mte_precision = Asymmetric) in
             if memtag_sync || Access.is_physical ac || pac then
               M.bind_ctrldata ma (mop ac)
@@ -2083,24 +2089,63 @@ Arguments:
           (rmw_to_read rmw)
           ii
 
-      let do_cas_fail_no_wb sz an rn ma mv mop ii =
-        (* CAS does not generate an Explicit Write Effect          *)
-        (* It is IMPLEMENTATION SPECIFIC if there is an update to  *)
-        (*                                the dirty bit of the TTD *)
-        let proc = ii.AArch64.proc in
-        let tthm = dirty.DirtyBit.tthm proc
-        and hd = dirty.DirtyBit.hd proc in
-        let may_update_db = tthm && hd in
-        let action updatedb =
-          (* Dir.W would force check for dbm bit:                  *)
-          (* - if set then either update or not db bit per R_TXGHB *)
-          (* - if unset raise Permission fault                     *)
-          lift_memop rn Dir.W updatedb memtag mop (to_perms "rw" sz) ma mv an ii
+      let do_cas_fail do_wb sz an rn ma mv mop ii =
+        let action checked ma =
+          let do_action updatedb checked ma =
+              (* Dir.W would force check for dbm bit:                  *)
+              (* - if set then either update or not db bit per R_TXGHB *)
+              (* - if unset raise Permission fault                     *)
+              lift_memop rn Dir.W updatedb checked mop (to_perms "rw" sz) ma mv an ii
+          in
+          if do_wb then
+            do_action true checked ma
+          else begin
+            (* When there is no writeback, It is IMPLEMENTATION SPECIFIC *)
+            (* if there is an update to the dirty bit of the TTD *)
+            let proc = ii.AArch64.proc in
+            let tthm = dirty.DirtyBit.tthm proc
+            and hd = dirty.DirtyBit.hd proc in
+            let may_update_db = tthm && hd in
+            if may_update_db then
+              M.altT (do_action true checked ma) (do_action false checked ma)
+            else
+              do_action false checked ma
+          end
         in
-        if may_update_db then
-          M.altT (action true) (action false)
+
+        if memtag && C.mte_store_only then
+          (* If FEAT_MTE_STORE_ONLY is implemented it is              *)
+          (* CONSTRAINED UNPREDICTABLE whether the Tag Check          *)
+          (* operation is performed.                                  *)
+          M.altT (
+            (* No Tag Check *)
+            (* Extract location without a tag from an address *)
+            let ma = ma >>= fun a -> loc_extract a in
+            action false ma
+          )(
+            (* Tag Check *)
+            action true ma
+          )
         else
-          action false
+          action memtag ma
+
+      let do_cas_fail_with_wb = do_cas_fail true
+      let do_cas_fail_no_wb = do_cas_fail false
+
+      let do_cas sz an rn ma mv mop_success mop_fail_with_wb mop_fail_no_wb ii =
+        M.altT (
+          (* CAS succeeds and generates an Explicit Write Effect *)
+          (* there must be an update to the dirty bit of the TTD *)
+          lift_memop rn Dir.W true memtag mop_success (to_perms "rw" sz) ma mv an ii
+        )( (* CAS fails *)
+          M.altT (
+            (* CAS generates an Explicit Write Effect              *)
+            do_cas_fail_with_wb sz an rn ma mv mop_fail_with_wb ii
+          )(
+            (* CAS does not generate an Explicit Write Effect      *)
+            do_cas_fail_no_wb sz an rn ma mv mop_fail_no_wb ii
+          )
+        )
 
       let cas sz rmw rs rt rn ii =
         let an = rmw_to_read rmw in
@@ -2141,24 +2186,9 @@ Arguments:
           M.aarch64_cas_ok (Access.is_physical ac) ma read_rs read_rt write_rs
                                 read_mem write_mem branch M.eqT
         in
-        M.altT (
-          (* CAS succeeds and generates an Explicit Write Effect *)
-          (* there must be an update to the dirty bit of the TTD *)
-          lift_memop rn Dir.W true memtag mop_success (to_perms "rw" sz)
-            (read_reg_addr rn ii) (read_reg_data_sz sz rt ii) an ii
-        )( (* CAS fails *)
-          M.altT (
-            (* CAS generates an Explicit Write Effect              *)
-            (* there must be an update to the dirty bit of the TTD *)
-            lift_memop rn Dir.W true memtag mop_fail_with_wb (to_perms "rw" sz)
-              (read_reg_addr rn ii) (read_reg_data_sz sz rt ii) an ii
-          )(
-            (* CAS does not generate an Explicit Write Effect          *)
-            let ma = read_reg_addr rn ii
-            and mv = read_reg_data_sz sz rt ii in
-            do_cas_fail_no_wb sz an rn ma mv mop_fail_no_wb ii
-          )
-        )
+        let ma = read_reg_addr rn ii
+        and mv = read_reg_data_sz sz rt ii in
+        do_cas sz an rn ma mv mop_success mop_fail_with_wb mop_fail_no_wb ii
 
       let casp sz rmw rs1 rs2 rt1 rt2 rn ii =
         let an = rmw_to_read rmw in
@@ -2217,28 +2247,9 @@ Arguments:
           M.aarch64_cas_ok (Access.is_physical ac) ma read_rs
               read_rt write_rs read_mem write_mem branch eqp
         in
-        M.altT (
-          (* CASP succeeds and generates Explicit Write Effects  *)
-          (* there must be an update to the dirty bit of the TTD *)
-          lift_memop rn Dir.W true memtag mop_success
-            (to_perms "rw" sz) (read_reg_addr rn ii)
-            (read_reg_data_sz sz rt1 ii >>> fun _ -> read_reg_data_sz sz rt2 ii)
-            an ii
-        )( (* CASP fails *)
-          M.altT (
-            (* CASP generates Explicit Write Effects               *)
-            (* there must be an update to the dirty bit of the TTD *)
-            lift_memop rn Dir.W true memtag mop_fail_with_wb
-              (to_perms "rw" sz) (read_reg_addr rn ii)
-              (read_reg_data_sz sz rt1 ii >>> fun _ -> read_reg_data_sz sz rt2 ii)
-              an ii
-          )(
-            (* CASP does not generate Explicit Write Effects           *)
-            let ma = read_reg_addr rn ii
-            and mv = read_reg_data_sz sz rt1 ii >>> fun _ -> read_reg_data_sz sz rt2 ii in
-            do_cas_fail_no_wb sz an rn ma mv mop_fail_no_wb ii
-          )
-        )
+        let ma = read_reg_addr rn ii
+        and mv = read_reg_data_sz sz rt1 ii >>> fun _ -> read_reg_data_sz sz rt2 ii in
+        do_cas sz an rn ma mv mop_success mop_fail_with_wb mop_fail_no_wb ii
 
       (* Temporary morello variation of CAS *)
       let cas_morello sz rmw rs rt rn ii =
--- a/herd-www/jerd.ml
+++ b/herd-www/jerd.ml
@@ -213,6 +213,7 @@ let run_herd bell cat litmus cfg =
     let variant = !variant
     let fault_handling = !Refs.fault_handling
     let mte_precision = !Refs.mte_precision
+    let mte_store_only = !Refs.mte_store_only
     let sve_vector_length = !Refs.sve_vector_length
     let sme_vector_length = !Refs.sme_vector_length
     let dumpallfaults = !dumpallfaults
--- a/herd/herd.ml
+++ b/herd/herd.ml
@@ -465,6 +465,7 @@ let () =
     let variant = !variant
     let fault_handling = !Refs.fault_handling
     let mte_precision = !Refs.mte_precision
+    let mte_store_only = !Refs.mte_store_only
     let sve_vector_length = !Refs.sve_vector_length
     let sme_vector_length = !Refs.sme_vector_length
   let byte = !byte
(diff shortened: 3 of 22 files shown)


relokin (commented):
Vladimir is this ready to review? Just checking because it's marked as draft.

herd/AArch64Sem.ml
```diff
@@ -1545,15 +1554,17 @@ module Make
           else if checked then
             lift_memtag_virt mop ma dir an ii
           else
+            let ma = if C.mte_store_only then ma >>= fun a -> loc_extract a else ma in
```
relokin, line 1557: I am a bit dubious about this but I am also wondering if this is better done inside `do_ldr` and `cas`.

herd/AArch64Sem.ml
```diff
@@ -646,15 +646,24 @@ module Make
...
+            choice c ma) in
+        match (dir,C.mte_store_only) with
+          | Dir.W,_
+          | Dir.R,false
+            -> check
+          | Dir.R,true
+            ->
+            (* The case of a failed CAS with no write, but with a tag check *)
+            let ma = ma >>= fun a -> loc_extract a in
+            M.altT (check) (m1 ma)
```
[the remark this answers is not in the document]
relokin: yes I think you would need to add another execution (yet another M.altT) with lift_memop.

relokin (commented):
LGTM, do we need a similar change for CASP?

herd/AArch64Sem.ml
```diff
@@ -2093,13 +2096,31 @@ module Make
...
+                  (*       performed                                          *)
+                  let ma = ma >>= fun a -> loc_extract a in
+                  action false false ma
+                )(
+                  (* Tag Check operation is performed, plus an update to the *)
+                  (* dirty bit of the TTD *)
+                  action true memtag ma
+                )
+               else
+                action true memtag ma
```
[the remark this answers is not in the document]
relokin: sorry ignore this, I was experimenting and I was planning to delete this.

herd/AArch64Sem.ml
```diff
@@ -2093,13 +2096,31 @@ module Make
...
+              (* It is IMPLEMENTATION SPECIFIC if there is an update to  *)
+              (*                                the dirty bit of the TTD *)
+              (* Note: the combination of dir=Dir.R and updatedb=true    *)
+              (*                    triggers an alternative in check_ptw *)
+              let ma = read_reg_ord rn ii
+              and action updatedb checked ma =
+                lift_memop rn Dir.R updatedb checked mop_fail_no_wb (to_perms "rw" sz)
+                  ma (read_reg_data sz rt ii) an ii in
+               if memtag && C.mte_store_only then
+                M.altT (
```
artkhyzha, line 2109: This is just checking if I understand the code correctly: would it consider two options when there is no Explicit Write Effect:
* no db bit update, no tagcheck
* with the db bit update, with the tagcheck

? If yes, do we already know if this actually needs to be four cases (yes/no to each part)?

artkhyzha (commented):
Looks nice --I am a bit surprised at how graceful this turned out being in comparison to the db bit logic: changing the `ma` before giving it to `lift_memop` looks easy to follow. Well done! :-)

Just checking one question about interactions between hardware updates and tagchecking. To that end, do we need to have a test with `CAS` and `-variant vmsa,mte` simultaneously?

herd/AArch64Sem.ml
```diff
@@ -955,26 +955,12 @@ module Make
...
-                    M.altT (
-                      add_setbits_db ipte m
-                    )( (* no need to check ha, because hd implies ha *)
-                      add_setbits_af ipte m
-                    )
-                  else if ha then
-                    add_setbits_af ipte m
-                 else m in
-            setbits in
+            (* Write or failed CAS with no write, but with a db update *)
```
relokin, line 963: Do you have to fix the code in lines 929-938 to follow a similar logic?

herd/AArch64Sem.ml
```diff
@@ -955,26 +955,12 @@ module Make
...
+            if hd && updatedb then
+              add_setbits_db ipte m
```
relokin, line 965: would it be worth to add
`assert dir = Dir.W`
and later
`assert dir = Dir.R`?

herd/AArch64Sem.ml
```diff
@@ -2101,10 +2093,40 @@ module Make
...
+            and action updatedb checked ma =
+              lift_memop rn Dir.R updatedb checked mop_fail_no_wb (to_perms "rw" sz)
+                ma (read_reg_data_sz sz rt ii) an ii in
+
+            if memtag && C.mte_store_only then
+              M.altT (
+                (* If FEAT_MTE_STORE_ONLY is implemented it is              *)
+                (* CONSTRAINED UNPREDICTABLE whether the Tag Check          *)
+                (* operation is performed.                                  *)
+                (*                                                          *)
```
relokin: I am nitpicking but this comment applies for both parts of altT so you might want to move it above the call to altT?

relokin (commented):
LGTM, just a few minor points and then I would be happy to merge this.

herd/AArch64Sem.ml
murzinv: I might be missing something but it seems that code in lines 929-938 is fine. IIUC the code produces set of flags for successful access for given `dir` and has nothing to do with `updatedb`. So far for CAS case we have

- CAS succeeds or fails and generates Explicit Write Effect which translates into `Dir.W`
- CAS fails and doesn't generate Explicit Write Effect which translates into `Dir.R`

the last one could update dirty bit (`dbm:1 && hd`) , but that not required for successful access with `Dir.R`

herd/AArch64Sem.ml
murzinv: Not for `hd`:
- CAS succeeds and generates an Explicit Write Effect -> `Dir.W` and `updatedb=true`
- CAS fails and generates an Explicit Write Effect -> `Dir.W` and `updatedb=true`
- CAS fails, does not generate an Explicit Write Effect and no update to the dirty bit -> `Dir.R` and `updatedb=false`
- CAS fails, does not generate an Explicit Write Effect and do update to the dirty bit -> `Dir.R` and `updatedb=true`

Thus `dir` for `hd` case could be both `Dir.W` and `Dir.R`. Meanwhile, I can add `assert dir = Dir.R` for `ha` case

herd/AArch64Sem.ml
murzinv: Sure!

herd/AArch64Sem.ml
relokin: I find this case a little odd:

> CAS fails, does not generate an Explicit Write Effect and do update to the dirty bit -> Dir.R and updatedb=true

Which is not to say why do we have it. It's clearly needed as per R_TXGHB. But the implications of modelling this with Dir.R is that when HAFDB is not enabled then this wouldn't raise a MMU:Permission FAULT. Is that also expected?
cc: @artkhyzha

herd/AArch64Sem.ml
```diff
@@ -2101,10 +2093,40 @@ module Make
             (* CAS does not generate an Explicit Write Effect          *)
```
relokin: According to RMNHCZ shouldn't the above case where CAS fails but generates an Explicit Write Effect also be allowed to not check tags?

herd/AArch64Sem.ml
```diff
@@ -2101,10 +2093,40 @@ module Make
...
+                (* Extract location without a tag from an address           *)
+                let ma = ma >>= fun a -> loc_extract a in
+                if may_update_db then
+                  M.altT (action true false ma) (action false false ma)
+                else
+                  action false false ma
+              )(
+                (* Tag Check *)
+                if may_update_db then
+                  M.altT (action true memtag ma) (action false memtag ma)
```
relokin: memtag is guaranteed to be true 

```suggestion
                  M.altT (action true true ma) (action false true ma)
```

herd/AArch64Sem.ml
```diff
@@ -2101,10 +2093,40 @@ module Make
...
-            (*                    triggers an alternative in check_ptw *)
-            lift_memop rn Dir.R true memtag mop_fail_no_wb (to_perms "rw" sz)
-              (read_reg_addr rn ii) (read_reg_data_sz sz rt ii) an ii
+            let proc = ii.AArch64.proc in
+            let tthm = dirty.DirtyBit.tthm proc
+            and hd = dirty.DirtyBit.hd proc in
+            let may_update_db = tthm && hd in
+            let ma = read_reg_addr rn ii
+            and action updatedb checked ma =
+              lift_memop rn Dir.R updatedb checked mop_fail_no_wb (to_perms "rw" sz)
```
relokin: I think when updatedb is true then dir has to be Dir.W. If HW is permitted to set the dirty state when there is no write it should also be permitted to raise a Permission fault if HAFDBS is not enabled.

herd/AArch64Sem.ml
murzinv: Hmm, perhaps we should allow to not check tags...

herd/AArch64Sem.ml
murzinv: Ack

herd/AArch64Sem.ml
```diff
@@ -2232,12 +2275,42 @@ Arguments:
             (* CASP does not generate Explicit Write Effects           *)
             (* It is IMPLEMENTATION SPECIFIC if there is an update to  *)
             (*                                the dirty bit of the TTD *)
-            (* Note: the combination of dir=Dir.R and updatedb=true    *)
-            (*                    triggers an alternative in check_ptw *)
-            lift_memop rn Dir.R true memtag mop_fail_no_wb
-              (to_perms "rw" sz) (read_reg_addr rn ii)
-              (read_reg_data_sz sz rt1 ii >>> fun _ -> read_reg_data_sz sz rt2 ii)
-              an ii
+            let proc = ii.AArch64.proc in
```
relokin: Is the code in lines 2259-2313 the same as cas? Can we put it in a common function called `do_cas`?

herd/AArch64Sem.ml
```diff
@@ -2143,16 +2135,67 @@ Arguments:
...
+                (* Extract location without a tag from an address *)
+                let ma = ma >>= fun a -> loc_extract a in
+                if may_update_db then
+                  M.altT (action Dir.R true false ma) (action Dir.R false false ma)
+                else
+                  action Dir.W false false ma
+              )(
+                (* Tag Check *)
+                if may_update_db then
+                  M.altT (action Dir.R true true ma) (action Dir.R false true ma)
```
relokin: ditto

herd/AArch64Sem.ml
```diff
@@ -2143,16 +2135,67 @@ Arguments:
...
+                else
+                  action Dir.W false true ma
+              )
+            else
+              if may_update_db then
+                M.altT (action Dir.R true memtag ma) (action Dir.R false memtag ma)
```
relokin: ditto

herd/AArch64Sem.ml
```diff
@@ -2143,16 +2135,67 @@ Arguments:
...
+              (* CONSTRAINED UNPREDICTABLE whether the Tag Check          *)
+              (* operation is performed.                                  *)
+              M.altT (
+                (* No Tag Check *)
+                (* Extract location without a tag from an address *)
+                let ma = ma >>= fun a -> loc_extract a in
+                if may_update_db then
+                  M.altT (action Dir.R true false ma) (action Dir.R false false ma)
+                else
+                  action Dir.W false false ma
```
relokin: If we stick to this rule:

> If it is permitted to set the dirty state when there is no write

Shouldn't this be

```suggestion
                if may_update_db then
                  M.altT (action Dir.R true false ma) (action Dir.W false false ma)
                else
                  M.altT (action Dir.R false false ma) (action Dir.W false false ma)
```

herd/AArch64Sem.ml
relokin: also since this if may_update_db then ... else ... appears three times can we move it inside the action function?

murzinv: Ack

herd/AArch64Sem.ml
murzinv: Hmm, `may_update_db` captures if `HAFDBS` is active, so for if arm
```
if may_update_db then
  M.altT (action Dir.R true false ma) (action Dir.W false false ma)
```
states that it could be a read access that updates the dirty bit, or a write access that doesn’t. The latter seems opposite to what `HAFDBS` implies. Is it actually peritted to have a write access without updating the dirty bit (assuming HD=1, DMB=1, and the direct permission model)?

The else arm
```
else
  M.altT (action Dir.R false false ma) (action Dir.W false false ma)
```
looks reasonable to me

herd/AArch64Sem.ml
murzinv: Ack

herd/AArch64Sem.ml
relokin: Sorry that's a typo I meant to write the opposite:
```suggestion
                if may_update_db then
                  M.altT (action Dir.W true false ma) (action Dir.R false false ma)
```

relokin: Can you also please add a note that we need to look somewhere close to the comment:
>     If it is permitted to set the dirty state when there is no write
Or if you know what rule in the Arm ARM says that add a reference to it instead.

murzinv: Isn't
```
action Dir.W true false ma
```
 already covered by
```
           (* CAS generates an Explicit Write Effect              *)
           lift_memop rn Dir.W true checked mop_fail_with_wb (to_perms "rw" sz) ma (read_reg_data_sz sz rt ii) an ii in
```

murzinv: > Or if you know what rule in the Arm ARM says that add a reference to it instead.

R_TXGHB
> For a CAS or CASP instruction to an address translated by a writable-clean descriptor, if the comparison fails and the location is therefore not updated, then it is CONSTRAINED UNPREDICTABLE whether hardware updates the dirty state.

relokin: Sorry this was in reference to the latter part of this comment

> If it is permitted to set the dirty state when there is no write  it should also be permitted to raise a Permission fault if HAFDBS is not enabled.

I think it's reasonable that a failing CAS is permitted to raise an MMU:Permission Fault but we need to cross check with Arm ARM.

murzinv: Reference to R_TXGHB was for
> If it is permitted to set the dirty state when there is no write

So yes it is permitted for CAS{P} to set dirty state if no write

> I think it's reasonable that a failing CAS is permitted to raise an MMU:Permission Fault but we need to cross check with Arm ARM.

I could not find reference to that in Arm ARM. 

Unless I'm missing something MMU:Permission Fault already covered by CAS fail with write-back, there we have
```
lift_memop rn Dir.W true ...
```
which in case of  HAFDBS not being enabled would be equivalent to
```
lift_memop rn Dir.W  false
```
that would generate MMU:Permission. Unless I've missed something `herd7` (build from master branch) seems to confirm that 
```
AArch64 CAS-Fail
variant=vmsa
{
 [PTE(x)]=(db:0);
 0:X9=x;
}
 P0             ;
MOV W0,#1       ;
CAS W0,WZR,[X9] ;
exists fault(P0,x)
```
result in
```
Test CAS-Fail Allowed
States 2
  ~Fault(P0,x);
 Fault(P0,x,MMU:Permission);
Ok
Witnesses
Positive: 2 Negative: 1
Condition exists (fault(P0,x))
Observation CAS-Fail Sometimes 2 1
Time CAS-Fail 0.02
Hash=4052cf9b612a03a633978bf2ad2686b9
```
In other words it seems CAS fail with write-back perhaps better place to consolidate all `Dir.W` cases, so CAS fail with no write-back would care of all `Dir.R`

herd/AArch64Sem.ml
```diff
@@ -2095,6 +2087,65 @@ Arguments:
...
+            (* Extract location without a tag from an address *)
+            let ma = ma >>= fun a -> loc_extract a in
+            action false ma;
+          )(
+            (* Tag Check *)
+            action true ma
+          )
+        else
+          action memtag ma
+
```
relokin: Genuine question. Is it worth consolidating these two into one?

```suggestion
      let do_cas_fail do_wb sz an rn ma mv mop ii =
        let action checked ma =
          let do_action updatedb checked ma =
              (* Dir.W would force check for dbm bit:                  *)
              (* - if set then either update or not db bit per R_TXGHB *)
              (* - if unset raise Permission fault                     *)
              lift_memop rn Dir.W updatedb checked mop (to_perms "rw" sz) ma mv an ii
          in
          if do_wb then
            do_action true checked ma
          else begin
            (* When there is no writeback, It is IMPLEMENTATION SPECIFIC *)
            (* if there is an update to the dirty bit of the TTD *)
            let proc = ii.AArch64.proc in
            let tthm = dirty.DirtyBit.tthm proc
            and hd = dirty.DirtyBit.hd proc in
            let may_update_db = tthm && hd in
            if may_update_db then
              M.altT (do_action true checked ma) (do_action false checked ma)
            else
              do_action false checked ma
          end
        in

        if memtag && C.mte_store_only then
          (* If FEAT_MTE_STORE_ONLY is implemented it is              *)
          (* CONSTRAINED UNPREDICTABLE whether the Tag Check          *)
          (* operation is performed.                                  *)
          M.altT (
            (* No Tag Check *)
            (* Extract location without a tag from an address *)
            let ma = ma >>= fun a -> loc_extract a in
            action false ma
          )(
            (* Tag Check *)
            action true ma
          )
        else
          action memtag ma
```

herd/AArch64Sem.ml
```diff
@@ -2134,25 +2185,19 @@ Arguments:
...
           )(
-            (* CAS does not generate an Explicit Write Effect          *)
-            (* It is IMPLEMENTATION SPECIFIC if there is an update to  *)
-            (*                                the dirty bit of the TTD *)
-            (* Note: the combination of dir=Dir.R and updatedb=true    *)
-            (*                    triggers an alternative in check_ptw *)
-            lift_memop rn Dir.R true memtag mop_fail_no_wb (to_perms "rw" sz)
-              (read_reg_addr rn ii) (read_reg_data_sz sz rt ii) an ii
+            (* CAS does not generate an Explicit Write Effect      *)
+            do_cas_fail_no_wb sz an rn ma mv mop_fail_no_wb ii
```
relokin: Is this a do_cas function share between cas and casp?

relokin (commented):
LGTM to more questions and we can merge this. Feel free to reject my suggestions if you don't agree that there are improvements, or if you think it makes the code harder to navigate.

herd/AArch64Sem.ml
murzinv: Good point! I think we indeed can have shared `do_cas`

relokin (approved):
Thanks for all the changes Vladimir!

Issue #1160, opened 2025-02-04
[litmus-kvm] final condition check


RemyCiterin:
if I write a program that expect a given address in a register, and this address doesn't match because of an offset then herd will return a fail without error, but litmus will return that the test pass:
```
AArch64 offset_cex
{ 0:x0=x; int64_t 0:x1=4; }
P0;
  add x0,x0,x1;
forall
  ( 0:x0=x )
```
herd will return:
```
States 1
0:X0=x+4;
No
Witnesses
Positive: 0 Negative: 1
```
but litmus will return:
```
Histogram (1 states)
4000000:>0:X0=x;
Ok

Witnesses
Positive: 4000000, Negative: 0
Condition forall (0:X0=x) is validated
```
btw if the offset is greater than 4096, litmus will fail with this error:
```
Failure: Cannot find symbol for address
Failure: Cannot find symbol for address
Failure: Cannot find symbol for address

EXIT: STATUS=3
```
I suppose this is because the `final_cond` function take as argument the result of the `idx_addr` function:
```C
static int idx_addr(intmax_t *v_addr,vars_t *p) {
  intmax_t *p_addr =
    (intmax_t *)((uintptr_t)v_addr & PAGE_MASK);
  if (p_addr == p->x) return DATA_SYMB_ID_X;
  if ((pteval_t *)v_addr == p->pte_x) return DATA_SYMB_ID_PTE_X;
  fatal("Cannot find symbol for address"); return -1;
}
```
but this function only return the ID of the variable it detect and doesn't search to find any information about the offset. So litmus fail to perform the final check and pretty print the wrong value (`x` instead of `x+4` here).
Idealy we may also expect a warning in herd if the offset is greater than `4095` or negative during a memop or in the final state to inform the user that the result may not be coherent with litmus? (I don't know if this is also correct without kvm)

PR #1166, opened 2025-02-06
Fix offsets in postcondition checks with litmus

RemyCiterin:
Fix the issue #1160 about errors in litmus7 in case of an offset between the expected value and the observed value in a litmus test, for example with the test:
```
AArch64 offset_cex
{ 0:x0=x; uint64_t 0:x1=4; }
P0;
  add x0,x0,x1;

forall
  ( 0:x0=x )
```

herd return 
```
States 1
0:X0=x+4;
No
Witnesses
Positive: 0 Negative: 1
```

but the current version of litmus-kvm return
```
4000000:>0:X0=x;
Ok

Witnesses
Positive: 4000000, Negative: 0
```

but with this PR it return:
```
4000000*>0:X0=x+4;
No

Witnesses
Positive: 0, Negative: 4000000
```

because `idx_addr` now return a structure of type `symb_t` with the information of the ID and the offset of the variable it match.

Now the generated `idx_addr` look like this:
```C
static symb_t idx_addr(intmax_t *v_addr,vars_t *p) {
  symb_t ret = { .id = -1, .offset = 0 };
  intmax_t *p_addr =
    (intmax_t *)((uintptr_t)v_addr & PAGE_MASK);
  ret.offset = (uintptr_t)v_addr - (uintptr_t)p_addr;
  if (p_addr == p->x) ret.id = DATA_SYMB_ID_X;
  if ((pteval_t *)v_addr == p->pte_x) {
    ret.offset = 0;
    ret.id = DATA_SYMB_ID_PTE_X; }
  if (ret.id == -1)
    fatal("Cannot find symbol for address");
  return ret;
}
```

### DIFF
--- a/litmus/libdir/_aarch64/kvm_symb.c
+++ b/litmus/libdir/_aarch64/kvm_symb.c
@@ -1,0 +1,34 @@
+
+/* Generic type of symbolic value generated from the final state of a litmus
+ * test or the FAR_EL1 register in case of a fault.
+ *
+ * This type register all the data used for the pretty printing and checks of
+ * the final values like the `id` of the variable, it's `offset` and the raw
+ * pointer for a future usage in presence of a non-canonical MTE or PAC field
+ */
+typedef struct {
+  intmax_t* raw; // raw pointer value
+  uintptr_t offset; // offset between the raw value and the symbolic value
+  int id; // id of the symbolic value
+} symb_t;
+
+static symb_t symbolic_of_id(int id) {
+  symb_t ret = { .id = id, .offset = 0 };
+  return ret;
+}
+
+static symb_t unknown_symbolic() {
+  return symbolic_of_id(DATA_SYMB_ID_UNKNOWN);
+}
+
+static int symbolic_equal(symb_t s1, symb_t s2) {
+  return s1.id == s2.id && s1.offset == s2.offset;
+}
+
+static void pp_symbolic(symb_t s) {
+  printf("%s", data_symb_name[s.id]);
+
+  if (s.offset != 0)
+    printf("+%ld", s.offset);
+}
+
--- a/litmus/libdir/_aarch64/kvm_symb.c
+++ b/litmus/libdir/_aarch64/kvm_symb.c
@@ -1,13 +1,8 @@
 
 /* Generic type of symbolic value generated from the final state of a litmus
  * test or the FAR_EL1 register in case of a fault.
- *
- * This type register all the data used for the pretty printing and checks of
- * the final values like the `id` of the variable, it's `offset` and the raw
- * pointer for a future usage in presence of a non-canonical MTE or PAC field
  */
 typedef struct {
-  intmax_t* raw; // raw pointer value
   uintptr_t offset; // offset between the raw value and the symbolic value
   int id; // id of the symbolic value
 } symb_t;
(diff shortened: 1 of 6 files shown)


maranget:
Hi @RemyCiterin. Nice work, I was able to run your example on top of kvm-unit-tests. Unfortunately the example does not work in the simplest modes of **litmus**. With file `Ex.litmus` containing your exemple, on some Apple/ARM machine:
```
% mkdir -p R
% litmus7 -mach M1 -driver C -o R Ex.litmus
% make -C R
gcc -Wall -std=gnu99 -O2 -O2 -c outs.c
gcc -Wall -std=gnu99 -O2 -O2 -c utils.c
gcc -Wall -std=gnu99 -O2 -O2 -c litmus_rand.c
gcc -DASS -Wall -std=gnu99 -O2 -S Ex.c
Ex.c:120:10: error: call to undeclared function 'symbolic_equal'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
  120 |   return symbolic_equal(_out_0_x0, x);
      |          ^
1 error generated.
make: *** [Ex.s] Error 1
```
In `-mode presi` (similar to kvm/vmsa mode), things are  better, since gcc compilation succeeds:
```
% litmus7 -mach M1 -mode presi -driver C -o R Ex.litmus
% make -C R
...
gcc  -Wall -std=gnu99 -O2  -o run.exe utils.o litmus_io.o platform_io.o litmus_rand.o topology.o @obj run.o
rm Ex.t Ex.s
```
However, the executable attempt to print an address with an non-zero offset does not succeed:
```
%  R/run.exe
Thu Feb  6 22:45:02 2025
%%%%%%%%%%%%%%%%%%%%%%%%%
% Results for Ex.litmus %
%%%%%%%%%%%%%%%%%%%%%%%%%
AArch64 offset_cex
{
 0:X0=x; uint64_t 0:X1=4;
}
 P0           ;
 ADD X0,X0,X1 ;

forall (0:X0=x)
Generated assembler
	;START _litmus_P0
	add	x8, x8, x10
Failure: Cannot find symbol for address
Failure: Cannot find symbol for address
Failure: Cannot find symbol for address
```

RemyCiterin:
Hi @maranget , I fixed the first issue, it was because `compCond.ml` is common with `presi.ml` and the other implementations, now I added a parameter to ensure that it only generate `symbolic_eq` if we use the `preSi.ml` backend.
The second issue was because in its actual state litmus only supports offset in the output state with `kvm-unit-tests`. So the only thing I changed was to keep the offset information in a struct, but without KVM the offset depends on how the variables are allocated and not knowing that I kept intact the way we detect the variables.

litmus/libdir/_aarch64/kvm_symb.c
```diff
@@ -0,0 +1,26 @@
...
+  return ret;
+}
+
+static int symbolic_equal(symb_t s1, symb_t s2) {
+  return s1.id == s2.id && s1.offset == s2.offset;
+}
+
+// Pretty print a symbolic data and return it as a string
+static const char* pp_symbolic(symb_t s) {
+  char* buffer = (char*)malloc(sizeof(char) * 1024);
```
maranget, line 21: Hi @RemyCiterin. I am worried about this call to malloc:
 1. Malloc error code is not checked, please use `malloc_check` from `utils.c` when present.
 2.  There is no mathcimg free. Consider that tests may run for a long time and that with mode `-driver C` many tests  are run in sequence. For these reasons, any memory allocated must be freed.

Here,  static allocation of the buffer is a possibility, or stack-allocation of the buffer by the caller.

litmus/libdir/_aarch64/kvm_symb.c
```diff
@@ -0,0 +1,26 @@
...
+} symb_t;
+
+static symb_t symbolic_of_id(int id) {
+  symb_t ret = { .id = id, .offset = 0 };
+  return ret;
+}
+
+static int symbolic_equal(symb_t s1, symb_t s2) {
+  return s1.id == s2.id && s1.offset == s2.offset;
+}
```
maranget, line 17: In some occasions those functions may be included and not called, which will yield a warning. As the functions are small, an easy workaround can be to declare them as `inline static`.

maranget:
Hi @RemyCiterin. Thanks for the explanations. If I  understand correctly the PR is useful for vmsa mode. This is fine by me.

It will be more useful if one could use the x+offset syntax in final conditions, as for instance in the following test:
```
AArch64 Post
{
int x=1;
0:X0=x;
}
  P0             ;
  LDR W1,[X0],#4 ;
forall 0:X1=1 /\ 0:X0=x+4
```

litmus/libdir/_aarch64/kvm_symb.c
RemyCiterin: I think it will be complex to remove this malloc call without adding extra complexity to the `pp_log` function because `pp_log` automatically generate the pretty printer using a list of C strings, and this function is already very complex. By the way are you sure that this is problem with `-driver C`? Because this code is only call at the pretty printing at the end of the execution of all the tests.

PR #1181, opened 2025-02-18
[herd] Add a little polymorphism to ASL bitvector setfield

maranget:
This patch fixes a problem in AArch64+ASL when ASL code operates ovec PSTATE encoded as an integer.

As a matter of fact if some entry register is a variable and not a constant, then no translation from integer to bitvector occurs. Something that should also be fixed...


maranget:
Set draft status as PR #1188 may be a more general alternative.

maranget:
Superseded by PR #1188.

PR #1185, opened 2025-02-19
[asl] Paired accessor syntax

hrutvik:
Implement paired syntax for accessors (getters/setters), so they must always be declared together as follows:
```
accessor Foo{params}(args) <=> ty
begin
  getter begin
    ...
  end;

  setter = value_in begin
    ...
  end;
end;
```
Now that pairing is syntactically enforced, remove the check that a setter always has a corresponding getter.


Roman-Manevich:
Can you please add an example of an accessor somewhere? Probably the Subprogram Declarations chapter is a good candidate.

PR #1191, opened 2025-02-21
Dot2desc implementation

TiberiuBucur:
`dot2desc` is a tool that translates execution graphs produced by `herd7` in the dot format to accurate English descriptions written in Markdown, for the purpose of inclusion in the Arm ARM (Section B2.3.2.1). It targets intrinsic data, order and control dependencies between effects and for each such dependency, it outputs a line describing it along with the 2 effects it connects. The tool specifically targets the `aarch64` architecture, meaning that describing effects that are not produced by `herd7` in the `aarch64` mode are outside the scope of the tool.

`dot2desc` produces Markdown with inlined LaTeX commands, and relies on the definition of these commands in a dictionary file, which is `herd/libdir/catdefinitions.tex`. The Markdown with inlined LaTeX is meant to be converted to pure Markdown using `pandoc` (which has been implemented in a script that is not part of this pull request), and thus the English descriptions are directly manipulable through this dictionary.

**Strengths:**

- The tool supports accesses to any kind of memory location (with or without offsets) and is compatible with MTE, SVE, VMSA and IFetch.
- Along with generic read/write accesses, it also supports cache (DC/IC) maintenance and TLBI effects, as well as any kind of Fault and Branching effect. I believe these are all the effects `herd7` produces that will be relevant for Section B2.3.2.1.

_Example:_

Let's consider the following test showcasing the SWP instruction:
```
AArch64 SWP
{
 0:X0=1; 0:X2=x;
}
 P0             ;
 SWP W0,W1,[X2] ;
``` 
Running `herd7 -show all -showevents all SWP.litmus -o .` produces the following execution graph:

Screenshot 2025-02-21 at 14 26 09

We are only interested in intrinsic data, control and order dependencies, so we discard everything else and end up with the following Markdown, where LaTeX commands described in `herd/libdir/catdefinitions.tex` have been inlined:

```
-   D~1~: \iicodata{\RREof{X2}}{\ExpMREof{\memlocAddrBy{x}{X2}}}.

-   D~2~: \iicodata{\RREof{X2}}{\ExpMWEof{\memlocAddrBy{x}{X2}}}.

-   O~1~: \iicoorder{\ExpMREof{\memlocAddrBy{x}{X2}}}{\ExpMWEof{\memlocAddrBy{x}{X2}}}.

-   D~3~: \iicodata{\RREof{X0}}{\ExpMWEof{\memlocAddrBy{x}{X2}}}.

-   D~4~: \iicodata{\ExpMREof{\memlocAddrBy{x}{X2}}}{\RWEof{X1}}.
```

This LaTeX commands inlined in this markdown will then get interpreted with `pandoc`, giving rise to the following final pure Markdown output:

```
-   D~1~: There is an Intrinsic Data Dependency from the Register Read Effect of X2 to the Explicit Memory Read Effect of the Memory Location x addressed by X2.

-   D~2~: There is an Intrinsic Data Dependency from the Register Read Effect of X2 to the Explicit Memory Write Effect of the Memory Location x addressed by X2.

-   O~1~: There is an Intrinsic Order Dependency from the Explicit Memory Read Effect of the Memory Location x addressed by X2 to the Explicit Memory Write Effect of the Memory Location x addressed by X2.

-   D~3~: There is an Intrinsic Data Dependency from the Register Read Effect of X0 to the Explicit Memory Write Effect of the Memory Location x addressed by X2.

-   D~4~: There is an Intrinsic Data Dependency from the Explicit Memory Read Effect of the Memory Location x addressed by X2 to the Register Write Effect of X1.
```

**Limitations**

- The tool relies on the dot input to be complete and contain all the information required to describe an effect. This has been developed by closely following the outputs produced by `herd7` and matching the structure of the description of each effect. However, in the case of predicate based branching effects, the graph sometimes misses important information. One example of this are VMSA operations (as highlighted below). When it detects an effect as part of a VMSA instruction, `dot2desc` replaces the missing bits with a default hardcoded based on the most popular observed behaviour in `herd7` and produces a sensible output. Outside of VMSA, if the branching predicate is missing, `dot2desc` assumes that it is part of the instruction (eg. the "EQ" condition in `CSEL X0, X1, X2, EQ`) and fails if it cannot find it.

_Example:_

Consider the following VMSA test:
```
AArch64 LDR-v1
Variant=vmsa,fatal
{
 [PTE(x)]=(oa:PA(x),valid:1);
 [x]=1;
 0:X0=x;
}
 P0          ;
L0:          ;
LDR W1,[X0] ;
```
This is the dot graph produced by `herd7` on this test, which very clearly specifies the predicate `valid:1 && af:1` for the branching effect:

Screenshot 2025-02-21 at 14 07 00

However, when the `valid` bit of the page table entry is not set, as in the following test
```
AArch64 LDR-v0
Variant=vmsa,fatal
{
 [PTE(x)]=(oa:PA(x),valid:0);
 [x]=1;
 0:X0=x;
}
 P0          ;
L0:          ;
 LDR W1,[X0] ;
```

then `herd7` produces the following graph, with the branching predicate missing:

Screenshot 2025-02-21 at 14 15 47

In this case, `dot2desc` will assume that the branching predicate is `valid:1 && af:1` and produce the following output:

```
-   D~1~: \iicodata{\RREof{X0}}{\ImpTagMREof{\taglocOf{x}{\memlocAddrBy{x}{X0}}}}.

-   D~2~: \iicodata{\ImpTagMREof{\taglocOf{x}{\memlocAddrBy{x}{X0}}}}{\IntrBranching{\PTECheck{x}{X0}{valid:1 && af:1}}}.

-   C~1~: \iicoctrl{\IntrBranching{\PTECheck{x}{X0}{valid:1 && af:1}}}{\genericExcEntry{MMU Translation}}.

-   C~2~: \iicoctrl{\IntrBranching{\PTECheck{x}{X0}{valid:1 && af:1}}}{\RWEof{`ELR_EL1`}}.
```


dot2desc/dotGraph.ml
```diff
@@ -0,0 +1,745 @@
...
+      let any_active = Str.regexp {|[a-zA-Z0-9_]*: Branching(pred)(AnyActive(\([A-Z_]+[0-9]*\)))|} in
+      let active_elem = Str.regexp {|[a-zA-Z0-9_]*: Branching(pred)(ActiveElem(\([A-Z_]+[0-9]*\), \([0-9]+\)))|} in
+      let branching_instr_cond = Str.regexp {|[a-zA-Z0-9_]*: Branching(pred)|} in
+      let bcc_branching = Str.regexp {|[a-zA-Z0-9_]*: Branching(bcc)|} in
+      let exc_return = Str.regexp {|[a-zA-Z0-9_]*: ExcReturn|} in
+      let fault = Str.regexp {|[a-zA-Z0-9_]*: Fault(\([a-zA-Z0-9_:,]*\))|} in
+      let exc_entry = Str.regexp {|[a-zA-Z0-9_]*: ExcEntry(\([a-zA-Z0-9_:,]*\))|} in
+      let empty = Str.regexp {|[a-zA-Z0-9_]*: \\|} in
+
+      let require_address_reg f name =
```
ShaleXIONG, line 174: This function is always called with `require_address_reg DescDict.tag_read`. This function returns another function `func`?

dot2desc/dotGraph.ml
```diff
@@ -0,0 +1,745 @@
...
+
+  match stmt with
+  | ParsedStmt.Node n ->
+    let value = get_label_value n.ParsedNode.attrs in
+    if is_init_event n then
+      (* Skip init events *)
+      acc
+    else begin
+      let symbol = "\\([a-zA-Z0-9_\\+]+\\)" in
+      let tag = Printf.sprintf "tag(%s)" symbol in
```
ShaleXIONG, line 143: Can you add an example string you want to match, in all the following `regexp`? For most variable I can understand based on the variable name, but I get a bit confused, for example, `tlbi` vs `generic_tlbi`.

dot2desc/dotGraph.ml
```diff
@@ -0,0 +1,745 @@
...
+        func in
+
+      let tag_read = require_address_reg DescDict.tag_read "tag read" in
+      let tag_write = require_address_reg DescDict.tag_read "tag write" in
+      let pte_read = require_address_reg DescDict.tag_read "PTE read" in
+      let pte_write = require_address_reg DescDict.tag_read "PTE write" in
+      let pa_read = require_address_reg DescDict.tag_read "PA read" in
+      let pa_write = require_address_reg DescDict.tag_read "PA write" in
+
+      let node = if Str.string_match mem_access value 0 then
```
ShaleXIONG, line 187: There are a few large "if-else if-else" block. Can you put a bit comments:
- are boolean conditions exclusive with each other?
- how do you decide the checking ordering and why it is correct?
Also just a general short description might help, alongside the variable names.

hugookeeffe: perhaps this could even be simplified using pattern matching? you can have the access type as the pattern to match against and use the 'when' statement to ensure your condition is met.

i.e.
```
let check a = Str.string_match a value 0 in
match access with
| mem_access when check access -> <code>
| tag_access when check access -> <code>
...
```

As the code within the if is quite similar in a lot of these it might be possible to simplify that too in some cases

dot2desc/dotGraph.ml
TiberiuBucur: Oops, my mistake here. There should be a different function for each case.

dot2desc/dotGraph.ml
TiberiuBucur: As far as I know, there isn't a way to combine regular expression matching with string pattern matching. You have to either use one or the other, and since the text inside the nodes can vary a bit, I think regexes are the way to go.

dot2desc/dotGraph.ml
```diff
@@ -0,0 +1,745 @@
...
+      let exc_entry = Str.regexp {|[a-zA-Z0-9_]*: ExcEntry(\([a-zA-Z0-9_:,]*\))|} in
+      let empty = Str.regexp {|[a-zA-Z0-9_]*: \\|} in
+
+      let require_address_reg f name =
+        let func loc reg is_explicit = match reg with
+        | Some r -> f loc r is_explicit
+        | None -> Warn.fatal "No address register found for %s access\n" name in
+        func in
+
+      let tag_read = require_address_reg DescDict.tag_read "tag read" in
```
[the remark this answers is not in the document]
TiberiuBucur: This is a mistake. They are not all meant to take `DescDict.tag_read` as input 😅

dot2desc/dotGraph.ml
```diff
@@ -0,0 +1,745 @@
...
+        let loc = Str.matched_group 2 value in
+
+        let reg_regex = Str.regexp {|[A-Z_]+[0-9]*|} in
+        let reg = try
+          ignore (Str.search_backward reg_regex value (String.length value - 1));
+          Str.matched_group 0 value
+        with Not_found ->
+          Warn.fatal "Could not find register as part of TLBI instruction" in
+
+        let loc = if Str.string_match tag_regex loc 0 then
```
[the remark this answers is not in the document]
TiberiuBucur: Resolving this after I implemented the suggestions we discussed on the call we had today.

dot2desc/dotGraph.ml
TiberiuBucur: And yes, the conditions are exclusive with each other. The if-else statements check what kind of effect the current dot graph node is. An effect can only be of one type (it cannot be both a memory access and a branching effect). The ordering mostly doesn't matter, except for the case of `branching_instr_cond`. A string that matches any other branching regex beside `bcc_branching` (eg. `branching`, `branching_mte_tag`, `branching_pte` etc.) would also match `branching_instr_cond`. This is why all the others are checked before this one, so that the more precise type of effects can be "ruled out" before we decide the branching condition is part of the instruction.

dot2desc/dotGraph.ml
TiberiuBucur: Done.

dot2desc/dotGraph.ml
```diff
@@ -0,0 +1,745 @@
...
+        let edge = { Edge.left=e.ParsedEdge.left; right=e.ParsedEdge.right; desc=desc; kind=kind } in
+        { acc with edges=edge::acc.edges }
+      with Not_found ->
+        (* Skip any other kind of edge *)
+        acc
+    end
+  | ParsedStmt.Subgraph _ ->
+    Warn.fatal "Found dot subgraph even after flattenning was performed"
+
+let convert_bnodes stmts =
```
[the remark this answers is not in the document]
TiberiuBucur: Done. Let me know if it's more clear now.

dot2desc/dotGraph.ml
```diff
@@ -0,0 +1,745 @@
...
+  ) stmts
+
+let flatten_subgraphs stmts =
+  let rec flatten stmt acc = match stmt with
+  | ParsedStmt.Subgraph s ->
+      List.fold_right flatten s.ParsedSubgraph.stmts acc
+  | stmt -> stmt :: acc in
+  List.fold_right flatten stmts []
+
+let get_param_maps stmts instr =
```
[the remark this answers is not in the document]
TiberiuBucur: Done. Does it make sense?

dot2desc/dotGraph.ml
TiberiuBucur: Added some comments. Let me know if they clear things up.

ShaleXIONG (approved):

ShaleXIONG:
Can you add the tutorial to use this tool as a readme.

dot2desc/debug.mli
ShaleXIONG, line 1: Probably remove this file as there is no different between the interface `debug.mli` and `debug.ml`, because you have expose every function.

dot2desc/descDict.mli
ShaleXIONG, line 1: You can move all descriptions and comments into the `descDict.ml` and remove the interface file `descDict.mli`.

dot2desc/opts.mli
ShaleXIONG, line 1: This interface file `opts.mli` is unnecessary.

dot2desc/opts.ml
ShaleXIONG, line 1: Probably find a better file name and merge `opts.ml` to `debug.ml`. Both of them are only small configurations related informations.

lib/dotLexer.mli
ShaleXIONG, line 1: Is there a way to merge this `dotLexer.mli`, to `dotLexer.mll`? Possbly inline the module interface declaration?

TiberiuBucur: My thought process when I developed this was to follow the convention for all the other lexers under `/lib`. For example `AArch64Lexer.mll`, `BellLexer.mll`, `CLexer.mll` etc. all have respective `.mli` files with minimal signatures that could be automatically inferred by the Ocaml compiler.

Either way, I decided to integrate the signature into the `dotLexer.mll` file as requested.

PR #1192, opened 2025-02-21
[diy] Code improvement: upgrade the `Code.v` type

ShaleXIONG:
Change the type of `Code.v` from plain `int` to a customer type `NoValue | Plain of int`.  We did not use the standard `option` type as we will introduce fold more value type in the future.

In cycle building process, there are too many different value related field in the `node` data structure. The use of those fields are based on the `node` type and mode/variant of the `diy7`, e.g. `kvm` and `mte`. Those cause confusion and unnecessary complications when maintaining the code.

It is now a relatively trivial change, where `NoValue` collapses to previously default value `-1` via a new function `Code.value_to_int`. However, this leads to a future change that folds other value-related types, such as pte value, instruction value (for iFetch) and memory tag value (for MTE), which means simpler and better-typed code for data type `node` in `cycle.ml`.

In general we assume that 
- in `xxxcompile`, `Code.v` will collapse to `int` and the compilation will work around `int` type
- all previous step work around `node` data structure will work around the more expressive `Code.v`. 

** Note that `Code.value_of_int` is NOT precisely the duality of `Code.value_to_int`, because `Code.value_of_int -1 = Plain -1`, which means `NoValue <> (Code.value_of_int @@ Code.value_to_int @@ NoValue)`.

### DIFF
--- a/gen/cycle.ml
+++ b/gen/cycle.ml
@@ -24,27 +24,37 @@ module type S = sig
   type atom
   module PteVal : PteVal_gen.S with type pte_atom = atom
 
+  (* TODO can be parametric by dir *)
   type event =
-      { loc : loc ; ord : int; tag : int ;
+      { loc : loc ; ord : int; 
+        (* TODO morello related value, fold into value *)
+        tag : int ;
         ctag : int; cseal : int; dep : int ;
         v   : v ; (* Value read or written *)
+        (* TODO fold into value *)
         vecreg: v list list ; (* Alternative for SIMD *)
+        (* TODO instruction fetch related value, fold into value *)
         ins : int ;
         dir : dir option ;
         proc : Code.proc ;
         atom : atom option ;
+        (* TODO why not put into dir ? *)
         rmw : bool ;
+        (* TODO, cell and tcell fold together *)
         cell : v array ; (* Content of memory, after event *)
         tcell : v array ; (* value of tag memory after event *)
         bank : SIMD.atom Code.bank ;
         idx : int ;
+        (* TODO this `pte` value should be fold into value `v`. *)
         pte : PteVal.t ; }
 
   val evt_null : event
   val make_wsi : int -> Code.loc -> event
 
+  (* TODO change v event *)
   val debug_evt : event -> string
 
+  (* TODO change v event *)
   module OrderedEvent : Set.OrderedType with type t = event
 
   module EventMap : MyMap.S with type key = event
@@ -167,12 +177,12 @@ module Make (O:Config) (E:Edge.S) :
     { loc=Code.loc_none ; ord=0; tag=0;
       ctag=0; cseal=0; dep=0;
       vecreg= [];
-      v=(-1) ; ins=0;dir=None; proc=(-1); atom=None; rmw=false;
+      v=Code.no_value ; ins=0;dir=None; proc=(-1); atom=None; rmw=false;
       cell=[||]; tcell=[||];
       bank=Code.Ord; idx=(-1);
       pte=pte_default; }
 
-  let make_wsi idx loc = { evt_null with dir=Some W ; loc=loc; idx=idx; v=0;}
+  let make_wsi idx loc = { evt_null with dir=Some W ; loc=loc; idx=idx; v=Plain 0;}
 
   module OrderedEvent = struct
     type t = event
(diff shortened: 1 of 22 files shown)


ShaleXIONG:
@relokin @maranget  Hi, this is just a small code improvement but there will be more on-going work in the future. I figure it is better to make a pull request once in a stable stage.

relokin:
@hugookeeffe have you looked at this PR? Can you please have a look if you haven't already?

gen/cycle.ml
```diff
@@ -24,27 +24,37 @@ module type S = sig
...
         proc : Code.proc ;
         atom : atom option ;
+        (* TODO why not put into dir ? *)
         rmw : bool ;
+        (* TODO, cell and tcell fold together *)
         cell : v array ; (* Content of memory, after event *)
         tcell : v array ; (* value of tag memory after event *)
         bank : SIMD.atom Code.bank ;
         idx : int ;
+        (* TODO PTE fold into value *)
```
hugookeeffe, line 48: is value referring to the v in this structure? it is a little unclear to me from the comment alone.

gen/cycle.ml
```diff
@@ -515,7 +526,9 @@ module CoSt = struct
          | Ord ->
             co_cell.(0) <- E.overwrite_value old e.atom cell2
          | Pair -> (* No Rmw for pairs *)
-            co_cell.(0) <- E.overwrite_value old e.atom (e.v-1);
+                 (* TODO check the variable name `width` *)
```
hugookeeffe, line 529: what needs to be checked about the variable name?

gen/top_gen.ml
```diff
@@ -422,6 +422,10 @@ let max_set = IntSet.max_elt
             vs in
         build_observers p i x vs in
 
+    (* string ->
```
hugookeeffe, line 425: whats the purpose of this comment and the todo after?

hugookeeffe (approved):
This PR seems good to me. I only have a few minor comments about some of the TODO's you put but nothing blocking.
thanks!

hugookeeffe:
> @hugookeeffe have you looked at this PR? Can you please have a look if you haven't already?

I have just had another look @relokin . I only had a few minor comments, overall it seems good to me

gen/cycle.ml
ShaleXIONG: Yes it means to fold into value type `v`. I will make it clear. I am working on it now as a side code improvement project but will make a separate pull once it is ready.

gen/cycle.ml
ShaleXIONG: It is no longer a valid comment. I will remove it.

gen/top_gen.ml
ShaleXIONG: Both these comments and the TODO are no longer used in this function. I will remove.

gen/topUtils.ml
```diff
@@ -281,7 +282,8 @@ module Make : functor (O:Config) -> functor (C:ArchRun.S) ->
         Some r
       with Not_found -> None
 
-    let is_load_init e = e.C.C.dir = Some R && e.C.C.v = 0
+    (* TODO check if the initial should be change to None? *)
```
maranget, line 285: Notice that. by litmus test convention, locations that are not explicitly initialized are in fact initialized to zero.

gen/topUtils.ml
```diff
@@ -54,9 +54,10 @@ module Make : functor (O:Config) -> functor (C:ArchRun.S) ->
     open Printf
     open Code
 
-    let pp_v =
-      if O.hexa then sprintf "0x%x"
-      else sprintf "%i"
+    let pp_v v =
+      let v = Code.value_to_int v in
+      if O.hexa then sprintf "0x%x" v
+      else sprintf "%i" v
```
maranget, line 60: Hi @ShaleXIONG. Would it be meaningfull to print `NoValue` as, say `**` instead of  `-1` ?

ShaleXIONG: That is a good idea ! I will do it.

maranget:
Hi @ShaleXIONG. The PR is fine and can be merged. Please rebase on top of master before merging for the sake of a as linear as possible history. Tanks.

maranget:
Merged, thanks @ShaleXIONG , @relokin and @hugookeeffe.

PR #1195, opened 2025-02-21
[Makefile] Add aarch64-ifetch catalogue to the regressions

relokin:


relokin:
@artkhyzha was there a reason we didn't add the catalogue to the Makefile regressions?

artkhyzha (approved):
Thank you, @relokin. I have not realised ifetch was not added into the regressions! Thank you for correcting the oversight.

I ran `make ifetch-test` and saw no surprises. I also ran `make cata-test` and see the ifetch tests in the output. Overall, `make test` fails for me due to ASL (`test-aarch64-asl`).

relokin:
Thanks for the review @artkhyzha. The regressions on my end run pass, if they pass on github as well I will merge because I think the failure you get for `test-aarch64-asl` seems unrelated to this PR.

PR #1076, opened 2024-11-26
[asl] Pending constrained integers

hrutvik:
This PR iterates on https://github.com/herd/herdtools7/pull/1056. In particular, that PR used the `integer{}` syntax (AST: `WellConstrained []`) to inherit integer constraints from right-hand sides to left-hand sides. However, `integer{}` could be confused with a hypothetical "empty" set of constraints, and the changes to the reference documents and AST did not help this.

This PR replaces this with the `integer{-}` syntax, which parses to a new AST node `PendingConstrained : constraint_kind`. Similarly, reference documentation refers to these integers as "pending constrained". This should avoid the confusion above. This therefore partially reverts the PR above, but not fully.

PR #1204, opened 2025-02-26
[asl] Pending constrained globals

hrutvik:
Build on https://github.com/herd/herdtools7/pull/1056 and https://github.com/herd/herdtools7/pull/1076 to apply the concept of "pending constrained" integers to global declarations. However, they are still forbidden for `config` declarations. For example:
```
var w : integer{-} = 1;      // OK
let x : integer{-} = 2;      // OK
constant y : integer{-} = 3  // OK
config z : integer{-} = 4    // ERROR
```


asllib/doc/GlobalStorageDeclarations.tex
```diff
@@ -236,14 +236,14 @@ \section{Typing}
   \item checking that $\name$ is not already declared in $\genv$ yields $\True$\ProseOrTypeError;
   \item applying $\withemptylocal$ to $\genv$ yields $\tenv$;
   \item define $\vtargettimeframe$ as $\timeframeconstant$ if $\keyword$ is $\GDKConstant$ or $\GDKConfig$, and $\timeframeexecution$ otherwise;
-  \item applying $\annotatetyoptinitialvalue$ to $\vtargettimeframe$, $\tyoptp$, and \\
-        $\initialvalue$ in $\tenv$ yields
+  \item applying $\annotatetyoptinitialvalue$ to $\keyword$, $\vtargettimeframe$, $\tyoptp$, \\
```
Roman-Manevich, line 239: I think `\annotatetyoptinitialvalue` has only four arguments, not five. If you're adding keyword, you need to change its signature in `TypingRule.AnnotateTyOptInitialValue`

Roman-Manevich (changes requested):
Only one comment about the possible change of signature for `TypingRule.AnnotateTyOptInitialValue`

asllib/tests/regressions.t/inherit-integer-constraints-bad-config.asl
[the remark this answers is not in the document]
hrutvik: Will remove

asllib/doc/GlobalStorageDeclarations.tex
hrutvik: I believe this is already done on line 314 - unless I have missed something?

Roman-Manevich: oh, okay, I missed that

Roman-Manevich (approved):
Looks good to me!

PR #1207, opened 2025-02-27
[CI] add test-all and test-menhir2bnfc workflows

HadrienRenaud:
This PR introduces 2 workflows:
1. "Run BNFC dependent tests", from `menhir2bnfc.yml`, runs bnfc dependent tests once a week and on every PR that edits the parsing of ASLv1, on ubuntu with OCaml5.2, with bnfc and carpenter dependencies installed.
2. "Make all tests", from `make-test-all.yml`, runs `make test-all` once a week, on:
   1. OCaml4.08 on Ubuntu, without carpenter dependencies
   2. OCaml4.14 on Ubuntu, with carpenter dependencies
   3. OCaml5.2 on Ubuntu, with carpenter dependencies

Both workflows also run on demand, and on PR where their workflow file is edited.

### DIFF
--- a/.github/workflows/build-asl-reference.yml
+++ b/.github/workflows/build-asl-reference.yml
@@ -9,7 +9,7 @@ on:
     paths:
       - 'internal/bento.mll'
       - 'asllib/doc/**'
-      - '.github/workflows/**'
+      - '.github/workflows/build-asl-reference.yml'
 
 permissions: read-all
 
--- a/.github/workflows/make-test-all.yml
+++ b/.github/workflows/make-test-all.yml
@@ -1,0 +1,56 @@
+name: Make all tests
+
+on:
+  pull_request:
+    paths:
+      - ".github/workflows/make-test-all.yml"
+  # push:
+  workflow_dispatch:
+  schedule:
+    # Prime the caches every Monday
+    - cron: 0 1 * * MON
+
+permissions: read-all
+
+# Copy-pasted from https://stackoverflow.com/questions/66335225/how-to-cancel-previous-runs-in-the-pr-when-you-push-new-commitsupdate-the-curre
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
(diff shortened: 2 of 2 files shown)


maranget:
> @maranget what would be a good setting of `TIMEOUT` for `make test-all` here?

No idea, will depend on the CI machines...

IGGeorgiev:
@HadrienRenaud those paths should suffice, the lexer code relies on utilities in `Lexer.mll` exclusively, and the parser data is only relevant in mly files. So the core menhir2bnfc functionality should be covered. I think some of the parser comparison tests technically touch ASTUtil/builder, though I'm not sure if we care to include those paths?

maranget:
LGTM, we can (and will) extend `make test-all` later.

asllib/carpenter/config.ml
```diff
@@ -251,7 +251,12 @@ let default_config =
   end : S)
 
 module Parse = struct
-  module Tbl = Hashtbl.Make (String)
+  module Tbl = Hashtbl.Make (struct
+    let hash = Hashtbl.hash
+    let _remove_warning_32 = hash
+
+    include String
```
maranget, line 258: I'd had written
```suggestion
     type t = string
     let equal = String.equal
     let hash = Hashtbl.hash
```
A matter of taste...

asllib/carpenter/dune
```diff
@@ -4,6 +4,8 @@
   (:standard -w -40-42-44-48))
  (modules :standard \ main)
  (optional)
+ (enabled_if
+  (>= %{ocaml_version} 4.14))
```
maranget, line 8: I am not very keen on it, but does this mean that carpenter code is not compatible with OCaml < 4.14?

maranget (approved):

asllib/carpenter/dune
maranget: No big deal.

asllib/carpenter/config.ml
maranget: Fine, a matter of taste...

PR #1213, opened 2025-02-28
[asl] Updates for primitives and "primitive subprograms"

hrutvik:
Various updates concerning primitives and "primitive subprograms". In particular:
- Support for `real * int` and `int * real` (both returning `real`)
- Ensure that `Log2()` accepts only powers of 2
- Implement various stdlib functions in ASL itself: `UInt()`, `SInt()`, `Log2()`, `Real()`, `Round{Down,Up,TowardsZero}()`
  - **NB** `SInt()` as implemented in ASL does not cooperate with zero-width bitvectors
  - Along the way, fix a bug in `Typing.extract_parameters`
- Remove the concept of "primitive subprograms" from the ASL Reference - these become specific to ASLRef's implementation
  - Allow ASLRef's primitives to "patch" ASL functions (e.g. the new stdlib functions above)
  - Add a `--no-primitives` flag to ASLRef which avoids "patching" ASL functions, to allow testing of the un-"patched" ASL functions


asllib/libdir/stdlib.asl
```diff
@@ -4,18 +4,36 @@
 //
 //-----------------------------------------------------------------------------
 
-
 //------------------------------------------------------------------------------
-// Externals
+// Standard integer functions and procedures
```
Roman-Manevich, line 8: Perhaps `// Standard integer functions` is better as there are no procedures.

PR #1221, opened 2025-03-06
[asl] Fix a typo in a listing

hrutvik:
- Fix a typo where `20 MOD 3` actually printed the result of `-20 MOD 3`
- Modify permissions on `doclint.py` and `extended_macros.py` to ensure they are directly executable


Roman-Manevich (approved):
Looks good to me! Thanks for spotting this.

PR #1222, opened 2025-03-06
ASL-762: left-to-right evaluation

hrutvik:
- Document left-to-right evaluation order in the reference, with examples
- Remove mentions of unspecified evaluation order
- Hide "non-conflicting unions" in side-effect analysis behind a flag


Roman-Manevich (approved):
Looks good to me!

PR #1225, opened 2025-03-07
[asl] Fixes for environments in evaluation of blocks

hrutvik:
Thanks to @solswords for a bug report.

- Delete block-local variables after evaluating a block, even when an exception is thrown.
- Add a test using `--no-typecheck`. This previously passed, exposing that block-local variables were persisting across their block boundaries. Now it correctly fails with an "undefined identifier" error.

### DIFF
--- a/asllib/doc/BlockStatements.tex
+++ b/asllib/doc/BlockStatements.tex
@@ -67,26 +67,84 @@ which drops back to the original local environment of $\env$ when the evaluation
 \SemanticsRuleDef{Block}
 See \ExampleRef{Block Statements}.
 
+We first define the helper function $\poplocalscope$:
+\hypertarget{def-poplocalscope}{}
+\[
+  \begin{array}{c}
+    \poplocalscope : \overname{\dynamicenvs}{\vouterenv} \cartimes \overname{\dynamicenvs}{\vinnerenv} \rightarrow \dynamicenvs \\
+    \poplocalscope(\vouterenv, \vinnerenv) \triangleq
+      (G^{\vinnerenv}, \restrictfunc{L^{\vinnerenv}}{{\dom(L^\vouterenv)}})
+  \end{array}
+\]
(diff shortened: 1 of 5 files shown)


solswords (approved):

asllib/doc/BlockStatements.tex
```diff
@@ -71,24 +71,74 @@ \section{Semantics\label{sec:BlockStatementsSemantics}}
...
+\end{mathpar}
+
+\begin{mathpar}
+\inferrule[continuing]{
+  \env \eqname (\tenv,\denv)\\
+  \evalstmt{\env, \stm} \evalarrow \vres \\\\
+  \commonprefixline \\\\
+  \vres = \Continuing(\newg, \envcont) \\
+  \envcont \eqname (\tenvone, \denvone) \\
+  \newenv \eqdef (\tenv, (G^{\denvone}, \restrictfunc{L^{\denvone}}{{\dom(L^\denv)}}))
```
Roman-Manevich, line 124: since we're using this more than once, I'm thinking that perhaps we should define something like
`\poplocalscope(\outerenv, \innerenv) \triangleq ((G^{\innerenv}, \restrictfunc{L^{\innerenv}}{{\dom(L^\outerenv)}}) `
and a corresponding
`\Prosepoplocalscope`

asllib/tests/regressions.t/throw-local-env.asl
```diff
@@ -0,0 +1,16 @@
+type Ex of exception{};
```
Roman-Manevich, line 1: Perhaps makes this `TypingRule.Catcher.asl` and explain it in `TypingRule.Catcher`?

Roman-Manevich (approved):
Looks good to me + a few comments.

asllib/tests/regressions.t/throw-local-env.asl
hrutvik: Ah but this test is ill-typed, it only works because it's run with type-checking turned off

PR #1228, opened 2025-03-10
ASL-786, ASL-787, ASL-788: update standard library

hrutvik:
- Add `ROL` and `ROL_C`
- Add `FloorLog2`, `CeilLog2`; remove `Log2`
- Update signatures of `UInt` and `SInt`: parameter `N` now has constraints `integer{1..128}`. This also requires introducing the same constraint to `Align{Up,Down}Size`

### DIFF
--- a/asllib/libdir/stdlib.asl
+++ b/asllib/libdir/stdlib.asl
@@ -609,21 +609,55 @@ begin
   return (ASR{N}(x, shift), x[Min(shift-1, N-1)]);
 end;
 
-// Rotate right.
-// This function shifts by [shift] bits to the right, the bits deleted are
-// reinserted on the left. This makes it operate effectively modulo N.
+// ROR()
+// =====
+// Rotate right by [shift] bits. The bits are deleted on the right and
+// reinserted on the left, so the operation is effectively modulo N.
+
 func ROR{N}(x: bits(N), shift: integer) => bits(N)
 begin
   assert shift >= 0;
+  if N == 0 then return x; end;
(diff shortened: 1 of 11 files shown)


asllib/libdir/stdlib.asl
```diff
@@ -609,21 +609,53 @@ begin
...
   let cshift = (shift MOD N) as integer{0..N-1};
   return x[0+:cshift] :: x[N-1:cshift];
 end;
 
+
+// ROR_C()
+// =======
 // Rotate right with carry out.
-// As ROR, the function effectively operates modulo N.
+// As with [ROR], the operation is effectively modulo N.
```
Roman-Manevich, line 647: Can you please add a comment on the intended role of the additional bit in the return type?

herd/ASLSem.ml
```diff
@@ -773,6 +773,8 @@ module Make (C : Config) = struct
       let pow_2 = binop `POW (lit 2) in
       let t_named x = T_Named x |> with_pos in
       let side_effecting = true in
+      let uint_sint_parameter_type =
+        Asllib.ASTUtils.integer_range' (lit 1) (lit 128) |> with_pos in
```
Roman-Manevich, line 777: shouldn't `lit 128` be 2^128?

hrutvik: I believe this is reflecting the constraint on parameter `N` of `UInt`/`SInt`, so `integer{1..128}`

PR #1231, opened 2025-03-14
[herd,asl] Fix path to UDF opn

yuxiliu-arm:


maranget:
Would it be possible to add a tests, so that the problem surfaces if this path is changes again?

PR #1237, opened 2025-03-21
[tools] New `-stdlatex <bool>` option for miaou

maranget:
Due to production constraints, **miaou** may now produce tex code that latex does not accept.

Use new option `-stdlatex true` when miaou output is for direct LaTeX consumption.


artkhyzha:
Thank you, @maranget !

PR #1243, opened 2025-03-24
[aarch64, cat, catalogue] IFB relaxation + tests

diaolo01:
This PR reflects the following architectural relaxation:
If `FEAT_ExS` is not implemented or if `SCTLR_Elx.EIS` is 1, then an exception entry caused by SVC, HVC, SMC, BKPT or BRK is a context synchronising event and it causes an IFB effect.
    
Currently, only SVC is supported in herd so the cat file is updated accordingly.
    
The result of this change is that tests with eis=0, which were forbidden prior to this change, are now allowed.

This PR introduces two litmus tests, SVC-pte.exs1eis0 and SVC-pte.exs1eis1, that terminate with the following warning: "unrolling limit exceeded at L0, legal outcomes may be missing.". The tooling used to run regression tests terminates with fatal error when warnings are present. For the purpose of this PR, these tests are excluded from the regression suite, until a more robust solution is in place.


catalogue/aarch64-faults/tests/SVCctrl.litmus
```diff
@@ -0,0 +1,13 @@
+AArch64 svc-ctrl
+variant=ifetch,exs
```
artkhyzha, line 2:
```suggestion
variant=exs
```

artkhyzha (commented):
Thank you, Diana. This looks good to me, just nitpicks.

Also, could you please rename the tests so that their filenames and their names inside of the text file coincided? Also, could `SVCctrl` be renamed into `MP+dmb.st+ctrl-svc`?

The two new tests, to the best of my understanding, are allowed with `exs` and forbidden with `exs,eis`. I think we may benefit from having four files. Not sure about the naming pattern... how about `MP+dmb.st+ctrl-svc.exs1eis1.litmus` and MP+dmb.st+ctrl-svc.exs1eis0.litmus`?

relokin (commented):
Thanks @diaolo01. Other than the small change Artem requested, this PR looks good to me.

artkhyzha:
Merged -- thank you, @diaolo01.

PR #1257, opened 2025-04-11
[diy] [draft] Some code improve

ShaleXIONG:
Improve code quality in diy.

    - Rewrite code using standard library such as `List.map` and `List.fold_left`.
    - Make some explanatory comments to single letter variable and key
      functions.
    - Remove some extra white space.


gen/top_gen.ml
```diff
@@ -432,69 +443,88 @@ let max_set = IntSet.max_elt
...
+        - `i` initial value, of type `init` (defined in archExtra_gen.ml).
+          It remains unchanged in the default configuration.
+          It is only updated via `call_build_observers`.
+        - `cs` pseudo code. It is empty in the default configuration.
+          It is only updated via `call_build_observers`.
+        Element of `vxs` is `(x, vs)`.
+        - `x` is the location represented by a string
+        - `vs` the final value of the location `x` *)
+      let _p,i,cs,fs = List.fold_left (fun (p, i, cs, fs) (x, vs) ->
+        (* `vs` : (IntSet.elt array * IntSet.t) list list *)
```
fsestini: Any reason why this type comment can't be an actual type annotation? That would give the extra advantage that the compile can check that the type annotation is actually correct and doesn't go out of sync with the code. In this case, for example, it seems like the type of `vs` is actually `(Code.v array * IntSet.t) list list` rather than the type in the comment.

fsestini (approved):
I left one small question, but otherwise, looks good to me!

I can see that there are quite a few formatting-related edits in this PR (white space, indendation adjustments, etc.). I'm sure this topic came up before, but I wonder what's your stance on using an automatic formatter (like `ocamlformat`) for `gen`-related modules.

gen/top_gen.ml
ShaleXIONG: Oh that is good comment, I will change it to type annotation. My previous thought is less invasive change. However since we gain full control on `diy` I think it is worth to do it.

fsestini:
> > I left one small question, but otherwise, looks good to me!
> > I can see that there are quite a few formatting-related edits in this PR (white space, indendation adjustments, etc.). I'm sure this topic came up before, but I wonder what's your stance on using an automatic formatter (like `ocamlformat`) for `gen`-related modules.
> 
> In short, it should be introduced at least for `gen` side. I kind of pending this since there are other big change, I am planning to commit a ocamlformat for `gen` probably around new year, which I believe less pull request around. Or we might event introduce ocaml format file by file, as many file, for example, other backend than `AArch64` are barely touched for a while and I expected limit new feature for those backend.
> 
> Luc voiced he do not want a large formatting change, yet he is shifting `diy` decision to us now. I think we simply need to find a plan to make it happen.

I agree we should be careful not to disrupt existing ongoing work. Introducing it gradually per-file is a good idea. `ocamlformat` can be [configured](https://ocaml.org/p/ocamlformat/latest/doc/configuration.html) to apply to a selected list of files via `.ocamlformat-enable`, for example.

ShaleXIONG:
> > > I left one small question, but otherwise, looks good to me!
> > > I can see that there are quite a few formatting-related edits in this PR (white space, indendation adjustments, etc.). I'm sure this topic came up before, but I wonder what's your stance on using an automatic formatter (like `ocamlformat`) for `gen`-related modules.
> > 
> > 
> > In short, it should be introduced at least for `gen` side. I kind of pending this since there are other big change, I am planning to commit a ocamlformat for `gen` probably around new year, which I believe less pull request around. Or we might event introduce ocaml format file by file, as many file, for example, other backend than `AArch64` are barely touched for a while and I expected limit new feature for those backend.
> > Luc voiced he do not want a large formatting change, yet he is shifting `diy` decision to us now. I think we simply need to find a plan to make it happen.
> 
> I agree we should be careful not to disrupt existing ongoing work. Introducing it gradually per-file is a good idea. `ocamlformat` can be [configured](https://ocaml.org/p/ocamlformat/latest/doc/configuration.html) to apply to a selected list of files via `.ocamlformat-enable`, for example.

Cool ! thanks for the link. I will study the configuration.

PR #1259, opened 2025-04-17
[herd,solver] Fix topological solver

maranget:
This PR fixes two problems found while writing  PR #1258.
  1. Iterate solver steps on dependency cycles.
  2. Handle independant equations that "define" the same variable several times.


maranget:
> LGTM, we can merge whenever you want @maranget

I'll merge after having fixed a few typos in commit messages.

PR #1262, opened 2025-04-24
[gen] A few small fixes on diy generation.

ShaleXIONG:
Minor changes on diy.
- Weaken validity check for a cycle containing only rmw, this allow "Amo.Cas Rfe" and "LxSx Rfe".
- Allocate the same register for locations such as `x` and `y` for all procedures. Because of the allocation strategy change, the hash values have change and the expected CI result, `herd/tests/diycross/` has been updated.
- Fix an issue in post-condition in `-oneloc` mode, this merge will generate a check to observe the oldest value, e.g.
```AArch64 A
"PosWW Coe"
Generator=diyone7 (version 7.57+1)
Com=Co
Orig=PosWW Coe
{
0:X1=x;
}
 P0          ;
 MOV W0,#2   ;
 STR W0,[X1] ;
 MOV W2,#1   ;
 STR W2,[X1] ;
exists ([x]=2) <--------- It used to be [x] = 1
```


ShaleXIONG:
> > this allow "Amo.LxSx Rfe"
> 
> Is it typo here?
> 
> It seems that there is no such thing as `Amo.LxSx`, at least `diy{one}7` doesn't advertise that
> 
> ```
> $ diyone7 -arch AArch64 -list 2>&1 | grep  Amo
> Amo.Cas Amo.LdAdd Amo.LdClr Amo.LdEor Amo.LdSet Amo.StAdd Amo.StClr Amo.StEor Amo.StSet Amo.Swp ...
> ```
> 
> Probably `Amo.Swp` would be closest to your intent, no?

Sorry it is a typo. I mean `Amo.Cas` and also `Amo.Swp` too.

ShaleXIONG:
@relokin @artkhyzha  This is the merge request that contains the change to the allocation strategy on register. The change to the code is small, but large file change is due to the expected CI result is different now. I have manually change the generated tests are actually the same, apart from different registers allocation.

maranget:
Hi @murzinv. I'll look at this PR as soon as possible.

relokin:
hey @ShaleXIONG thanks for this!
As discussed offline, I think there are several things we need to address:
* It's very useful to be able to generate coherence tests correctly, line you example:
```
"PosWW Coe"
Generator=diyone7 (version 7.57+1)
Com=Co
Orig=PosWW Coe
{
0:X1=x;
}
 P0          ;
 MOV W0,#2   ;
 STR W0,[X1] ;
 MOV W2,#1   ;
 STR W2,[X1] ;
exists ([x]=2
```
But the cycle in this test is  `PosWW Coi` and not `PosWW Coe`

* I think it would be nice to be able to generate coherence tests with atomics and indeed your change enables that, but the cycles are incorrect. For example:
```
$> diyone7 -oneloc -arch AArch64 Amo.Swp Rfe
AArch64 A
"Amo.Swp Rfe"
Com=Rf
Orig=Amo.Swp Rfe
{
0:X0=x;
}
 P0             ;
 MOV W2,#1      ;
 SWP W2,W1,[X0] ;
exists (0:X1=0)
```
I think a forbidden test would be checking for `0:X1=1`. At its current shape the test doesn't have a cycle.

gen/archExtra_gen.ml
```diff
@@ -257,6 +260,19 @@ and type special3 = I.special3
       friends = RegMap.empty;
       next_addr = 0; }
 
+  let used_register init =
+    List.filter_map ( function
+      | (Reg (_,r),Some (S _)) -> Some r
```
relokin, line 265: IIUC, used register creates a list of registers already used in the init list. Why do you exclude registers that are initialised to values other than `S`? For example, PTE values (`AArch64PteVal.t`)

gen/archExtra_gen.ml
```diff
@@ -257,6 +260,19 @@ and type special3 = I.special3
...
+      | _ -> None ) init
+
+  let remove_reg_allocator st remove =
+    let regs = st.regs
+      |> List.filter ( fun reg ->
+          (* Keep those `reg` NOT equals to any in `remove` *)
+          not ( List.exists ( fun r -> reg = r ) remove )
```
relokin, line 272: Doesn't that work?
```suggestion
          not (List.mem reg remove)
```

relokin (commented):
Some comments on the code.

gen/genUtils.ml
```diff
@@ -38,18 +38,31 @@ module Make(Cfg:Config)(A:Arch_gen.S)
...
-           | [] ->
-               let r,st =
-                 if Extra.use_symbolic then
-                   A.symb_reg (Printf.sprintf "%s%i" loc p),st
-                 else A.alloc_reg st in
-               r,(Reg (p,r),Some (A.S loc))::init,st in
-         find_rec init
+         let exact_match = List.find_opt ( function
+           | (Reg (p0,_),Some (A.S loc0))
+             when Misc.string_eq loc0 loc && Misc.int_eq p p0 -> true
```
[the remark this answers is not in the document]
ShaleXIONG: Done

gen/genUtils.ml
```diff
@@ -38,18 +38,31 @@ module Make(Cfg:Config)(A:Arch_gen.S)
...
+           | _ -> false ) init in
+         (* Despite `st` is local for procedure, `p`, we assume `st.regs`,
+            i.e. the allocation pool, contains no `r0`.
+            This means, `r0` is reserved for location `loc0` for all procedures,
+            hence `A.alloc_reg st` will never return `r0`. Given this assumption,
+            it is safe to bind `r0` to `loc` (`loc` = `loc0`) here. *)
+         let same_reg_match = List.find_opt ( function
+           | (Reg (_,_),Some (A.S loc0))
+             when Misc.string_eq loc0 loc -> true
```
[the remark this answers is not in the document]
ShaleXIONG: Done

gen/genUtils.ml
```diff
@@ -38,18 +38,31 @@ module Make(Cfg:Config)(A:Arch_gen.S)
...
+         let exact_match = List.find_opt ( function
+           | (Reg (p0,_),Some (A.S loc0))
+             when Misc.string_eq loc0 loc && Misc.int_eq p p0 -> true
+           | _ -> false ) init in
+         (* Despite `st` is local for procedure, `p`, we assume `st.regs`,
+            i.e. the allocation pool, contains no `r0`.
+            This means, `r0` is reserved for location `loc0` for all procedures,
+            hence `A.alloc_reg st` will never return `r0`. Given this assumption,
+            it is safe to bind `r0` to `loc` (`loc` = `loc0`) here. *)
+         let same_reg_match = List.find_opt ( function
```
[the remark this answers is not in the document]
ShaleXIONG: change to `same_loc_match`.

gen/archExtra_gen.ml
```diff
@@ -73,6 +73,9 @@ module type S = sig
 
   val st0 : st
 
+  val used_register : init -> arch_reg list
+  val remove_reg_allocator : st -> arch_reg list -> st
```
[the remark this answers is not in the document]
ShaleXIONG: Done

gen/archExtra_gen.ml
ShaleXIONG: I made a reserve decision here to not involved in `PTE` value. Let me explore how to generalise to also for `PTE` value. I feel it should be simple. The idea here is those used registers will be removed from the allocation pool and other thread will just reuse the same register without much check. I need to ensure these greedy way does not lead to problem.

gen/archExtra_gen.ml
ShaleXIONG: yes it should work better, thanks.

gen/archExtra_gen.ml
ShaleXIONG: I update the PR, so it will assign the same register to the same PTE value cross process/thread as well.

ShaleXIONG:
> hey @ShaleXIONG thanks for this! As discussed offline, I think there are several things we need to address:
> 
> * It's very useful to be able to generate coherence tests correctly, line you example:
> 
> ```
> "PosWW Coe"
> Generator=diyone7 (version 7.57+1)
> Com=Co
> Orig=PosWW Coe
> {
> 0:X1=x;
> }
>  P0          ;
>  MOV W0,#2   ;
>  STR W0,[X1] ;
>  MOV W2,#1   ;
>  STR W2,[X1] ;
> exists ([x]=2
> ```
> 
> But the cycle in this test is `PosWW Coi` and not `PosWW Coe`
> 
> * I think it would be nice to be able to generate coherence tests with atomics and indeed your change enables that, but the cycles are incorrect. For example:
> 
> ```
> $> diyone7 -oneloc -arch AArch64 Amo.Swp Rfe
> AArch64 A
> "Amo.Swp Rfe"
> Generator=diyone7 (version 7.58+1)
> Com=Rf
> Orig=Amo.Swp Rfe
> {
> 0:X0=x;
> }
>  P0             ;
>  MOV W2,#1      ;
>  SWP W2,W1,[X0] ;
> exists (0:X1=0)
> ```
> 
> I think a forbidden test would be checking for `0:X1=1`. At its current shape the test doesn't have a cycle.

Regarding `PosWW Coi` vs `PosWW Coe`, I widen the accepted cycle here, which affects less on the overall functionality of `diy`. That is, both `PosWW Coi` and `PosWW Coe` are allowed and yield to the same litmus test. Due to `diy` technical assumption that a cycle must contain at least one external edges, in the case of `PosWW Coi`, we will rewrite to `PosWW Coe`.

The post condition of `Amo.Swp Rfe` has been fixed too.

gen/cycle.ml
```diff
@@ -1266,15 +1288,16 @@ let merge_changes n nss =
...
             [],[]
         end else do_rec m.next in
       if same_proc m.edge then
         m::k1,k2
       else
         [m],cons_not_nil k1 k2 in
+
+
+
+
```
relokin: Is this change intentional?

gen/archExtra_gen.ml
```diff
@@ -257,6 +265,19 @@ and type special3 = I.special3
...
+      | (Reg (_,r),Some _) -> Some r
+      | _ -> None ) init
+
+  let remove_reg_allocator st remove =
+    let regs = st.regs
+      |> List.filter ( fun reg ->
+          (* Keep those `reg` NOT equals to any in `remove` *)
+          not ( List.mem reg remove )
+        ) in
+    { st with regs }
```
relokin:
```suggestion
    (* Keep those `reg` NOT equals to any in `remove` *)
    let regs = List.filter (fun reg -> not (List.mem reg remove)) st.regs in
    { st with regs }
```

gen/cycle.ml
```diff
@@ -980,17 +982,17 @@ let set_same_loc st n0 =
       with
       | Not_found ->
         (*check if node is preceded by a non com/rmw node and is itself a com/rmw node*)
```
relokin: This comment is now stale, can you please fix this?

gen/cycle.ml
```diff
@@ -980,17 +982,17 @@ let set_same_loc st n0 =
...
-        let to_com_rmw n0 = not (is_com_rmw n0.prev) && is_com_rmw n0 in
+        let to_com n0 = not (E.is_com n0.prev.edge) && E.is_com n0.edge in
         fold (fun n0 _ -> if E.is_id n0.edge.E.edge then assert false) n ();
         try
           (* check for R ensures that we start on Fr or Rmw if possible*)
```
relokin: This comment is stale as well.

gen/cycle.ml
```diff
@@ -1106,8 +1108,12 @@ let finish n =
 (* Set locations *)
   let sd,n =
     let no =
-      try Some (find_edge_prev diff_loc (find_edge_prev diff_proc n))
-      with Not_found -> None in
+      try begin
+        let start_n =
+          try find_edge_prev diff_proc n
```
relokin, line 1118: Can you please motivate this change. I don't see why it helps you with these single threaded tests. Before and after the change `no` is `None`.

gen/cycle.ml
```diff
@@ -1165,6 +1171,32 @@ let extract_edges n =
...
+  if
+    diff_proc p.edge
+  then p.next
+  else
+    try begin
+      let ext_n = find_edge diff_proc n in
+      try find_edge same_proc ext_n
+      with Not_found -> ext_n
+    (* in the case of all internal edges,
+       we convert one to external to form a cycle *)
```
relokin: This comment applies to the code that starts in line 1191 can you please move it there? Also, we don't really convert the node we just return one with an internal edge.

gen/genUtils.ml
```diff
@@ -61,7 +73,24 @@ module Make(Cfg:Config)(A:Arch_gen.S)
...
+         | Some (Reg (_,r),Some _), _ -> r,init,st
+         | None, Some (Reg (_,r),Some _) ->
+           r,(Reg (p,r),Some k)::init,st
+         | None, None ->
+           (* no previous register assignment, so add new *)
+           let r,st = A.alloc_reg st in
+           r,(Reg (p,r),Some k)::init,st
+         | _,_ -> Warn.user_error "Unexpected error in `next_const."
+
+(*
```
relokin: Can we remove find_rec?

gen/normaliser.ml
```diff
@@ -225,6 +230,11 @@ module Make : functor (C:Config) -> functor (E:Edge.S) ->
       let find_non_pseudo_prev n = find_node_rev (fun n -> E.is_non_pseudo n.edge.E.edge) n
 
       let find_start_proc n =
+        let ext_count =
+          map ( fun n -> if ext_com n.edge then 1 else 0 ) n
+          |> List.fold_left (+) 0 in
+        (* Reject cycle with precisely one external edge *)
+        if ext_count = 1 then raise(CannotNormalise "only one external edge");
```
relokin, line 237: Do we really need to check twice (here and in `cycle.ml`) if the number of external com edges is 1?

gen/cycle.ml
ShaleXIONG: it should be removed. Done.

gen/archExtra_gen.ml
ShaleXIONG: Done.

gen/cycle.ml
ShaleXIONG: Done.

gen/cycle.ml
ShaleXIONG: I add a comment on top. In general this change allows a single thread test contains only internal edges but different locations like `PodRW Rfi PodRW Rfi`.

gen/cycle.ml
ShaleXIONG: Done.

gen/genUtils.ml
ShaleXIONG: Done.

gen/normaliser.ml
ShaleXIONG: yes we have to do it twice. This is a bit unfortunate as `split_procs` function, and anything functions it calls, in `normaliser.ml` and `cycle.ml`, must maintain the same behaviours; there is an existing comment suggesting that. The are effectively doing the same thing however operating on different data structures.

gen/cycle.ml
ShaleXIONG: Done

gen/cycle.ml
relokin: You mean that without this change you can't generate the test that corresponds to the relax list `PodRW Rfi PodRW Rfi`. Are you sure? If I am not messing up something locally, I can generate this test without this diff.

ShaleXIONG: I assume you generate with `-oneloc` that is `diyone7 -arch AArch64 -oneloc PodRW Rfi PodRW Rfi` and the result is 
```
AArch64 A
"PodRW Rfi PodRW Rfi"
Com=Rf
Orig=PodRW Rfi PodRW Rfi
{
0:X0=x;
}
 P0          ;
 LDR W1,[X0] ;
 MOV W2,#2   ;
 STR W2,[X0] ;
 LDR W3,[X0] ;
 MOV W4,#1   ;
 STR W4,[X0] ;
exists ([x]=1 /\ 0:X1=1 /\ 0:X3=2)
```
This litmus test is incorrect, as there should be two locations here. and we should not need to use `-oneloc` and still generate a test. This changes basically allows such: `diyone7 -arch AArch64  PodRW Rfi PodRW Rfi`, and the result is
```
AArch64 A
"PodRW Rfi PodRW Rfi"
Com=Rf
Orig=PodRW Rfi PodRW Rfi
{
0:X0=y; 0:X3=x;
}
 P0          ;
 LDR W1,[X0] ;
 MOV W2,#1   ;
 STR W2,[X3] ;
 LDR W4,[X3] ;
 MOV W5,#1   ;
 STR W5,[X0] ;
exists (0:X1=1 /\ 0:X4=1)
```

relokin: I see, I missed that point thanks Shale!

PR #1265, opened 2025-04-25
[herd] Complete tracking of the NZCV "register"

maranget:
This PR fixes a bug found by @murzinv while reviewing PR #1258.

Namely, herd crashes on the following test:
```
AArch64 NZCV-03
Variant=sve
{
uint64_t 0:X3;
0:X0=0; 0:X1=1;
}
   P0               ;
 CMP W0,W1          ;
 WHILELT P0.S,W0,W1 ;
 MRS X3,NZCV        ;

forall 0:X3=0xa0000000
```

The root cause of the crash is a discrepancy between the interpretation time  and solver time values of the `NZCV` flags. More precisely, the `MRS X3,NZCV ` uses the interpretation time value; while the `WHILELT` instruction does not signal that it destroys the previous values of the flags.

Instructions that write into the NZCV "register" must signal it as follows:
  1. Include the `NZCV` into the kill list of the instruction, (function `killed` in AArch64Arch_herd.ml`)
  2. Include the new binding as a result of instruction interpretation.

Notice that 1. or 2. alone are enough, 1. is simpler; while 2: leads to potential optimisations while interpreting the instructions that follow. Having both enhances code robustness, especially in the presence of alternative means of instruction interpretation such as ASL.

A quick way to implement 1. above is to kill all registers as we do for all morello instructions.

As far as we have checked, instructions that write the NZCV flags are: Some operations, namely ADDS, SUBS, ANDS, BICS, the SVE WHILE.. instructions, the Morello CSEAL instruction. and the MSR NZCV,... instruction.

### DIFF
--- a/herd/AArch64Arch_herd.ml
+++ b/herd/AArch64Arch_herd.ml
@@ -327,7 +327,7 @@ module Make (C:Arch_herd.Config)(V:Value.AArch64) =
           -> None
 
     let all_regs =
-      all_gprs@vregs (* Should be enough, only those are tracked *)
+      nzcv_regs@all_gprs@vregs (* Should be enough, only those are tracked *)
 
     let all_streaming_regs =
       zregs@pregs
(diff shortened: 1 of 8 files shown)


maranget:
> Hi @maranget,
> 
> That looks very good to me.
> 

Thanks

> 2 follow-up questions:
> 
> 1. Can we merge this quickly, so that I can rebase [[herd,aarch64] Split NZCV into 4 1bit registers #1258](https://github.com/herd/herdtools7/pull/1258) and we can merge it too?

I'll merge this PR soon. PR #1258 will probably need to be adapted a bit.

> 2. Probably long-term, at least for another PR, is there a way we could systematically test that the correct kill registers are used? Could we generate automatically the set of killed registers?

I do not see how to do better that specific tests that basically create the situation where interpretation time and solving time resolution of the rf-reg relation compete. See for instance `NZCV-03` above, test [V31](
https://github.com/herd/herdtools7/blob/ec472ed58d0a0ab5da14e3f9522dc2f130b7b297/herd/tests/instructions/AArch64.sve/V31.litmus) in this PR.

PR #1270, opened 2025-04-28
[gen] Bye-bye Joker

murzinv:
J (joker) relaxation has been disabled for a while after 0b068a7e84ef ("gen: Avoid jokers (J) in generated edges, could be commanded by command line switch or arch, if needed."). #648 attempted to re-enable it yet it turned out it does not bring any benefits in current form. We might revisit it in future if we find legitimate use case, but so far it is dead code and seems it is time to say goodbye to joker.


relokin (approved):
👋  👋 

@murzinv feel free to rebase and merge when you are ready.

PR #1272, opened 2025-04-29
[test] Disable looping test

maranget:
Disable some (catalogue) test. This test contains an (unresolved) loop resulting in a warning message on standard error. As a consequence `make mte-test` may fail on fast machines.

At the moment, tests from the catalogue that output onto standard error trigger a failure of the test
driver `internal/herd_catalogue_regression_test.ml`. This can be seen as a shortcomming of this driver.

By lack of time, simply disable the problematic test.

Hi @artkhyzha. Another solution to this problem may be to rewrite the test `catalogue/aarch64-MTE/tests/to-tagcheck-async/MP+dmb.st+acqrel_and_ctrl--async.litmus` so that it does not loop anymore. As the author of the test, you may be iin the best position to consider this alternative.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -517,6 +517,7 @@ mte-test:
 		-libdir-path ./herd/libdir \
 		-kinds-path catalogue/aarch64-MTE/tests/kinds.txt \
 		-shelf-path catalogue/aarch64-MTE/shelf.py \
+		-conf-path catalogue/aarch64-MTE/cfgs/test-MTE.cfg \
 		$(REGRESSION_TEST_MODE)
 	@ echo "herd7 catalogue aarch64-MTE tests: OK"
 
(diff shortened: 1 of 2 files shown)


maranget:
I fact I'd favor the solution of adapting the test by avoiding the backward jump. Namely, it is not obvious to compare stderr output to some reference, as the tests  in a catalogue series are (usually) executed concurrently. As a result, stderr outputs are concatenated in some arbitrary order.

artkhyzha:
Thank you, @maranget. On the one hand, happy for the test to be changed. On the other, I see that this test came up in a discussion with architects, so my hunch is not to change it, but instead replacing it in regressions with a rewrite while keeping the original.

I've drafted a batch of updates to tests (#1293). There appears to be another issue, but I will only have a chance to investigate further next week, so sharing this for visibility more than anything.

artkhyzha:
Thank you, @HadrienRenaud  and @maranget, for the troubleshooting.

@maranget, I have added your fix as a commit for this PR -- would this be okay with you?

@relokin and @murzinv, I just wanted to check with your if this reorganisation of the catalogue for MTE looks good to you. Essentially, I am using this window of opportunity to update some MTE tests to remove all the warnings we get upon `make mte-test`.

maranget:
Superseded by PR #1293

PR #1275, opened 2025-04-29
[asl] Forbid storage declarations that discard all names

hrutvik:
Forbid storage declarations that discard all names, both locally and globally - e.g.:
```
var - = ...;
let (-, -, -) = ...;
```
However, still permit discarding some names but not others:
```
let (-, x, -) = ...;
```
Guard this change behind a `--allow-storage-discards` flag for backwards compatibility.

### DIFF
--- a/asllib/doc/LocalStorageDeclarations.tex
+++ b/asllib/doc/LocalStorageDeclarations.tex
@@ -88,6 +88,11 @@ Declaring a local storage element is done via the following grammar rules:
   |\ & \LDITuple(\identifier^*) &
 \end{flalign*}
 
+\RequirementDef{DiscardingLocalStorageDeclarations}
+Local storage declarations must bind at least one name.
+All the local storage declarations in \listingref{local-storage-discards} are illegal, as they discard all declared storage elements.
+\ASLListing{Illegal local storage declarations that discard all storage elements}{local-storage-discards}{\syntaxtests/GuideRule.DiscardingLocalStorageDeclarations.asl}
+
 \ASTRuleDef{LocalDeclKeyword}
 \hypertarget{build-localdeclkeyword}{}
 The function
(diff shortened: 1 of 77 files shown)


Roman-Manevich:
Please update the ALP3.1 changelog.

asllib/doc/LocalStorageDeclarations.tex
```diff
@@ -118,13 +119,16 @@ \section{Abstract Syntax\label{sec:LocalStorageDeclarationsAbstractSyntax}}
...
-  \overname{\astof{\vignoredoridentifier}}{\vastnode}
+  \builddeclitem(\Ndeclitem(\Tidentifier(\name))) \astarrow
+  \overname{\name}{\vastnode}
 }
 \end{mathpar}
 
 \begin{mathpar}
 \inferrule[tuple]{
+  i \in \listrange(\vids): \vb_i \eqdef \vids_i = \Tminus \\
+  \vb \eqdef \bigwedge_i \vb_i \\
```
Roman-Manevich, line 130: nit:  you can inline the individual check like so `\vb \eqdef \bigwedge_i \vids_i = \Tminus`

Roman-Manevich (commented):
Perhaps this change deserves a Guide rule, or maybe two - one for local declarations and one for global declarations.

hrutvik:
I've added a guide rule about local storage declarations, as these incur a separate check. I have not added one for global storage declarations, as these are "by construction" (discarding global storage declarations are now nonsensical according to the ASL grammar, so no extra check required). Let me know if you feel strongly that one should be included.

As discussed separately, I will batch additions to the changelog in a separate PR if this is OK.

Roman-Manevich:
Batching the PRs for the changelog is fine.
For the global storage declarations: I see your point, but consider that not every reader of the reference will be well versed with the grammar details so perhaps it is a good idea to do the same as for the local declarations.

hrutvik:
Makes sense - added that in now!

Roman-Manevich (approved):
Great, thanks!

PR #1276, opened 2025-04-29
[asl] Base values for bit vectors

hrutvik:
Relax restrictions on base values for bit vectors, instead producing the following:
```
var x: bits(e);
-- base value -->
var x: bits(e) = 0[:e];
```


Roman-Manevich (approved):
The reference update looks good to me!

PR #1279, opened 2025-04-29
[asl] Forbid expression-level `elsif`

hrutvik:
Remove support for the `elsif` keyword at the expression level. Guard this change behind a `--allow-expression-elsif` flag for backwards compatibility.


Roman-Manevich (approved):
LGTM!

PR #1283, opened 2025-04-30
[asl] Add `IsAligned{Size,P2}` to stdlib

hrutvik:
For both integer- and bit vector-flavoured versions.


asllib/libdir/stdlib.asl
```diff
@@ -542,6 +564,17 @@ begin
     return AlignUpSize(UInt(x), size)[:N];
 end;
 
+// IsAlignedSize()
+// ===============
+// Return TRUE if when viewed as an unsigned integer, bit vector x is aligned
```
Roman-Manevich, line 569: bit: "bit vector" -> bitvector.

asllib/libdir/stdlib.asl
```diff
@@ -542,6 +564,17 @@ begin
...
+// to a multiple of size (not necessarily a power of 2).
+// Otherwise, return FALSE.
+
+func IsAlignedSize{N: integer{1..128}}(x: bits(N), size: integer {1..2^N}) => boolean
```
Roman-Manevich, line 573: Do we really need `size` to be constrained?
In case, we do, shouldn't it be `integer{0..2^N-1}`?

asllib/tests/stdlib.t/align.asl
```diff
@@ -25,6 +25,19 @@ begin
   assert AlignUpP2('001000', 5) == '100000';
   assert AlignUpP2('001000', 6) == '000000';
 
+  assert !IsAlignedP2('110111', 1);
```
Roman-Manevich, line 28: Please add a test for an empty bitvector and for size==0.
Also for `IsAlignedSize`, perhaps one simple case and for edge cases x==0 and negative x, if we decide we need to support them.
Same for `IsAlignedP2`

Roman-Manevich (commented):
Left some comments.

asllib/libdir/stdlib.asl
```diff
@@ -161,6 +161,16 @@ begin
     return AlignDownSize(x + (size - 1), size);
 end;
 
+// IsAlignedSize()
+// ===============
+// Return TRUE if integer x is aligned to a multiple of size (not necessarily a
+// power of 2). Otherwise, return FALSE.
+
+func IsAlignedSize(x: integer, size: integer) => boolean
+begin
```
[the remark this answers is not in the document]
hrutvik: Confirmed with @nigelito and added

asllib/tests/stdlib.t/align.asl
hrutvik: Added tests for empty bitvector and `size == 0`, and simple cases for `IsAligned{Size,P2}`. No edge cases for `x == 0` and `x < 0` as we will not support these.

Roman-Manevich: then perhaps add `assert x > 0`?

PR #1287, opened 2025-05-01
[herd] Semantics for address translation instruction with system registers as a new constant

diaolo01:
This PR introduces the semantics of AT instructions and enables the PAR_EL1 register to be represented as a multi-field structure, similar to PTEs.
 
AT is the Address Translating instruction. It takes two arguments, the translation regime and virtual address to be translated. It performs translation and returns the output address to the PAR_EL1 system register. If there is a fault, F bit is set to 1 and the bit fields of PAR_EL1 change, such that there isn't an output address field, and FSC (fault status code) is captured instead.
 
The AT translation regimes implemented exclude exception levels, effects of PAN, and effects of Address Translation operations that ignore stage 1 permissions (FEAT_ATS1A). Stage 2 translation regimes are also out of scope, as herd does not support stage 2 translation.
 
The new constant type, AddrReg is used to embed PAR_EL1. The intention is that the general structure would be able used for more that just PAR_EL1, for registers such as TTBR_Elx, VTTBR_Elx, PFAR_EL1, but potentially could be used as backbone for GPTs and DPTs (SMMU).
 
The current syntax for litmus tests of the PAR_EL1 register is parel1_t:(oa:PA(x),f:0). Note that the field name associated with oa is PAR_EL1.pa.
 
To do:
* The patch contains AddrReg as a generalised concept, but the code is designed only to work for PAR_EL1.
* As part of this work, it has been uncovered that litmus tests with asynchronous exceptions and fault handlers still execute the fault handler, which is not architecturally correct. This impacts both this PR and MTE.
* Fault reporting mechanism currently doesn't set the f bit to 1. Do we want the oa field to return something meaningful, an agreed upon error code, if there is a fault?
* Litmus7 side of work is not complete.
* A comprehensive list of tests needs to be put together.


herd/tests/instructions/AArch64.kvm/A023.litmus
```diff
@@ -0,0 +1,13 @@
+AArch64 A023
+Variant=vmsa
+(*this needs to show that the test is asynchronous; WIP*)
```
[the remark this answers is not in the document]
diaolo01: I was dissatisfied that an asynchronous fault would be captured the same as a synchronous one. There has to be a distinction, and I would be happy with your proposal `0:PAR_EL1=(fault:"MMU:Permission")`.
No, PAR_EL1 does not contain the faulting address.

I will capture this decision on the PR message and investigate when I get to the litmus side of things.

lib/AArch64AddrReg.ml
```diff
@@ -0,0 +1,155 @@
...
+  pa : OutputAddress.t;
+  f : int;
+  }
+
+(* Let us abstract... *)
+let is_f {f; _} = f <> 0
+
+and same_pa {pa=pa1; _} {pa=pa2; _} = OutputAddress.eq pa1 pa2
+
+(* For ordinary tests not to fault, the dirty bit has to be set. *)
```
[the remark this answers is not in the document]
diaolo01: I tried to the best of my abilities to remove all the files that are currently not being used. It wasn't clear to me whether keeping the option for multiple fields in AddrReg was desirable. This hinges on the direction of how abstract should AddrReg be.

lib/AArch64AddrReg.mli
```diff
@@ -0,0 +1,33 @@
...
+
+val eq : t -> t -> bool
+val compare : t -> t -> int
+
+val default : int -> t
+
+val is_f : t -> bool
+val same_oa : t -> t -> bool
+
+val prot_default : t
```
artkhyzha, line 27: Just a minor question for my own education: what does this value mean?

lib/dune
```diff
@@ -1,7 +1,28 @@
-(ocamllex AArch64Lexer ARMLexer BPFLexer BellLexer CLexer lexHint
-  lexOutMapping lexRename lexSplit lexUtils MIPSLexer modelLexer PPCLexer
-  RISCVLexer scopeLexer splitter stateLexer X86Lexer X86_64Lexer echo
-  dirtyBit cacheType lexScan JavaLexer)
+(ocamllex
```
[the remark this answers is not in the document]
artkhyzha: My mild suggestion is not to change the file unless the change is needed.

lib/AArch64AddrReg.mli
diaolo01: It was an artefact from AArch64PteVal. In that context it is used for defining a PTE with its properties set to a default state, without setting the output address.
That is not relevant for AddrReg, so I have removed it. Thank you.

herd/AArch64Sem.ml
```diff
@@ -1381,12 +1381,21 @@ module Make
             fun (r,_) -> M.unitT r
         else m
 
-      let lift_kvm dir updatedb mop ma an ii mphy =
-        let lbl_v = get_instr_label ii in
+      let lift_kvm dir updatedb mop ma an ii mphy instr_at =
```
artkhyzha: A minor nit. I am curious if this `instr_at` is avoidable. Note how `ii` here has the type of `inst_instance_id`, which is a record including a field `inst : instr`. Won't that include the information you need? Using something like:
```
        let instr_at = let open AArch64Base in
          match ii.A.inst with
          | I_AT _ -> true
          | _ -> false in
...
```
At the beginning of `lift_kvm`? If that works, it may simplify calls to this function as you won't have that extra boolean argument.

lib/AArch64AddrReg.ml
```diff
@@ -0,0 +1,85 @@
...
+        ((if old_oa then OutputAddress.pp_old
+          else OutputAddress.pp) p.oa)
+    else "parel1_t:(f:1)" in
+  sprintf "(%s)" field
+
+(* By default pp does not list fields whose value is default *)
+let pp hexa = do_pp hexa false
+let pp_v = pp false
+
+let my_int_of_string s v =
```
artkhyzha, line 43: Just a comment, no action point. I tripped over there being a function called `my_*`, but then I realised it's not in the module interface and is probably "inherited" from AArch64PteVal.

litmus/AArch64Compile_litmus.ml
```diff
@@ -1951,6 +1951,12 @@ module Make(V:Constant.S)(C:Config) =
         cache (sprintf "dc %s" (Misc.lowercase (DC.pp_op op))) r::k
     | I_TLBI (op,r) ->
         tlbi op r::k
+    | I_AT (op,r) ->
+      let op = Misc.lowercase (AT.pp_op op) in 
+        { empty_ins with
+        memo = sprintf "at %s,^i0" op; inputs=[r];
+        reg_env=add_p [r]; } (*TODO: not convinced we need reg_env here. what does it do?*)
```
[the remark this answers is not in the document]
artkhyzha: (Marking "unresolved" just because of the "TODO" comment and in case if anything needs doing.)

herd/AArch64Sem.ml
diaolo01: Thanks, @artkhyzha, very elegant solution! It worked well.

herd/tests/instructions/AArch64.kvm/A023.litmus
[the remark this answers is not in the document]
diaolo01: This is no longer relevant.

litmus/AArch64Compile_litmus.ml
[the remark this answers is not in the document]
diaolo01: Thanks for catching it, I've fixed now!

lib/AArch64AddrReg.ml
diaolo01: That is right.

artkhyzha:
In the PR description:

> Limitation: Currently, the test below is not running on hardware:
> ...
> Potential fix: perform AT as part of the initialisation sequence or adjust the comparison to mask the real PAR_EL1.

This makes sense. I am not yet sure what litmus test would require initialising a register with a parel1_t value; perhaps, this functionality could be implemented as a part of another PR unless we come up with a motivating example.

That said, I wonder if we need some sort of a check, so that unsupported values did not appear in litmus tests. (No suggestions for the time being.)

artkhyzha (approved):
Thank you, @diaolo01. I understand that after recent updates, litmus tests with unsupported values will result in an error message, which is helpful and I think was the only worry I've had.

LGTM. I'd say this is ready for merging. @murzinv, @relokin - would do you think?

herd/AArch64Sem.ml
```diff
@@ -1382,11 +1382,21 @@ module Make
         else m
 
       let lift_kvm dir updatedb mop ma an ii mphy =
-        let lbl_v = get_instr_label ii in
         let mfault ma a ft =
-          insert_commit_to_fault ma
-            (fun _ -> set_elr_el1 lbl_v ii >>| mk_fault (Some a) dir an ii ft None)
-            None ii >>! B.fault [AArch64Base.elr_el1, lbl_v] in
+          let lbl_v = get_instr_label ii in
```
relokin: Rather than adding more complexity to lift_kvm can we modify it to accept an mfault argument? That way the function implementing at can provide an mfault path that makes sense.

diaolo01: Done.

artkhyzha:
To @murzinv's point, I agree that an `ISB` would be expected:

- It should be expected in the litmus-test body, but I would expect that to be mandated by the memory model rather than the herd7's generation of candidate executions. Do we have a precedent for devising special semantics for accesses to specific system registers? If not, it could be okay to defer this until the cat model considers AT, and in the meantime avoid reading PAR_EL1 directly in litmus tests.
- When PAR_EL1 is not read directly in the body of the test, but is read in the postcondition, indeed we could assume that there is an ISB before the postcondition is validated. However, this is an assumption that `litmus7` then must implement. Is it implemented already?

(While specific for this PR, this question may apply to other cases like TFSR_ELx in async MTE.)

artkhyzha:
Merged -- thank you, @diaolo01 and @murzinv.

PR #1289, opened 2025-05-01
Fix more MTE tests

murzinv:
Some MTE tests in catalogue are mixing fault expression and custom fault handler which may confuse litmus7. Also it has been discovered that some MTE sync tests are in async's clothing. Let's fix them all.


relokin (approved):
LGTM but I would like @artkhyzha to have a look as well.

artkhyzha (approved):
LGTM, thanks, @murzinv!

PR #1292, opened 2025-05-02
fix(lib): wait for process before calling channel functions

yuxiliu-arm:
Running `make test` directly will result in the following failure:

```
Fatal error: exception Failure("Herd returned stderr: Warning: File \"catalogue/aarch64-MTE/tests/to-tagcheck-async/MP+dmb.st+acqrel_and_ctrl--async.litmus\": unrolling limit exceeded at L0, legal outcomes may be missing.")
```

`make test DUNE_PROFILE=dev` on the other hand succeeds (as is used in the CI).

I suspect the reason is the `dev` profile doesn't wait for the spawned `herd` process to finish before calling the function to collect the `stderr` result.

This commit moves the function calls to after the process is finished or an exception is raised.

### DIFF
--- a/internal/lib/command.ml
+++ b/internal/lib/command.ml
@@ -93,14 +93,14 @@ let do_run must_succeed ?stdin:in_f ?stdout:out_f ?stderr:err_f bin args =
         bin (String.concat " " args) in
   let pid =
     Fun.protect
-      ~finally:(fun _ -> close_in_pipe () ; close_out_pipe () ; close_err_pipe ())
-      (fun _ ->
-        let pid = Unix.create_process bin (Array.of_list (bin :: args)) in_fd out_fd err_fd in
+      ~finally:(fun _ ->
         in_f () ;
+        close_in_pipe () ;
         out_f () ;
+        close_out_pipe () ;
         err_f () ;
-        pid
-      )
+        close_err_pipe ())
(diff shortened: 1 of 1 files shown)


maranget:
Hi @yuxiliu-arm. I had a different point of view on this problem : the test driver for catalogue tests considers that a non-empty stderr is a failure. As a temporary patch,  PR #1272 disables the faulty test.

Your solutions looks interesting. However having the test to output on stderr could perhaps be handled, either with some reference file (as for ordinary regression tests) or by rewriting the test without a backward jump.

maranget:
My analysis of the problem is as follows: the problem usually does not occur because the test `catalogue/aarch64-MTE/tests/to-tagcheck-async/MP+dmb.st+acqrel_and_ctrl--async.litmus` times out. However, by increasing the timeout limit, or by running the tests on a fast machine, the failures occurs:
```
% make mte-test PROFILE=dev TIMEOUT=60

Fatal error: exception Failure("Herd returned stderr: Warning: File \"catalogue/aarch64-MTE/tests/to-tagcheck-async/MP+dmb.st+acqrel_and_ctrl--async.litmus\": unrolling limit exceeded at L0, legal outcomes may be missing.")
make: *** [Makefile:513: mte-test] Error 2
```

maranget:
Superseded  by PR #1293.

PR #1293, opened 2025-05-02
[AArch64,catalogue] Rewritten tests to avoid loops

artkhyzha:
A draft of a follow-up to #1272 .

The main reason to keep this "draft" is that currently `make test` fails with a mysterious error message:
```
...
Fatal error: exception Base.Fun.Finally_raised(_)
make: *** [Makefile:513: mte-test] Error 2
```


maranget:
Notice that merging this PR will void PR #1272 and #1292.

maranget:
> I may have found the problem. Closing the output end of the stdin pipe may fail raising `Sys_blocked_io` , maybe because the reading counterpart (forked process) is not ready. The following fix that delays the actual closing to a further loop iteration apparently works:

Having looked at the code of OCaml standard library, I understand how `close_out` can raise `Sys_blocked_io`, which is not unexpected. Namely, `close_out` attempts flushing the channel buffer before actually closing the underlying file descriptor. Flushing means writing onto this file descriptor, an attempt that can fail with error code `WOULDBLOCK` on a non-blocking descriptor. In that case OCaml system code raises `Sys_blocked_io`.

maranget:
> @maranget, I have added your fix as a commit for this PR -- would this be okay with you?
> 

This is fine by me.

maranget:
Hi @artkhyzha. I look as everybody is ok with this PR. Would you rebase and merge. Then we can close PR #1272 and #1292.

relokin (approved):
Thanks Artem, should we merge this?

PR #1294, opened 2025-05-02
[asl] Update signatures of `UInt` and `SInt`

hrutvik:
The new signatures are:
```
func UInt{N} (x: bits(N)) => integer{0..2^N-1}
func SInt{N} (x: bits(N))
  => integer{(if N == 0 then 0 else -(2^(N-1))) .. (if N == 0 then 0 else 2^(N-1)-1)}
```
As a result, the constraints on the parameters `N` of `Align{Down,Up}Size` and `IsAlignedSize` are also removed.


Roman-Manevich (approved):
LGTM!

PR #1295, opened 2025-05-02
[asl] Update syntax for empty records and elided parameters

hrutvik:
- For empty records: rather than empty braces (`{}`), add a hyphen (`{-}`)
- For elided parameters: allow lone empty braces, e.g.
  `let x: bits(64) = Zeros{};`
  This is possible because now this form does not clash with empty records.


hrutvik:
As discussed offline, it's not clear how to support the feature flag without trying to significantly refactor the grammar

PR #1301, opened 2025-05-08
[asl] Fix under approximation bug with empty constraints

HadrienRenaud:


Roman-Manevich (approved):
LGTM!
Agreed that updating the reference will be done in a following PR.

PR #1310, opened 2025-05-20
Cat2table

TiberiuBucur:
This tool takes 2 effects e1 and e2 as command line parameters and outputs all the relations and chains that can order e1 and e2 in the aarch64 memory model. It starts from a top level relation which can be specified ('ob' by default) and recurses into all of its clauses. A clause is a solution if it can be traversed from left to right starting with e1, such that the last effect in the chain has a non-empty intersection with e2 (eg. if e2=Exp & R, we want to check that the last effect is "some kind of Exp & R").


lib/modelParser.mly
```diff
@@ -84,6 +86,9 @@ main:
   ModelOption.set_arch a
     (ModelOption.set_catdep catdep ModelOption.default), id, $3 }
 
+base_start:
```
relokin, line 89: Naive question, why can you start from main?

tools/cat2table.ml
```diff
@@ -0,0 +1,883 @@
...
+    | e1, e2 when is_empty e1 && is_empty e2 -> true
+    | Konst (_, (Universe _)), Konst (_, (Universe _)) -> true
+    | Var (_, id1), Var (_, id2) -> id1 = id2
+    | Op (_, op1, _), Op (_, op2, _) -> do_op op1 op2 e1 e2
+    | Op1 (_, op1, e1), Op1 (_, op2, e2) -> do_op op1 op2 e1 e2
+    | _, _ -> false
+
+  let barriers = ["DMB.ISH"; "DMB.ISHLD"; "DMB.ISHST"; "DMB.LD"; "DMB.OSH";
+    "DMB.OSHLD"; "DMB.OSHST"; "DMB.ST"; "DMB.SY"; "DSB.ISH"; "DSB.ISHLD";
+    "DSB.ISHST"; "DSB.LD"; "DSB.OSH"; "DSB.OSHST"; "DSB.ST"; "DSB.SY"; "ISB"]
```
relokin, line 161: can you at least get this list via lib/AArch64Base.ml?

tools/cat2table.ml
```diff
@@ -0,0 +1,883 @@
...
+    ("SupervisorCall", ["EXC-ENTRY"; "FAULT"]);
+    ("T", ["M"]);
+    ("TagCheck", ["FAULT"]);
+    ("Translation", ["MMU"; "EXC-ENTRY"; "FAULT"]);
+    ("TLBIIS", ["TLBI"]);
+    ("TLBInXS", ["TLBI"]);
+    ("UndefinedInstruction", ["EXC-ENTRY"; "FAULT"]);
+    ("W", ["M"]);
+  ] @ List.map (fun b -> b, ["F"]) barriers
+    @ List.map (fun e -> e, ["PTE"; "M"; "PTEMemAttr"]) pte_accesses
```
relokin, line 202: I think I understand why you need this but it would be really good if there was a way to get it programmatically rather than having it hardcoded.

tools/cat2table.ml
```diff
@@ -0,0 +1,883 @@
...
+
+  let intersects = [
```
relokin, line 204: ditto.

relokin (commented):
Some quick comments.

tools/cat2table.ml
relokin: I understand, we might have to do some refactoring elsewhere, but as it stands, it's very difficult to ensure this correct/complete and it will be impossible to keep it in sync with changes in herd7.

tools/cat2table.ml
```diff
@@ -0,0 +1,104 @@
...
+]
+
+let args = ref []
+let get_cmd_arg s = args := s :: !args
+
+let () =
+  try
+    Arg.parse options
+      get_cmd_arg
+      (Printf.sprintf "Usage %s [options] [cat_file], output all chains of \
```
murzinv: Just a suggestion. Can we manage cat file the same way as `herd7`? In other words via `-cat` option, which defaults to `aarch64.cat` for `AArch64` when omitted (for other arches it can be `unsupported` or like that).  IIRC, `herd7` infers arch from litmus test, so for your tool we might want to introduce `-arch` option (similar to `diy7`/`diyone7`), again it is not required to support arches other than `AArch64`  we can just fail those fast.

TiberiuBucur: I addressed this. The tool now takes a `-model` optional parameter to pass the cat file. If omitted, the tool defaults to `herd/libdir/aarch64.cat`.

tools/cat2table.ml
TiberiuBucur: I did the refactoring we discussed. We now have an `aarch64axioms.cat` file, which encodes all of these assumptions. `cat2table` parses this file and transforms it into the data structure we used to have before. The rest of the tool is unchanged.

lib/modelParser.mly
TiberiuBucur: I needed a parser to transform effects passed as strings (from the command line) into AST expressions. `base` is the required entry point for that, as `main` returns a full AST, not an expression.

tools/libdir/aarch64_wrapped_primitives.cat
ShaleXIONG, line 1: Can you put small explanation maybe at the top of this file on why do you want to do this?

Also we need a licensing information on this file too. You can copy paste from other similar file if necessary.

internal/herd_axioms_test.ml
```diff
@@ -0,0 +1,195 @@
...
+  libdir : path;
+  dirs_and_confs_file: path;
+  axioms_file: path;
+  debug: bool;
+}
+
+let prog = Filename.basename Sys.argv.(0)
+let usage = Printf.sprintf "Usage: %s [options]" prog
+
+let string_starts_with ~prefix s =
```
ShaleXIONG: The function has the same behaviour as `String.starts_with`?

tools/cat2table.ml
```diff
@@ -0,0 +1,952 @@
...
+    | _ , Eq _ -> 1
+    | Subset (e11, e12), Subset (e21, e22) -> do_compare_expr_pairs (e11, e12) (e21, e22)
+    | Subset _, _ -> -1
+    | _, Subset _ -> 1
+    | In (e11, e12), In (e21, e22) -> do_compare_expr_pairs (e11, e12) (e21, e22)
+    | In _, _ -> -1
+    | _, In _ -> 1
+    | VariantCond vc1, VariantCond vc2 -> do_compare_variant_cond vc1 vc2
+
+    and do_compare_variant_cond vc1 vc2 = match vc1, vc2 with
```
fsestini: Not quite sure why this is defined mutually recursively with the other `do_compare_*` functions. While `do_compare_variant_cond` is recursive (with itself), it does not actually depend on the other functions, and so it should be possible to just move its definition above this big mutual recursive section.

Maybe it's just me, and this is a simple example anyway. But generally, to my eyes, a mutual recursion block signals to the reader that all the defs involved depend on each other in non-trivial ways, and therefore need to be understood together. When a mutual recursion block includes defs that could just as easily be defined on their own, it may risk adding unnecessary burden to the reader.

tools/cat2table.ml
```diff
@@ -0,0 +1,952 @@
...
+      do_compare_expr_pairs (e11, e12) (e21, e22)
+    | Try _, _ -> -1
+    | _, Try _ -> 1
+    | If (_, c1, e11, e12), If (_, c2, e21, e22) ->
+      do_compare_pairs (c1, (e11, e12)) (c2, (e21, e22))
+        do_compare_cond
+        do_compare_expr_pairs
+    | If _, _ -> -1
+    | _, If _ -> 1
+    | _, _ -> 0 (* We don't care about the rest, since they are not effects *)
```
ShaleXIONG: You can fall back to the `Stdlib.compare`. 

In fact many of `do_compare_*` function you introduce here probably can use the default `Stdlib.compare`, if you do not care a particular order but just need a total order over data structure.

tools/cat2table.ml
```diff
@@ -0,0 +1,952 @@
...
+    | OpAnd (vc11, vc12), OpAnd (vc21, vc22) ->
+      do_compare_pairs (vc11, vc12) (vc21, vc22)
+        do_compare_variant_cond do_compare_variant_cond
+    | OpAnd _, _ -> -1
+    | _, OpAnd _ -> 1
+    | OpOr (vc11, vc12), OpOr (vc21, vc22) ->
+      do_compare_pairs (vc11, vc12) (vc21, vc22)
+        do_compare_variant_cond do_compare_variant_cond
+
+    and do_compare_expr_list es1 es2 = match es1, es2 with
```
fsestini: I think it might be worth extracting this pattern into a generic utility function `compare : ('a -> 'a -> int) -> 'a list -> 'a list -> int`, and then instantiating it here as appropriate. In fact I would expect this `compare` pattern to be useful to other parts of the project too, so I would even put it in `Misc` or some other common location in `lib/`.

ShaleXIONG: Use the existing `List.compare` or `Stdlib.compare`?

tools/cat2table.ml
```diff
@@ -0,0 +1,952 @@
...
+    and sort_cond = function
+    | Eq (e1, e2) as e ->
+      let e1 = sort_subexprs e1 in
+      let e2 = sort_subexprs e2 in
+      if do_compare_expr e1 e2 > 0 then Eq (e2, e1) else e
+    | Subset (e1, e2) -> Subset (sort_subexprs e1, sort_subexprs e2)
+    | In (e1, e2) -> In (sort_subexprs e1, sort_subexprs e2)
+    | VariantCond vc -> VariantCond (sort_variant_cond vc)
+
+    and sort_variant_cond = function
```
fsestini: Same as `do_compare_variant_cond`.

tools/cat2table.ml
```diff
@@ -0,0 +1,952 @@
...
+    else
+      let id = from_alias id in
+      if id = "same-instance" then universe_effect
+      else begin
+        try
+          let pairs = StringMap.find id rel_effects in
+          List.fold_left (fun acc (eff, _) ->
+            unite defs acc eff
+          ) empty_effect pairs
+        with Not_found ->
```
fsestini: Is this exception possibly thrown by the `fold_left` too?
If not, then I would reduce the scope of the try to just the `StringMap.find`, or even better just use `find_opt` instead.

tools/cat2table.ml
```diff
@@ -0,0 +1,952 @@
...
+    if results = [] then empty_effect
+    else List.fold_left (func defs) (List.hd results) (List.tl results)
+  | Try (_, e1, e2) | If (_, _, e1, e2) ->
+    let d1 = range looked_up_ids defs e1 in
+    let d2 = range looked_up_ids defs e2 in
+    unite defs d1 d2
+  | _ -> Warn.fatal "Range called on an invalid expression"
+
+  and get_functions =
+    let lazy_map = lazy (List.fold_left (fun map (key, value) ->
```
fsestini: I don't understand why this needs to be a lazy `StringMap.t` (or even a `StringMap.t` at all). I don't see any expensive computation or side effect that needs to be delayed at a later stage.
I think you could instead define `get_functions` as a simple OCaml function that takes the string map's key as its first argument and pattern matches on it. In fact, `get_functions` is only ever called by `apply_func` above, and that's all `apply_func` does, essentially. So you could even just inline the contents of `get_functions` into `apply_func`.

internal/herd_axioms_test.ml
TiberiuBucur: Yes, but `String.starts_with` is not available in Ocaml 4.08

tools/cat2table.ml
```diff
@@ -0,0 +1,952 @@
...
+      if do_compare_variant_cond vc1 vc2 > 0 then OpAnd (vc2, vc1) else e
+    | OpOr (vc1, vc2) as e ->
+      let vc1 = sort_variant_cond vc1 in
+      let vc2 = sort_variant_cond vc2 in
+      if do_compare_variant_cond vc1 vc2 > 0 then OpOr (vc2, vc1) else e
+    | Variant _ as e -> e in
+
+    do_compare_expr (sort_subexprs e1) (sort_subexprs e2)
+
+  let expr_equal e1 e2 = compare_expr e1 e2 = 0
```
ShaleXIONG, line 218: Given inline 148 will have effect on this function as we might have situation when `e1 <> e2` but `expr_equal e1 e2 = true`?

tools/cat2table.ml
TiberiuBucur: I cannot use those, since the comparison function on individual expressions has a different behaviour. My comparison function also allows for possible reorderings of expressions, as long as they are semantically identical.

tools/cat2table.ml
```diff
@@ -0,0 +1,952 @@
...
+
+  let get_axioms ast =
```
ShaleXIONG: Can I ask some explanation for this function? I think the return are three `StringMap` but what they represent for?

tools/cat2table.ml
```diff
@@ -0,0 +1,952 @@
...
+    | (If (_, _, e1, e2) | Try (_, e1, e2)), e
+    | e, (If (_, _, e1, e2) | Try (_, e1, e2)) ->
+      let union = unite defs e1 e2 in
+      intersect defs union e
+    | e1, e2 ->
+      Printf.eprintf "Intersection for %s and %s is not supported\n"
+        (pp_expr e1) (pp_expr e2);
+      Konst (TxtLoc.none, (Empty SET))
+
+  (* TODO: Write a version of this that also takes care of optimisations
```
ShaleXIONG, line 590: Is this `TODO` meant to be kept?

tools/cat2table.ml
```diff
@@ -0,0 +1,952 @@
...
+    | Tag (_, _) | Bind (_, _, _) | BindRec (_, _, _) | Fun (_, _, _, _, _)
+    | ExplicitSet (_, _) | Match (_, _, _, _) | MatchSet (_, _, _, _) as e ->
+      Printf.eprintf "Matching %s not supported\n" (pp_expr e);
+      []
+    | e ->
+      List.filter (fun e -> not (is_empty e))
+        (List.map (intersect defs e) e1s) in
+
+    (* TODO: Rewrite this so it filters out duplicates in the result *)
+    let rec do_tr_chain looked_up_ids ret_sol sol e1s is_inv = function
```
ShaleXIONG: Can you put some explanation on the input arguments? It is not straightforward to understand by the name.

dirs-and-confs.txt
ShaleXIONG, line 1: What is this file for?

Makefile
ShaleXIONG, line 1: What will be the impact of this extra target for `test` regarding CI-time wise? if making sense, I think you can just add an extra CI run only trigger for change to Cat2table.

tools/cat2table.ml
TiberiuBucur: Thanks for pointing this out. In general, this can cause problems. In my case, since I do not care about the ordering of effects not treated by compare_expr, it wouldn't be much of an issue. Still, for the sake of completeness, I addressed the inconsistencies of line 148.

tools/libdir/aarch64_wrapped_primitives.cat
TiberiuBucur: I copied the same explanation I had in `aarch64axioms.cat` and copied the license in both files. Feel free to resolve this if you're happy to.

Makefile
TiberiuBucur: The axioms test runner takes approximately 42 seconds, taking the total `make test` time from 340 to about 382 seconds.

Triggering the runner only for changes in `cat2table` is not enough. The axioms rely on the way the primitives are defined, which can change whenever something in `/herd` or `/lib` changes. In fact, the runner does not test `cat2table`, it effectively tests `herd`. More precisely, it tests whether the assumptions we make based on the primitives that `herd` defines still hold.

tools/cat2table.ml
TiberiuBucur: The function parses the axioms given inside the ast and returns the 3 data structures that used to be hardcoded before, namely the `supersets` (representing pairs of effects where one is a superset of the other), `intersects` (representing pairs of effects that are intersectable) and `rel_effects` (representing the effects that can sit on each side of a primitive relation).
This is also shown by the usage of the function a few lines below:
`let supersets, intersects, rel_effects = get_axioms ast`

dirs-and-confs.txt
TiberiuBucur: This file takes the directories that contain the tests `herd_axioms_test.ml` will run to check the axioms defined in `aarch64axioms.cat`, and the respective `herd7` configuration files that go along with the tests in those directories. Since the entire test suite and catalogue were too big to run (it would take very long), I only selected the ones that were enough to check our assumptions.

If you have any idea on how to make this better or get rid of this file, I'd be curious to know.

tools/cat2table.ml
TiberiuBucur: It is a possible improvement for the future, although not crucial and with no impact other than readability. It would be more involved to implement it, so decided to postpone it.

dirs-and-confs.txt
HadrienRenaud: Can this live in `herd/tests/AArch64-axioms`?

Makefile
```diff
@@ -87,6 +88,17 @@ test:: dune-no-missing-file-in-runt
...
+test:: test.aarch64axioms
+test-local:: test.aarch64axioms
+test.aarch64axioms:
+	@ echo
+	$(HERD_AXIOMS_TEST) \
+		-herd-path $(HERD) \
+		-libdir-path ./herd/libdir \
+		-dirs-and-confs-path ./dirs-and-confs.txt \
+		-axioms-path ./herd/libdir/aarch64axioms.cat
+	@ echo "cat2table AArch64 axioms: OK"
```
HadrienRenaud: Why is this different to `herd_regression_test` with a conf that would specify `model = aarch64axioms.cat`?

tools/cat2table.ml
TiberiuBucur: Done. Please resolve this if you feel the comment clarifies this.

dirs-and-confs.txt
TiberiuBucur: @HadrienRenaud do you mean to have a separate directory that includes this and the 2 new cat files?

Makefile
TiberiuBucur: Because the crucial part of `herd_axioms_test.ml` is that it checks that all of the flags in `aarch64axioms.cat` (the ones used to test whether 2 individual effects intersect) are raised at least once.

dirs-and-confs.txt
HadrienRenaud: Yes for example

Makefile
HadrienRenaud: Ah I had not understood that those axioms are meant to be negated! Could you explain this in the cat file? I fear that people could get confused and think that those intersection need to be empty at all time (which is what "axiom" means for me at least)

tools/cat2table.ml
TiberiuBucur: Thanks. This actually exists in `Misc`, it's called `list_compare`. Thanks for pointing this out.

Makefile
TiberiuBucur: Done

herd/libdir/aarch64axioms.cat
```diff
@@ -0,0 +1,226 @@
+(*
+ * The Armv8 Application Level Memory Model.
+ *
+ * This is a machine-readable, executable and formal artefact, which aims to be
+ * the latest stable version of the Armv8 memory model.
+ * If you have comments on the content of this file, please send an email to
+ * memory-model@arm.com
+ * For a textual version of the model, see section B2.3 of the Armv8 ARM:
+ * https://developer.arm.com/documentation/ddi0487/
```
relokin: Can we please remove this?

tools/libdir/aarch64assumptions.cat
relokin, line 1: Shouldn't this file live closer to the tool? `tools/libdir`?

herd/libdir/aarch64_wrapped_primitives.cat
```diff
@@ -0,0 +1,112 @@
+(*
+ * The Armv8 Application Level Memory Model.
+ *
+ * This is a machine-readable, executable and formal artefact, which aims to be
+ * the latest stable version of the Armv8 memory model.
+ * If you have comments on the content of this file, please send an email to
+ * memory-model@arm.com
+ * For a textual version of the model, see section B2.3 of the Armv8 ARM:
+ * https://developer.arm.com/documentation/ddi0487/
```
relokin: Can we remove this, and explain what this file really about?

tools/libdir/aarch64assumptions.cat
```diff
@@ -0,0 +1,226 @@
...
+* here in order to avoid "undefined" errors when running herd, as the
+* aarch64 model has to be included only after the axioms have been checked.
+* This is because of shadowing (redefinition) of the primitives in the
+* model (eg. let A = try A & (Exp & M) with emptyset). cat2table requires
+* knowledge of the bare primitives as exported by herd, and not as
+* redefined by the cat model, as the tool unwraps the redefinitions
+* itself when checking effect intersection.
+*)
+
+include "aarch64_wrapped_primitives.cat"
```
relokin, line 35: Rather than trying to do this on your own why can't you include aarch64utils.cat and/or aarch64fences.cat and/or aarch64memattrs.cat?

herd/libdir/aarch64axioms.cat
```diff
@@ -0,0 +1,226 @@
...
+    PTEiWT |
+    PTEiNC |
+    PTEoWB |
+    PTEoWT |
+    PTEoNC |
+    PTENormal |
+    PTETaggedNormal |
+    PTEXS
+) \ PTEMemAttr as Subsets-of-PTEMemAttr **)
+
```
relokin: Can you explain why this is disabled?

tools/libdir/aarch64assumptions.cat
```diff
@@ -0,0 +1,226 @@
...
+
+(* Effect superset-subset relations.
+*  These expressions should ALWAYS be empty, hence they are asserted.
+*)
+
+assert empty (R | W | Instr | PTE | PA | T) \ M as Subsets-of-M
```
relokin: This check won't help much as it should be that `M = R | W`

tools/libdir/aarch64assumptions.cat
relokin, line 1: How did you generate this file? I think we have to find a more systematic way to generate this. I am not sure I follow the logic and the axioms seem to be adhoc. See below for some initial concerns.

tools/libdir/aarch64assumptions.cat
```diff
@@ -0,0 +1,226 @@
...
+    PTEXS
+) \ PTEMemAttr as Subsets-of-PTEMemAttr **)
+
+(* Pairs of effects that are intersectable.
+* These expressions are not meant to be non-empty all the time. Rather,
+* all of these flags should be raised AT LEAST ONCE. This is checked by
+* the herd_axioms_test.ml runner.
+*)
+
+flag ~empty A & R as A-and-R
```
relokin, line 82: I am curious about this, shouldn't you be checking that A is subset of R (or better of Exp & R)?

tools/libdir/aarch64assumptions.cat
```diff
@@ -0,0 +1,226 @@
...
+flag ~empty A & Exp as A-and-Exp
+flag ~empty L & Exp as L-and-Exp
+flag ~empty L & W as L-and-W
+flag ~empty Q & R as Q-and-R
+flag ~empty Q & Exp as Q-and-Exp
+flag ~empty EXC-ENTRY & TagCheck as EXC-ENTRY-and-TagCheck
+flag ~empty Exp & R as Exp-and-R
+flag ~empty Exp & W as Exp-and-W
+flag ~empty Exp & Instr as Exp-and-Instr
```
relokin: Why do you need this? There is nowhere in the cat file that we say Exp & Instr?

tools/libdir/aarch64assumptions.cat
```diff
@@ -0,0 +1,226 @@
...
+flag ~empty W & PTEV as W-and-PTEV
+flag ~empty W & PTEINV as W-and-PTEINV
+flag ~empty W & PTEAF0 as W-and-PTEAF0
+flag ~empty W & PTEDB0 as W-and-PTEDB0
+flag ~empty DATA & PTEV as DATA-and-PTEV
+flag ~empty DATA & PTEINV as DATA-and-PTEINV
+flag ~empty DATA & PTEAF0 as DATA-and-PTEAF0
+flag ~empty DATA & PTEDB0 as DATA-and-PTEDB0
+flag ~empty Wreg & PTEV as Wreg-and-PTEV
+flag ~empty Wreg & PTEAF0 as Wreg-and-PTEAF0
```
relokin: Why do you have to check this PTEV and PTEAF0 are subsets of M and M and Wreg are disjoint sets. There are much more fundamental expressions to assert and these two should follow from these.

tools/libdir/aarch64assumptions.cat
TiberiuBucur: In `aarch64fences.cat` we have:
`let A = try A & (Exp & M) with emptyset`
This implies that the _primitive_ set "A" as exposed by `herd` (not the A as overriden by the cat file) is not restricted to `Exp & R`.
Indeed, if I include `assert empty A \ R as A-subset-of-R` in `aarch64axioms.cat`, then the runner fails.

tools/cat2table.ml
```diff
@@ -0,0 +1,1107 @@
...
+          exit 0),
+      " show installation directory and exit" );
+    ( "-set-libdir",
+      Arg.String (fun s -> libdir := s),
+      "<path> set installation directory to <path>" );
+    ArgUtils.parse_string "-model" model !model;
+    ArgUtils.parse_string "-axioms_file" axioms_file !axioms_file;
+    ("-e1", Arg.String (fun s -> e1 := Some s), "<str> first effect");
+    ("-e2", Arg.String (fun s -> e2 := Some s), "<str> second effect");
+    ("-start", Arg.String (fun s -> start_rel := s), "<str> top level relation");
```
fsestini: The help message here should indicate that `ob` (or, whatever `start_rel` holds by default) is the default value.

tools/cat2table.ml
```diff
@@ -0,0 +1,1107 @@
...
+    ( "-libdir",
+      Arg.Unit
+        (fun () ->
+          print_endline !libdir;
+          exit 0),
+      " show installation directory and exit" );
+    ( "-set-libdir",
+      Arg.String (fun s -> libdir := s),
+      "<path> set installation directory to <path>" );
+    ArgUtils.parse_string "-model" model !model;
```
fsestini: The help message for this shows:
```
  -model <string> herd/libdir/aarch64.cat, default herd/libdir/aarch64.cat
 ```
I don't think it's very helpful. Maybe the part after `<string>` should just describe what this flag does.

File: tools/cat2table.ml [hunk 28]
```diff
@@ -0,0 +1,1107 @@
...
+    ArgUtils.parse_string "-axioms_file" axioms_file !axioms_file;
```
fsestini: Same as `-model`. Also, let's use `-axioms-file` for consistency.
```suggestion
    ArgUtils.parse_string "-axioms-file" axioms_file !axioms_file;
```

tools/cat2table.ml
```diff
@@ -0,0 +1,1107 @@
...
+    | Some e1, Some e2 ->
+        let e1 = parse_effect_from_string "e1" e1 |> ASTUtils.flatten in
+        let e2 = parse_effect_from_string "e2" e2 |> ASTUtils.flatten in
+        let sol = tr_rel e1 e2 start_rel defs in
+        if sol = ExprGraph.empty then Printf.printf "No solutions\n"
+        else
+          ignore
+            (ExprGraph.dfs_fold
+               ~per_vertex:(fun rel lvl _ ->
+                 Printf.printf "%s%s\n" (String.make lvl '-') (pp_expr rel))
```
fsestini: Why not use a space to indent the output here? I think `-` makes it unnecessarily distracting.

tools/cat2table.ml
```diff
@@ -0,0 +1,1107 @@
...
+        let defs = tr_bds defs (kont false) bds in
+        (defs, parsed_files)
+    | Rec (_, bds, _) ->
+        let defs = tr_bds defs (kont true) bds in
+        (defs, parsed_files)
+    | Include (_, fname) when not (StringSet.mem fname parsed_files) ->
+        tr_ast defs kont parsed_files fname
+    | _ -> (defs, parsed_files)
+
+  and tr_bds defs kont bds =
```
fsestini: As with other function I've commented on previously, this `tr_bds` doesn't need to be mutually recursively defined with the others, as it doesn't depend on them.

tools/cat2table.ml
```diff
@@ -0,0 +1,1107 @@
...
+    and run_bds bds = List.map (fun (loc, pat, e) -> (loc, pat, run e)) bds
+    and run_cond = function
+      | Eq (e1, e2) -> Eq (run e1, run e2)
+      | Subset (e1, e2) -> Subset (run e1, run e2)
+      | In (e1, e2) -> In (run e1, run e2)
+      | VariantCond _ as c -> c
+    in
+    run expr
+
+  let rec tr_ast defs kont parsed_files fname =
```
fsestini, line 1023: This function is only used once in the body of `tr_ins`, and it's also re-defined/shadowed only a few lines later. If think it would be much cleaner if you just inline the definition of this `tr_ast` into its use site within `tr_ins`, and thus make `tr_ins` non-recursive.

fsestini (commented):
As the PR is not quite ready yet, I'm taking the opportunity to leave some further comments.

tools/cat2table.ml
TiberiuBucur: It was @ShaleXIONG 's suggestion. I'm sure different people would have different preferences. I don't mind either.

tools/libdir/aarch64assumptions.cat
relokin: But if you included aarch64fences.cat as I suggested in a different comment, would it still fail?

tools/cat2table.ml
TiberiuBucur: Not sure I understand what you mean. `tr_ast` also calls `tr_ins`, so they are mutually recursive. Also, `tr_ast` was called from inside the redefinition of `tr_ast` (the one you mentioned is a few lines later), now split into `count_defs` and `get_defs`.

fsestini: My bad, didn't realise `tr_ast` was also used inside the redefinition.

tools/cat2table.ml
TiberiuBucur: Let me know if the current version is better.

fsestini: Much better, thank you.

tools/cat2table.ml
TiberiuBucur: I replaced the `-` characters with whitespaces. Now the printing should be more user-friendly.

tools/libdir/aarch64assumptions.cat
TiberiuBucur: I tried to do this myself. Unfortunately, `cat2table` unwraps the definitions inside the files you mentioned down to the terminal symbols, and only then checks intersection of `Var`s in the lookup table. In the case of shadowing, this might cause problems, as the redefinition of a primitive might change its `supersets` and `intersects` sets. For example, `let A = try A & (Exp & M) with emptyset` is defined in `aaarch64fences.cat`, effectively restricting the set "A" as exposed by `herd7` to just explicit memory effects (which I was surprised to see it was not by default).

`cat2table` could change its algorithm so it does not expand redefinitions / shadowing, but that would be extra complexity and would be less future-proof (the cat file can change much more often than the herd internals).

tools/libdir/aarch64assumptions.cat
TiberiuBucur: This is indeed hand-written. I am not sure there is a reliable way to auto-generate these axioms. They have been written after analysing the code of `herd7` and the sets it exposes. We already discussed the possibility of generating the lookup tables from the AArch64 semantics implemented in `herd7` and we deemed it infeasible.

For what it's worth, this file only references primitives, which shouldn't change much in general. Once I strip it down to only the most basic needed assertions, it would be easier to maintain.

herd/libdir/aarch64axioms.cat
TiberiuBucur: Agreed to remove this for now as the subsets of PTEMemAttr are not fully covered by the catalogue.

tools/libdir/aarch64assumptions.cat
TiberiuBucur: It wouldn't, but as I explained above, `cat2table` right now unwraps all of the (re-)definitions it finds until it reaches terminal symbols. As such, these assumptions encoded in the lookup tables need to refer to the original primitives (not the redefined ones). This is the whole reason why we needed `aarch64_wrapped_primitives.cat` in the first place.

tools/libdir/aarch64assumptions.cat
TiberiuBucur: As discussed in our meeting, the catalogue does indeed find at least one occurrence of an Exp & Instr effect. We agreed to not support this until the catalogue is revised.

tools/libdir/aarch64assumptions.cat
TiberiuBucur: As discussed in our meeting, we discovered that PTEV and PTEAF0 are actually (surprisingly so) not subsets of M. They refer to effects that use PTEVals, which can indeed be memory effects, but also register effects. Again, we agreed not to support these until after this is clarified, hence they will be removed.

tools/libdir/aarch64assumptions.cat
TiberiuBucur: I moved it to a new directory `tools/libdir` as requested. In order to do this, I had to modify the test runner so that it copies all files of this directory and `herd/libdir` into a new directory inside `/tmp` (via the `get_temp_dir_name` function, which is cross-platform compatible) and deleting this at the end, since all cat files must be in the same libdir. I also copied the files inside `tools/libdir` into the herd installation directory upon `make install`, for the same reason.

TiberiuBucur:
> I wonder if this can be done with `herd7`'s `-I` option without the need to copy cat files. I believe `-I` allows to specify additional search paths. Can you check that this is the case and that it works for your use case?

You are absolutely right and it does work. Had I known this yesterday I would have saved myself a lot of time. Thanks

tools/cat2table.ml
```diff
@@ -0,0 +1,1220 @@
...
+          let id = if num = total_num then id else to_alias id num in
+          let defs = StringMap.add id converted_def defs in
+          (defs, crt_nums))
+        StringSet.empty fname
+    in
+    (defs, crt_nums)
+
+  let execute fname e1 e2 start_rel =
+    let valid_primitives =
+      StringSet.of_list
```
ShaleXIONG: Can we automatically generate this list from your extra cat file?

ShaleXIONG (approved):

tools/cat2table.ml
TiberiuBucur: I could. It would just be a bit more code, but it would certainly be more robust. I can write it now if you want.

ShaleXIONG: Can you do it, since it sounds like a small code? Thanks.

lib/txtLoc.mli
```diff
@@ -29,5 +29,6 @@ val pp : out_channel -> t -> unit
 
 module Extract() :
 sig
+ val cache : string -> string -> unit
```
fsestini: Can we please add a comment to explain what this function does in the context of this module?
I think it's clear what "extraction" means in the context of source code locations (i.e., `extract t : string` is the substring being referenced by `t`), however it's not clear to me what a caching operation is supposed to achieve in the context of `Extract`.

I would also add named arguments to the type signature so that it's a bit more clear which parameter is which. (feel free to use more appropriate/informative names instead of `key`/`value`)
```suggestion
 val cache : key:string -> value:string -> unit
```

dirs-and-confs.txt
```diff
@@ -0,0 +1,10 @@
+./herd/tests/instructions/AArch64
```
fsestini, line 1: Can I ask the reasons why this list needs to be in a separate file (and even at the root of the project repo), as opposed to just be part of the `herd_assumptions_test.ml` source code? It doesn't look like any other tool or test suite would need this list anyway.

tools/cat2table.ml
TiberiuBucur: Done. It was not as straightforward as we thought. There were things to consider to do with shadowing, once again.

dirs-and-confs.txt
TiberiuBucur: I just went away from hardcoding things in the source code as a rule of thumb, as I have been told in the past. It makes no difference in terms of functionality and it is easier to change if one wants to (no recompilation needed).

lib/txtLoc.mli
TiberiuBucur: The purpose of the function is to add an entry to the corresponding Hashtbl in the extract module, which prevents the extract function from attempting to open a file. Since e1 and e2 come from the command line, there would be no file to open, so attempting to print them using a `TxtLoc` would result in an error.

In fact, I added this because when I looked at the `txtLoc.ml` file, it seemed like it was not ever added to, only queried, so this would imply that files are always re-read whenever printing based on location is attempted.

dirs-and-confs.txt
fsestini: I see. I guess my point is that there is value in making it clear which test data/config pertains which test script (and vice versa) by way of keeping things _local_/close to each other (be it in terms of filesystem location, code, etc.). Especially for internal test scripts like this one, where avoiding recompilation isn't that much of an issue.
Perhaps we can revisit this point when we add a test suite for `cat2table`.

lib/txtLoc.mli
fsestini: > when I looked at the txtLoc.ml file, it seemed like it was not ever added to

Indeed it looks like `read_fname` is missing a cache insertion after the `Misc.input_protect`.

> Since e1 and e2 come from the command line, there would be no file to open, so attempting to print them using a TxtLoc would result in an error.

Thanks, I now understand the reasoning behind this. I now wonder, however, if your use case would then better served by a pretty-printing function/module that is not tied to source files in the file system in the way `TxtLoc` is. The addition of this `cache` function makes your use case possible, at the cost of exposing some internal implementation details of `TxtLoc`/`Extract` from their public interface. Perhaps we can leave this to a follow-up PR.

TiberiuBucur: @ShaleXIONG and I discussed this much earlier in the development process (when I made the decision to implement this caching function) and we decided that having a complete pretty printer is not the highest priority.

fsestini: > having a complete pretty printer is not the highest priority

Understandable. Relying on the existing `TxtLoc` _logic_ is a fine approach. If I can add what _I_ would have done in this case, I would have simply copied the `extract` function from `TxtLoc.Extract` needed for pretty-printing, and put it in the `cat2table` source file (possibly with some adjustments). This way:
- we don't need to implement a full-fledged pretty printer right now
- cat2table can use the `extract` copy to do pretty printing
- the public interface of `lib/txtLoc` does not need to change one bit

Makefile
```diff
@@ -87,6 +88,17 @@ test:: dune-no-missing-file-in-runt
 dune-no-missing-file-in-runt:
 	asllib/tests/check-no-missing-file-in-run.sh ./
 
+test:: test.aarch64assumptions
```
HadrienRenaud, line 91: I know I'm arriving very late after merging, but I'm wondering what the rationale to add those to regression test is, compared to tests that are run every week (e.g. `make test-all`). Have those been broken before?

PR #1312, opened 2025-05-21
[tools,miaou] Use simple gensym for events names.

maranget:
Allocation of events names by *miaou* was wrong. As an example `let erel = (brel & (crel ; drel)) ; rel` was translated as:
```
There is a *e-rel* from an Effect E~1~ to an Effect E~2~ if all of the following apply:
-   E~3~ b-follows E~1~.
-   E~1~ c-rel E~3~.
-   There is a d-rel from E~3~ to E~3~.
-   E~3~ relates E~2~.
```
Which is wrong, as `E~3`. stands for two different events. A simple allocating function with global state fixes the issue, yielding:
```
 There is a *e-rel* from an Effect E~1~ to an Effect E~2~ if all of the following apply:
-   E~3~ b-follows E~1~.
-   E~1~ c-rel E~4~.
-   There is a d-rel from E~4~ to E~3~.
-   E~3~ relates E~2~.

```

### DIFF
--- a/tools/miaou.ml
+++ b/tools/miaou.ml
@@ -147,9 +147,14 @@ module Make
 
     let next_indent indent = indent ^ "  "
 
-    let next_event = function
-      | 1 -> 3
-      | n -> n+1
+    module Next : sig
+      val reset : unit -> unit
+      val next : unit -> int
+    end = struct
+      let c = ref 1
+      let reset () = c := 1
+      let next () = let r = !c in incr c ; r
+    end
 
     let pp_evt k = sprintf "E\\textsubscript{%d}" k
 
@@ -427,7 +432,7 @@ module Make
            match es with
            | [] -> e1,e2,e2
            | _::_ ->
-              e1,next_event e1,e2 in
+              e1,Next.next (),e2 in
          Item
            (sprintf
             "\\%s{%s}{either  %s} or \\%s{an Effect which}{%s}"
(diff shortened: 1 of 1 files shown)


tools/miaou.ml
```diff
@@ -147,9 +147,14 @@ module Make
...
 
-    let next_event = function
-      | 1 -> 3
-      | n -> n+1
+    module Next : sig
+      val reset : unit -> unit
+      val next : unit -> int
+    end = struct
+      let c = ref 3
+      let reset () = c := 3
```
relokin, line 155: should we start from 1 instead and make a call to `next ()` in lines 672, 673, 677? That way miaou7 will also handle correctly the indices for definitions of events like
```
let evts1 = range([M]; po; [M])
```

relokin (approved):
Thanks for this Luc, I just ran into the same problem earlier today. The patch looks good to me, I just have one suggestion for improvement.

tools/miaou.ml
maranget: Thanks for your suggestion. Will fix now,.

maranget:
The current simple scheme does not re-allocate event numbers when it could. As of now, **miaou**  will format `(a;b)|(c;d)`
```
One of the following applies:
  - All of the following apply:
  -    E1 A-rel E3
    -  E3 B-rel E2
 - All of the following apply:
    -   E1 C-rel E4
    -   E4 D-rel E2
 ```
One may think that  `E4` above could be `E3` . However, some flattening occurs when formatting  `(a;b)&(c;d)`:
```
 - All of the following apply:
  -  E1 A-rel E3
  -  E3 B-rel E2
  -   E1 C-rel E4
  -   E4 D-rel E2
 ```
As a consequence,having a one-to-one correspondance between numbers and relation interfaces probably is the right option.

PR #1313, opened 2025-05-21
[asl][menhir2bnfc] Rework installation and dune files

HadrienRenaud:
People have reported crashes with `dune < 3.17`.
This PR moves menhir2bnfc into its own installation.
Minimum required version for dune is up to 3.0 (released Feb 2022) for conditional compilation in carpenter.

### DIFF
--- a/.github/workflows/make-test.yml
+++ b/.github/workflows/make-test.yml
@@ -23,13 +23,11 @@ jobs:
         os:
           - ubuntu-latest
         cfg:
-          - ocaml-version: "4.08.1"
-            with-min-versions: true
+          - ocaml-version: "ocaml-base-compiler.4.08.1,dune.3.10.0,zarith.1.13,menhir.20220210"
             with-fmt: false
             job-name: "OCaml v4.08"
             dune-cache: "enabled"
           - ocaml-version: "5.2"
-            with-min-versions: false
             with-fmt: true
             dune-cache: "enabled-except-user-rules"
             job-name: "OCaml v5.2, with formatting check"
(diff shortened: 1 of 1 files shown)


.github/workflows/make-test.yml
```diff
@@ -47,6 +51,12 @@ jobs:
           ocaml-compiler: ${{ matrix.cfg.ocaml-version }}
           dune-cache: true
 
+      - run: |
+          opam pin add dune 2.9.3
+          opam pin add menhir 20200123
```
murzinv: JFYI, I noticed that some old version of `menhir` could produce asserts or could make build process stall (consuming CPU and memory). After some version bisecting turns to be `20220210` the minimal working version for me  (which also depends on `dune`  >= 2.8.0)

HadrienRenaud: Thank you! It looks like 2023something is required for menhir2bnfc anyways

asllib/menhir2bnfc/CvtGrammar.ml
```diff
@@ -31,11 +31,31 @@ end = struct
   let eof_token = "EOF"
 
   (* Create useful mapping structures *)
-  module ProductionMap = Map.Make (Production)
-  module NonterminalMap = Map.Make (Nonterminal)
+  module P = struct
```
Roman-Manevich: Perhaps `ComparableProduction`, `ComparableNonterminal`, `ComparableTerminal`?

maranget (approved):
LGTM, I have not looked at the `menhir2bnc` part at all.

I will complete the minimal version check by pinning later, adding a periodic check that follows the technique of PR #1423. This may help to maintain the `.opam` file.

PR #1315, opened 2025-05-22
[asl] Simplifying local constants

Roman-Manevich:


Roman-Manevich:
The typechecker seems to be working for the new test as well.

PR #1316, opened 2025-05-22
[asl] Fix bug in lhs type-checking

HadrienRenaud:
When type-checking assignable elements, the type-checker would do too much work with un-typed slices, and so the sub-typing algorithm would not work correctly.

TODO:
- [x] Add tests
- [x] Implementation
- [x] Add documentation

Documentation after:
Screenshot 2025-05-22 at 16 55 31
Screenshot 2025-05-22 at 16 57 16

### DIFF
--- a/asllib/doc/ASLmacros.tex
+++ b/asllib/doc/ASLmacros.tex
@@ -2372,6 +2372,7 @@
 \newcommand\slicesone[0]{\texttt{slices1}}
 \newcommand\slicesp[0]{\texttt{slices'}}
 \newcommand\slicestwo[0]{\texttt{slices2}}
+\newcommand\slicesannotated[0]{\texttt{slices\_annotated}}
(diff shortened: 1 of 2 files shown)


hrutvik (approved):

Roman-Manevich (approved):
Overall looks good.
However, I would add the new test to `SemanticsRule.LESlice.asl` instead of creating the new `HExample19.asl`.

asllib/Typing.ml
```diff
@@ -2581,21 +2581,21 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
         match t_le1_anon.desc with
         | T_Bits _ ->
             let le2, ses1 = annotate_lexpr env le1 t_le1 in
+            let slices2, ses2 =
```
Roman-Manevich, line 2584: ideally `slices2` would be named something like `slices_annotated` and `ses2` something like `ses_slices`.

PR #1318, opened 2025-05-26
ASL-716 accessibility enhancements phase 20: examples and fixes

Roman-Manevich:
* Added examples to the Semantics Utilities chapter.
* Fixed `SemanticsRule.SlicesToPositions` and removed `SemanticsRule.AsBitvector` to more closely match the implementation.
* Fixed `TypingRule.ExplodeConstraint` to output the precision loss indicator.
* Added examples to the Relations on Types chapter.


asllib/doc/ASLmacros.tex
```diff
@@ -275,6 +275,7 @@
 \newcommand\emptylist[0]{\hyperlink{def-emptylist}{[\ ]}}
 \newcommand\Proseemptylist[1]{#1 is the empty list}
 \newcommand\choice[3]{\hyperlink{def-choice}{\textsf{choice}}(#1,#2,#3)}
+%\newcommand\choice[3]{\hyperlink{def-choice}{\textbf{if }}#1\textbf{ then }#2\textbf{ else }#3}
```
HadrienRenaud: Shouldn't this be removed? I'm worried about noise making `doclint.py` having some false-negative.

PR #1320, opened 2025-05-29
[tools] Fix the pretty-printing of the transitive closure operator '+' by miaou.

maranget:
Yet another problem with **miaou**: useless and disturbing expansion of transitive closure. Currently, **miaou** only accepts transitive closures when applied to disjunctions and conjunctions of names. But it does so in a contrived way, applying some expansion.

For instance, the tool **miaou** described `let frel = (erel|drel)+` as:
```
There is a *f-rel* from an Effect E~1~ to an Effect E~2~ if one of the following applies:
-   There is a d-rel from E~1~ to E~2~.
-   There is a e-rel from E~1~ to E~2~.
-   There exists a chain of d-rel or e-rel from E~1~ to E~2~.
```
After this PR, output is:
```
There is a *f-rel* from an Effect E~1~ to an Effect E~2~ if There exists a chain of d-rel or e-rel from E~1~ to E~2~.
```

### DIFF
--- a/tools/miaou.ml
+++ b/tools/miaou.ml
@@ -394,25 +394,17 @@ module Make
            match ASTUtils.as_vars es with
            | None -> raise Exit
            | Some ids ->
-              let itms =
-                List.map
-                  (fun (loc,id) -> tr_rel_id e1 e2  loc id)
-                  ids
-              and lst =
-                let names =
-                  List.map (fun (_,id) -> id_name id) ids
-                  |> String.concat pp_op in
-                pp_rel_id e1 e2 (pp_transitive names) in
-              itms,lst in
+               let names =
+                 List.map (fun (_,id) -> id_name id) ids
+                 |> String.concat pp_op in
+               pp_rel_id e1 e2 (pp_transitive names) in
          begin
            try
              match op with
              | Union ->
-                let itms,lst = tr_es "{} or " in
-                mk_list Union (itms@[lst])
+                tr_es "{} or "
              | Inter ->
-                let itms,lst = tr_es "{} and " in
-                mk_list Union [mk_list Inter itms;lst;]
+                tr_es "{} and "
              | _ -> tr_fail loc
            with Exit -> tr_fail loc
          end


relokin:
@artkhyzha are you happy with this? It simplifies the definition of Pick Data Dependency where one of the conditions reads:

> One of the following applies:
> * There is an Intrinsic Data Dependency from E3 to E2.
> * There is an Intrinsic Control Dependency from E3 to E2.
> * There exists a chain of Intrinsic Data Dependency or Intrinsic Control Dependency
from E3 to E2.

and it would be updated to read:

> There exists a chain of Intrinsic Data Dependency or Intrinsic Control Dependency
from E3 to E2.

relokin:
> @relokin, a few concerns:
> 
>     * miaou7 should make the difference between `(erel|drel)+` and `(erel|drel)*` clear.
> 

I agree

>     * when transliterating `aarch64.cat`, the generated English for the pick-data-dep looks a bit broken, and I am not sure how to avoid that generically. Currently, "a chain of Intrinsic Data Dependency or Intrinsic Control Dependency" does not read beautifully. Alas, I do not have suggestions that do not modify the cat file.
> 

I agree but I think the solution would be to avoid (iico_data|iico_ctrl)+ altogether.

>     * when transliterating `aarch64.cat`, should "(erel|drel)+" and "let rec frel = erel | drel | frel;frel" be transliterated the same way?
> 

Is it possible?

`let rec frel = erel | drel | frel;frel` would be transliterated as

```
There is a frel between an Effect E₁ and an Effect E₂ if one of the
following applies:
- There is a erel between E₁ and E₂.
- There is a drel between E₁ and E₂.
- All of the following apply:
  - There is a frel between E₁ and E₃.
  - There is a frel between E₃ and E₂.
```
`let rec frel =  (erel|frel)+` is transliterated as

```
There is a frel between an Effect E₁ and an Effect E₂ if one of the
following applies:
- There is a erel between E₁ and E₂.
- There is a drel between E₁ and E₂.
- There exists a chain of erel or drel from E₁ to E₂.
```

Maybe there is no good solution and the answer is that we need to avoid the cat operators `*` and `+`.

> 
> I suppose `(erel|drel)+` could be a "non-empty chain" and `(erel|drel)*` could be a "possibly empty chain"? We do not use the latter in the cat files, I think.

Yes

PR #1323, opened 2025-06-05
[asl] removed '-' from pending constrained integer syntax

Roman-Manevich:
This PR changes the syntax of pending integer constraints from `integer{-}` to `integer{}`. 
The change does not incur any conflicts during parser building.
More specifically:
* The old `integer{-}` syntax is hidden behind the new `--allow-hyphenated-pending-constraint` command line flag.
* Updated `Parser.mly`
* Updated ASL Reference.
* Updated the relevant tests.
* Added the test `regressions.t/hyphenated-pending-constraint.asl` to test the behaviour of the new command line flag.

### DIFF
--- a/asllib/Parser.mly
+++ b/asllib/Parser.mly
@@ -292,7 +292,11 @@ let expr :=
 let constraint_kind_opt := constraint_kind | { UnConstrained }
 let constraint_kind :=
   | cs=braced(clist1(int_constraint)); { WellConstrained (cs, Precision_Full) }
-  | braced(MINUS); { PendingConstrained }
+  | braced(MINUS); {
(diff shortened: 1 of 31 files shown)


asllib/Parser.mly
```diff
@@ -292,7 +292,7 @@ let expr :=
 let constraint_kind_opt := constraint_kind | { UnConstrained }
 let constraint_kind :=
   | cs=braced(clist1(int_constraint)); { WellConstrained (cs, Precision_Full) }
-  | braced(MINUS); { PendingConstrained }
+  | LBRACE; RBRACE; { PendingConstrained }
```
hrutvik, line 299: I think we may want to support the old form, guarded behind a suitable flag

hrutvik (approved):

PR #1336, opened 2025-06-18
[asl] Refine collections (ASL-806)

HadrienRenaud:
The main goal of this PR is to refine the syntax of collections. They now can
only be used in global variable declarations, as anonymous types.

For example:
```
var my_collection: collection { field1: bits(3), field2: bits(5) };

func main () => integer
begin
  print(my_collection.field1);

  return 0;
end;
```

This PR also fixes a few holes in `check_is_not_collection` and its usage that
were missed in the previous PR.

### DIFF
--- a/asllib/Typing.ml
+++ b/asllib/Typing.ml
@@ -1356,8 +1356,7 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
         let ses = SES.union ses_t ses_index in
         (T_Array (index', t') |> here, ses) |: TypingRule.TArray
     (* Begin TStructuredDecl *)
-    | (T_Record fields | T_Exception fields | T_Collection fields) when decl
-      -> (
+    | T_Record fields | T_Exception fields | T_Collection fields -> (
         let+ () =
           match get_first_duplicate (List.map fst fields) with
           | None -> ok
@@ -1374,10 +1373,13 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
         let ses = SES.unions sess in
         match ty.desc with
         | T_Record _ ->
+            assert decl;
             (T_Record fields' |> here, ses) |: TypingRule.TStructuredDecl
         | T_Exception _ ->
+            assert decl;
             (T_Exception fields' |> here, ses) |: TypingRule.TStructuredDecl
         | T_Collection _ ->
+            assert (not decl);
             let+ () =
               check_true
                 (List.for_all (fun (_, t) -> has_structure_bits env t) fields)
@@ -1386,7 +1388,8 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
             (T_Collection fields' |> here, ses) |: TypingRule.TStructuredDecl
         | _ -> assert false
         (* Begin TEnumDecl *))
-    | T_Enum li when decl ->
+    | T_Enum li ->
+        assert decl;
         let+ () =
           match get_first_duplicate li with
           | None -> ok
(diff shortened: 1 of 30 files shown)


asllib/doc/GlobalStorageDeclarations.tex
```diff
@@ -392,6 +392,7 @@ \section{Typing}
     \item $\tyoptp$ is the singleton set for the type $\vt$;
     \item $\initialvalue$ is $\None$;
     \item annotating the type $\vt$ in $\tenv$ yields $(\vtp, \vsest)$\ProseOrTypeError;
+    \item \Prosecheckisnotcollection{$\vt$}{$\tenv$};
```
hrutvik: Should this be in the `None`/`Some` case below instead?

asllib/doc/Types.tex
```diff
@@ -61,6 +61,17 @@ \subsection{Abstract Syntax}
...
+\hypertarget{build-tyorcollection}{}
+The function
+\[
+  \buildtyorcollection(\overname{\parsenode{\Ntyorcollection}}{\vparsednode}) \;\aslto\; \overname{\ty}{\vastnode}
+  \cup \overname{\TBuildError}{\BuildErrorConfig}
+\]
+transforms a type annotation parse node $\vparsednode$ into a type AST node $\vastnode$.
+\ProseOtherwiseBuildError
+
+We define $\buildtyorcollection$ per the corresponding type in the following sections.
```
hrutvik, line 73: I'm not sure I follow where the rules for this are - I can see the `collection` case, but not the `ty` case

asllib/tests/ASLTypingReference.t/TypingRule.AnnotateOneArg.bad2.asl
```diff
@@ -1,5 +1,3 @@
-type MyCollection of collection;
```
hrutvik, line 1: Should we also have a negative regression test for this syntax in isolation?

asllib/tests/ASLTypingReference.t/TypingRule.TCollection.asl
```diff
@@ -1,25 +1,21 @@
...
-var z :CollectionWithoutFields;
+// var x of collection { a: bits(8), b: bits(16) } = MyCollection {a = Zeros{8}, b = Zeros{16}};
+// var - : collection { a: bits(8), b: bits(16) } = MyCollection {a = Zeros{8}, b = Zeros{16}};
 
 func main() => integer
 begin
     // The next declaration in comment is illegal:
     // local storage elements of collection types are forbidden.
-    // var - : MyCollection;
+    // var - : collectioin { a: bits(8), b: bits(16) };
```
hrutvik: typo: `collectioin`

asllib/Typing.ml
```diff
@@ -1351,9 +1353,31 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
         in
         let ses = SES.union ses_t ses_index in
         (T_Array (index', t') |> here, ses) |: TypingRule.TArray
+    (* Begin TCollection *)
+    | T_Collection _ when decl -> assert false
+    | T_Collection fields ->
```
hrutvik: Could we not leave this in with the other structured types?

asllib/doc/Types.tex
HadrienRenaud: I've put it with the other rule for `\buildtyorcollection`, do you think it's ok there?

hrutvik (approved):

PR #1337, opened 2025-06-21
[ASL] A control-flow analysis that takes the `noreturn` qualifier into account

Roman-Manevich:
When a function includes the `noreturn` qualifier, the static analysis must ensure that evaluating the function's body results in either a thrown exception or a dynamic error.


asllib/Typing.ml
```diff
@@ -3625,81 +3625,133 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
...
-    (* End *)
-
-    (* Begin ControlFlowFromStmt *)
+    type abs_config =
+      | Abs_Throwing
+        (* evaluation of a statement yielded an a thrown exception. *)
+      | Abs_Returning (* evaluation of a return statement completed normally. *)
+      | Abs_Continuing
+        (* evaluation of a non-return statement completed normally. *)
+      | Abs_DynError
```
hrutvik: Do we need to distinguish between `Abs_Throwing` and `Abs_DynError`? As far as I can see the checks always pair the two together

asllib/Typing.ml
```diff
@@ -3625,81 +3625,133 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
...
+    (* evaluation of a statement (possible [Unreachable]) resulted in a dynamic error *)
+
+    let abs_config_equal a b = a == b
```
hrutvik: The referential equality concerns me slightly - does OCaml make guarantees about uniqueness of enumerations like this?

asllib/Typing.ml
```diff
@@ -3625,81 +3625,133 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
...
+
+    (** Removes repeating configurations in [configs]. *)
+    let normalize configs =
```
hrutvik: It feels like we should use a `Set` of configurations

asllib/Typing.ml
```diff
@@ -3625,81 +3625,133 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
...
+
+    (** [approx_stmt senv s] returns a superset of the abstract configurations
+        that evaluating [s] with any environment consisting of [senv] would yield.
+        The analysis conservatively assumes that evaluating any expression
+        results in either a value or a thrown exception.
+    *)
+    let rec approx_stmt senv s =
+      let configs =
+        match s.desc with
+        | S_Pass | S_Decl _ | S_Assign _ | S_Assert _ | S_Print _ | S_Pragma _
```
hrutvik: - Could `S_Pass` be only `Abs_Continuing`?
- `S_Pragma` is interesting, as these are removed by type-checking. Would it be sound to make this the same as `S_Pass`, or are other side-effects possible?

asllib/Typing.ml
```diff
@@ -3625,81 +3625,133 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
...
+        | S_Throw _ -> [ Abs_Throwing; Abs_DynError ]
+        | S_Seq (s1, s2) ->
+            let configs1 = approx_stmt senv s1 in
+            let configs2 = approx_stmt senv s2 in
+            List.map
+              (fun c1 ->
+                match c1 with Abs_Continuing -> configs2 | _ -> [ c1 ])
+              configs1
+            |> List.flatten
+        | S_Cond (_, s1, s2) ->
```
hrutvik, line 3702: This needs to account for the condition expression I think

asllib/Typing.ml
```diff
@@ -3625,81 +3625,133 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
...
+            let configs1 = approx_stmt senv s1 in
+            let configs2 = approx_stmt senv s2 in
+            union configs1 configs2
+        | S_Repeat (body, _, _) | S_For { body } | S_While (_, _, body) ->
+            let configs_body = approx_stmt senv body in
+            union configs_body [ Abs_Throwing; Abs_DynError ]
```
hrutvik: nit: I wonder if it is worth factoring out something like `expr_configurations = [ Abs_Throwing; Abs_DynError ]` and reusing this when expressions are involved

asllib/Typing.ml
Roman-Manevich: fixed to do without physical equality

asllib/Typing.ml
Roman-Manevich: Initially I used only `Abs_Continuing` for `S_Pass`, but the extra precision is never needed to I merged it with the other cases. I'm now asserting that `S_Pragma` doesn't appear.

asllib/Typing.ml
Roman-Manevich: Sure, I'll merge them together.

asllib/Typing.ml
Roman-Manevich: I've merged these two configurations together.

asllib/Typing.ml
```diff
@@ -3625,82 +3625,158 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
...
+          (* evaluation of a return statement completed normally. *)
+        | Abs_Continuing
+      (* evaluation of a non-return statement completed normally. *)
+
+      let abs_config_id = function
+        | Abs_Abnormal -> 1
+        | Abs_Returning -> 2
+        | Abs_Continuing -> 3
+
+      let compare a b = abs_config_id b - abs_config_id a
```
HadrienRenaud: I believe this is what `Stdlib.compare` does. I think because the sum type is not recursive and not nested you can just use `Stdlib.compare` here.

asllib/Typing.ml
```diff
@@ -3625,82 +3625,158 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
...
+        The approximation of each statement is independent of the input environment,
+        which means it is also the fixpoint result, which in turn justifies the
+        soundness of approximating the loop statement and the, potentially recursive
+        subprogram calls.
+    *)
+    let rec approx_stmt senv s : AbsConfigSet.t =
+      let configs =
+        match s.desc with
+        | S_Pass | S_Decl _ | S_Assign _ | S_Assert _ | S_Print _ ->
+            abs_of_list [ Abs_Continuing; Abs_Abnormal ]
```
HadrienRenaud: Why would `S_Pass` be an `Abs_Abnormal`? Why don't we look into the expressions that make up the body of `S_Decl` or `S_Assign`?

asllib/Typing.ml
```diff
@@ -3625,82 +3625,158 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
...
-      match (t1, t2) with
-      | MayNotInterrupt, _ | _, MayNotInterrupt ->
-          MayNotInterrupt |: TypingRule.ControlFlowJoin
-      | AssertedNotInterrupt, t | t, AssertedNotInterrupt ->
-          t (* Assertion that the condition always holds *)
-      | Interrupt, Interrupt -> Interrupt
+      type t =
+        | Abs_Abnormal
+          (* evaluation of a statement yielded an a thrown exception or a dynamic error
+             (possibly due to calling Unreachable). *)
```
HadrienRenaud, line 3657: Why do we need to track down the dynamic errors except for Unreachable?

asllib/Typing.ml
```diff
@@ -3625,82 +3625,158 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
...
+      let () =
+        if false then
+          Format.eprintf "check_control_flow %s = %a@." f.name AbsConfigSet.pp
+            abs_configs
+      in
+      let allowed_abs_configs =
+        if ASTUtils.is_noreturn f then abs_of_list [ Abs_Abnormal ]
+        else
+          match f.return_type with
+          | None -> abs_of_list [ Abs_Continuing; Abs_Returning; Abs_Abnormal ]
```
HadrienRenaud: I find it strange not to put it in a variable but fine I guess

asllib/tests/ASLTypingReference.t/run.t
```diff
@@ -643,6 +643,7 @@ ASL Typing Tests / annotating types:
   0xffffffffffffffff
   $ aslref TypingRule.Subprogram.asl
   0x04
+  $ aslref --no-exec TypingRule.CheckControlFlow.noreturn.asl
```
HadrienRenaud, line 639: Can we add a test where a `noreturn` function does return something?

asllib/Typing.ml
Roman-Manevich: `Abs_Abnormal` also represents thrown exceptions, not just dynamic errors. 
In terms of naming, I it conveys the set of configurations it is representing and helps to show that `AbsConfig` is.a sound abstraction of the set of all configurations that `eval_stmt` may return.

asllib/Typing.ml
Roman-Manevich: This is an overapproximation. It's just that there's no need for the extra precision for `S_Pass`.
Regarding expressions, we always overapproximate them as returning either a value or an abnormal configuration, which is precise enough for our needs.

asllib/Typing.ml
Roman-Manevich: Put what in a variable?

asllib/tests/ASLTypingReference.t/run.t
Roman-Manevich: I want to postpone this to the point where the front-end supports `noreturn`.

asllib/Typing.ml
HadrienRenaud: I think pedagogically it makes sense to have `S_Pass` as continuing, but it's just an opinion.

For the expressions, do you think we could add something like `abs_of_expr` that would take an expression and return `abs_of_list [ Abs_Continuing; Abs_Abnormal ]`?

asllib/Typing.ml
HadrienRenaud: Ok - I guess I didn't see the point of representing faulting configurations but ok

asllib/Typing.ml
```diff
@@ -3625,82 +3625,158 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
...
+              IMap.find_opt call.name senv.global.subprograms
+            in
+            match opt_subprogram_entry with
+            | Some (f, _) ->
+                if ASTUtils.is_noreturn f then abs_of_list [ Abs_Abnormal ]
+                else abs_of_list [ Abs_Continuing; Abs_Abnormal ]
+            | None ->
+                assert (s.version = V0);
+                (* V0 subprograms in the shared pseudo-code. Return Top. *)
+                abs_of_list [ Abs_Continuing; Abs_Abnormal; Abs_Returning ])
```
HadrienRenaud: Can we keep them in the same order all the time?

```suggestion
                abs_of_list [ Abs_Continuing; Abs_Returning; Abs_Abnormal ])
```

asllib/Typing.ml
HadrienRenaud: Here `abs_of_list [ Abs_Continuing; Abs_Returning; Abs_Abnormal ]`, but some other similar are used a few times, could be named `abs_all` or `abs_top`

asllib/tests/ASLTypingReference.t/run.t
HadrienRenaud: Why? I think that's a recipe for forgetting some tests

asllib/Typing.ml
Roman-Manevich: Okay, I've added a variable for each subset of abstract configurations used.

asllib/Typing.ml
Roman-Manevich: done, in the new variables in `AbsConfigSet`.

asllib/tests/ASLTypingReference.t/run.t
Roman-Manevich: Okay, I'm adding a test.

asllib/Typing.ml
Roman-Manevich: Okay, I'll have `S_Pass` return only continuing and use `abs_of_expr`.

asllib/Typing.ml
HadrienRenaud: Thank you!

asllib/Typing.ml
```diff
@@ -3625,82 +3625,158 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
...
+        | S_Return _ -> abs_of_list [ Abs_Returning; Abs_Abnormal ]
```
hrutvik: Perhaps we should be a little more precise here - an `S_Return None` inside a procedure could effectively be considered `Abs_Continuing` if I understand correctly (as it's not a "true" return of a value)

asllib/Typing.ml
hrutvik: Similarly here, I think we should not permit `Abs_Returning` for procedures - that way we check for a bad `return` statement inside a procedure as part of control-flow analysis (I suppose this needs a new error message).

Edit: I see this is handled already during type-checking of `S_Return` before control-flow analysis triggers.

asllib/Typing.ml
hrutvik: Discussed offline as a possible future enhancement

asllib/doc/Syntax.tex
```diff
@@ -323,7 +323,7 @@ \section{ASL Grammar\label{sec:ASLGrammar}}
 
 \hypertarget{def-nqualifier}{}
 \begin{flalign*}
-\Nqualifier \derivesinline\ & \emptysentence \;|\; \Tpure \;|\;\Treadonly &
+\Nqualifier \derivesinline\ & \emptysentence \;|\; \Tpure \;|\;\Treadonly \;|\; \Tnoreturn &
```
hrutvik, line 326: I think this edit to `\Nqualifier` needs copying to `SuprogramDeclarations.tex` too

hrutvik (approved):

PR #1338, opened 2025-06-23
[asl] Introduce `pure` and `readonly` keywords

hrutvik:
These keywords optionally annotate subprograms:
- A `readonly` subprogram does not modify global storage elements or throw exceptions, but can read global storage elements and use non-determinism (`ARBITRARY`). It can only call `readonly` or `pure` subprograms.
- A `pure` subprogram does not read or modify global storage elements, throw exceptions, or use non-determinism. It can only call `pure` subprograms. It can therefore be considered `readonly` too.
- Both `pure` and `readonly` subprograms may use assertions (including asserting type conversions).

These keywords are used as the basis for a more coarse-grained side-effect analysis. In particular, they can be generalised to expressions (e.g. a `readonly` expression calls only `readonly` subprograms, and is otherwise subject to the same restrictions as `readonly` subprograms). Then:
- Assertions, `for`-loop bounds, and slices must involve only `readonly` subprograms.
- Static evaluation requires `pure` expressions.

This PR keeps the existing fine-grained side-effect analysis behind a flag, and weakens it in two specific ways:
- `for`-loop bounds may be non-deterministic
- assertions may be used in symbolically evaluable expressions


asllib/ASTUtils.ml
```diff
@@ -456,6 +456,12 @@ and bitfield_equal eq bf1 bf2 =
   | BitField_Type _, BitField_Simple _ ->
       false
 
+let func_qualifier_equal q1 q2 =
+  match (q1, q2) with
+  | Pure, Pure | Readonly, Readonly -> true
+  | (Pure | Readonly), _ -> false
```
HadrienRenaud: I think for this here `Stdlib.equal` is good enough

asllib/SideEffect.ml
```diff
@@ -161,13 +161,33 @@ let is_pure = function
...
 (* End *)
 
 (* SES = Side Effect Set *)
 module SES = struct
+  type purity = SE_Pure | SE_Readonly | SE_Impure
+
+  let purity_equal p1 p2 =
+    match (p1, p2) with
+    | SE_Pure, SE_Pure | SE_Readonly, SE_Readonly | SE_Impure, SE_Impure -> true
+    | (SE_Pure | SE_Readonly | SE_Impure), _ -> false
```
HadrienRenaud: Same here, I think `Stdlib.equal` is good enough

asllib/SideEffect.ml
```diff
@@ -161,13 +161,33 @@ let is_pure = function
...
+
+  let purity_satisfies desired input =
+    match (desired, input) with
+    | SE_Pure, SE_Pure -> true
+    | SE_Readonly, (SE_Pure | SE_Readonly) -> true
+    | _, _ -> false
```
HadrienRenaud, line 179: Why not `| SE_Impure, _ -> true`? Or maybe `| SE_Impure, _ -> assert false`?

asllib/tests/ASLTypingReference.t/run.t
```diff
@@ -144,8 +144,7 @@ ASL Typing Tests / annotating types:
   File TypingRule.AnnotateConstraint.bad.asl, line 4, characters 17 to 18:
     let t: integer{x..x+1} = 2; // illegal as 'x' is not constrained.
```
HadrienRenaud: Probably we need to edit this comment? Or change the keyword for `let`?

asllib/tests/side-effects.t/run.t
```diff
@@ -76,49 +76,58 @@
 
   $ aslref constant-func.asl
   $ aslref constant-func-read.asl
-  File constant-func-read.asl, line 9, characters 0 to 21:
-  constant C = foo (4);
-  ^^^^^^^^^^^^^^^^^^^^^
-  ASL Type error: expected constant-time expression, got foo(4), which produces
-    the following side-effects: [WritesGlobal "X"].
+  File constant-func-read.asl, line 5, character 2 to line 6, character 19:
+    X = 3;
```
HadrienRenaud: Looks like the test is quite badly named, or that this line should be transformed into a read of X

asllib/tests/side-effects.t/run.t
```diff
@@ -76,49 +76,58 @@
...
-  constant C = foo(3);
-  ^^^^^^^^^^^^^^^^^^^^
-  ASL Type error: expected constant-time expression, got foo(3), which produces
-    the following side-effects: [ReadsGlobal "K"].
+  File constant-func-sig-let.asl, line 3, character 0 to line 6, character 4:
+  pure func foo(x: integer {0..K}) => integer
+  begin
+    return x;
+  end;
+  ASL Type error: expected a pure expression/subprogram.
```
HadrienRenaud, line 130: Can I double check that this is the desired behaviour? Like 'pure' means 'constant time-frame'?

hrutvik: I believe so! As you say, purity is used in place of time frames, and then a `pure`-annotated subprogram is one that preserves purity (if given pure inputs, outputs are also considered pure). So in this case the read of `K` is problematic.

HadrienRenaud: Ok this makes sense thank you

asllib/SideEffect.ml
hrutvik: I've gone for `| SE_Impure, _ -> true`

asllib/AST.mli
```diff
@@ -421,6 +421,14 @@ type subprogram_body =
   | SB_ASL of stmt  (** A normal body of a subprogram *)
   | SB_Primitive of bool  (** Whether or not this primitive is side-effecting *)
 
+type func_qualifier =
+  | Pure
+      (** A `pure` subprogram does not read or modify mutable state. It can be
```
Roman-Manevich, line 426: A `pure` function...

asllib/Parser.mly
```diff
@@ -635,13 +645,14 @@ let decl :=
...
             recurse_limit;
+            qualifier;
             override;
             builtin = false;
           }
         }
     (* End *)
     (* Begin procedure_decl *)
-    | ~=override; FUNC; name=IDENTIFIER; ~=params_opt; ~=func_args; body=func_body;
+    | ~=qualifier; ~=override; FUNC; name=IDENTIFIER; ~=params_opt; ~=func_args; body=func_body;
```
Roman-Manevich, line 655: Does it make sense for a procedure to be pure or reasonly?
That would mean it has no effect, except maybe for some diagnostics.

asllib/AST.mli
hrutvik: Sorry just to check - do you mean replace "subprogram" by "function" in the comment here?

asllib/Parser.mly
hrutvik: I think as assertions are still permitted, it should be possible to call a `pure`/`readonly` procedure from a `pure`/`readonly` function - e.g. something like the following:
```
pure func MyProcedure()
begin
  assert ...;
end;

pure func MyFunc() => integer
begin
  MyProcedure();
  ...
  return ...;
end;
```

Roman-Manevich: Okay

asllib/AST.mli
Roman-Manevich: Given your answer to my second question - please ignore this one.

asllib/Typing.ml
```diff
@@ -2977,7 +2987,7 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
     (* Begin SAssert *)
     | S_Assert e ->
         let t_e', e', ses_e = annotate_expr env e in
-        let+ () = check_is_pure e ses_e in
+        let+ () = check_is_readonly e ses_e in
```
HadrienRenaud, line 2990: Is this intended?
What would be the behaviour of `assert (ARBITRARY: boolean)` then? Is that something that we are fine with?

asllib/Typing.ml
```diff
@@ -3007,10 +3017,8 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
         and limit', ses_limit =
           annotate_limit_expr ~warn:false ~loc env limit
         in
-        let+ () = check_is_pure start_e ses_start in
-        let+ () = check_is_deterministic start_e ses_start in
```
HadrienRenaud, line 3011: Is this intended?
Are we fine with the following:
```
for i = (ARBITRARY: integer {2, 3}) to (ARBITRARY: integer {0..10}) do
  print(i);
end;
```
Or is this a consequence of left-to-right?

asllib/Typing.ml
```diff
@@ -3480,6 +3488,18 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
...
+    match qualifier with
+    | None -> ok
+    | Some keyword -> (
+        match keyword with
+        | Pure ->
+            check_true (SES.is_pure ses) (fun () ->
+                fatal_from ~loc (Error.MismatchedPurity "pure"))
+        | Readonly ->
+            check_true (SES.is_readonly ses) (fun () ->
+                fatal_from ~loc (Error.MismatchedPurity "readonly")))
```
HadrienRenaud: Just stylistic:
```suggestion
    | Some Pure ->
        check_true (SES.is_pure ses) (fun () ->
          fatal_from ~loc (Error.MismatchedPurity "pure"))
    | Some Readonly ->
        check_true (SES.is_readonly ses) (fun () ->
          fatal_from ~loc (Error.MismatchedPurity "readonly"))
```

asllib/Typing.ml
hrutvik: This is intended, as a consequence of left-to-right as you say.

asllib/Typing.ml
hrutvik: This is intended - this behaviour was also permitted previously I believe, and this change only reflects the renaming of `check_is_pure` to `check_is_readonly`. In particular, the previous `SES.is_pure` was as follows, which seems in line with the new `readonly`:
```
    ISet.is_empty ses.local_writes
    && ISet.is_empty ses.global_writes
    && ISet.is_empty ses.thrown_exceptions
    && ISet.is_empty ses.calls_recursives
```

HadrienRenaud: Ok that makes sense thank you

asllib/Typing.ml
HadrienRenaud: Great

asllib/doc/LexicalStructure.tex
```diff
@@ -602,15 +603,15 @@ \subsection{Regular Tokens Tables}
...
-\end{center}
+% The following list represents keywords that are reserved for future use.
+% \begin{center}
+% \begin{tabular}{ll}
+% \textbf{Lexical Regular Expressions} & \textbf{Lexeme Action}\\
+% \hline
+% \texttt{"pure"}, \texttt{"readonly"} & $\lexicalerror$ \\
+% \hline
+% \end{tabular}
+% \end{center}
```
HadrienRenaud: Dead code?

hrutvik: Ah yes, I was thinking just in case we needed this table again - but that's what version control is for I suppose! Will delete unless @Roman-Manevich feels differently

Roman-Manevich: I'm okay with removing it.

asllib/tests/ASLDefinition.t/CaseStatement.where.asl
```diff
@@ -1,4 +1,4 @@
-func f(d: bits(2), a: integer) => integer
+readonly func f(d: bits(2), a: integer) => integer
```
Roman-Manevich, line 1: could be pure

asllib/tests/ASLSemanticsReference.t/SemanticsRule.CheckRecurseLimit.asl
```diff
@@ -1,4 +1,4 @@
-func factorial(n: integer) => integer recurselimit 11
+readonly func factorial(n: integer) => integer recurselimit 11
```
Roman-Manevich, line 1: could be pure

asllib/tests/ASLSemanticsReference.t/SemanticsRule.CheckRecurseLimit.limit_reached.asl
```diff
@@ -1,4 +1,4 @@
-func factorial(n: integer) => integer recurselimit 10
+readonly func factorial(n: integer) => integer recurselimit 10
```
Roman-Manevich, line 1: could be pure

asllib/tests/ASLSemanticsReference.t/SemanticsRule.CheckRecurseLimit.no_limit.asl
```diff
@@ -1,4 +1,4 @@
-func factorial(n: integer) => integer
+readonly func factorial(n: integer) => integer
```
Roman-Manevich, line 1: could be pure

asllib/tests/ASLSemanticsReference.t/SemanticsRule.FCall.asl
```diff
@@ -1,4 +1,4 @@
-func foo (x : integer) => integer
+readonly func foo (x : integer) => integer
```
Roman-Manevich, line 1: could be pure

asllib/tests/ASLSemanticsReference.t/run.t
```diff
@@ -251,7 +251,7 @@ ASL Semantics Tests:
   $ aslref SemanticsRule.CheckRecurseLimit.no_limit.asl
   File SemanticsRule.CheckRecurseLimit.no_limit.asl, line 1, character 0 to
     line 4, character 4:
-  func factorial(n: integer) => integer
+  readonly func factorial(n: integer) => integer
```
Roman-Manevich, line 254: could be pure

asllib/tests/ASLTypingReference.t/TypingRule.LDTyped.asl
```diff
@@ -1,6 +1,6 @@
 type MyT of integer;
 
-func foo (t: MyT) => integer
+readonly func foo (t: MyT) => integer
```
Roman-Manevich, line 3: could be pure

asllib/tests/ASLTypingReference.t/TypingRule.SESForSubprogram.asl
```diff
@@ -0,0 +1,14 @@
+pure func foo() // GlobalEffect(SE_Pure), Immutability(TRUE)
+begin
+  pass;
+end;
+
+readonly func bar() // GlobalEffect(SE_Readonly), Immutability(FALSE)
```
Roman-Manevich, line 6: could be pure

asllib/tests/ASLTypingReference.t/TypingRule.TIntParameterized.asl
```diff
@@ -1,9 +1,9 @@
-func foo {N} (x: bits(N)) => integer
+readonly func foo {N} (x: bits(N)) => integer
```
Roman-Manevich, line 1: could be pure

asllib/tests/ASLTypingReference.t/TypingRule.TIntParameterized.asl
```diff
@@ -1,9 +1,9 @@
...
 begin
   return N;
 end;
 
-func bar{M: integer}() => bits(M)
+readonly func bar{M: integer}() => bits(M)
```
Roman-Manevich, line 6: could be pure

asllib/tests/control-flow.t/if-return-return.asl
```diff
@@ -1,4 +1,4 @@
-func sign (n: integer) => integer
+readonly func sign (n: integer) => integer
```
Roman-Manevich, line 1: could be pure

asllib/tests/division.t/param-div-2.asl
```diff
@@ -1,4 +1,4 @@
-func foo {N} (bv: bits(N)) => bits(N)
+readonly func foo {N} (bv: bits(N)) => bits(N)
```
Roman-Manevich, line 1: could be pure

Roman-Manevich (commented):
Some readonly qualifiers could be turned to pure.

Roman-Manevich (approved):
Looks good to me!

PR #1341, opened 2025-06-25
Introduce diymicro

augustin64:
This PR introduces `diymicro`, a small sibling of `diyone` (separate codebase apart from `AArch64_compile.ml`) with a support for register events between edges. This allows us to introduce new edges that target iico dependencies.

For instance, `diymicroenum 'iico[cas:ok]' 'iico[cas:no]' -output generated_cas` is able to generate a certain amount of [catalogue's CAS tests](https://github.com/herd/herdtools7/tree/master/catalogue/aarch64-cas/tests).


diymicro/lexer.mll
```diff
@@ -0,0 +1,42 @@
...
+  | "Ws" { WS }
+  | "Po" { PO }
+  | "Dp" { DP }
+  | "DMB"    { DMB }
+  | "Rf-reg" { RF_REG }
+  | "iico"   { IICO }
+
+  | 'R' { RM }
+  | 'W' { WM }
+  | "r" { R  }
```
ShaleXIONG: inconsistent use of " and '

diymicro/compile.ml
```diff
@@ -0,0 +1,319 @@
...
+  dump_code instructions channel;
+  dump_final stl channel
+
+let to_channel annot_edges ?(name = "test") channel =
+  C.reset_proc_count ();
+  A.reset_loc_count ();
+  let cycle = Cycle.make_cycle annot_edges in
+  let prog = prog_of_cycle cycle in
+
+  let instructions = List.map (fun (a, _) -> a) prog in
```
ShaleXIONG: `List.split` will do here.

diymicro/iico.ml
```diff
@@ -0,0 +1,549 @@
...
+          let rt, st = A.next_reg st in
+          [A.mov rs write_value], rs, rt, st
+      | _ -> (
+          let src_reg, v_opt =
+            match dep with
+            | DepReg (r, v) -> r, v
+            | _ -> Warn.fatal "Event has not forwarded any register"
+          in
+          match src with
+          | "Rt" ->
```
HadrienRenaud: Why can `SWP` start with `Rt`? I guess why not try this but is there a specific reason?

diymicro/iico.ml
```diff
@@ -0,0 +1,549 @@
...
+       SWP(Rs, Rt, Rn)
+         Rs=write_value (to be stored in [Rn])
+         Rt=no specific value (0) (will get the value of [Rn])
+        [Rn]=read_value needs no action
+    *)
+    let pre_ins, rs, rt, st =
+      match src with
+      | "M" ->
+          (match dep with
+          | DepNone -> ()
```
HadrienRenaud: Style nitpick:

```suggestion
      match src, dep with
      | "M", DepNone ->
```

and then later:
```ocaml
      | "Rt", DepReg (src_reg, v_opt) ->
```

diymicro/utils.ml
```diff
@@ -0,0 +1,22 @@
+(** Print something if verbose level is high enough (0:Verbose, 1:Debug) *)
+let verbose_print level string =
+  if level <= !Config.verbose then output_string stderr string
+
+(** Removes element from the option monad, raises Not_found if appropriate *)
+let unsome = function
```
ShaleXIONG: There is a std function `Optional.get`.

ShaleXIONG (commented):
Look good to me in general, I add a few detail comments.

diymicro/iico.ml
```diff
@@ -0,0 +1,549 @@
...
+    ins, DepReg (rd, Some final_expected_val), st
+
+  let repr ok = "csel:" ^ if ok then "ok" else "no"
+end
+
+module Swp = struct
+  let repr = "swp"
+
+  let compile src dst st dep src_evt dst_evt =
+    let rn, _, read_value, write_value, st =
```
HadrienRenaud: I don't understand why the second argument is needed

diymicro/iico.ml
```diff
@@ -0,0 +1,549 @@
...
+      let pre_ins, rs, rt, rn, st =
+        match src with
+        | "M" ->
+            let rs, st = A.next_reg st in
+            let rt, st = A.next_reg st in
+            let st = A.add_condition st rs read_value in
+            (* if the src event is memory, the read is significant in all cases *)
+            [A.mov rt rt_value; A.mov rs rs_value], rs, rt, rn, st
+        | _ -> (
+            let src_reg, v_opt =
```
HadrienRenaud: Same, can we merger the 2 pattern matchings?

diymicro/iico.ml
```diff
@@ -0,0 +1,549 @@
...
+  List.iter
+    (fun ok ->
+      add_iico
+        {
+          instruction_name = CasNoMem.repr ok;
+          to_edge =
+            (fun src dst ->
+              {
+                repr = "";
+                compile_edge = CasNoMem.compile ok src dst;
```
HadrienRenaud: Why can `CasNoMem` take a `ok` boolean?

diymicro/iico.ml
```diff
@@ -0,0 +1,549 @@
...
+
+module C = struct
+  open Cycle
+end
+
+module A = struct
+  include AArch64_compile
+end
+
+module CasNoMem = struct
```
HadrienRenaud: Can we add a little comment explaining what the difference between `Cas` and `CasNoMem` is?

diymicro/edge.ml
```diff
@@ -0,0 +1,168 @@
...
+    A.state ->
+    node_dep ->
+    event_data option ->
+    event_data option ->
+    int A.kinstruction A.kpseudo list * node_dep * A.state;
+  direction : direction * direction;
+  ie : int_ext;
+  sd : sd;
+  significant_source : bool;
+  significant_dest : bool;
```
HadrienRenaud, line 50: What do those correspond to?

diymicro/edge.ml
```diff
@@ -0,0 +1,168 @@
...
+      (* Expected to be "" in an add_iico call, replaced later by get_iico *)
+  compile_edge :
+    A.state ->
+    node_dep ->
+    event_data option ->
+    event_data option ->
+    int A.kinstruction A.kpseudo list * node_dep * A.state;
+  direction : direction * direction;
+  ie : int_ext;
+  sd : sd;
```
HadrienRenaud: Can we add a little comment to explain what those are?

diymicro/edge.ml
```diff
@@ -0,0 +1,168 @@
...
+  significant_source : bool;
+  significant_dest : bool;
+}
+
+type iico = {
+  instruction_name : string;
+  to_edge : string -> string -> iico_edge;
+  inputs : string list;
+  outputs : string list;
+}
```
HadrienRenaud: Can you write a little comment on why you're not using variants? (like polymorphic variants)

diymicro/iico.ml
augustin64: `NoMem` just means no memory event:
`CasNoMem` is the same as `Cas`, before `Rm true` and `Wm true` were introduced (hence direction is always `RegEvent, RegEvent`). As the tests each generates detect the same herd variants, I just removed `CasNoMem` in the end.

diymicro/edge.ml
augustin64: If the src or dst event is a memory read, we mark it as significant to add a condition in the finals. I don't know if the denomination is the same in diy, but I think those are called "significant reads" in _Fences in Weak Memory Models_.

I will add a comment too.

Makefile
```diff
@@ -26,6 +27,9 @@ HERD_DIYCROSS_REGRESSION_TEST = _build/default/internal/herd_diycross_regression
 HERD_CATALOGUE_REGRESSION_TEST = _build/default/internal/herd_catalogue_regression_test.exe
 BENTO                         = _build/default/tools/bento.exe
 
+DIYMICRO_EDGES = $(shell $(DIYMICROENUM) -list-iico | sed -n 's/^iico\[\([^ ]*\).*/iico[\1]/p')
```
HadrienRenaud: Will this work even if `diymicroenum` is not built yet? Maybe yes because of the assignment style but if you could confirm + leave a comment that'd be great

diymicro/compile.ml
```diff
@@ -0,0 +1,318 @@
...
+
+type prog = int A.kinstruction A.kpseudo list * A.state
+(** Instructions and meta (eg initials & finals) relevant to a single proc *)
+
+(** Returns a list of lists of nodes (cycles), splitted by proc *)
+let split_by_proc (cycle : C.t) =
+  let first_node =
+    C.find_first cycle (fun c ->
+        c.C.prev.C.source_event.C.proc <> Some (C.Proc 0)
+        && c.C.source_event.C.proc = Some (C.Proc 0))
```
HadrienRenaud: Maybe would be good to open C here? at least locally?

diymicro/compile.ml
```diff
@@ -0,0 +1,318 @@
...
+  include AArch64_compile
+end
+
+module C = struct
+  include Cycle
+end
+
+module E = struct
+  include Edge
+end
```
HadrienRenaud: Why not simply declaring them as such?

```suggestion
module A = AArch64_compile
module C = Cycle
module E = Edge
```

Or why not using them without any alias in the rest?

diymicro/edge.ml
```diff
@@ -0,0 +1,170 @@
...
+
+let iico_ht = Hashtbl.create 10
+let add_iico iico = Hashtbl.add iico_ht iico.instruction_name iico
+
+(** look-up an iico by instruction name *)
+let get_iico s = Hashtbl.find iico_ht s
+
+let iico_to_edge iico src dst =
+  let _ = List.find (( = ) src) iico.inputs in
+  let _ = List.find (( = ) dst) iico.outputs in
```
HadrienRenaud: Can we use the non polymorphic equality here?

Also I guess it would be more explicit to say
```ocaml
  let () = assert (List.exists (String.equal src) iico.inputs) in
```
Or am I mis-understanding the intend here?

Makefile
augustin64: For now, it works in the CI because `make build` is ran before `make test` but I need a workaround for the general case (and `make build` raises a warning for now).

diymicro/compile.ml
augustin64: Thank you for the syntax. I find it a bit heavy to have the full module name (for `AArch64_compile` at least), but I'm still happy to know where things come from.

I will replace that with local `open` where needed

HadrienRenaud: Great thank you!

PR #1343, opened 2025-06-25
[asl][reference] fixes

Roman-Manevich:
Fixed for comments raised during a PDF review.

### DIFF
--- a/asllib/tests/ASLDefinition.t/Accessor.asl
+++ b/asllib/tests/ASLDefinition.t/Accessor.asl
@@ -4,7 +4,7 @@ var R : array [[32]] of bits(64);
 // Accessor, X
 accessor X(regno: integer{0..31}) <=> value: bits(64)
 begin
-  getter
+  readonly getter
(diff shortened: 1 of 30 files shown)


asllib/tests/ASLDefinition.t/Accessor.asl
```diff
@@ -4,7 +4,7 @@ var R : array [[32]] of bits(64);
 // Accessor, X
 accessor X(regno: integer{0..31}) <=> value: bits(64)
 begin
-  getter
+  readonly getter
```
hrutvik, line 7: Just to check - what was the reason for introducing all the `readonly` getters here?

hrutvik (approved):

asllib/tests/ASLDefinition.t/Accessor.asl
Roman-Manevich: I think most getters only read and so this is a good example and a rule of thumb for getters.

PR #1345, opened 2025-06-30
[asl] Permit declarations of multiple uninitialised global variables

hrutvik:
Mirroring declarations of multiple uninitialised local variables:
```
var a, b, c : integer;

func foo()
begin
  var d, e, f : integer;
end;
```


asllib/Parser.mly
```diff
@@ -76,6 +76,13 @@ let make_ldi_vars (xs, ty) =
   in
   List.map make_one xs |> stmt_from_list |> desc
 
+let make_global_vars (names, ty) =
```
Roman-Manevich: I wonder if we should move this and similar helper functions to `desugar.mli`/`desugar.ml`

asllib/doc/GlobalStorageDeclarations.tex
```diff
@@ -129,6 +130,27 @@ \section{Abstract Syntax\label{sec:GlobalStorageDeclarationsAbstractSyntax}}
...
+\begin{mathpar}
+\inferrule[global\_uninit\_multiple\_vars]{
+  \buildclist[\buildidentity](\vids) \astarrow \vidasts \\
+  {
+  i \in \listrange(\vidasts): \vdecl_i \eqdef
+    \DGlobalStorage\left(
+      \left\{
+      \begin{array}{rcl}
+        \GDkeyword&:& \GDKVar,\\
+        \GDname&:& \vidasts_i,\\
```
Roman-Manevich: `\vidasts_i` => `\vidasts[i]`

asllib/Parser.mly
hrutvik: I've moved the `make_ldi_vars` and `make_global_vars` - some of the others don't feel like a good fit, but perhaps we can follow up in a future PR.

PR #1346, opened 2025-06-30
[asl] Syntax for empty structured type declarations

hrutvik:
Require declarations of empty structured types to explicitly declare an empty field list:
```
type Foo of record;    // ERROR
type Bar of exception; // ERROR
var x : collection;    // ERROR

type Foo of record{-};    // OK
type Bar of exception{-}; // OK
var x : collection{-};    // OK
```
Gate this change behind a flag: `--allow-empty-structured-type-declarations`


Roman-Manevich (approved):
Looks good to me (Bonga Bonga :-)

PR #1347, opened 2025-06-30
[asl] Convert function-like statements into keywords

hrutvik:
In particular:
- `Unreachable();` -> `unreachable;`
- `print(a, b, c);` -> `print a, b, c;`
- `println(d, e, f);` -> `println d, e, f;`

Allow backwards compatibility via a flag: `--allow-function-like-statements`


Roman-Manevich (approved):
Looks good to me!
- `StandardLibrary.tex` has `\verb|Unreachable()|` and there are occurrences of `\texttt{Unreachable` in `ErrorCodes.tex`, `Specifications.tex`, and `Statements.tex`, which should probably be replaced by the `\unreachablestatementterm` macro.
- Also `LexicalStructure.tex` has `\texttt{"Unreachable`
- Rerun `doclint.py` with the `-m` flag to update the console outputs that have error messages referring to `Unreachable`.

PR #1350, opened 2025-07-01
[asl][reference] Always make with -m option

Roman-Manevich:
Always use the `-m` flag to make sure that the console outputs are synchronized with the ASL Reference.
This requires building and installing aslref so that it can be run from `doclint.py`.
Also a `--aslref` option was added to `doclint.py` to point to the aslref binary path.

### DIFF
--- a/asllib/doc/extended_macros.py
+++ b/asllib/doc/extended_macros.py
@@ -3,9 +3,10 @@
 import os, fnmatch, subprocess, shlex
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
+from pathlib import Path
 import re
 
-ASLREF_EXE = "aslref"
+ASLREF_EXE: str = "aslref"
 
 debug = True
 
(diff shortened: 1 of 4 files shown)


asllib/doc/extended_macros.py
```diff
@@ -215,23 +216,25 @@ def transform(cls, block_lines: list[str]) -> list[str]:
...
             .replace(cls.CONSOLE_CMD, "")
-            .replace("aslref", ASLREF_EXE)
             .replace("\\definitiontests", "../tests/ASLDefinition.t")
             .replace("\\syntaxtests", "../tests/ASLSyntaxReference.t")
             .replace("\\typingtests", "../tests/ASLTypingReference.t")
             .replace("\\semanticstests", "../tests/ASLSemanticsReference.t")
             .replace("\n", "")
             .strip()
         )
+        executable_command = command.replace("aslref", ASLREF_EXE)
```
hrutvik, line 238: Why are two `replace` calls needed?

hrutvik (approved):

asllib/doc/extended_macros.py
Roman-Manevich: We want the `command` version to be included in the LaTeX output without any user-specific path and `executable_command` to be the one that is actually run. I added a comment explaining this.

PR #1353, opened 2025-07-03
[asl][reference] fixed SemanticsRule.GetIndex and SemanticsRule.SetIndex

Roman-Manevich:
The rules missed the cases where the index could be out of bounds, resulting in a dynamic error.

### DIFF
--- a/asllib/doc/SemanticsUtilities.tex
+++ b/asllib/doc/SemanticsUtilities.tex
@@ -554,13 +554,14 @@ In reference to \listingref{semantics-leslice}, we have the following applicatio
 \end{mathpar}
 
 \SemanticsRuleDef{GetIndex}
-\ProseParagraph
 The relation
 \hypertarget{def-getindex}{}
 \[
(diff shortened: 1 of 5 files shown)


asllib/doc/SemanticsUtilities.tex
```diff
@@ -575,59 +576,101 @@ \chapter{Semantics Utility Rules\label{chap:SemanticsUtilityRules}}
 \end{itemize}
 \ASLListing{Reading a vector at an index}{GetIndex}{\semanticstests/SemanticsRule.GetIndex.asl}
 
+\ProseParagraph
+\OneApplies
+\begin{itemize}
+  \item \AllApplyCase{ok}
+  \begin{itemize}
+    \item $\vi$ is less than the number of elements in $\vvec$;
```
hrutvik, line 584: One thing that is maybe worth re-emphasising: the number of elements in `\vvec` is `k + 1`, so the distinctions between `<`/`\leq` and `>`/`\geq` get a bit confusing. One suggestion could be using `|\vvec|` across these rules, rather than numbering elements `0..k`

PR #1354, opened 2025-07-03
[asl] Remove unique identifiers for parameters

hrutvik:
These do not seem to be used significantly, and are not documented in the ASL Reference.


hrutvik:
Potentially of interest to @slobodova and @solswords

PR #1357, opened 2025-07-08
[all] Add support for AArch64 ST2G instruction

relokin:


murzinv:
LGTM. I can run `B015` with #1042, `B016` would require some extra work to make it run due to offsets in final clause, but that is not show stopper. Once PR merged adding support in #1042 would be just oneliner :wink:

PR #1363, opened 2025-07-11
[asl] Infer coarse grained side-effects for v0 functions

HadrienRenaud:
Motivation comes from this example (that I've added to the testsuite):

Say you have a readonly function `f0` written in ASL0:
```
integer f0(x: integer)
  return x + (integer UNKNOWN);
```

And you want to call it from a readonly function `f1` written in ASL1:
```
readonly func f1(x: integer) => integer
begin
  return f0(x);
end;
```

This PR adapts the readonly/pure function annotations to work in this context

### DIFF
--- a/asllib/tests/lexer.t
+++ b/asllib/tests/lexer.t
@@ -139,6 +139,7 @@ Forbidden patterns
 
   $ cat >forbiddenhex02.asl <<EOF
   > let x = 0x_12;
+  > EOF
   $ aslref forbiddenhex02.asl
   File forbiddenhex02.asl, line 1, characters 8 to 11:
   let x = 0x_12;
@@ -148,6 +149,7 @@ Forbidden patterns
 
   $ cat >forbiddenhex03.asl <<EOF
   > let x = 0x12h12;
+  > EOF
   $ aslref forbiddenhex03.asl
   File forbiddenhex03.asl, line 1, characters 8 to 13:
   let x = 0x12h12;
@@ -158,6 +160,7 @@ Forbidden patterns
   $ cat >forbiddenhex04.asl <<EOF
   > let x_foo = 213;
   > let x = 0x_foo;
+  > EOF
   $ aslref forbiddenhex04.asl
   File forbiddenhex04.asl, line 2, characters 8 to 11:
   let x = 0x_foo;
@@ -177,6 +180,7 @@ Forbidden patterns
 
   $ cat >forbiddenreal02.asl <<EOF
   > let x = 1._12;
+  > EOF
   $ aslref forbiddenreal02.asl
   File forbiddenreal02.asl, line 1, characters 8 to 11:
   let x = 1._12;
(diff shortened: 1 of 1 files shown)


asllib/SideEffect.ml
```diff
@@ -184,6 +184,11 @@ module SES = struct
     | SE_Readonly, _ | _, SE_Readonly -> SE_Readonly
     | SE_Pure, SE_Pure -> SE_Pure
 
+  let purity_to_string = function
```
hrutvik, line 187: Perhaps we could share this function between this file and `PP.ml`?

asllib/StaticEnv.ml
```diff
@@ -185,8 +185,12 @@ let mem_constants env x =
 let add_subprogram name func_def ses env =
   let () =
     if false then
-      Format.eprintf "@[Adding func %s with side effects:@ @[%a]@]@." name
-        SideEffect.SES.pp_print ses
+      Format.eprintf
+        "@[Adding func %s with side effects:@ @[%a]@ and@ qualitifier@ %s@]@."
```
hrutvik: Nit: typo, "qualitifier"

asllib/Typing.ml
```diff
@@ -3818,21 +3818,28 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
...
     let () =
       if false then
         Format.eprintf "@[<v 2>For program %s, I got side-effects:@ %a@]@."
           f.name SES.pp_print ses
     in
+    let qualifier =
+      if C.fine_grained_side_effects then f.qualifier
+      else if func_version f == V0 then
+        if SES.fine_grained_is_pure ses then
+          if SES.max_time_frame ses = TimeFrame.Constant then Some Pure
```
hrutvik: I'm wondering whether it would be worth trying to use the coarse-grained analysis here to derive the keyword - could we take e.g. the minimum global purity at this point?

asllib/Typing.ml
```diff
@@ -3818,21 +3818,28 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
...
-    let ses = SES.set_purity_for_subprogram f.qualifier ses in
     let () =
       if false then
         Format.eprintf "@[<v 2>For program %s, I got side-effects:@ %a@]@."
           f.name SES.pp_print ses
     in
+    let qualifier =
+      if C.fine_grained_side_effects then f.qualifier
+      else if func_version f == V0 then
+        if SES.fine_grained_is_pure ses then
```
Roman-Manevich: I think it's better to move this to a new function, e.g., `infer_v0_qualifier`

Roman-Manevich (approved):
Looks good to me!

asllib/SideEffect.ml
HadrienRenaud: I guess we could, but why?

Roman-Manevich (commented):
Maybe also include a negative example? A call from a readonly/pure function to a function that is not inferred to be pure/readonly?

hrutvik (approved):

PR #1367, opened 2025-07-15
[asl] Implement total po in ASL

HadrienRenaud:
Following #1222, evalation order in ASL is left-to-right. This means that the program order in ASL is total.

We keep the old `partial_po` infrastructure to implement this total program order for technical reasons: we need a structure to keep track of PO during the evaluation of a single pseudo-instruction (indeed an ASL program is a single instruction as far as herd's machinery is concerned).


herd/monad.mli
```diff
@@ -253,6 +253,11 @@ module type S =
         [para_bind_output_right]. Unlike the later, a [partial_po] arrow is put
         between the two structures. *)
 
+    val asl_para : 'a t -> 'b t -> ('a * 'b) t
+    (** [asl_para s1 s2] returns the parralel composition of the event
```
Roman-Manevich, line 257: parallel

asllib/Interpreter.ml
```diff
@@ -287,9 +287,9 @@ module Make (B : Backend.S) (C : Config) = struct
   (* To give name to rules *)
   let ( |: ) = C.Instr.use_with
 
-  (* Product parallel *)
-  (* ---------------- *)
-  let ( and* ) = B.prod_par
+  (* Product in po order *)
```
Roman-Manevich, line 290: "po order" stands for "program order order"?
If so, how about "Product in program order (po)"?

PR #1371, opened 2025-07-18
Fix array index

murzinv:
Depending on context array indexing can be handled differently - in
some cases as byte offset in others as element offset.

Consider an example

```
AArch64 T
{
  int x[2];
  0:X0=x[1]; (* Expect offset to 2nd element of x *)
}
P0;
MOV W1,#1  ;
STR W1,[X0];

forall x[1]=1
```

which upsets herd7

```
$ herd7 T.litmus
Warning: File "T.litmus": mixed-size test rejected (symbol [x+1]), consider option -variant mixed (User error)
```

However, we can make it happy again with a slight change

```
AArch64 T
{
  int x[2];
  0:X0=x[4]; (* Now byte offset *)
}
P0;
MOV W1,#1  ;
STR W1,[X0];

forall x[1]=1 (* Note here it is still element offset *)
```

which produces expected result

```
$ herd7 T.litmus
Test T Required
States 1
x[1]=1;
Ok
Witnesses
Positive: 1 Negative: 0
Condition forall (x[1]=1)
Observation T Always 1 0
Time T 0.02
Hash=0d36b5ef2a3f2027bfb76f23a9c9a2fc
```

It seems that byte offset comes from the rule in `lib/stateParser.mly`

```
maybev_notag:
...
| NAME LBRK NUM RBRK { mk_sym_with_index $1 $3 }
```

At that stage we do not have type information yet, so
`mk_sym_with_index` just set `.offset` as is.

The way to handle it is to fixup offset when type information become
available. That also means that we need to build type environment
ahead of the state.


maranget:
Thanks @murzin. Would you please rebase on master, so that I can merge this PR?

maranget:
Merged, thanks @murzinv.

PR #1375, opened 2025-07-22
[herd] [litmus] Refactoring the help command

hazel394:
This PR improves the herd -help command to make it easier for users to discover and understand available options when using herd.
- Options are grouped into separate lists (setup, runmode, filter, graph content, graph presentation) to support displaying them by category in the help output.
- Each group of options and details about their behavior can be displayed using the command `-help <category>` .
- The -help command displays a short list of commonly used options and a summary of all option categories.
- The --help command (or -help all) displays a full list of all options separated by category.
- If -help or --help appears in the command line, all other arguments are ignored and only the help message is shown.

This PR also updates the presentation of variant tags in the herd and litmus -help commands.
- Missing variants from the help output tag list have been added, along with an identity function to prompt developers to add new variants into the list.
- Variants are now grouped into two categories: modes of operation and architectural features.
- In herd there is a dedicated help page (-help variant) that lists all variants separated by category. The option to display the list of variants with their descriptions is guarded by a conditional, which should be removed when the blank descriptions are populated.

### DIFF
--- a/herd/herd.ml
+++ b/herd/herd.ml
@@ -47,7 +47,7 @@ let gen_model_opt s =
     | None -> false
     | Some _ as m -> model :=  m ; true)
     Model.tags
-    (sprintf " select model, defaults %s, %s, %s, %s, %s, %s, %s"
+    (sprintf " Select model, defaults %s, %s, %s, %s, %s, %s, %s"
        (pp_default_model x86)
        (pp_default_model x86_64)
        (pp_default_model ppc)
@@ -56,76 +56,30 @@ let gen_model_opt s =
        (pp_default_model riscv)
        (pp_default_model Archs.c))
 
-let options = [
-(* Basic *)
-  ("-version", Arg.Unit
-     (fun () -> printf "%s, Rev: %s\n" Version.version Version.rev ; exit 0),
-   " show version number and exit") ;
-  ("-libdir", Arg.Unit (fun () -> print_endline !Opts.libdir; exit 0),
-    " show installation directory and exit");
-  ("-set-libdir", Arg.String (fun s -> Opts.libdir := s),
-    "<path> set installation directory to <path>");
-  ("-v", Arg.Unit (fun _ -> incr verbose),
-   "<non-default> show various diagnostics, repeat to increase verbosity");
-  ("-q", Arg.Unit (fun _ -> verbose := -1; debug := Debug_herd.none),
-   "<default> do not show diagnostics");
-  ("-I", Arg.String (fun s -> includes := !includes @ [s]),
-   "<dir> add <dir> to search path");
-  parse_bool "-exit" Opts.exit_if_failed "exit in case of failure";
-  parse_float_opt "-timeout" Opts.timeout "timeout (CPU time)";
-  ("-conf",
-   Arg.String load_config,
-   "<name> read configuration file <name>") ;
-  ("-bell",
-   Arg.String (fun x -> Opts.bell := (Some x)),
-   "<name> read bell file <name>") ;
-  ("-macros",
-   Arg.String (fun x -> Opts.macros := (Some x)),
-   "<name> read macro (.def) file <name>") ;
-  ("-o", Arg.String
-     (fun s -> match s with
-     | "-" -> outputdir := PrettyConf.StdoutOutput
-     | _ -> outputdir := PrettyConf.Outputdir s),
-   "<dir> generated files will go into <dir>, default: do not generate") ;
-  ("-suffix", Arg.String (fun s -> suffix := s),
-   "<suf> add <suf> at the end of the base of generated files") ;
-  parse_bool "-dumpes" Opts.dumpes "dump event structures";
-  begin let module ParseView = ParseTag.Make(View) in
-  ParseView.parse_opt "-view" PP.view
-    "fork specified viewer to show output graphs" end ;
-  ( "-gv",
-    Arg.Unit (fun _ -> PP.view := Some View.GV),
-    "<non-default>  alias for -view gv") ;
-  ( "-evince",
-    Arg.Unit (fun _ -> PP.view := Some View.Evince),
-    "<non-default>  alias for -view evince") ;
-  ( "-preview",
-    Arg.Unit (fun _ -> PP.view := Some View.Preview),
-    "<non-default>  alias for -view preview") ;
-  ("-unroll",
-   Arg.Int (fun x -> unroll := Some x),
-   sprintf "<int> branch unrolling upper limit, default ASL: %i, others: %i"
-     (unroll_default `ASL)  (unroll_default `Others));
-  parse_bool "-hexa" PP.hexa "print numbers in hexadecimal";
-(* undocumented *)
-  ("-switch",
-   Arg.Unit (fun () -> Misc.switch := true),
-   "switch something") ;
-  ("-web",
-   Arg.Unit (fun () -> load_config "web.cfg")," alias for -conf web.cfg");
-  ("-c11",
-   Arg.Unit (fun () -> load_config "cpp11.cfg")," alias for -conf cpp11.cfg");
-  parse_tags
-    "-debug"
-    (fun tag -> match Debug_herd.parse !debug tag with
+
+(*Common options*)
+let variant_option =
+  begin
+    let module ParseVariant = ParseTag.MakeS(Opts.OptS) in
+    ParseVariant.parse "-variant" variant
+      "<tags> select an architecture variation. Use '-help variant' for the list of tags" end 
+let showevents_option =
+  parse_tag "-showevents"
+    (fun tag -> match PrettyConf.parse_showevents tag with
     | None -> false
-    | Some t -> debug := t ; true)
-    Debug_herd.tags
-    "show debug messages for specific parts" ;
-  parse_bool "-morefences" (ref false) "does nothing (deprecated)" ;
-(* Engine control *)
-  gen_model_opt "-model";
-  gen_model_opt "-cat";
+    | Some t -> PP.showevents := t ; true)
+     PrettyConf.tags_showevents
+     (sprintf "Select events shown in figures, default %s"
+       (PrettyConf.pp_showevents !PP.showevents)) 
+let show_option =
+  parse_tag "-show"
+    (fun tag -> match PrettyConf.parse_show tag with
+    | None -> false
+    | Some t -> show := t ; true)
+    PrettyConf.tags_show
+    (sprintf "Executions shown in figure, default %s"
+       (PrettyConf.pp_show !show)) 
+let through_option =
   parse_tag
     "-through"
     (fun tag -> match Model.parse_through tag with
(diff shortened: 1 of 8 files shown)


herd/herd.ml
```diff
@@ -354,18 +286,215 @@ let options = [
...
+  ("", Arg.Unit (fun () -> ()), "\n");
+  ("-version", Arg.Unit
+     (fun () -> printf "%s, Rev: %s\n" Version.version Version.rev ; exit 0),
+   " show version number and exit") ;
+  ("-libdir", Arg.Unit (fun () -> print_endline !Opts.libdir; exit 0),
+    " show installation directory and exit");
+  ("-set-libdir", Arg.String (fun s -> Opts.libdir := s),
+    "<path> set installation directory to <path>");
+  ("-v", Arg.Unit (fun _ -> incr verbose),
+   "<non-default> show various diagnostics, repeat to increase verbosity");
```
relokin:
```suggestion
   "show various diagnostics, repeat to increase verbosity");
```
I find `<non-default>` a little confusing, I think the idea was to indicate that verbosity is not enabled by default. But this syntax is also used to indicate that the option take arguments. Same applies for `-gv`, `-evince` and `-preview`

herd/herd.ml
```diff
@@ -354,18 +286,215 @@ let options = [
...
+
+let setup_options = Arg.align ~limit:40 [
+   "-help", Arg.Unit (fun () -> ()), "" ;
+  "--help", Arg.Unit (fun () -> ()), "";
+  ("", Arg.Unit (fun () -> ()), "\n");
+  ("-version", Arg.Unit
+     (fun () -> printf "%s, Rev: %s\n" Version.version Version.rev ; exit 0),
+   " show version number and exit") ;
+  ("-libdir", Arg.Unit (fun () -> print_endline !Opts.libdir; exit 0),
+    " show installation directory and exit");
```
relokin: I know you're copying over the old text, but I think the old text is not very helpful. What about?
```suggestion
    " show the path used to search for the top level model and included .cat files");
```
Is there anything else other than cat file this is used for? @artkhyzha what do you think?

herd/herd.ml
```diff
@@ -354,18 +286,215 @@ let options = [
...
+  ("-set-libdir", Arg.String (fun s -> Opts.libdir := s),
+    "<path> set installation directory to <path>");
```
relokin: I think this needs to change accordingly.

herd/herd.ml
```diff
@@ -354,18 +286,215 @@ let options = [
...
+   Arg.String (fun x -> Opts.bell := (Some x)),
+   "<name> read bell file <name>") ;
+  ("-macros",
+   Arg.String (fun x -> Opts.macros := (Some x)),
+   "<name> read macro (.def) file <name>") ;
+  ("-o", Arg.String
+     (fun s -> match s with
+     | "-" -> outputdir := PrettyConf.StdoutOutput
+     | _ -> outputdir := PrettyConf.Outputdir s),
+   "<dir> generated files will go into <dir>, default: do not generate") ;
```
relokin:
```suggestion
   "<dir> generated files will go into <dir>. No default, if not set herd7 does not generate any files.") ;
```

herd/herd.ml
```diff
@@ -354,18 +286,215 @@ let options = [
...
+  ("-libdir", Arg.Unit (fun () -> print_endline !Opts.libdir; exit 0),
+    " show installation directory and exit");
+  ("-set-libdir", Arg.String (fun s -> Opts.libdir := s),
+    "<path> set installation directory to <path>");
+  ("-v", Arg.Unit (fun _ -> incr verbose),
+   "<non-default> show various diagnostics, repeat to increase verbosity");
+  ("-q", Arg.Unit (fun _ -> verbose := -1; debug := Debug_herd.none),
+   "<default> do not show diagnostics");
+  ("-I", Arg.String (fun s -> includes := !includes @ [s]),
+   "<dir> add <dir> to search path");
```
relokin: Can we rephrase this, similarly to -set-libdir, -I adds more paths and takes precedence over libdir.

herd/herd.ml
```diff
@@ -56,74 +56,57 @@ let gen_model_opt s =
...
-  ("-I", Arg.String (fun s -> includes := !includes @ [s]),
-   "<dir> add <dir> to search path");
-  parse_bool "-exit" Opts.exit_if_failed "exit in case of failure";
+
+
+let runmode_options = Arg.align ~limit:40 [
+  "-help", Arg.Unit (fun () -> ()), "" ;
+  "--help", Arg.Unit (fun () -> ()), "" ;
+ ("", Arg.Unit (fun () -> ()), "\n");
+ parse_tag "-speedcheck"
```
relokin: nit
```suggestion
  "", Arg.Unit (fun () -> ()), "\n";
  parse_tag "-speedcheck"
```

herd/herd.ml
```diff
@@ -56,74 +56,57 @@ let gen_model_opt s =
...
+    (fun tag -> match Speed.parse tag with
+    | None -> false
+    | Some t -> speedcheck := t ; true)
+    Speed.tags
+    "aim at checking condition in place of listing final states" ;
+  begin
+    let module ParseVariant = ParseTag.MakeS(Opts.OptS) in
+    ParseVariant.parse "-variant" variant
```
relokin: I think variants deserves a help menu on its own.

herd/herd.ml
```diff
@@ -174,134 +137,82 @@ let options = [
...
-       (PrettyConf.pp_dotmode !PP.dotmode)) ;
-  parse_tag "-dotcom"
-    (fun tag -> match PrettyConf.parse_dotcom tag with
-    | None -> false
-    | Some _ as t -> PP.dotcom := t ; true)
-    PrettyConf.tags_dotcom
-    "select command to translate dot, default depends on other modes" ;
+  "-nshow",
+  Arg.Int (fun n -> nshow := Some n),
+  "<n> collect at most <n> pictures, default is to collect all (specified) pictures";
```
relokin:
```suggestion
  "<n> generate at most <n> execution graph. By default all execution graphs specified by -show are generated.";
```

herd/herd.ml
```diff
@@ -174,134 +137,82 @@ let options = [
...
+   parse_bool "-checkfilter" check_filter "discard outcomes that negate filter proposition (if any)" ;
+  "-showone",
```
relokin, line 209: can we move this closer to -nshow?

herd/herd.ml
```diff
@@ -174,134 +137,82 @@ let options = [
...
+  "<bool> alias for -nshow 1";
+  ]
+
+let graph_content_options = Arg.align ~limit:40 [
+  "-help", Arg.Unit (fun () -> ()), "" ;
+  "--help", Arg.Unit (fun () -> ()), "" ;
+  (* Edges *)
+  ("", Arg.Unit (fun () -> ()), "\n");
+  ("Edge options", Arg.Unit (fun () -> ()), " ");
+  parse_bool "-showpo" PP.showpo "show po edges in pictures" ;
```
relokin: should we move this after -doshow?

herd/herd.ml
```diff
@@ -174,134 +137,82 @@ let options = [
...
-            | _ ->
-                raise
-                  (Arg.Bad
-                     (sprintf "bad argument for option -shift: '%s'" tag)))
-          fs in
-      PP.shift := Array.of_list fs),
-  "<float,...,float> add vertical space at thread start (column mode only)";
+  parse_bool "-showinitwrites" PP.showinitwrites
+    "show init write events in pictures" ;
+ parse_stringset "-cycles" cycles  "<name1,...,nameN> show failing checks as cycles, cumulates" ;
```
relokin: nit
```suggestion
  parse_stringset "-cycles" cycles  "<name1,...,nameN> show failing checks as cycles, cumulates" ;
```

herd/herd.ml
```diff
@@ -174,134 +137,82 @@ let options = [
...
+ "-initwrites", Arg.Bool (fun b -> initwrites := Some b),
+    "<bool> represent init writes as write events, this option should not be used except for debugging model options";
+  parse_string_opt "-classes" PP.classes "show classes of this equivalence (no not cumulate)" ;
+]
+let graph_presentation_options = Arg.align ~limit:40 [
+  "-help", Arg.Unit (fun () -> ()), "" ;
+  "--help", Arg.Unit (fun () -> ()), "" ;
+  ("", Arg.Unit (fun () -> ()), "\n");
+  parse_bool "-showobserved" PP.showobserved
+    "highlight observed memory reads in pictures" ;
```
relokin:
```suggestion
    "highlight observed memory reads in execution graphs" ;
```

herd/herd.ml
```diff
@@ -174,134 +137,82 @@ let options = [
...
   parse_bool "-edgemerge" PP.edgemerge "merge edges, cppmem style" ;
-  parse_bool "-labelinit" PP.labelinit "show labels on the init node" ;
-(* Legend *)
   parse_bool "-showlegend" PP.showlegend  "show legend in pictures" ;
   parse_bool "-showkind" showkind  "show test kind in legends" ;
   parse_bool "-shortlegend" shortlegend "show test name only in legends";
-(* Nodes *)
- parse_bool "-squished" PP.squished "limit information in graph nodes" ;
+  parse_bool "-labelinit" PP.labelinit "show labels on the init node" ;
+  parse_bool "-showthread" PP.showthread "show thread numbers in figures" ;
```
relokin:
```suggestion
  parse_bool "-showthread" PP.showthread "show thread numbers in execution graphs" ;
```

herd/herd.ml
```diff
@@ -174,134 +137,82 @@ let options = [
...
-  Arg.Bool (fun b -> if b then nshow := Some 1),
-  "<bool> alias for -nshow 1";
-  parse_int_opt
-    "-maxphantom" maxphantom "maximum phantom update (per variable)";
   "-statelessrc11",
   Arg.Bool (fun b -> if b then statelessrc11 := true),
   "<bool> enable stateless RC11 model checking, use with -variant normw, SC check can be skipped";
   "-dumpallfaults",
   Arg.Bool (fun b -> dumpallfaults := b),
   "Dump final states with all faults that that happenned regardless of the post-condition";
```
relokin: sorry noticed that this doesn't work as expected. Perhaps something along the lines of:
```suggestion
  "<bool> show all faults in the output regardless of the post-condition";
```

herd/herd.ml
```diff
@@ -56,74 +56,57 @@ let gen_model_opt s =
...
-   "<non-default> show various diagnostics, repeat to increase verbosity");
-  ("-q", Arg.Unit (fun _ -> verbose := -1; debug := Debug_herd.none),
-   "<default> do not show diagnostics");
-  ("-I", Arg.String (fun s -> includes := !includes @ [s]),
-   "<dir> add <dir> to search path");
-  parse_bool "-exit" Opts.exit_if_failed "exit in case of failure";
+
+
+let runmode_options = Arg.align ~limit:40 [
+  "-help", Arg.Unit (fun () -> ()), "" ;
```
ShaleXIONG: Not sure if `Fun.id` can replace `fun () -> ()`. The former is a standard library function in Ocaml.

herd/herd.ml
artkhyzha: My understanding is that this means to point to a folder like `$PREFIX/share/herdtools7/herd` (where PREFIX is as defined in the Makefile). I see that this folder in my installation contains mostly cat files, but also a folder `asl-pseudocode`.

I wonder if the text "show the default search path for .cat files and ASL pseudocode files" could work.

I would sync up the text for this option and `-I`.

herd/variant.ml
```diff
@@ -117,17 +117,98 @@ type t =
...
-   "fullscdepend";"splittedrmw";"switchdepscwrite";"switchdepscresult";"lrscdiffok";
-   "mixed";"dontcheckmixed";"weakpredicated"; "lkmmv1"; "lkmmv2"; "memtag";"vmsa";"kvm";]@
-    Precision.tags @ Fault.Handling.tags @
-   ["CutOff"; "deps"; "morello"; "instances"; "noptebranch"; "pte2";
-   "pte-squared"; "PhantomOnLoad"; "OptRfRMW"; "ConstrainedUnpredictable";
-    "exp"; "self"; "cos-opt"; "test"; "T[0-9][0-9]"; "asl"; "strict";
-    "warn"; "S128"; "ASLType+Warn";    "ASLType+Silence"; "ASLType+Check";
-    "ASL+AArch64+UDF"; "telechat"; "OldSolver"; "oota";
-    "pac"; "fpac"; "const-pac-field";]
+(* Identity function to prompt user to insert variant into list used for help message when adding a new variant*)
```
diaolo01: This is to prompt developers, not users.

herd/variant.ml
```diff
@@ -297,6 +378,107 @@ let pp = function
   | ConstPacField -> "const-pac-field"
   | FPac -> "fpac"
 
+  let descriptions_on = false
```
diaolo01: Can we flip the polarity for this? I suggest renaming to "under_development = true"

herd/variant.ml
```diff
@@ -297,6 +378,107 @@ let pp = function
...
+    (List.map variant_descriptions mode_variants)
+
+ let arch_help_tags =
+  List.map2 
+    (fun variant  description ->
+      variant ^ ": " ^ description)
+    (List.map pp arch_variants)
+    (List.map variant_descriptions arch_variants)
+
+let help_page =
```
diaolo01: Please can this be renamed to "variants_help_page"?

lib/parseTag.ml
```diff
@@ -105,10 +105,9 @@ module MakeS (O:OptS)
 
       let parse_tag_set opt add =  do_parse_tag_set opt (add_tag add)
 
-      let parse opt add msg =
+      let parse opt add (msg:string) =
```
diaolo01: Is the msg:string required? What function does it provide? Does the compiler not assign string to msg by default?

herd/herd.ml
```diff
@@ -354,18 +306,217 @@ let options = [
...
+  runmode_options @
+  filter_options @
+  graph_content_options @
+  graph_presentation_options @
+  help_options
+
 (* Parse command line *)
 let () =
+  let argv_len = Array.length Sys.argv in
+  let arg = Sys.argv.(argv_len -1) in
```
diaolo01: This seems to only use the last argument passed after help. This means that if a user inputs `herd -help variant runmode` then it returns `herd -help runmode`. I would expect if a user passes two arguments for help then it should give the `herd -help` output, because they're doing something wrong. 

This also means that if a user passes `herd <option> -help foobar` then the output of `-help foobar` is shown. My expectation is the output to be the same as `herd -help foobar <option>`.

herd/herd.ml
```diff
@@ -354,18 +306,217 @@ let options = [
...
+        Printf.printf "%s" (Arg.usage_string graph_presentation_options usg_graph_content); exit 0
+    | "variant" ->
+        Printf.printf "%s" (Variant.help_page); exit 0
+    | "all" ->
+        show_long_help ();
+    |  _ -> 
+        if Array.exists (fun s -> s = "-help") Sys.argv then
+          show_short_help ()
+        else 
+          show_long_help ();
```
diaolo01: I wonder why would we make a difference between which helper is used when the user isn't inputting reasonable things in the CLI?

herd/herd.ml
```diff
@@ -354,18 +306,217 @@ let options = [
...
+    | "setup" ->
+        Printf.printf "%s" (Arg.usage_string setup_options usg_setup); exit 0
+    | "runmode" ->
+        Printf.printf "%s" (Arg.usage_string runmode_options usg_runmode); exit 0
+    | "filter" ->
+        Printf.printf "%s" (Arg.usage_string filter_options usg_filter); exit 0
+    | "graph_content" ->
+        Printf.printf "%s" (Arg.usage_string graph_content_options usg_graph_content); exit 0  
+    | "graph_presentation" ->
+        Printf.printf "%s" (Arg.usage_string graph_presentation_options usg_graph_content); exit 0
```
diaolo01: This should be calling usg_graph_presentation.

herd/herd.ml
diaolo01: Help handling is now order-dependent. Please make help parsing order-insensitive and treat -help vs --help as follows:
•	-help → short help.
•	--help → long help.
•	-help <one known category> → that category help.
•	-help <two or more known categories> → short help (user error fallback).
•	Any extra non-category options anywhere on the line should be ignored in favour of help behavior.
•	--help <anything> → still long help (ignore extras).
•	Order never matters.
When help is requested, ignore unrelated options. `herd -help <category> <options>` and `herd <options> -help <category>` should both show category help.

herd/herd.ml
```diff
@@ -354,18 +306,217 @@ let options = [
...
+        if Array.exists (fun s -> s = "-help") Sys.argv then
+          show_short_help ()
+        else 
+          show_long_help ();
+    else
+      let current = ref 0 in
+      match Arg.parse_argv ~current Sys.argv options get_cmd_arg usg_default with
+      | () -> ()
+      | exception Arg.Bad msg -> 
+          begin
```
diaolo01: What functionality does this code offer? Is it not possible to handle this exception in the "with" part of the code? Why can we not use the original Arg.parse here?

herd/herd.ml
diaolo01: Another idea here is to take the helper function out of this main into its own function. It will help with readability of code.

herd/variant.ml
```diff
@@ -297,6 +378,107 @@ let pp = function
...
+  | Telechat -> ""
+  | NV2 -> ""
+  | OldSolver -> ""
+  | OOTA -> ""
+  | Pac -> ""
+  | ConstPacField -> ""
+  | FPac -> ""
+
+let mode_tags =
+    List.map( fun x-> if x = "T00" then "T [0-9]" else x) (List.map pp mode_variants)
```
diaolo01: Since this is used twice, I think it deserves its own function. Although, my understanding is that it should be T[0-9][0-9]. Can you please investigate what this variant does? Can we encode better that this variant requires a double digit number? Is it not really `T[0,1][0-9]`? Is it possible to make this not look like a regex, which is what it seems like now?

herd/variant.ml
```diff
@@ -297,6 +378,107 @@ let pp = function
...
+    (fun variant  description ->
+      variant ^ ": " ^ description)
+    (List.map pp arch_variants)
+    (List.map variant_descriptions arch_variants)
+
+let help_page =
+  if descriptions_on then
+    Printf.sprintf "\n Use -variant <tags> where:\n Mode of Operation tags:\n  %s 
+    \n Architecture Feature tags:\n  %s \n" 
+    (String.concat "\n  " mode_help_tags) (String.concat "\n  " arch_help_tags)
```
diaolo01: There is duplication here which can be avoided.

herd/variant.ml
```diff
@@ -297,6 +378,107 @@ let pp = function
...
+  | OldSolver -> ""
+  | OOTA -> ""
+  | Pac -> ""
+  | ConstPacField -> ""
+  | FPac -> ""
+
+let mode_tags =
+    List.map( fun x-> if x = "T00" then "T [0-9]" else x) (List.map pp mode_variants)
+let arch_tags =
+  List.map pp arch_variants
```
diaolo01, line 456: There are a few List.map pp list computed multiple times - please can we only do them once?

herd/herd.ml
```diff
@@ -56,74 +56,57 @@ let gen_model_opt s =
...
+    | None -> false
+    | Some t -> speedcheck := t ; true)
+    Speed.tags
+    "aim at checking condition in place of listing final states" ;
+  begin
+    let module ParseVariant = ParseTag.MakeS(Opts.OptS) in
+    ParseVariant.parse "-variant" variant
+      "<tags> select an architecture variation, use '-help variant' for the list of tags" end ;
+  begin 
+    let module ParseMachSize = ParseTag.Make(MachSize.Tag) in
```
diaolo01, line 150: There are a few options that are duplicated. Please, can you make sure options appear only once?

herd/herd.ml
```diff
@@ -354,18 +306,217 @@ let options = [
...
+     | "-" -> outputdir := PrettyConf.StdoutOutput
+     | _ -> outputdir := PrettyConf.Outputdir s),
+   "<dir> generated files will go into <dir>. No default, if not set herd7 does not generate any files.") ;
+  
+  begin let module ParseView = ParseTag.Make(View) in
+  ParseView.parse_opt "-view" PP.view
+    "fork specified viewer to show output graphs" end ;
+  ( "-gv",
+    Arg.Unit (fun _ -> PP.view := Some View.GV),
+    "alias for -view gv") ;
```
diaolo01: This appears in the option column rather than the description column.

herd/herd.ml
```diff
@@ -354,18 +306,217 @@ let options = [
...
     "apply various tricks to enhance edge label placement in pictures" ;
+  ]
+
+let setup_options = Arg.align ~limit:40 [
+   "-help", Arg.Unit Fun.id, "" ;
+  "--help", Arg.Unit Fun.id, "";
+  ("", Arg.Unit Fun.id, "\n");
+  ("-version", Arg.Unit
+     (fun () -> printf "%s, Rev: %s\n" Version.version Version.rev ; exit 0),
+   " show version number and exit") ;
```
diaolo01: I think this is self explanatory and can be removed.

herd/herd.ml
```diff
@@ -354,18 +306,217 @@ let options = [
...
+  ("-libdir", Arg.Unit (fun () -> print_endline !Opts.libdir; exit 0),
+    "show the default search path for .cat files and ASL pseudocode files");
```
diaolo01: show appears in the option column rather than the description column

herd/herd.ml
```diff
@@ -354,18 +306,217 @@ let options = [
...
+  ("-set-libdir", Arg.String (fun s -> Opts.libdir := s),
+    "<path> set the default search path for .cat files and ASL pseudocode files to <path>");
+  ("-v", Arg.Unit (fun _ -> incr verbose),
+   "show various diagnostics, repeat to increase verbosity");
```
diaolo01: This show appears in the options column rather than the description column. I am seeing multiple instances of this issue, please may you check all of them?

herd/herd.ml
```diff
@@ -354,18 +306,217 @@ let options = [
...
+    Arg.Unit (fun _ -> PP.view := Some View.GV),
+    "alias for -view gv") ;
+  ( "-evince",
+    Arg.Unit (fun _ -> PP.view := Some View.Evince),
+    "alias for -view evince") ;
+  ( "-preview",
+    Arg.Unit (fun _ -> PP.view := Some View.Preview),
+    "alias for -view preview") ;
+  ("-web",
+   Arg.Unit (fun () -> load_config "web.cfg")," alias for -conf web.cfg");
```
diaolo01: Please can these be ordered after the -conf option?

herd/herd.ml
```diff
@@ -354,18 +306,217 @@ let options = [
...
+  ("-c11",
+   Arg.Unit (fun () -> load_config "cpp11.cfg")," alias for -conf cpp11.cfg");
+  parse_tag "-dotcom"
+    (fun tag -> match PrettyConf.parse_dotcom tag with
+    | None -> false
+    | Some _ as t -> PP.dotcom := t ; true)
+    PrettyConf.tags_dotcom
+    "select command to translate dot, default depends on other modes " ;
+  parse_tags
+    "-debug"
```
diaolo01, line 389: Could we have `debug <tag>' and then list the tags in the description instead and suggest that there can be multiple tags chosen as long as they are separated by comma?

herd/herd.ml
```diff
@@ -354,18 +306,217 @@ let options = [
...
+  ("Input options", Arg.Unit Fun.id, " ");
+
+  ( "-kinds",
+    Arg.String (fun s -> kinds := !kinds @ [s]),
+    "<name> specify kind of tests (can be repeated)");
+  ( "-conds",
+    Arg.String  (fun s -> conds := !conds @ [s]),
+    "<name> specify conditions of tests (can be repeated)");
+    ]
+   @ parse_noselect @ [
```
diaolo01, line 407: The strings under this list aren't formatted correctly, please can they be updated?

herd/herd.ml
```diff
@@ -354,18 +306,217 @@ let options = [
...
+  (* Output *)
+  ("", Arg.Unit Fun.id, "\n");
+  ("Output options", Arg.Unit Fun.id, " ");
```
diaolo01, line 410: The options descriptions under this bucket aren't formatted, please can you do that?

herd/herd.ml
diaolo01: Please, can we use this principle for all the options that can take multiple tags which are fixed? I would rather the tags appear in the description rather than enumerated in the option column. This might fix a few formatting issues too!

herd/herd.ml
```diff
@@ -354,18 +306,217 @@ let options = [
...
+]
+(*Usage messages*)
+let usg_setup = (sprintf "\n Control the command environment: paths, config, input and output handling")
+let usg_runmode = (sprintf "\n Configure herd")
+let usg_filter = (sprintf"\n Select model and control which executions are filtered or allowed")
+let usg_graph_content = (sprintf "\n Control what events and edges are included in the graph")
+let usg_graph_presentation = (sprintf "\n Control graph layout, sizing, annotations and visual clarity")
+
+let usg_default = String.concat "\n"
+  ["These are common herd options:
```
diaolo01: Please can this say "Common herd options"

herd/herd.ml
```diff
@@ -354,18 +306,217 @@ let options = [
...
+  -variant <variant_name>       Activate variations of models | use '-help variant' for the list of variants
```
diaolo01: "Select model variants. Use '-help variant' for the supported list of variants."

herd/herd.ml
```diff
@@ -478,45 +458,62 @@ let options = setup_options @
...
+  let inputs = Array.to_list Sys.argv |> List.tl in
+  let has_long_help = List.exists ((=) "--help") inputs in
+  let has_short_help = List.exists ((=) "-help") inputs in
+
+  if has_long_help then  show_long_help () else
+    if has_short_help then 
+      let rest = List.filter (fun s -> s <> "-help") inputs in
+      let (categories, others) = List.partition is_known_category rest in
+      match (categories, others) with
+      | ([], []) -> show_short_help ()
```
diaolo01: This case is already covered by `| _`, so I don't think we need it.

lib/parseTag.ml
```diff
@@ -38,11 +38,11 @@ module Make (O:Opt) =
       match d with
       | Some d ->
           sprintf
-            "<%s> %s, default %s"
+            "<tags> where tags is <%s> %s, default %s"
             (String.concat "|" O.tags) msg
             (O.pp d)
       | None ->
-          sprintf "<%s> %s" (String.concat "|" O.tags) msg
+          sprintf "<tags> where tags is <%s> %s" (String.concat "|" O.tags) msg
```
diaolo01: All the options that use the `<tags>` format have no separation between what the tags are and what the option does. For example:

> -speedcheck <tags>     where tags is <false|true|fast> aim at checking condition in place of listing final states

Could this, and all the other options affected, be written as:
> -speedcheck <tags>     where tags is <false|true|fast>. Aim at checking condition in place of listing final states

herd/variant.ml
```diff
@@ -448,34 +448,36 @@ let pp = function
   | FPac -> ""
 
 let mode_tags =
-    List.map( fun x-> if x = "T00" then "T [0-9]" else x) (List.map pp mode_variants)
+  let normalise_T00 = function
+    | "T00" -> "T[0-9][0-9]"
+    | s -> s in
+    List.map( normalise_T00) (List.map pp mode_variants)
```
diaolo01: Minor, but I think the indentation here isn't quite right.

herd/herd.ml
```diff
@@ -411,8 +391,8 @@ let usg_graph_content = (sprintf "\n Control what events and edges are included
 let usg_graph_presentation = (sprintf "\n Control graph layout, sizing, annotations and visual clarity")
 
 let usg_default = String.concat "\n"
-  ["These are common herd options:
-  -variant <variant_name>       Activate variations of models | use '-help variant' for the list of variants
+  ["Common herd options:
+  -variant <variant_name>       Select model variants, Use '-help variant' for the supported list of variants
```
diaolo01: Full stop between the two sentences.

PR #1378, opened 2025-07-29
[gen] Remove unused code in instruction fetch `ifetch` feature

ShaleXIONG:
Fix a few problems in instruction fetch feature (ifetch):

- A mistake in assign initial value in the case of code location.
- Remove code that are no longer used, including (1) internal data structure of `IFF` and `FIF` in `AArch64Arch_gen.ml`, and (2) `ins` field in `node` in `cycle.ml`.

### DIFF
--- a/gen/cycle.ml
+++ b/gen/cycle.ml
@@ -1211,7 +1211,8 @@ let do_set_read_v init =
     (fun ns k -> match ns with
     | [] -> k
     | n::_  ->
-        let init = List.assoc_opt (Code.as_data n.evt.loc) initvals
+        let init = if not (Code.is_data n.evt.loc) then 0
+                 else List.assoc_opt (Code.as_data n.evt.loc) initvals
                   |> Option.map Code.value_to_int
                   |> Option.value ~default:0 in
         let vf = do_set_read_v init ns in


gen/cycle.ml
```diff
@@ -1179,7 +1179,8 @@ let do_set_read_v init =
     (fun ns k -> match ns with
     | [] -> k
     | n::_  ->
-        let init = List.assoc_opt (Code.as_data n.evt.loc) initvals
+        let init = if not (Code.is_data n.evt.loc) then 0
```
murzinv, line 1214: JFYI, I do not know If it can give you any clue yet it started asserting in `Code.is_data` since c2dd8bdf4e0aeae5255b16b21251a7cc6f5b59ac ("Add the fault check functionality related to pte value")

ShaleXIONG: Oh thanks ! I will look into it.

ShaleXIONG: I remove the commit for "Allowed standalone annotation of `I`. For example `PodWR I` (previous disallowed) which is equal to `PodWRPI`." This gives back the old behaviour. This pull should be ready again for review.

PR #1381, opened 2025-07-30
[WIP] [gen] Fix the atomic read-modify-write edges

ShaleXIONG:
This pull request contains a series of fixes to atomic read-modify-write (rmw) instructions/edges, such as `LDADD`/`Amo.LdAdd`, with the goal to ensure those operations give observable behaviours.

- Finish the implementation of signed and unsigned, max and min read-modify-write instructions/edges, such as `LDUMIN`/`Amo.LdUmin`.
- Given any read-modify-write edges, it now assigns distinct values. That is, assuming `op1` and `op2` for a location and `op1` is not the same type of bit-wise or max-min operation as `op2`, distinct values will be assigned to the following:
    * initial value
    * after `op1`
    * after `op2`
    * after `op1` then `op2`
    * after `op2` then `op1`
This means the observed values will uniquely reflect one execution order.
- Reject any cycle contains two edges of the same type of atomic operation, e.g. `Amo.LdClr Rfe Amo.LdClr Rfe` will be rejected.
- Because of the strategy change, there are several update to CI.


gen/AArch64Arch_gen.ml
```diff
@@ -904,6 +913,93 @@ let compute_rmw r old co =
...
+         that is, if `op1` and `op2` are not same type of rmw,
+         applying `op2` followed by `op1` gives a different value from
+         applying `op1` followed by `op2.
+*)
+let init_rmw rmw =
+  let value = match rmw with
+    | LdOp op|StOp op -> begin match op with
+      (* Set the spacial initial values as described above. *)
+      | A_ADD|A_CLR|A_EOR|A_SET -> 0x07_20
+      | A_SMIN|A_UMIN -> 0x5_00_00
```
murzinv: Shouldn't it be `0x50_00`?

gen/cycle.ml
```diff
@@ -595,14 +600,37 @@ module CoSt = struct
...
+              |> E.to_rmw_operand rmw init in
+      (* NOTE st.map is NOT updated upon `rmw` with calculation
+         because `Ord` only tracks the value used for
+         various `STR` or rmw instruction *)
+      v,{ st with map=M.add Ord counter st.map; }
+    (* Only support rmw on ordinary, `Ord`, memory access *)
+    | E.Rmw _, _ -> assert false
+    (* The following is for none RMW operation *)
+    | _, VecReg n ->
+      let counter = (find_no_fail Ord st.map) + E.SIMD.nregs n in
```
murzinv, line 612: `nregs` could be up to 4, so couple SIMD annotations like `Ne4`, `Nei` (`-variant neon`) and/or `Sv4` (`-variant sve`) would easily exhaust `counter` for `CLR`, `SET` and `EOR`

gen/cycle.ml
```diff
@@ -1265,6 +1345,8 @@ let finish n =
     | Diff -> set_diff_loc st n
     | Same -> set_same_loc st n in
 
+  if not (valid_cycle n) then Warn.fatal "Invalid cycle.\n" ;
```
murzinv: It is too little information to the user explaining why cycle is invalid. Can we turn it into say `check_cycle` and fail on callee site where we perform actual check and have  all information about why cycle is not valid?

gen/common/AArch64Arch_gen.ml
```diff
@@ -904,6 +913,93 @@ let compute_rmw r old co =
...
+  (List.length atomic_list) = (RmwSet.cardinal @@ RmwSet.of_list atomic_list)
+
+(* NOTE Assuming all variants of STR manipulates on
+   bit 4 --- bit 0 (both included), the bit-wise
+   add and max-min manipulates the rest.
+    In summary the initial value given by `init_rmw`
+    and bit assignment, hard-coded in `to_rmw_operand`, is:
+    |<-MINMAX->|  |<--CLR,SET,EOR-->|
+    ------------------------------------------------
+    | 0  1  0  1, 0  1  1  1, 0  0  1  0, 0  0  0  0
```
murzinv, line 899: Can you please, explain how these init values have been chosen? I assume for `MINMAX` we want to start in the middle of the range so we can add/subtract for `MAX` and `MIN` operations. Is that correct? What about bit-wise operations and ADD? Why they overlap in such specific way?

murzinv (commented):
It looks like it doesn't play well with arrays
```
$ diyone7 -arch AArch64 "DMB.SYdWW Pa Rfe Amo.StClr DMB.LDdWR Fre"
AArch64 A
"DMB.SYdWWPPa RfePaP Amo.StClr DMB.LDdWR Fre"
Com=Rf Fr
Orig=DMB.SYdWWPPa RfePaP Amo.StClr DMB.LDdWR Fre
{
int y[2];
 y=1824;
0:X1=x; 0:X2=y;
1:X2=y; 1:X1=x;
}
 P0             | P1            ;
 MOV W0,#1      | MOV W0,#2528  ;
 STR W0,[X1]    | STCLR W0,[X2] ;
 DMB SY         | DMB LD        ;
 MOV W3,#1826   | LDR W3,[X1]   ;
 SUB W4,W3,#1   |               ;
 STP W4,W3,[X2] |               ;
exists (y={1537,1826} /\ 1:X3=0)
```
but it seems similar issue existed even before your change with `-variant kvm`
```
$ diyone7 -arch AArch64 -variant kvm "DMB.SYdWWPaP Rfe DMB.LDdRRPPa Fre"
AArch64 A
"DMB.SYdWWPaP Rfe DMB.LDdRRPPa FrePaPa"
Com=Rf Fr
Orig=DMB.SYdWWPaP Rfe DMB.LDdRRPPa FrePaPa
{
int x[2];
 int y=0; x=4;
0:X0=x; 0:X4=y;
1:X4=y; 1:X0=x;
}
 P0             | P1             ;
 MOV W1,#6      | LDR W1,[X4]    ;
 SUB W2,W1,#1   | DMB LD         ;
 STP W2,W1,[X0] | LDP W2,W3,[X0] ;
 DMB SY         | ADD W2,W2,W3   ;
 MOV W3,#1      |                ;
 STR W3,[X4]    |                ;
exists (1:X1=1 /\ 1:X2=8)
```

gen/common/AArch64Arch_gen.ml
ShaleXIONG: I wrote a small program to search all the possible value (up to a range) on  (1) initial value, if certain `Amo.*` exists,  (2) the addition part in `bit_mask`, (3) `shift_bit`, (4) `shift_add` and (5) the offset number of `compare_offset`. 
The idea is when `counter` from 1 to 5. all possible operations op1 and op2, for example `op1 = StAdd` and `op2 = StSub` the following execution gives different result
- op1 counter -> op2 counter +1
- op1 counter +1 -> op2 counter
- op2 counter -> op1 counter +1
- op2 counter+1 -> op1 counter

Especially the observable value of the `op1` and `op2` in the above execution and the initial value must be different.
Note that we also need to ensure some lower bit is reserved for the basic `str *`.

I made a practice choice here as I only explore counter until 5. Basically means in the generate litmus test there are up to 5 writes per location which already exceed the default `diy` upper bound.


To keep a record, the program I wrote as the follow, it should be standalone and run `ocaml main.ml > result.log` will dump all valid combinations that satisfies above.
 ```ocaml
type atomic_op =
  | A_ADD | A_EOR | A_SET | A_CLR
  | A_SMAX | A_SMIN | A_UMAX | A_UMIN
  | MOVE

let pp = function
  | A_ADD -> "ADD"
  | A_EOR -> "EOR"
  | A_SET -> "SET"
  | A_CLR -> "CLR"
  | A_SMAX -> "SMAX"
  | A_SMIN -> "SMIN"
  | A_UMAX -> "UMAX"
  | A_UMIN -> "UMIN"
  | MOVE -> "MOVE"

(* - co is the counter *)
let compute_rmw init offset shift_bit shift_add shift_compare op co old =
  let bit_mask = ((1 lsl (co + 1)) - 1) + offset in
(*
  let bit_mask = ((1 lsl (co))) + offset in
*)
  let compare_max = init + (co lsl shift_compare) in
  let compare_min = init - (co lsl shift_compare) in
  match op with
    | A_ADD -> old + (co lsl shift_add)
    | A_CLR -> old land (lnot (bit_mask lsl shift_bit))
    | A_EOR -> old lxor (bit_mask lsl shift_bit)
    | A_SET -> old lor (bit_mask lsl shift_bit)
    | A_SMAX -> if old > compare_max then old else compare_max
    | A_UMAX ->
       let o = Int32.of_int old and c = Int32.of_int compare_max in
       if Int32.unsigned_compare o c >  0 then old else compare_max
    | A_SMIN -> if old < compare_min then old else compare_min
    | A_UMIN ->
       let o = Int32.of_int old and c = Int32.of_int compare_min in
       if Int32.unsigned_compare o c <  0 then old else compare_min
    | MOVE -> (init + co)

let unique_pairs lst =
  List.flatten (
    List.map (fun x ->
      List.filter (fun y -> x <> y) lst
      |> List.map (fun y -> (x, y))
    ) lst
  )

let apply init offset shift_bit shift_add shift_compare (a_op,b_op) a_value b_value =
  let config_compute_rmw = compute_rmw init offset shift_bit shift_add shift_compare in
  let apply_a = config_compute_rmw a_op a_value in
  let apply_b = config_compute_rmw b_op b_value in
  let a = apply_a init in
  let a_b = apply_b a in
  let b = apply_b init in
  let b_a = apply_a b in
  let values = [init;a;a_b;b;b_a] in
  let result =    init <> a
               && init <> a_b
               && init <> b
               && init <> b_a
               && a <> a_b
               && b <> b_a
               && a_b <> b_a
               (* a = b is allowed *)
               && (List.for_all ( fun v -> v >= 0 && v < 0x10000 ) values) in
(*
  Printf.eprintf "apply: (%s %d) (%s %d) (%d %d %d) -> 0x%x 0x%x 0x%x 0x%x 0x%x %b\n"
          (pp a_op) a_value (pp b_op) b_value shift_bit shift_add shift_compare
          init a a_b b b_a result;
*)
  result

let range a b =
  if a > b then []
  else List.init (b - a + 1) (fun i -> a + i)

let combinations lst =
  let rec aux acc = function
    | [] -> acc
    | x::xs ->
      let new_pairs = List.map (fun y -> (x, y)) xs in
      aux (acc @ new_pairs) xs
  in
  aux [] lst

let int_range = range 1 5
let counters = combinations int_range
let op_list = [(MOVE, MOVE)] @ unique_pairs ([ MOVE; A_ADD ; A_EOR ; A_SET ; A_CLR ] @ [ A_UMAX ; A_UMIN ](* @ [A_SMAX ; A_SMIN] *))

let run_counters init offset shift_bit shift_add shift_compare =
  List.for_all (
    fun (first,second) -> List.for_all (
      fun ops ->
        apply init offset shift_bit shift_add shift_compare ops first second
      ) op_list
    ) counters


(*
let init_range = range 0x1 0x7ff |> List.map (fun e -> e lsl 5)
*)
let init_range = range 0x2b9 0x7ff |> List.map (fun e -> e lsl 5)
let offset_range = range 0x1 0xfff
let shift_range = range 5 6
let shift_range = range 5 15

let () =
  List.iter ( fun init ->
  List.iter ( fun offset ->
    List.iter ( fun shift_bit ->
      List.iter ( fun shift_add ->
        List.iter ( fun shift_compare ->
            Printf.eprintf "try: 0x%x 0x%x %d %d %d\n" init offset shift_bit shift_add shift_compare;
            let t = run_counters init offset shift_bit shift_add shift_compare in
            if t then Printf.printf "0x%x 0x%x %d %d %d\n" init offset shift_bit shift_add shift_compare;
        ) shift_range
      ) shift_range
    ) shift_range
  ) offset_range
  ) init_range;
  List.iter ( fun (a,b) -> Printf.eprintf "%d,%d\n" a b ) counters;
(* The best value we pick, relatively easier to understand and small values *)
(*
let () =
  let t = run_counters 0x5720 0x40 5 7 12 in
  Printf.eprintf "final result: %b\n" t;
*)

```

File: gen/cycle.ml (hunk 3 above)
Comment (ShaleXIONG) [reply gen/cycle.ml]: Will change it, and add more informative warning message.

File: gen/AArch64Arch_gen.ml (hunk 1 above)
Comment (ShaleXIONG) [reply gen/AArch64Arch_gen.ml]: You are right. Thanks for find it out. I will update again on the expected test results. The reason of such value strategy has been presented in the following comment and experiment program also provides.

File: gen/cycle.ml (hunk 2 above)
Comment (ShaleXIONG) [reply gen/cycle.ml:612]: Yes, you are right. I decide to put `SIMD` out of the scope of this new change, hence, it will NOT apply here if it is a `SIMD` hence the old behaviour of counter `1,2,...` applies to `SIMD`. It certainly will lead to unexpected allowed litmus tests if applying atomic operations on `SIMD` as previously. In fact I think the `rmw` is not supported in `SIMD` as previously, so I also rule it out in line 615 in `cycle.ml`.

**Comment by ShaleXIONG** [comment]
@murzinv I upstreamed separate commits regarding the suggestions and mistakes you found, which will be merge to the existing commits once you think it makes sense to you. Thank you so much for reviewing.

File: gen/cycle.ml (hunk 2 above)
Comment (murzinv) [reply gen/cycle.ml:612]: > I also rule it out in line 615 in cycle.ml

Do you refer to
```
    (* Only support rmw on ordinary, `Ord`, memory access *)
    | E.Rmw _, _ -> assert false
```
?
If so then it would not assert
```
$ diyone7 -arch AArch64 -variant neon "DMB.STdWWPNeP RfePA Amo.LdAddAL PodWWLP Coe"  
AArch64 A
"DMB.STdWWPNeP RfeNePA Amo.LdAddAL PodWWLP Coe"
Com=Rf Co
Orig=DMB.STdWWPNeP RfeNePA Amo.LdAddAL PodWWLP Coe
{ int y=1824;
0:X1=x; 0:X2=y;
1:X2=y; 1:X1=x;
}
 P0            | P1                 ;
 MOV W0,#2     | MOV W3,#256        ;
 STR W0,[X1]   | LDADDAL W3,W0,[X2] ;
 DMB ST        | MOV W4,#1          ;
 MOVI V0.4S,#1 | STR W4,[X1]        ;
 STUR S0,[X2]  |                    ;
exists ([x]=2 /\ [y]=257 /\ 1:X0=1)
```
I think the reason for that is for `SIMD` we reduce vector into single value and place it into `Ord`. Also note that we have `SIMD` annotations which work on single element.

ShaleXIONG: Sorry I confused myself. You are correct here the `SIMD` will not work. I have a second thought here, I think I can make further change on `SIMD` so it only pick counter value, 1,2,3, and so on, for the atomic operation. Let me try to fix this. Thanks for spotting this.

ShaleXIONG:
> > I think similar behaviour before this change. The atomic operation only apply to the first element in the array, here it executes STCLR on y[1], the result 1537=0x601 which is the value I expected. I think it is actually the expected behaviour?
> 
> I'm sorry I should have been clear on what problem is. It is
> 
> ```
> int y[2];
>  y=1824;
> ```
> 
> which upsets `herd7`
> 
> ```
> $ herd7 T.litmus 
> Warning: File "T.litmus": Location y is initialized more than once (User error)
> ```
> 
> However, I agree that it can be separate PR

Oh that is interesting. I will invest. Do you know the correct syntax for `herd7` to initialise array? `int y[2] = {1824,1824}`?

murzinv:
> Oh that is interesting. I will invest. Do you know the correct syntax for `herd7` to initialise array? `int y[2] = {1824,1824}`?

Yes `int y[2] = {1824,1824}` is correct syntax

gen/cycle.ml
ShaleXIONG: Update a patch to revert the value choice strategy in neon, sme, sve back the basically counter mode.

ShaleXIONG:
> > Oh that is interesting. I will invest. Do you know the correct syntax for `herd7` to initialise array? `int y[2] = {1824,1824}`?
> 
> Yes `int y[2] = {1824,1824}` is correct syntax

I have another separate pull request ready to fix this now https://github.com/ShaleXIONG/herdtools7/commit/c9bfc04f494b218c7484699df6048497dbb423aa I will put separate pull request later once I put more words in the commit message and do a bit more tests.

PR #1382, opened 2025-07-31
[litmus] Change handling of the self variant

maranget:
The variant `self` impacts (1) the C code of a test and  (2) gcc compilation flags. Compilation flags are now changed as soon as one test is compiled in self mode.

This departs from previous behaviour where compilation  flags were changed only when the variant was specified on the command line.
    
Change is minor and only applies when
     1. `-variant self` is *not* specified on the command line.
     2. Some of the compiled tests meta-data specifies `Variant=self`.
    
In that situation the previous behaviour was *not* changing the compilation flags, while the new behaviour is changing them.

As a technical note, the functor  `DumpRun.Make` configuration argument does not longer include a `variant` member.

### DIFF
--- a/litmus/answer.mli
+++ b/litmus/answer.mli
@@ -20,13 +20,15 @@
 type hash = { filename : string ;  hash : string ; }
 type hash_env = hash StringMap.t
 
+type completed = (* Answer for completed test *)
+  { arch : Archs.t ; (* Arch of test *)
+    doc : Name.t   ; (* Name of test *)
+    src : string   ; (* Name of emitted source file *)
+    fullhash : hash ; nprocs : int ; (* hash and numbre of threads *)
+    pac : bool     ; (* Requires pointer authentification *)
+  }
+
 type answer =
-  | Completed of
-      Archs.t * Name.t   (* Test arch and name *)
-        * string         (* C source file of test *)
-        * StringSet.t    (* cycles *)
-        * hash_env       (* name -> hash *)
-        * int            (* number of threads *)
-        * bool           (* require Pointer Authentication Code *)
-  | Interrupted of Archs.t * exn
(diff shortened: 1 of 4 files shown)


litmus/answer.mli
```diff
@@ -20,13 +20,16 @@
 type hash = { filename : string ;  hash : string ; }
 type hash_env = hash StringMap.t
 
+type completed = (* Answer for completed test *)
+  { arch : Archs.t ; (* Arch of test *)
+    doc : Name.t   ; (* Name of test *)
+    src : string   ; (* Name of emitted source file *)
+    fullhash : hash ; nprocs : int ; (* hash and numbre of threads *)
+    pac : bool     ; (* Requires pointer authentification *)
+    self : bool    ; (* Self modying code *)
```
murzinv, line 29: I'm just wondering if we want isolate it in it's own type, say, `flags`?

maranget: Hi @murzinv, you mean having just one field `flags` of type `Flags.t` that would be isomorphic to a pair of booleans?

litmus/dumpRun.ml
murzinv, line 1: Nitpick: there are few places (L360 and L506) where `run_tests` still return `pac` while that now de-facto `flags`

litmus/answer.mli
murzinv: Yes, then we could isolate flag management from other bits... that's just nitpick, so feel free to ignore

maranget: As there are two flags only, I am tempted to leave things as they are.

litmus/dumpRun.ml
maranget: Nice catch.

maranget:
Merged , thanks @murzinv.

PR #1385, opened 2025-08-01
[ASL-828] Improve error messages for primitives

hrutvik:
Rather than using "type mismatch" errors for primitives that expect e.g. positive inputs, use a new kind of error to report an appropriate message.

### DIFF
--- a/asllib/error.ml
+++ b/asllib/error.ml
@@ -106,6 +106,7 @@ type error_desc =
   | TooManyOverrideCandidates of func annotated list
   | PrecisionLostDefining
   | UnexpectedCollection
+  | BadPrimitiveArgument of identifier * string
 
 type error = error_desc annotated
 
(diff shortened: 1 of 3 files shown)


asllib/error.ml
```diff
@@ -607,7 +609,11 @@ module PPrint = struct
         fprintf f
           "ASL Type error:@ multiple@ `impdef`@ candidates@ for@ \
            `implementation`:@ %a"
-          (pp_print_list pp_pos) impdefs);
+          (pp_print_list pp_pos) impdefs
+    | BadPrimitiveArgument (name, reason) ->
+        pp_print_text f
+          ("ASL Execution error: " ^ name ^ " (primitive) expected an argument "
+         ^ reason));
```
HadrienRenaud, line 616: For next time, using printf is probably more consistent with the rest of the file:

```suggestion
        fprintf f
          "ASL Execution error: %s (primitive) expected an argument %s"
          name reason);
```

PR #1388, opened 2025-08-04
[ASL-858] Ensure annotated type declarations persist in typed AST

hrutvik:
Type declarations (`D_TypeDecl` constructor of `decl`) were previously thrown away after type-checking, so the untyped versions persisted in typed AST. This meant that e.g. `Slice_Single` slice forms that should have been desugared into `Slice_Length` were still found in annotated bitfield declarations.
This PR ensures that typed versions replace the untyped ones after type-checking.


asllib/Typing.ml
```diff
@@ -3988,7 +3988,7 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
               | Some _ -> conflict ~loc [ T_Record []; T_Exception [] ] ty
               | None -> undefined_identifier ~loc super
           and env = add_subtype name super env in
-          (env, new_ty)
+          (env, new_ty, Some (super, []))
```
HadrienRenaud, line 3993: Shouldn't the extra fields stay also?

hrutvik: I deleted them because `new_ty` has incorporated the extra fields into the underlying `T_Record`/`T_Exception` - this felt more consistent to me, but I'm happy to change (I think either way these fields will not be relied on)

HadrienRenaud: Ok sounds good to me - maybe this is something that we need to document somewhere though, for example as a comment in the code here?

PR #1390, opened 2025-08-04
[ASL-851] Remove various compatibility flags

hrutvik:
Remove the following outdated flags:
```
--allow-no-end-semicolon
--allow-expression-elsif
--allow-double-underscore
--allow-unknown
--allow-local-constants
--allow-single-arrows
--allow-hyphenated-pending-constraint
--allow-function-like-statements  
--allow-storage-discards
--allow-empty-structured-type-declarations
--no-control-flow-analysis
```

Rename the following flag to clarify it is not relevant for ASL1.0:
```
--use-field-getter-extension  --> --v0-use-field-getter-extension
```


asllib/SideEffect.ml
```diff
@@ -21,7 +21,7 @@ module TimeFrame = struct
 
   let of_ldk =
```
Roman-Manevich:
```suggestion
  let of_ldk = Execution
```

Roman-Manevich (approved):
Looks good.
Only one minor comment and a suggestion made offline.

asllib/Parser.mly
```diff
@@ -247,13 +242,6 @@ let binop ==
 
 let field_assign := separated_pair(IDENTIFIER, EQ, expr)
 
-let e_else :=
```
HadrienRenaud, line 250: Is this a change w.r.t. the reference? Doesn't the inlining of else needs to be reflected? Or is that something that was already happening?

asllib/StaticModel.ml
```diff
@@ -190,20 +190,18 @@ let rec to_ir env (e : expr) =
   match e.desc with
   | E_Literal l -> of_lit l
   | E_Var s -> (
-      try StaticEnv.lookup_constant env s |> of_lit
```
HadrienRenaud, line 190: I don't understand here, we're not removing global constants right? Maybe this is unused, in which case fine by me to remove, but this looks a bit too simple to me

asllib/Parser.mly
Roman-Manevich: It is already reflected in the reference: 
Image

hrutvik: The inlining has already been reflected in the reference I believe, in #1279

asllib/StaticModel.ml
hrutvik: Ah this is indeed a mistake, thanks for spotting

asllib/Parser.mly
HadrienRenaud: Ah I see, and because we need it to be a proper node because of recursion, we couldn't inline it at that time. Would have been good to document, but ok.

asllib/Parser.mly
```diff
@@ -90,10 +90,11 @@ let check_is_associative ~loc (op : AST.binop) =
   | `ADD | `AND | `BAND | `BEQ | `BOR | `MUL | `OR | `XOR | `BV_CONCAT
   | `STR_CONCAT ->
       ()
-  | _ -> Error.(fatal_from loc CannotParse)
+  | _ -> Error.(fatal_from loc (CannotParse (Some "Non-associative binary operators.")))
```
Roman-Manevich: How about something like "Operator %s is associative. Add parenthesis to disambiguate the compound expression"?

asllib/Parser.mly
```diff
@@ -90,10 +90,11 @@ let check_is_associative ~loc (op : AST.binop) =
...
 
 let check_not_same_prec loc op op' =
-  if prec op = prec op' then Error.(fatal_from loc CannotParse)
+  if prec op = prec op' then
+    Error.(fatal_from loc (CannotParse (Some "Binary operators of equal precedence.")))
```
Roman-Manevich: How about something like "Operators %s and %s have the same precedence. Add parenthesis to disambiguate the compound expression"?

asllib/error.ml
```diff
@@ -409,7 +409,10 @@ module PPrint = struct
           (pp_comma_list pp_type_desc)
           expected
     | AssertionFailed e -> pp_err dynamic "Assertion failed:@ %a." pp_expr e
-    | CannotParse -> pp_err parse "Cannot parse."
+    | CannotParse s -> (
+        match s with
+        | None -> pp_err parse "Cannot parse."
+        | Some s -> pp_err parse "Cannot parse.@ %s" s)
```
HadrienRenaud: Nit pick: I would print it as text (i.e. `Format` interprets each space in `s` as a breaking space)
```ocaml
        | Some s -> pp_err parse "Cannot parse.@ %a" pp_print_text s)
```

asllib/builder.ml
```diff
@@ -68,49 +43,31 @@ let lexbuf_set_filename lexbuf pos_fname =
   lexbuf.lex_curr_p <- { lexbuf.lex_curr_p with pos_fname };
   lexbuf.lex_start_p <- { lexbuf.lex_start_p with pos_fname }
 
-let from_lexbuf ast_type parser_config version (lexbuf : lexbuf) =
+let from_lexbuf ast_type _parser_config version (lexbuf : lexbuf) =
```
Roman-Manevich: Do you not need this argument? If so, why not remove it?

Roman-Manevich (approved):
Looks good to me. Just one minor question.

asllib/builder.ml
hrutvik: We don't currently, as `parser_config` will be of type `unit` now that this PR removes all the flags - but I've left all the structure in place in the code, in case we ever need to introduce any flags in future

PR #1391, opened 2025-08-05
[litmus] Fix array index anomalies in litmus

maranget:
This PR is inspired  PR #1371.  PR #1371 fixes array index problems with **herd7**, this PR will fix similar problems with **litmus7**.

The typical test is now compiled correctly in all litmus modes:
```
AArch64 T
{
  int x[2];
  0:X0=x[1]; (* Expect offset to 2nd element of x *)
}
P0;
MOV W1,#1  ;
STR W1,[X0];

forall x[1]=1
```

### DIFF
--- a/lib/stateParser.mly
+++ b/lib/stateParser.mly
@@ -186,9 +186,17 @@ maybev_notag:
 | NUM COLON NUM { Concrete ($1 ^ ":" ^ $3) }
 | NAME LBRK NUM RBRK { mk_sym_with_index $1 $3 }
 
+maybev_amper:
+| AMPER NAME { Constant.mk_sym $2 }
+| AMPER NAME LBRK NUM RBRK { mk_sym_with_index $2 $4 }
+
 maybev:
 | maybev_notag { $1 }
 | COLON NAME  { Tag $2 }
+
+%inline maybev_or_amper:
+| maybev { $1 }
+| maybev_amper { $1 }
 
 maybev_list:
 | maybev_notag COMMA maybev_list { $1::$3 }
@@ -200,9 +208,12 @@ maybev_label:
 | NUM COLON NAME { mk_lab (Misc.string_as_int $1, $3) }
 | l=LABEL { mk_lab l }
 
-%inline location_reg:
+%inline std_reg:
 | PROC COLON reg  {Location_reg ($1,$3)}
 | NUM COLON reg   {Location_reg (Misc.string_as_int $1,$3)}
+
+%inline location_reg:
+| std_reg { $1 }
 | PROC COLON SYMB_REG  {Location_reg ($1,$3)}
 | NUM COLON SYMB_REG   {Location_reg (Misc.string_as_int $1,$3)}
 | SYMB_REG        {Location_sreg $1 }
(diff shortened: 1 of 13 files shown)


lib/stateParser.mly
```diff
@@ -170,10 +170,19 @@ maybev_notag:
 | NUM COLON NUM { Concrete ($1 ^ ":" ^ $3) }
 | NAME LBRK NUM RBRK { mk_sym_with_index $1 $3 }
 
+maybev_amper:
+| AMPER NAME { Constant.mk_sym $2 }
+| AMPER NAME LBRK NUM RBRK { mk_sym_with_index $2 $4 }
+| NAME PLUS NUM { mk_sym_with_index $1 $3 }
```
murzinv: Would it make sense to reserve `t+4` purely for byte offset, so we can express both aligned and unaligned addresses?
For instance, test like bellow
```
AArch64 T
{
 int x[3];
 0:X0=x[1];
}
 P0           ;
 ADD X0,X0,#1 ;
forall 0:X0=x
```
produces
```
$ herd7 T.litmus 
Test T Required
States 1
0:X0=x+5;  <<< note that we use address+offset to represent both aligned ad unaligned addresses 
No
Witnesses
Positive: 0 Negative: 1
Condition forall (0:X0=x)
Observation T Never 0 1
Time T 0.02
Hash=3aaee2f5bc3104f9720ebef611a61876
```
Although I might be missing other ways to express the same...

maranget: Oups I did not realise that this `+` notation is that confusing, let us forget about it!

```suggestion
```

litmus/litmus.ml
```diff
@@ -340,7 +340,7 @@ let () =
         match !mode with
         | Mode.Std -> !memory
         | Mode.(Kvm|PreSi) ->
-            (* Indirect memory does not applies here *)
+            (* Indirect memory does not apply here *)
```
murzinv, line 343: Nitpick: this change belongs  to `[litmus] Fix small output anomalies `

litmus/outUtils.ml
```diff
@@ -52,13 +52,21 @@ module Make(O:Config)(V:Constant.S) = struct
   open Memory
   open Constant
 
-  let dump_addr a = match O.memory with
+  let dump_addr a   = match O.memory with
```
murzinv: Nitpick: unrelated white-space change

maranget:
> Nitpick: we seem still use `addr+offset` syntax in prelude, for instance `herd/tests/instructions/AArch64/A149.litmus` result in
> 
> ```
> %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
> % Results for herd/tests/instructions/AArch64/A149.litmus %
> %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
> AArch64 A149
> 
> {
>  uint16_t x[4]={1,2,3,4};
>  0:X0=x+2; <<< here
> }
> ```
> 
> I'm fine either keeping it `address + byte offset` (i.e. scale index to size of type) or `array index ` (i.e. wrap index into `[` `]`), former seem to be generic enough to cover all cases, yet I have no strong preference...

Fair enough, I'll make an attempt to change the output to 0:X0=x[2];

litmus/litmus.ml
maranget: Thanks!

litmus/outUtils.ml
maranget: Thanks!

maranget:
Merged, thanks @murzinv.

PR #1393, opened 2025-08-07
[asl] Some superficial fixes

hrutvik:
- Error messages
  - Increase precision in some messages (e.g. "typing" vs. "grammar" errors)
  - Change error message for non-symbolic base values
  - Correct a previous error message where I had used `pp_print_text` instead of `fprintf`
  - Primitive error messages for `HexStr` and `AsciiStr` were being reported as `DecStr`
- Fix a rule that overflowed into page margins
- Fix `AsciiStr` to make the right check on its input (`0 <= input <= 127`)

### DIFF
--- a/asllib/Native.ml
+++ b/asllib/Native.ml
@@ -261,7 +261,7 @@ module NativeBackend (C : Config) = struct
       | [ v ] -> mismatch_type v [ integer' ]
       | li ->
           Error.fatal_unknown_pos
-          @@ Error.BadArity (Dynamic, "DecStr", 1, List.length li)
+          @@ Error.BadArity (Dynamic, "HexStr", 1, List.length li)
 
     let ascii_integer = integer_range' !$0 !$127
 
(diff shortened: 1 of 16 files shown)


asllib/Native.ml
```diff
@@ -269,7 +269,7 @@ module NativeBackend (C : Config) = struct
       let open! Z in
       function
       | [ NV_Literal (L_Int i) ] ->
-          if geq zero i && leq ~$127 i then
+          if geq i zero && leq i ~$127 then
```
HadrienRenaud: What about using the mathematical notations?
```
Z.(~$0 <= i && i <= ~$127)
```
(note that the `open! Z` just above would become useless I think)

asllib/tests/regressions.t/asciistr.asl
```diff
@@ -0,0 +1,5 @@
+func main() => integer
+begin
+  println "AsciiStr(42) = ", AsciiStr(42);
```
HadrienRenaud: Why not generalize a bit more to the whole 0 to 127 range:

What about:
```
for i = 0 to 127 do
  println "AsciiStr(", i, ") = ", AsciiStr(i);
end;
```

hrutvik: I've added this without the printing (i.e. just `- = AsciiStr(i)`) as the output goes a bit funny with the various non-text characters in ASCII 0-31

PR #1397, opened 2025-08-11
Fix semantics of the LDG instructuction

murzinv:
Consider litmus test
```
AArch64 LDG
variant=memtag,sync
{
  int x=1;
  int y=2;

  tag(x)=:red;
  tag(y)=:red;
  
  0:X1=x:green;
  0:X3=y:green;
}
P0;
LDG X1,[X3]; (* load tag of y (:red) and merges into x *)
LDR W0,[X1]; (* access with x:red *)

forall 0:X0=1 /\ ~fault(P0,x)
```
which indicates that the destination register of `LDG` is intended to be merged with the tag as described in ASL:
```
bits(64) address;
bits(4) tag;

if n == 31 then
    CheckSPAlignment();
    address = SP[64];
else
    address = X[n, 64];

constant boolean stzgm = FALSE;
constant AccessDescriptor accdesc = CreateAccDescLDGSTG(MemOp_LOAD, stzgm);

address = AddressAdd(address, offset, accdesc);
address = Align(address, TAG_GRANULE);

tag = AArch64.MemTag[address, accdesc];
X[t, 64] = AArch64.AddressWithAllocationTag(X[t, 64], tag);
```
Currently `herd7` merges tag with address, so it would access to `y` which is incorrect
```
$ herd7 LDG.litmus 
Test LDG Required
States 1
0:X0=2;  ~Fault(P0,x);
No
Witnesses
Positive: 0 Negative: 1
Condition forall (0:X0=1 /\ not (fault(P0,x)))
Observation LDG Never 0 1
Time LDG 0.01
Hash=b59b82001921be752ea8a01e478e6f4c 
```
The fix is pretty straight forward but uncovers issue of destination register not being initialized in some tests, so those are fixed as well.

### DIFF
--- a/herd/AArch64Sem.ml
+++ b/herd/AArch64Sem.ml
@@ -3305,22 +3305,20 @@ Arguments:
 
       let ldg rt rn k ii =
         let ma = get_ea rn (AArch64.K k) AArch64.S_NOEXT ii in
-        let do_ldg a_virt ac ma =
-          let _do_ldg a =
+        let mv = read_reg_ord rt ii >>= loc_extract in
+        let do_ldg ac ma mv =
+          let _do_ldg (a, v) =
             let* atag = M.op1 Op.TagLoc a in
             let* tag = do_read_tag atag ii in
-            let* v = M.op Op.SetTag a_virt tag in
+            let* v = M.op Op.SetTag v tag in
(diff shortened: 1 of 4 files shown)


relokin:
Merged thanks @murzinv!

PR #1399, opened 2025-08-13
[herd] Simplify the calls to lift_memop in AArch64Sem.ml

relokin:


herd/AArch64Sem.ml
```diff
@@ -3362,16 +3317,8 @@ Arguments:
...
           (fun a_virt ma ->
              let do_ldg = do_ldg a_virt in
-             let an = Annot.N in
-             let dir = Dir.R in
-             let lbl_v = get_instr_label ii in
-             let mfault ma a ft =
-              insert_commit_to_fault ma
-                (fun _ -> set_elr_el1 lbl_v ii >>| mk_fault (Some a) dir an ii ft None)
-                None ii >>! B.fault [AArch64Base.elr_el1, lbl_v] in
-            (* CMO by VA other than DC ZVA are Tag Unchecked *)
```
diaolo01, line 3372: Is it not worth preserving this comment?

herd/AArch64Sem.ml
```diff
@@ -107,6 +111,7 @@ module Make
       module Mixed = M.Mixed(SZ)
 
       let (>>=) = M.(>>=)
+      let (let*) = M.(>>=)
```
artkhyzha, line 114: Just a question for future reference: do we now prefer let-binders in new code? Would the idea be to define them globally upon use?

herd/AArch64Sem.ml
```diff
@@ -1399,6 +1404,14 @@ module Make
                  mop ac ma >>= M.ignore >>= B.next1T
         )
 
+      let lift_kvm dir updatedb mop ma an ii mphy =
```
artkhyzha: It appears as though `lift_kvm` raises a synchronous fault by default, while `do_lift_kvm` permits passing a fault as a parameter. Seems neat.

herd/AArch64Sem.ml
```diff
@@ -45,6 +45,10 @@ module Make
     let const_pac_field = C.variant Variant.ConstPacField
     let fpac = C.variant Variant.FPac
 
+    let check_kvm ins =
```
artkhyzha, line 48: I wonder why not to do this check at the beginning of `lift_kvm`. The instruction is known via `ii` there.

artkhyzha (approved):
Thank you, @relokin -- just sharing a few small comments/questions that occurred to me.

herd/AArch64Sem.ml
relokin: I think that all cases except for AT will call lift_kvm only if kvm is true so the check wouldn't add much.

herd/AArch64Sem.ml
relokin: I like them more but I don't think we have to convert existing code, not sure if @maranget has a strong opinion on this.

herd/AArch64Sem.ml
relokin: The comment as it was, it didn't make sense. It's refers to CMOs by VA in the implementation of LDG.

relokin (commented):
Thank you both for your reviews.

herd/AArch64Sem.ml
maranget: Hi @relokin, I have no strong opinion on the subject. Generally, I am not found of constraints in coding beyond decent indentation and variable names....

herd/AArch64Sem.ml
```diff
@@ -1564,32 +1573,37 @@ Arguments:
...
-          (to_perms "r" MachSize.Word) (read_reg_ord rd ii) mzero Annot.N ii mfault
+        let mop _ac ma =
+            let* a = ma in
+            let* v = M.op1 (Op.ArchOp1 AArch64Op.SetOA) a in
+            write_reg_dest sreg v ii in
+        let mphy ma a_virt =
+            let ma = get_oa a_virt ma in
+            mop Access.PHY ma >>= M.ignore >>= B.next1T in
+        let ma = read_reg_ord rd ii in
+        let maccess a ma =
```
relokin, line 1595: Actually rather than calling do_lift_kvm here, it's much cleaner if we called directly check_ptw. So I went ahead and implemented this change.

relokin (commented):
I've made a small change. I'm planning to merge this by the end of tomorrow. Please let me know if you have any objections or need more time to review this.

PR #1402, opened 2025-08-14
[asl0] Allow ASL1 to call ASL0 empty getters/setters

hrutvik:

PR #1403, opened 2025-08-14
[asl] Fix an omission in #1402

hrutvik:
Desugaring of read-modify-write produces an `ST_Getter` call, which was missed in #1402.

### DIFF
--- a/asllib/Typing.ml
+++ b/asllib/Typing.ml
@@ -1644,7 +1644,7 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
            looks like a function call may really be a getter call *)
         | _, ST_Getter, ST_Function -> true
         (* V0 compatibility: support for calling empty getters/setters *)
-        | V0, ST_EmptyGetter, ST_Function -> true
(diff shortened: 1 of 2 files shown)


asllib/Typing.ml
```diff
@@ -1644,7 +1644,7 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
            looks like a function call may really be a getter call *)
         | _, ST_Getter, ST_Function -> true
         (* V0 compatibility: support for calling empty getters/setters *)
-        | V0, ST_EmptyGetter, ST_Function -> true
+        | V0, ST_EmptyGetter, (ST_Getter | ST_Function) -> true
         | V0, ST_EmptySetter, ST_Setter -> true
```
HadrienRenaud, line 1648: Does this one needs to accept also `ST_Procedure`?

hrutvik: I don't think so - my reasoning is that I can't see how it's possible to parse an ASL1 call with `ST_Procedure` in a case where it should refer to a setter. In particular, all calls on left-hand sides should be parsed with `ST_Setter` - unless I am missing something?

PR #1405, opened 2025-08-18
[litmus] Add a time-based barrier for AArch64

relokin:


litmus/libdir/_aarch64/timebase.c
```diff
@@ -0,0 +1,21 @@
...
+/* This software is governed by the CeCILL-B license under French law and   */
+/* abiding by the rules of distribution of free software. You can use,      */
+/* modify and/ or redistribute the software under the terms of the CeCILL-B */
+/* license as circulated by CEA, CNRS and INRIA at the following URL        */
+/* "http://www.cecill.info". We also give a copy in LICENSE.txt.            */
+/****************************************************************************/
+
+inline static tb_t read_timebase(void) {
+  tb_t val;
+  asm volatile("mrs %0, cntpct_el0" : "=r" (val));
```
murzinv, line 19: IIUC, this will be picked up by both `std` and `presi` modes, so wouldn't `CNTVCT_EL0` be the safer option?

relokin: I am not sure, the difference between the physical and virtual timers is that the physical gives a count of wall-clock time whereas the virtual time gives a count of the cycles that have passed while the current thread was scheduled on a vCPU.

Why would we use the virtual timer?

maranget: I have tried Vladimir's suggestion and the virtual counter looks readable on my Raspberry PI4B while the physical counter is not. My attempts of reading the documentation did not help me to understand the nature of those counters and of their differences.

From what you write above I am afraid the virtual counter would not be monotonically increasing... What we need is an increasing counter that has the same value for all threads.

relokin: I think the virtual timer is guaranteed to be monotonically increasing (unless it overflows) much like the physical timer. But the way I understand, the primary difference is the following:
Let say we have a thread T0 running on vCPU 0:
* T0 Reads timer
* T0 gets preempted for 1000 cycles (another thread is scheduled).
* T0 gets scheduled
* T0 Reads timer again
If we use the physical timer in this theoretical sequence, we would count 1000 cycles. If we use the virtual timer,  we would count 0 cycles.

@murzinv please chime in if I misunderstood.

relokin: FWIW, we use the physical timer in kvm mode.

murzinv: What you explained is correct and it makes perfect sense to keep physical counter for `kvm` mode. However, as @maranget found physical counter could not be exposed to user space (aka `std` mode), so only virtual one remains as an option, in case of `presi` mode we assume we run baremetal so offset to virtual counter is zero which makes it not different to physical counter.

maranget: > I think the virtual timer is guaranteed to be monotonically increasing (unless it overflows) much like the physical timer. But the way I understand, the primary difference is the following: Let say we have a thread T0 running on vCPU 0:
> 
> * T0 Reads timer
> * T0 gets preempted for 1000 cycles (another thread is scheduled).
> * T0 gets scheduled
> * T0 Reads timer again
>   If we use the physical timer in this theoretical sequence, we would count 1000 cycles. If we use the virtual timer,  we would count 0 cycles.
> 
> @murzinv please chime in if I misunderstood.

From the above description I infer that different threads will read different values. This is not compatible with current code that works as follows:
  + All treads read the timebase counter and compete to save the value read in the same shared location `next_tb`
  + Threads synchronise with a classical sense reversing barrier.
  + Threads spin reading their time base counter until the increase from `next_tb` is more than some delay.
Hence, the code assumes that threads share the same timebase counter.

maranget: > FWIW, we use the physical timer in kvm mode.

Obvously it should stay that way :)

relokin: Would this logic work for the virtual timer if we did something along the lines of:

* Threads synchronise with a classical sense reversing barrier.
* All treads read the timebase counter and save the value read in a thread local location next_tb
* Threads spin reading their time base counter until the increase from their next_tb is more than some delay.

maranget:
Here are a few suggestions for addition to this PR
  1. More adequate default value for the time-based delay: 42825fb48d56354714feba0eaa70c04cdd36bc15
  2. A new variant `-variant VirtualCounter`  replaces the (default) physical counter by the virtual one: bec07a7a54dcd02a01d70e39d288f8320d146f27
  
This is not completely satisfactory, a nice error message when the physical counter is required and not available is still lacking.

PR #1408, opened 2025-08-18
[lib] Improve transitive closure, generic code.

maranget:
Generic version of PR #1406 by @HadrienRenaud: transitive closure and strongly connected component scanning (`ssc_kont`) now share the same generic code for performing  dfs and scc computation.

Given code complexity,  writing it once looks  better, in particular as regards maintenance.

### DIFF
--- a/lib/innerRel.ml
+++ b/lib/innerRel.ml
@@ -293,8 +293,6 @@ and module Elts = MySet.Make(O) =
 
       let add x y m = ME.add x (Elts.add y (succs x m)) m
 
-      let adds x ys m = ME.add x (Elts.union ys (succs x m)) m
-
       let exists_path (e1, e2) r =
         try ignore (do_is_reachable op_succs r e1 e2) ; false
         with Found -> true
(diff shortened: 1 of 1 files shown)


lib/innerRel.ml
```diff
@@ -330,21 +328,121 @@ and module Elts = MySet.Make(O) =
...
+          match scc with
+          | [] -> assert false
+          | [_] -> res
+          | ms ->
+              let vn =
+                try ME.find n res with Not_found -> assert false in
+              List.fold_left
+                (fun res m -> ME.add m vn res)
+                res ms in
+        FullSCC.scan_map kont kont_scc ME.empty m
```
HadrienRenaud, line 445: Why is here `res` just the final result? You could carry around the successors of the current SCC? That would avoid having to do updates to res all the time?

maranget: I do not see what do you mean exactly.

HadrienRenaud: So instead of just `res` being the final result (or its work-in-progress form), you could have `res` being the final result and the neighbours of the current SCC.
The only problem is that `res` is a folding value, in the sense that it is folded among the successors of a node, instead of just being collected.
I guess what I'm suggesting is what is implemented in #1406, and here not having it is the cost of factorisation.

PR #1409, opened 2025-08-19
[asl][draft] Create ASLRef JavaScript Library

pedro-fusco:
**Summary**
- Added a new `aslref-www` directory.
- Introduced `jaslref.ml` and `dune` files to generate the `jaslref.bc.js` file.
- Initial version of the JavaScript library provides type checking support.

**Implementation Notes**
- Implemented a custom `safe_exit` function.
  - In a js_of_ocaml runtime, calling `exit` raises an `Invalid_argument` because there is no underlying OS process.
  - To allow the JavaScript module to initialize correctly, this exception is swallowed so that the runtime continues without terminating.
- The current approach works, but the implementation may need refinement or review to ensure it doesn't interfere with ASL core functionality.

**Next Steps / Future Work**
- Extend functionality beyond type checking.

### DIFF
--- a/asllib/asllib-www/dune
+++ b/asllib/asllib-www/dune
@@ -1,0 +1,17 @@
+(copy_files
+ (files ../*.{ml,mli}))
+
+(executable
+ (name jaslref)
+ (modes js)
+ (libraries
+   js_of_ocaml
+   asllib
+   zarith_stubs_js)
+ (flags (:standard -w -40-42-44-48))
+ (preprocess (pps js_of_ocaml-ppx))
+ (modules
+   jaslref
+   aslref
+   (:standard \ jaslref))
+(modules_without_implementation aST backend parserConfig))
--- a/asllib/asllib-www/jaslref.ml
+++ b/asllib/asllib-www/jaslref.ml
@@ -1,0 +1,96 @@
+(* asllib-www/jaslref.ml *)
+
+open Js_of_ocaml
+
+let dbg = false
+
+(* Redirect OCaml’s stdout/stderr into the web page *)
+let install_channel_flushers () =
+    let out s =
+    Js.Unsafe.fun_call (Js.Unsafe.pure_js_expr "asl_output")
+        [| Js.Unsafe.inject (Js.string s) |] in
+    Sys_js.set_channel_flusher stdout out ;
+    if not dbg then
+    let err s =
+        Js.Unsafe.fun_call (Js.Unsafe.pure_js_expr "asl_stderr")
+        [| Js.Unsafe.inject (Js.string s) |] in
+    Sys_js.set_channel_flusher stderr err
+
+(* The single entry point visible from JavaScript                 *)
+(*   – code   : string that contains the ASL program              *)
+(*   – version: "v0" | "v1" etc.                                  *)
+(*   – exec   : boolean, run or just parse/type-check             *)
+let run_asl code version exec =
+    install_channel_flushers ();
+
+    let code = Js.to_string code
+    and version = Js.to_string version
+    and exec = Js.to_bool   exec in
+
+    (* Build the “args” record expected by AslRef *)
+    let open Aslref in
+    let argv =
+    { exec ;
+        files =
+          [
+            ((if String.equal version "v0" then NormalV0 else NormalV1),
+             "web-input.asl")
+          ];
+        opn = None;
+        (* … rest of the boolean flags left to their defaults … *)
+        allow_no_end_semicolon = false;
+        allow_expression_elsif = false;
+        allow_double_underscore = false;
+        allow_unknown = false;
+        allow_storage_discards = false;
+        allow_hyphenated_pending_constraint = false;
+        allow_local_constants = false;
+        allow_single_arrows = false;
+        print_ast = false;
+        print_lisp = false;
+        print_serialized = false;
+        print_typed = false;
+        show_rules = false;
+        strictness =
+        (if String.equal version "v0" then Silence else TypeCheck);
+        output_format = Asllib.Error.HumanReadable;
+        use_field_getter_extension = false;
+        use_fine_grained_side_effects = false;
+        use_conflicting_side_effects_extension = false;
+        override_mode = Permissive;
+        no_primitives = false;
+        control_flow_analysis = true;
+        allow_empty_structured_type_declarations = false;
+        allow_function_like_statements = false;
+    }
+    in
+
+    let () =
+      try
+        Js_of_ocaml.Sys_js.create_file ~name:"web-input.asl" ~content:code
+      with Sys_error _ ->
+        Js_of_ocaml.Sys_js.update_file ~name:"web-input.asl" ~content:code
+    in
+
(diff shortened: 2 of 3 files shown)


asllib/aslref.ml
```diff
@@ -54,6 +54,14 @@ type args = {
   allow_function_like_statements : bool;
 }
 
+let safe_exit n =
+  (* In a js_of_ocaml runtime [exit] raises [Invalid_argument] because
+     there is no underlying OS process. Swallow that exception so that the
+     JavaScript module keeps initialising. *)
+  try exit n with Invalid_argument _ -> ()
```
HadrienRenaud: I'm not very convinced by this - would you not dropped some errors and then continue silently and get into very strange states?

What about converting the use of `safe_exit` in this file to an exception pattern:
```
exception Exit of integer

let safe_exit n =
  throw Exit(n)

(* ... *)

(* probably wrap this in a function *)
let () =
  try
    (* ... current code ... *)
  with Exit n ->
    if runnin_in_jsoo then () else exit n
```

asllib/aslref.ml
```diff
@@ -54,6 +54,14 @@ type args = {
...
+
+let running_in_jsoo : bool = false
```
HadrienRenaud: I don't really understand how this would be set to true?

What about something like:
```suggestion
let running_in_jsoo : bool =
  match Sys.backend_type with
  | Other "Javascript" -> true
  | _ -> false
```

(the `"Javascript"` is not correct - or at least I haven't tried)

asllib/aslref.ml
```diff
@@ -279,7 +288,134 @@ let or_exit f =
...
+      allow_hyphenated_pending_constraint    =
+        args.allow_hyphenated_pending_constraint;
+      allow_local_constants                  = args.allow_local_constants;
+      allow_single_arrows                    = args.allow_single_arrows;
+      allow_empty_structured_type_declarations =
+        args.allow_empty_structured_type_declarations;
+      allow_function_like_statements         =
+        args.allow_function_like_statements;
+    }
+  in
```
HadrienRenaud: Why can't that be reused from the main below?

asllib/asllib-www/dune
```diff
@@ -0,0 +1,17 @@
...
+ (libraries
+   js_of_ocaml
+   asllib
+   zarith_stubs_js)
+ (flags (:standard -w -40-42-44-48))
+ (preprocess (pps js_of_ocaml-ppx))
+ (modules
+   jaslref
+   aslref
+   (:standard \ jaslref))
```
HadrienRenaud: Isn't this just standard? then could you just remove this?

asllib/aslref.ml
```diff
@@ -253,14 +262,15 @@ let parse_args () =
...
+      else if Sys.file_exists s then ()
+      else (
+        Printf.eprintf "%s cannot find file %S\n%!" prog s;
         (* Arg.usage speclist usage_msg; *)
-        exit 1
+        exit 1)
     in
     List.iter (fun (_, s) -> ensure_exists s) args.files;
-    Option.iter ensure_exists args.opn
+    if not running_in_jsoo then Option.iter ensure_exists args.opn
```
HadrienRenaud: I think this is probably redundant with the line 266 check that it's not running in jsoo?

asllib/asllib-www/jaslref.ml
```diff
@@ -0,0 +1,96 @@
...
+    let module WebFS = struct
+      module StringMap = Map.Make (String)
+
+      let files = ref (StringMap.singleton "web-input.asl" code)
+
+      let find name =
+        match StringMap.find_opt name !files with
+        | Some s -> s
+        | None -> failwith ("File not found: " ^ name)
+    end in
```
HadrienRenaud: Why are both of those needed? I would assume that if I use `WebFS` I don't need to use the real file system with `Js_of_ocaml.Sys_js.read_file`?

asllib/aslref.ml
```diff
@@ -279,138 +289,236 @@ let or_exit f =
...
-  in
+  (* Dump instrumentation (rule usage) if requested. *)
+  if args.show_rules then
+    Format.printf "@[<v 3>Used rules:@ %a@]@."
+      (Format.pp_print_list
+          ~pp_sep:Format.pp_print_cut
+          Instrumentation.SemanticsRule.pp)
+      used_rules
+
+let () =
```
HadrienRenaud, line 442: I the only difference between this and `run_with` that one uses `Builder.from_string` and the other uses `Builder.from_file`?

If so could we refacto this to be just a call to `run_with` with a special builder function?

asllib/asllib-www/jaslref.ml
```diff
@@ -0,0 +1,95 @@
+(* asllib-www/jaslref.ml *)
+
```
HadrienRenaud, line 2: What about adding `open Aslref` here?

asllib/aslref.ml
```diff
@@ -276,21 +284,19 @@ let parse_args () =
         Printf.eprintf
           "No files supplied! Run `aslref --help` for information on usage."
       in
-      exit 1
+      if running_in_jsoo then raise (Exit 1) else exit 1
```
HadrienRenaud: I think because you're catching it at the end (line 446), you can always throw `Exit`

```suggestion
      raise (Exit 1)
```

asllib/aslref.ml
```diff
@@ -276,21 +284,19 @@ let parse_args () =
...
   in
 
   let () =
     if !show_version then
       let () =
         Printf.printf "aslref version %s rev %s\n%!" Version.version Version.rev
       in
-      exit 0
+      if running_in_jsoo then raise (Exit 0) else exit 0
```
HadrienRenaud: Similarly:
```suggestion
      raise (Exit 0)
```

asllib/aslref.ml
```diff
@@ -331,7 +337,7 @@ let () =
             let output_format = args.output_format
           end) in
           EP.eprintln e;
-          exit 1
+          if running_in_jsoo then raise (Exit 1) else exit 1
```
HadrienRenaud: And also:

```suggestion
          raise (Exit 1)
```

asllib/aslref.ml
```diff
@@ -431,5 +437,11 @@ let () =
         (pp_print_list ~pp_sep:pp_print_cut Instrumentation.SemanticsRule.pp)
         used_rules
   in
+  ()
```
HadrienRenaud: What about still using the exit code passed down from above?

```suggestion
  raise (Exit exit_code)
```

asllib/aslref.ml
```diff
@@ -431,5 +437,10 @@ let () =
         (pp_print_list ~pp_sep:pp_print_cut Instrumentation.SemanticsRule.pp)
         used_rules
   in
+  match exit_code with 0 -> () | _ -> raise (Exit exit_code)
```
HadrienRenaud, line 440: Just a suggestion:

```suggestion
  if exit_code != 0 raise (Exit exit_code)
```

PR #1410, opened 2025-08-21
A few ASL performance improvements

HadrienRenaud:
This PR improves the run time of the ASL regression tests.

For example, the `aslref asllib/tests/stdlib.t/sqrt.asl` goes from 7s to 2.5s
on my machine.

- **[asl] Add an option not to generate branching commit messages**
- **[asl] Use static modules instead of values**


asllib/Interpreter.ml
```diff
@@ -66,6 +66,7 @@ module type Config = sig
 
   val unroll : int
   val error_handling_time : Error.error_handling_time
+  val empty_branching_effects : bool
```
Roman-Manevich: Nit: How about`empty_branching_effects_optimization`?

asllib/env.ml
```diff
@@ -161,28 +158,39 @@ module RunTime (C : RunTimeConf) = struct
       local =
         {
           env.local with
-          storage = Storage.add x v env.local.storage;
+          storage = IMap.add x v env.local.storage;
           declared = x :: env.local.declared;
         };
     }
 
+  let _assign x v =
```
hrutvik, line 166: nit: why do these names begin with underscores?

hrutvik (approved):

asllib/env.ml
HadrienRenaud: I did that because they were the old `Storage.assign` function, to differentiate them from `Env.Make.assign`, usually beginning with underscore signifies that this is a private function.

PR #1412, opened 2025-08-25
[litmus,vmsa] Less radical attribute change check

maranget:
According to some comment in code introduced by PR #1117, some cache flush after test run is performed when there is a risk of writing to some location with non-default attributes.

This risk is identified the predicate `memattrs_change` that checks the following two properties:
 1. The location PTE value is set to a non-default initial value. Notice that this check does not take attributes into account.
 2. This initial value has some attributes.

Running a test on a Raspberry P5B, we experienced some deadlock, or extreme slowdown due to the extra cache flush. Notice that the test ignores memory attributes:
```
AArch64 MP+UpdateAF
TTHM=P0:HD
{
0:X1=x; 1:X1=x;
0:X3=y; 1:X3=y;
1:X2=1;
pte_x=(af:0);
1:X6=pte_x;
1:X4=(oa:phy_x,af:0)
}
P0          | P1              ;
LDR W0,[X1] | STR W2,[X3]     ;
DMB LD      | SWPL X4,X5,[X6] ;
LDR W2,[X3] |                 ;

exists
  0:X2=0 /\ 1:X5=(oa:phy_x,af:0) /\ [PTE(x)] = (oa:PA(x))

```

We attempt to fix the problem by changing the predicate `memattrs_change` that now checks, as its name suggests, that a page table entry attributes are initialised to non-default values. As a result no corrective cache flush is emitted and the test runs in decent time.

### DIFF
--- a/lib/AArch64PteVal.ml
+++ b/lib/AArch64PteVal.ml
@@ -174,6 +174,9 @@ module Attrs = struct
   let as_kvm_symbols = function
     | Normal n -> NormalAttrs.as_kvm_symbols n
     | Device d -> DeviceAttrs.as_kvm_symbols d
+
+  let is_default = eq default
+
 end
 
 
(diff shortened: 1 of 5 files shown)


maranget:
Thanks a lot @relokin. Merge is on its way.

PR #1413, opened 2025-08-26
ASL-872: shadow banning bug

HadrienRenaud:
This PR fixes a subtle bug that appears when trying to overide a global variable without using the local variable.
It was previously allowed, and now forbidden.

### DIFF
--- a/asllib/doc/Makefile
+++ b/asllib/doc/Makefile
@@ -21,16 +21,23 @@ CONTROLS=ifempty.tex ifcode.tex
 .PHONY: control
 control: control.tex $(CONTROLS)
 
+LATEXMK=latexmk
+
 ASLReference.pdf: control\
 				ASLASTLines.tex ASLTypeSatisfactionLines.tex ASLStaticModelLines.tex ASLStaticInterpreterLines.tex\
 				ASLStaticEnvLines.tex ASLEnvLines.tex ASLASTUtilsLines.tex\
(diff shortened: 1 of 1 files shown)


asllib/tests/ASLTypingReference.t/TypingRule.UseLDI.asl
```diff
@@ -0,0 +1,8 @@
+func main () => integer
+begin
+  let i = 32; // { Other(x) }
```
hrutvik:
```suggestion
  let i = 32; // { Other(i) }
```

asllib/tests/ASLTypingReference.t/TypingRule.UseLDI.asl
```diff
@@ -0,0 +1,8 @@
...
+  var (x, y, z) = (0, 1, 2); // { Other(x), Other(y), Other(z) }
+  var (a, -, b) = (0, 1, 2); // { Other(a), Other(b) }
```
hrutvik: I wonder if this line should be removed, due to the added complication of desugaring the discard into a fresh variable name - but I'll leave it up to you

HadrienRenaud: Indeed ok, I had forgotten about this

asllib/tests/ASLTypingReference.t/TypingRule.UseLDI.asl
HadrienRenaud: Done thanks

PR #1416, opened 2025-08-29
GitHub Actions for prebuilt tools

hrutvik:
This PR adds two GitHub Actions workflows:

- `docker.yml`: builds and publishes an Ubuntu-based Docker image with `herd` tools already built and installed.
  - The image is built using `.github/Dockerfile`, which derives from an [`ocaml/opam`](https://hub.docker.com/r/ocaml/opam) image.
  - It is published to the GitHub Container Registry (appears as "[Packages](https://github.com/orgs/herd/packages?repo_name=herdtools7)" on the repository landing page).
- `static_binaries.yml`: builds statically-linked binaries for `herd` tools.
  - The binaries are published as an artifact of the GitHub Actions run.
  - The process roughly follows this [blog post](https://ocamlpro.com/blog/2021_09_02_generating_static_and_portable_executables_with_ocaml/), building on an [`ocaml/opam`](https://hub.docker.com/r/ocaml/opam) [Alpine](https://www.alpinelinux.org/) image.
  - A `static` profile was added to the top-level `dune` file to enable this. The ASL library also required a `(no_dynlink)` clause in `asllib/dune`, but this has been enabled only for the `static` profile.

Both workflows are manually triggered, and given as input a specific tag to build.

### DIFF
--- a/.github/workflows/static-binaries.yml
+++ b/.github/workflows/static-binaries.yml
@@ -1,0 +1,52 @@
+name: Build static binaries
+
+on:
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: 'herdtools7 tag to build'
+        required: true
+        default: 'master'
+
+permissions: read-all
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    container:
+      image: ocaml/opam:alpine-3.21-ocaml-5.2
+      options: --user root # GitHub Actions seems to require root
+    env:
+      OPAMROOT: '/home/opam/.opam' # Ensure opam directory is found as root
+      OPAMROOTISOK: 1 # Suppress warnings about running opam as root
+
+    steps:
+      - name: Checkout tree
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.inputs.tag }}
+
+      - name: Install prerequisites
+        run: |
+          apk add --quiet gmp-dev pkgconfig
+          opam install . --deps-only
+
+      - name: Enable verbose dune output
+        run: |
+          mkdir -p ~/.config/dune
+          echo $'(lang dune 2.7)\n(display short)' > ~/.config/dune/config
+
+      - name: Build
+        run: |
+          mkdir binaries
+          opam exec -- make build PREFIX=binaries DUNE_PROFILE=static
+          opam exec -- make install PREFIX=binaries DUNE_PROFILE=static
+
+      # NOTE: --set-libdir will need to be passed to various binaries, including `herd7` and `litmus7`
+      - name: Upload binaries
+        uses: actions/upload-artifact@v4
+        with:
+          name: binaries
+          path: |
+            binaries/bin
+            binaries/share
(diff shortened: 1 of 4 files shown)


asllib/dune
```diff
@@ -48,6 +48,24 @@
 
 (library
  (name asllib)
+ (enabled_if
+  (or
+   (= %{profile} "dev")
+   (= %{profile} "release")))
```
HadrienRenaud, line 54: Do we know that `profile` can only take 3 values?
Why no say `(no (= %{profile} "static"))`?

.github/workflows/static-binaries.yml
```diff
@@ -0,0 +1,46 @@
...
+          apk add --quiet gmp-dev pkgconfig
+          opam install . --deps-only
+
+      - name: Enable verbose dune output
+        run: |
+          mkdir -p ~/.config/dune
+          echo $'(lang dune 2.7)\n(display short)' > ~/.config/dune/config
+
+      - name: Build
+        run: opam exec -- make build DUNE_PROFILE=static
```
HadrienRenaud: Wouldn't that be more `make install PREFIX=artifacts DUNE_PROFILE=static`?

And then you upload only the installed artifacts?

.github/workflows/static-binaries.yml
```diff
@@ -0,0 +1,46 @@
...
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    container:
+      image: ocaml/opam:alpine-3.21-ocaml-5.2
+      options: --user root # GitHub Actions seems to require root
+    env:
+      OPAMROOT: '/home/opam/.opam'
+      OPAMROOTISOK: 1
```
HadrienRenaud: Can we put a comment explaining what this this?

asllib/dune
hrutvik: I tried this initially, but got the following error:
```Error: 'not' is only available since version 3.2 of the dune language. . Please update your dune-project file to have (lang dune 3.2)```

**Comment by hrutvik** [comment]
- Regarding `--set-libdir`: I will add a comment indicating this. Is there anywhere else it might be useful?
- Regarding the Docker image: yes I think the hope would be to allow people to use with OCaml without having to first set it up.

File: .github/workflows/static-binaries.yml (hunk 2 above)
Comment (hrutvik) [reply .github/workflows/static-binaries.yml]: There was a slight issue appearing:
```cp: cannot create directory 'binaries/share/herdtools7/herd': No such file or directory```
But I've fixed this by modifying `dune-install.sh`

**Comment by maranget** [comment]
Additionally, `make all` does not work:
```
$ make all
File "asllib/dune", lines 64-76, characters 0-350:
64 | (library
65 |  (name asllib)
66 |  (enabled_if
....
74 |  (flags
75 |   (:standard -w -40-42-48))
76 |  (libraries menhirLib zarith))
Error: Library "herdtools7.asllib" appears for the second time in this
directory
make: *** [just-build] Error 1          
```
I am wondering how such a situations was not signaled by CI.

hrutvik:
Hi @akiyks, we have reverted these changes now - please do let us know if this fixes the issue for you!

PR #1417, opened 2025-08-29
[ASL-873] Forbid empty function bodies

hrutvik:
Delete the `maybe_empty_stmt_list` non-terminal, and replace its occurrences with `stmt_list`.


Roman-Manevich (approved):
Looks good to me!
Please add `GuideRule.NoEmptyBody` somewhere in the `Subprogram Declarations` chapter, exaplining that subprogram bodies must not be empty and that in such cases `pass` should be used.

PR #1419, opened 2025-09-02
[ci] Use a TeX Live container to build ASL reference PDF

hrutvik:


.github/workflows/build-asl-reference.yml
```diff
@@ -21,30 +18,39 @@ concurrency:
...
-      - run: opam exec -- make build DUNE_PROFILE=dev
+      - name: Mark working directory as "safe"
+        run: git config --global --add safe.directory $GITHUB_WORKSPACE
 
-      - run: opam exec -- make install
+      - name: Build and install herdtools7
+        run: |
+          opam install . --deps-only --yes
+          opam exec -- make build DUNE_PROFILE=dev
+          opam exec -- make install
```
HadrienRenaud: Technically `make asldoc` that's on the next line should build `bento`, which is the only needed executable from the herd toolsuite to build the documentation.
So probably those lines are not necessary.

hrutvik: Ah I thought there are bits of the reference with verbatim output from running ASLRef that assume `aslref` exists on `PATH`?

HadrienRenaud: Oh indeed at least doclint.py needs it...
Probably we should edit the Makefile then

hrutvik: As discussed, will rebase on top of #1420 once it's merged.

PR #1422, opened 2025-09-04
[ASL] Improve handling of symbolic recursion

HadrienRenaud:
- **[asl] Rename stack-size to pending-calls**
- **[asl] Improve handling of recursive call in symbolic cases**

### DIFF
--- a/asllib/Interpreter.ml
+++ b/asllib/Interpreter.ml
@@ -151,12 +151,19 @@ module Make (B : Backend.S) (C : Config) = struct
 
   let fatal_from_no_env pos = Error.fatal_from pos
 
-  let fatal_from_genv pos genv e =
+  let display_call_stack pos genv =
     if C.display_call_stack_on_error then
-      Format.eprintf "@[<2>%a:@ @]" pp_pos (pos, genv);
+      Format.eprintf "@[<2>%a:@ @]" pp_pos (pos, genv)
+
+  let fatal_from_genv pos genv e =
+    display_call_stack pos genv;
     Error.fatal_from pos e
 
   let fatal_from pos env e = fatal_from_genv pos env.IEnv.global e
+
+  let fail_from pos env e =
+    display_call_stack pos env.IEnv.global;
+    B.fail (Error.ASLException (add_pos_from pos e)) Cutoff
 
   (*****************************************************************************)
   (*                                                                           *)
@@ -1272,7 +1279,7 @@ module Make (B : Backend.S) (C : Config) = struct
         return_continue new_env |: SemanticsRule.SPrint
     (* End *)
     | S_Pragma _ -> assert false
-    | S_Unreachable -> fatal_from s env Error.UnreachableReached
+    | S_Unreachable -> fail_from s env Error.UnreachableReached
 
   (* Evaluation of Blocks *)
   (* -------------------- *)
--- a/herd/ASLSem.ml
+++ b/herd/ASLSem.ml
@@ -912,6 +912,7 @@ module Make (C : Config) = struct
         let bitvector_length = bitvector_length
         let v_unknown_of_type = v_unknown_of_type
         let primitives = make_extra_funcs ii_env
+        let fail = M.failT
       end in
       let module ASLInterpreter =
         Asllib.Interpreter.Make (ASLBackend) (ASLInterpreterConfig)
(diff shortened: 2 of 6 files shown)


hrutvik (approved):
Minor suggestion: as the `Cutoff` is not in the reference I believe, perhaps worth commenting somewhere about what it does as an implementation-specific thing

asllib/Interpreter.ml
```diff
@@ -304,7 +319,8 @@ module Make (B : Backend.S) (C : Config) = struct
               match eval_res with
               | Normal (v, env2) -> return (v, env2)
               | Throwing (_, ty, _env2) ->
-                  fatal_from d env (UnexpectedInitialisationThrow (ty, name)))
+                  fatal_from d env (UnexpectedInitialisationThrow (ty, name))
+              | Cutoff -> assert false)
```
maranget: Out of curiosity, why are `Throwing` and `Cutoff` handled differently here?

asllib/Interpreter.ml
```diff
@@ -1567,7 +1598,8 @@ module Make (B : Backend.S) (C : Config) = struct
         Error.(fatal_unknown_pos (MismatchedReturnValue (Dynamic, main_name)))
     | Throwing ((v, _, _), ty, _genv) ->
         let msg = Format.asprintf "%a %s" PP.pp_ty ty (B.debug_value v) in
-        Error.fatal_unknown_pos (Error.UncaughtException msg))
+        Error.fatal_unknown_pos (Error.UncaughtException msg)
+    | Cutoff -> return (B.v_of_int ~-1))
```
maranget: Magic constant `-1`?

asllib/StaticInterpreter.ml
```diff
@@ -29,7 +29,8 @@ let ( |: ) = Instrumentation.TypingNoInstr.use_with
 module InterpConf = struct
   module Instr = Instrumentation.SemanticsNoInstr
 
-  let unroll = 0
+  let unroll = 1 lsl 12
+  let recursive_unroll _ = Some (1 lsl 12)
```
maranget, line 33: Why this change did has `unroll = 0` has some special meaning that changed?  Wouldn't `recursive_unroll _ = None` be equivalent in practice and would convey the intent (no limit) better?

herd/ASLSem.ml
```diff
@@ -938,7 +943,7 @@ module Make (C : Config) = struct
         | Ok m -> m
         | Error err -> Asllib.Error.error_to_string err |> Warn.fatal "%s"
       in
-      assert (V.equal i V.zero);
+      assert (V.equal i V.zero || V.equal i (V.intToV ~-1));
```
maranget: Why not return zero ? Does `-1` have a particular meaning?

maranget:
The following files can be deleted:
```
herd/tests/instructions/ASL-pseudo-arch/recurse-arbitrary.litmus.expected-failure
herd/tests/instructions/ASL-pseudo-arch/recurse-read-memory-too-far.litmus.expected-failure
herd/tests/instructions/ASL-pseudo-arch/recurse-too-far.litmus.expected-failure
```

maranget:
FYI, rebasing asl-vmsa on top of this PR is very easy,  Currently running on the extended VMSA test suite.

asllib/Interpreter.ml
HadrienRenaud: Replaced with 0

herd/ASLSem.ml
HadrienRenaud: I've replaced this with 0 as suggested.

asllib/Interpreter.ml
HadrienRenaud: Well they can't be treated in the same way (difficult to define `ty` and `name` for `Cutoff`), and then, there is the question: who is going to read this error? In the `Throwing` case, it could be anybody who tries to execute a program with `aslref`. In the `Cutoff` case, it needs to be someone who tries an ASL litmus test with `herd` (roughly you and me). I think we can leave the poor explanation messages for people running `herd`.
Nevertheless, I've added an explanation message before the `assert false`.

asllib/StaticInterpreter.ml
HadrienRenaud: So the intent with `recursive_unroll` to be something very big but fixed is to say that at some point your expression is not statically evaluable, and this will result in an `UnsupportedExpresssion` error, which will probably be caught by the type-checker

PR #1423, opened 2025-09-04
[build] Update minimal versions in the .opam file

maranget:
Tested with script:
```
opam switch create  TST 4.08.1
eval $(opam env --switch=TST)
opam install conf-which
opam pin dune 2.9.3
opam pin menhir 20200123
opam pin zarith  1.13
make all test PREFIX=/opt/TST
```

### DIFF
--- a/.github/workflows/make-test.yml
+++ b/.github/workflows/make-test.yml
@@ -25,9 +25,13 @@ jobs:
         cfg:
           - ocaml-version: "4.08.1"
             with-fmt: false
+            with-mins: true
+            dune-cache: "enabled"
             job-name: "OCaml v4.08"
           - ocaml-version: "5.2"
             with-fmt: true
+            with-mins: false
+            dune-cache: "enabled-except-user-rules"
             job-name: "OCaml v5.2, with formatting check"
 
     name: Make test on ${{ matrix.cfg.job-name }}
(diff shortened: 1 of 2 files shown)


maranget:
Hi @HadrienRenaud. How to use this with CI? I do not know! See PR #1313?

.github/workflows/make-test.yml
```diff
@@ -25,9 +25,12 @@ jobs:
         cfg:
           - ocaml-version: "4.08.1"
             with-fmt: false
+            with-mins: true
+            dune-cache: enabled
```
HadrienRenaud: I think this is unused here.

maranget: I do not understand, `dune-cache` is not used?

HadrienRenaud: Well, at line 41 below, the environment variable is set to the static `'enabled-except-user-rules'` and not to the github workflow variable `${{ matrix.cfg.dune-cache }}`, so indeed I don't think the variable `${{ matrix.cfg.dune-cache }}` is used in this workflow

maranget: Ok you mean that `dune-cache` is unused. This was not clear from the start.

maranget:
> "above comment": https://github.com/herd/herdtools7/pull/1423/files#r2330029961
> 
> "one week before": time we would need to merge #1313, which restores compatibility of our code with `dune>=3.10.0`.

I see it now. If #1313 supersedes both PR #1423 and #1424, we'd better forget about those two. I have a look at #1313.

maranget:
Superseded by PR #1313.

PR #1424, opened 2025-09-04
[ci] Test PR with minimal version

HadrienRenaud:


.github/workflows/make-test.yml
```diff
@@ -47,6 +51,12 @@ jobs:
           ocaml-compiler: ${{ matrix.cfg.ocaml-version }}
           dune-cache: true
 
+      - run: |
+          opam pin add dune 3.5.0
+          opam pin add menhir 20220210
```
maranget: Increasing `memhir` minimal version to "20240715" should fix the run failure. However, I'd like to have your opinion on comparing this PR with PR #1423.

HadrienRenaud: Looks like a little bit more is needed, see [ci run](https://github.com/herd/herdtools7/actions/runs/17546382551/job/49828703781?pr=1424).

maranget: We may need a more recent version of `dune` to deal with  the more recent version of `menhir`.

HadrienRenaud: I don't think so, I think it's the hand-written modules in `asllib/tests` that make `make test` fail. 

```
File "asllib/tests/dune", line 1, characters 0-0:
Error: Module "Helpers" is used in several stanzas:
- asllib/tests/dune:8
- asllib/tests/dune:24
To fix this error, you must specify an explicit "modules" field in every
library, executable, and executables stanzas in this dune file. Note that
each module cannot appear in more than one "modules" field - it must belong
to a single library or executable.
make: *** [Makefile:78: dune-test] Error 1
```

maranget: You mean that the need for recent versions originates in  one of our dune file?

PR #1425, opened 2025-09-08
[check] New action that checks the .opam file

maranget:
More precisely we test the opam installation as much as possible with minimal versions.


maranget:
The technique functions. Unfortunately, we still suffer from a 10 minutes timeout before finding the right place to download ocamlfind,1.9.1 from...

.github/workflows/check-opam-minimals.yml
```diff
@@ -0,0 +1,37 @@
+name: Test build with mininal versions
+
+on:
+  push:
```
HadrienRenaud, line 4: Maybe it only needs to run every week? In which case I would suggest removing this line

.github/workflows/check-opam-minimals.yml
```diff
@@ -0,0 +1,37 @@
...
+
+      - name: Set-up OCaml 4.08
+        uses: ocaml/setup-ocaml@v3
+        with:
+          ocaml-compiler: 4.08.1
+      - run: |
+          opam install . -v -v --deps-only --with-test --solver=builtin-mccs+glpk --criteria="-new"
+          opam exec -- make build DUNE_PROFILE=dev
+          opam exec -- make test DUNE_PROFILE=dev
+          opam exec -- make install
```
HadrienRenaud, line 37: Why not?

```suggestion
          opam exec -- make build test install DUNE_PROFILE=dev
```

.github/workflows/check-opam-minimals.yml
maranget: `push:` was here for testing, will remove it. Thanks

.github/workflows/check-opam-minimals.yml
maranget: Nice. will change.

.github/workflows/check-opam-minimals.yml
HadrienRenaud: Cool, you should be able to trigger it manually if needed also

maranget:
> Hi @maranget, Did you merge with the `push` on? Github shows it as still on the patch now

Oups I have not merge the final code :( Thanks @HadrienRenaud. I create a new PR for fixing this.

PR #1421, opened 2025-09-04
Revert build changes from #1416

hrutvik:
Attempt to fix a [bug](https://github.com/herd/herdtools7/pull/1416#issuecomment-3251217698) reported by @akiyks and a [bug](https://github.com/herd/herdtools7/pull/1416#issuecomment-3252155225) reported by @maranget.

PR #1426, opened 2025-09-08
[ci] Attempt to fix static binaries build

hrutvik:
In particular, attempt to reintroduce the functionality reverted by #1421. Tested locally with minimum versions from the `.opam` file (except OCaml 4.10, as I can't install any lower locally).

### DIFF
--- a/.github/workflows/static-binaries.yml
+++ b/.github/workflows/static-binaries.yml
@@ -38,6 +38,9 @@ jobs:
           mkdir -p ~/.config/dune
           echo $'(lang dune 2.7)\n(display short)' > ~/.config/dune/config
 
+      - name: Patch asllib/dune
+        run: sed -i -e 's/(name asllib)/(name asllib)(no_dynlink)/g' asllib/dune
+
       - name: Build
         run: |
           mkdir /artifacts
--- a/.github/workflows/static-binaries.yml
+++ b/.github/workflows/static-binaries.yml
@@ -20,6 +20,8 @@ jobs:
       OPAMROOT: '/home/opam/.opam' # Ensure opam directory is found as root
       OPAMROOTISOK: 1 # Suppress warnings about running opam as root
 
+    name: Static binaries for tag ${{ github.event.inputs.tag }}
+
     steps:
       - name: Checkout tree
         uses: actions/checkout@v4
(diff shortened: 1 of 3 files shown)


.github/workflows/static-binaries.yml
```diff
@@ -40,13 +42,13 @@ jobs:
         run: |
           mkdir binaries
           opam exec -- make build PREFIX=binaries DUNE_PROFILE=static
-          opam exec -- make install PREFIX=binaries DUNE_PROFILE=static
+          opam exec -- make install DUNE_PROFILE=static
```
HadrienRenaud: Why having 2 different ones? (with different `PREFIX` on top of this)


```suggestion
         opam exec -- make build install PREFIX=binaries DUNE_PROFILE=static
```

hrutvik: The `make install` step with a relative `PREFIX` fails I believe, as `dune-install.sh` changes directory before copying files. The "fix" I reverted [here](https://github.com/herd/herdtools7/pull/1421/commits/86d2bf0ee10c8f89e8825a22b928198f0bf8ace7) solves the issue locally/in Alpine CI, but seems to fail on other users' systems. So my proposed solution is to use `PREFIX` in the build stage to set a relative `libdir` prefix for the binaries, and then use the default `make install` behaviour and copy the binaries from home directory. What do you think?

HadrienRenaud: Oh I see. Why not set prefix to `"$PWD/binaries"`? I think `$PWD` is posix compliant so maybe that's good enough?

HadrienRenaud: Or even have a volume for it? Like `/artifacts`?

hrutvik: I've gone for the `/artifacts` suggestion!

.github/workflows/static-binaries.yml
```diff
@@ -36,17 +38,20 @@ jobs:
...
+      - name: Patch asllib/dune
+        run: sed -i -e 's/(name asllib)/(name asllib)(no_dynlink)/g' asllib/dune
+
       - name: Build
         run: |
-          mkdir binaries
+          mkdir /artifacts
           opam exec -- make build PREFIX=binaries DUNE_PROFILE=static
-          opam exec -- make install PREFIX=binaries DUNE_PROFILE=static
+          opam exec -- make install PREFIX=/artifacts DUNE_PROFILE=static
```
HadrienRenaud, line 48: Sorry to still be problematic with this. Could we use the same PREFIX for the build and the install?


```suggestion
          opam exec -- make build install PREFIX=/artifacts DUNE_PROFILE=static
```

hrutvik: Ah sorry perhaps I have misunderstood - I thought the two `PREFIX`es could serve different purposes, in that:
- the `build` one sets the `libdir` visible from `herd7 -libdir`
- the `install` one copies the built files to the given location, but must be an absolute path.

So I was hoping to set a relative path for the `libdir` using the `build` `PREFIX`, and then `install` to an absolute directory. The thinking is that perhaps users of these binaries can more easily set up their directory structure locally to avoid passing `-set-libdir`. What do you think?

HadrienRenaud: Oh I see. I think we can try this out and if at some point it doesn't work we'll revert to my suggested behavior?

PR #1427, opened 2025-09-10
Subprogram types bug

hrutvik:
This PR fixes unexpected behaviour when overloading accessors. In particular, a getter of one accessor could clash with a setter of another accessor, producing internal assertion failures. The fix is to filter out call candidates when fetching matching subprograms, rather than checking we have found a valid call candidate afterwards. For example, when fetching call candidates for a setter call on a left-hand side, we can ignore all matching getters/functions (and vice versa).


asllib/Typing.ml
```diff
@@ -295,8 +295,20 @@ module FunctionRenaming (C : ANNOTATE_CONFIG) = struct
...
+       looks like a function call may really be a getter call *)
+    | _, ST_Getter, ST_Function -> true
+    (* V0 compatibility: support for calling empty getters/setters *)
+    | V0, ST_EmptyGetter, (ST_Getter | ST_Function) -> true
+    | V0, ST_EmptySetter, ST_Setter -> true
+    | _ -> false
+
   (* Begin SubprogramForName *)
-  let subprogram_for_name ~loc env version name caller_arg_types =
+  let subprogram_for_name ~loc env version name caller_arg_types call_type =
```
Roman-Manevich: perhaps rename to `subprogram_for_name_and_call_type`?

asllib/doc/SubprogramCalls.tex
```diff
@@ -1453,19 +1446,36 @@ \chapter{Subprogram Calls\label{chap:SubprogramCalls}}
...
   \funcdef \eqdef G^\tenv.\subprograms(\name) \\
-  \hasargclash(\tenv, \formaltypes, \funcdef.\funcargs) \typearrow \vb \OrTypeError\\\\
-  \filtercallcandidates(\tenv, \formaltypes, \candidatesone) \typearrow \matchesone \OrTypeError\\\\
-  \matches \eqdef \choice{\vb}{\;[(\name,\funcdef)] \concat \matchesone}{\;\matchesone}
+  \hasargclash(\tenv, \formaltypes, \funcdef.\funcargs) \typearrow \vbone \OrTypeError\\\\
+  {
+    \vbtwo \eqdef
+    \left(
+    \begin{array}{l}
+      \funcdef.\funcsubprogramtype = \calltype \; \lor \\
```
Roman-Manevich: perhaps move this condition to a new function to match the code (`call_type_matches`)?

asllib/Typing.ml
hrutvik: This would be more accurately be something like `subprogram_for_name_and_args_and_call_type` though right? I worry this gets quite a long name, particularly for the reference. How about something more generic, like `fetch_subprogram` or similar?

Roman-Manevich: Sounds good. Here's another suggestion: `subprogram_for_signature`.

hrutvik: Even better! I'll go with that

PR #1429, opened 2025-09-11
[asl] Minor improvements, including error message bugs

hrutvik:
- Add a `--no-stdlib` flag to avoid loading the standard library. This can condense output for printing/debugging small examples.
- Fix an `End_of_file` exception when printing errors for ASL0 mode, caused by trying to input lines outside of file limits.
- Remove a rogue square bracket in pretty-printing code.

### DIFF
--- a/asllib/error.ml
+++ b/asllib/error.ml
@@ -259,7 +259,10 @@ open struct
       if end_bol > start_bol then really_input_string chan (end_bol - start_bol)
       else ""
     in
-    let last_line = input_line chan in
+    let last_line =
+      (* [input_line] raises [End_of_file] if EOF is at the start of the line *)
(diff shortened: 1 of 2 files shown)


asllib/error.ml
```diff
@@ -259,7 +259,7 @@ open struct
       if end_bol > start_bol then really_input_string chan (end_bol - start_bol)
       else ""
     in
-    let last_line = input_line chan in
+    let last_line = try input_line chan with End_of_file -> "" in
```
HadrienRenaud: Can we add a comment saying that `input_line` only raises `End_of_file` if the EOF is found at the beginning of the line? I know it's in the documentation but I had to check

PR #1431, opened 2025-09-11
[ASL-868] Improved error messages for no `main`

hrutvik:
Improve error reporting in two cases:
- No files are passed to `aslref`.
- No `main` function can be found.

For the second one, refine the way `main` is found by exposing a new entry point in `Typing.ml`, which calls `subprogram_for_signature`.

### DIFF
--- a/asllib/Typing.ml
+++ b/asllib/Typing.ml
@@ -149,6 +149,7 @@ end
 module type S = sig
   val type_check_ast : AST.t -> AST.t * global
   val type_check_ast_in_env : global -> AST.t -> AST.t * global
+  val find_main : global -> identifier
 end
 
 module Property (C : ANNOTATE_CONFIG) = struct
@@ -4452,6 +4453,19 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
       (List.rev ast_rev, env)
 
   let type_check_ast ast = type_check_ast_in_env empty_global ast
+
+  (* Note: produces a *dynamic* error if the main function cannot be found *)
+  let find_main env =
+    let env = with_empty_local env in
+    let _, main_name, func, _ =
+      try
+        Fn.subprogram_for_signature ~loc:dummy_annotated env V1 "main" []
+          ST_Function
+      with Error.(ASLException _) -> Error.(fatal_unknown_pos NoEntryPoint)
+    in
+    match func.return_type with
+    | Some { desc = T_Int UnConstrained } -> main_name
+    | _ -> Error.(fatal_unknown_pos NoEntryPoint)
 end
 (* End *)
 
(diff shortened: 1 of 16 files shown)


asllib/error.ml
```diff
@@ -217,6 +218,7 @@ let error_label = function
   | PrecisionLostDefining -> "PrecisionLostDefining"
   | UnexpectedCollection -> "UnexpectedCollection"
   | BadPrimitiveArgument _ -> "BadPrimitiveArgument"
+  | NoEntryPoint -> "NoEntryPoint"
```
Roman-Manevich, line 221: This is missing from the reference. Currently, the type system does not check for the existence of a `main` and the semantics just assumes a `main` is there.

hrutvik: I've now added this to the reference - please could you re-review?

asllib/doc/ErrorCodes.tex
```diff
@@ -464,4 +465,10 @@ \section{Dynamic Errors\label{sec:DynamicErrors}}
   \textit{Negative array length.}
   The expression used to determine the length of the array evaluates to a negative integer value
   (see \SemanticsRuleRef{EArray}).
+
+\hypertarget{def-noentrypoint}{}
+\item[$\NoEntryPoint$]
+  \textit{No entry point.}
+  The entry point \texttt{main} could not be found
```
Roman-Manevich: You can use `\vmain` for `\texttt{main}`

asllib/Typing.mli
```diff
@@ -50,6 +50,8 @@ module type S = sig
 
   val type_check_ast_in_env :
     StaticEnv.global -> AST.t -> AST.t * StaticEnv.global
+
+  val find_main : StaticEnv.global -> AST.identifier
```
Roman-Manevich, line 54: Returns the identifier of the function with the signature `fun main() => integer`, if one exists, and otherwise raises an exception.

asllib/aslref.ml
```diff
@@ -383,26 +383,24 @@ let () =
...
 
-      let control_flow_analysis = args.control_flow_analysis
-    end in
-    let module T = Annotate (C) in
-    or_exit @@ fun () -> T.type_check_ast ast
-  in
+    let control_flow_analysis = args.control_flow_analysis
+  end in
+  let module T = Annotate (C) in
+  let typed_ast, static_env = or_exit @@ fun () -> T.type_check_ast ast in
```
HadrienRenaud, line 403: I'm not against this refacto, maybe just curious why this was needed?

asllib/Typing.ml
```diff
@@ -149,6 +149,7 @@ end
 module type S = sig
   val type_check_ast : AST.t -> AST.t * global
   val type_check_ast_in_env : global -> AST.t -> AST.t * global
+  val find_main : global -> identifier
```
HadrienRenaud, line 152: Can we document here and in `Typing.ml` that this function raises a dynamic error if the correct `main` is not found? I find it counterintuitive to have a function in `Typing` raising a dynamic error.

asllib/aslref.ml
hrutvik: Ah this was so I could reference `T.find_main` below in the interpretation section. As for why `find_main` is in `Typing.ml` - it depends on `FunctionRenaming.subprogram_for_signature`, and I was reluctant to add a dependency on `Typing.ml` to `Interpret.ml` or similar

PR #1432, opened 2025-09-11
[tests] Two enhancements for catalogue regression test

maranget:
1. Specify tests with the new option -index-path. Using this option overrides the default behaviour of reading test names from the shelf file.
  2. Better formatted error message when some errors occurred while running herd7.


maranget:

PR #1434, opened 2025-09-12
[herd] New effect selection in diagrams

maranget:
The new selector `nobranches` selects all effecsts except branch effects.

I can be useful to disaplay diagrams in AArch64+ASL mode, as those include many such effects.


maranget:

PR #1437, opened 2025-09-12
[lib] Add optional data to symbolic values

HadrienRenaud:
This PR adds an extra field to the representation of symbolic variables, i.e.
variables which are yet to be determined by the solver.

This helps with ASL interpretation as this can make some operation possible on
those variables even if they are yet un-determined.

For example, if we know that S1 is a page table descriptor, then reading some bits of
this descriptor has some bits fixed to zero or one.
The implementation of this example will be done for now in the #916 PR.


maranget:
Hi @HadrienRenaud. This PR abstracts value over "symbolic data" by adding an extra argument to the `SymbValue.Make` unctor. However the argument to functor application is always the same : `SymData.No` . I assume there must be some additional code somewhere.... Could we at least have another module of the same signature that would be specific to the intended usage case ?

herd/AArch64ASLSem.ml
```diff
@@ -816,11 +816,11 @@ module Make (TopConf : AArch64Sig.Config) (V : Value.AArch64ASL) :
       | V.Val cst -> (tr_cst (ASLScalar.convert_to_bv sz) cst, [])
 
     let aarch64_to_asl = function
-      | V.Var v -> ASLS.A.V.Var v
+      | V.Var (v, _) -> ASLS.A.V.Var (v, ASLS.A.V.SData.default)
```
maranget, line 1048: We may rather aim at keeping the symbolic data identical in both worlds:
```suggestion
      | V.Var (v, s) -> ASLS.A.V.Var (v, d)
```
This would mean exposing some type equality for symbolic data in both "V" modules.

Commit 94a517e07815ea17117bac7128504b26565d0746 may help to understand the suggestion.

HadrienRenaud: Hi @maranget, I struggle to see why we would do this. For the moment, this is only used during interpretation of ASL, and not in the solver. What would be improved by propagating the ASL symbolic annotation to the AArch64 world?

PR #1438, opened 2025-09-12
[asl][refefrence] preparation for new spec DSL and fixes

Roman-Manevich:
Various fixes and changes meant to prepare the reference to the upcoming LaTeX generation of types, relations, and inference rules from a new DSL.
* Fixed types of literals defined in the the AST definition of `literal`.
* Changed literals-related macros to resemble existing naming convention, e.g., `LInt` instead of `lint`.
* Use \overtext' consistently in defining AST (rather than `\overname`).
* Defined `bit` type.
* Standardized relation macros - they all take 0 arguments now, e.g., `\annotateexpr[0]`
* Removed `check_teletype_in_rule_math_mode` from `doclint.py`, since in the future the generated LaTeX will contain teletype font in math mode.
* Fixed some incorrect example references.
* Introduced new macor for some expression types.


asllib/doc/RelationsOnTypes.tex
```diff
@@ -2315,7 +2314,7 @@ \section{Relations Over Types\label{sec:RelationsOnTypes}}
   \overname{\intconstraint}{\vc}) \aslto
   \overname{\langle\intconstraint\rangle}{\vcopt}
 \]
-returns $\None$ if $\vc$ is an exact constraint for a binary expression for dividing two integer literals
+returns $\None$ if $\vc$ is an exact constraint for a b\binopexpressionterm{} for dividing two integer literals
```
hrutvik:
```suggestion
returns $\None$ if $\vc$ is an exact constraint for a \binopexpressionterm{} for dividing two integer literals
```

hrutvik (approved):

PR #1442, opened 2025-09-15
[catalogue] Fix types for some MTE tests

murzinv:
Some tests pass a tag as if it were a plain value. Since a tag travels with the address, these variables are effectively pointers. Two patterns show up:

- `y = x:green` where `y` is (implicitly) an integer but `x:green` is a tagged pointer;
- (an attempted workaround?) `int64_t y = z` to make the storage large enough for a pointer.

Switch the left-hand side to a pointer type, which handles both cases:

- `int *y = x:green;`
- `int *y = z;`

Update affected MTE tests accordingly.

### DIFF
--- a/catalogue/aarch64-MTE/tests/asorted/LB+dmb.sypp+datapt-lrstt.litmus
+++ b/catalogue/aarch64-MTE/tests/asorted/LB+dmb.sypp+datapt-lrstt.litmus
@@ -1,12 +1,13 @@
 AArch64 LB+dmb.sypp+datapt-lrstt
 Variant=memtag,sync
 {
-    y=x:green;
+    int *y=x:green;
     0:X1=x:green; 0:X2=x:red; 0:X3=y;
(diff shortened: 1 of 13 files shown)


catalogue/aarch64-MTE/tests/asorted/LB+dmb.sypp+datapt-lrstt.litmus
```diff
@@ -1,7 +1,7 @@
 AArch64 LB+dmb.sypp+datapt-lrstt
 Variant=memtag,sync
 {
-    y=x:green;
+    int *y=x:green;
```
relokin, line 4: since you're editing this, can you please also add in P1 and before line 11
MOV W0,#1

murzinv: Good catch!

PR #1443, opened 2025-09-16
[herd] Fixing bug with -through all option in herd web interface

hazel394:
This PR fixes a bug with the `-through all` option in the herd web interface. When the option is added, it runs as expected. However, when it is subsequently removed, the output behaves as though it is still selected.

- The -through option is reset to its default value (none) every time the simulation runs.
- Other options with default values are grouped together with -through in a record.
- A record instance with the defaults is created, and a function sets these options back to defaults whenever a simulation is run.

### DIFF
--- a/herd-www/jerd.ml
+++ b/herd-www/jerd.ml
@@ -39,6 +39,30 @@ let load_config s =
   LexConf_herd.lex found
 
 
+
+type webOpts = {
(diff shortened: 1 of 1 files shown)


maranget:
Merged, thanks @hazel394 and @diaolo01.

PR #1445, opened 2025-09-18
Align the tranliteration of aarch64.cat with the Arm ARM issue L.b

relokin:
* Fixes the way miaou7 handles indices and whenever possible reuses the same indices as per the behaviour prior to PR #1312.
* Minor fix in the definition of va-loc to improve its transliteration in English.
* Rename the Explicitly-Observed-by and Explicitly-Hazard-ordered-before to Explicit-Observed-by and Explicit-hazard-ordered-before

### DIFF
--- a/herd/libdir/catdefinitions.tex
+++ b/herd/libdir/catdefinitions.tex
@@ -199,7 +199,10 @@
 
 \newcommand{\sameoa}[2]{#1 and #2 have the Same Output Address}
 \newcommand{\sameloworderbits}[2]{#1 and #2 have the Same Low Order Bits}
+
+\newcommand{\valocemph}[2]{#1 and #2 are to the \emph{Same Virtual Address}}
 \newcommand{\valoc}[2]{#1 and #2 are to the Same Virtual Address}
+\newcommand{\loc}[2]{#1 and #2 are to the same Physical Address}
 \newcommand{\lrsname}{a Local read successor of}
 \newcommand{\lrs}[2]{#2 is a Local read successor of #1}
 \newcommand{\lwfsname}{a Local write or MMU Fault successor of}
--- a/herd/libdir/catdefinitions.tex
+++ b/herd/libdir/catdefinitions.tex
@@ -284,11 +284,11 @@
 \newcommand{\localhwreqsemph}[2]{#1 is \emph{Locally-hardware-required-ordered-before} #2}
 
 \newcommand{\hazobname}{Hazard-ordered-before}
-\newcommand{\hazob}[2]{#1 is Hazard-ordered-before #2}
-\newcommand{\hazobemph}[2]{#1 is \emph{Hazard-ordered-before} #2}
-\newcommand{\Exphazobname}{Explicitly-hazard-ordered-before}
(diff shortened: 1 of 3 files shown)


relokin:
Thanks for the review Luc!

> I just noticed that my example above is the same as one of the examples discussed in PR #1312. Hence, you may already have decided that re-using `E3` is acceptable. However, the question is worth another look, IMHO.

I've missed this case. I've tried `let erel = (brel & (crel ; drel)) ; rel` which was in PR #1312 and I think it works as expected.

Tbh, the way miaou7 handles indices right now is a problem for us, so we need a fix. I can try and fix the case you raised as well. But we need this or a better fix.

herd/libdir/catdefinitions.tex
```diff
@@ -286,9 +289,9 @@
 \newcommand{\hazobname}{Hazard-ordered-before}
 \newcommand{\hazob}[2]{#1 is Hazard-ordered-before #2}
 \newcommand{\hazobemph}[2]{#1 is \emph{Hazard-ordered-before} #2}
-\newcommand{\Exphazobname}{Explicitly-hazard-ordered-before}
-\newcommand{\Exphazob}[2]{#1 is Explicitly-hazard-ordered-before #2}
-\newcommand{\Exphazobemph}[2]{#1 is \emph{Explicitly-hazard-ordered-before} #2}
+\newcommand{\Exphazobname}{Explicit-hazard-ordered-before}
+\newcommand{\Exphazob}[2]{#1 is Explicit-hazard-ordered-before #2}
+\newcommand{\Exphazobemph}[2]{#1 is \emph{Explicit-hazard-ordered-before} #2}
```
maranget: Nitpicking: it could be more rational to write:
```suggestion
\newcommand{\Exphazob}[2]{#1 is \hazobname{} #2}
\newcommand{\Exphazobemph}[2]{#1 is \emph{\hazobname} #2}
```

herd/libdir/catdefinitions.tex
```diff
@@ -307,9 +310,9 @@
 \newcommand{\obsname}{Observed-by}
 \newcommand{\obs}[2]{#1 is Observed-by #2}
 \newcommand{\obsemph}[2]{#1 is \emph{Observed-by} #2}
-\newcommand{\Expobsname}{Explicitly-Observed-by}
-\newcommand{\Expobs}[2]{#1 is Explicitly-Observed-by #2}
-\newcommand{\Expobsemph}[2]{#1 is \emph{Explicitly-Observed-by} #2}
+\newcommand{\Expobsname}{Explicit-Observed-by}
+\newcommand{\Expobs}[2]{#1 is Explicit-Observed-by #2}
+\newcommand{\Expobsemph}[2]{#1 is \emph{Explicit-Observed-by} #2}
```
maranget: As above:
```suggestion
\newcommand{\Expobs}[2]{#1 is \Expobsname{} #2}
\newcommand{\Expobsemph}[2]{#1 is \emph{Expobsname} #2}
```

maranget (approved):
With two minor suggestions.

relokin:
I'm planning to merge this by the end of today @artkhyzha let me know if you need more time to review.

PR #1446, opened 2025-09-18
[litmus] Multi-level page tables, contiguous bit and TLB conflict aborts

diaolo01:
This PR introduces initial support for three features:
1. Contiguous bit handling in page tables
2. TLB conflict aborts
3. Level 2 descriptors (PMD) which contain either a table or a block (pmdval)

Syntax changes:
- Contiguous bit in pteval: `[PTE(x)]=(contig:1)`.
- The syntax for multiple levels of translation is:
  - TTD(x,S1,LV3) - equivalent to PTE(x).
  - TTD(x,S1,LV2) - for level 2 (PMD) descriptors.

Notes on PMDs:
- The kvm-unit-tests scaffolding initialises PMD values as tables.
- Encoding a pmdval as a block is required to enable descriptor transformation (table → block). For this purpose, a new field dt has been introduced, used as follows: `TTD(x,S1,LV2)=(oa:PA(x), dt:Block)`.

Limitations:
- No mechanism to translate block->table. Based on the agreed test scenarios, this was not required.


relokin:
@murzinv can you please help with reviewing this?

litmus/libdir/_aarch64/kvm_fault_type.c
```diff
@@ -11,6 +11,7 @@ enum fault_type_t {
   FaultMMUTranslation,
   FaultMMUAccessFlag,
   FaultMMUPermission,
+  FaultMMUTLBConflictAbort,
```
murzinv: TLB conflict abort encoded in `DFSR` as `0b110000`. Since it is data abort we would fall under `else` branch on L50. There we apply trick to calculate offset into this enum: `FaultMMUAddressSize` is used as base and other fault types are addition to that, if we apply this logic we would get offset `12 + FaultMMUAddressSize` which is out of bound

litmus/libdir/_aarch64/kvm-headers.h
```diff
@@ -24,10 +24,19 @@
...
+#define BLOCK_SHIFT      21u
+#define BLOCK_SIZE       (1ULL << BLOCK_SHIFT)     // 0x0020_0000
+#define BLOCK_MASK       (BLOCK_SIZE - 1)          // 0x001F_FFFF
+
 static inline pteval_t *litmus_tr_pte(void *p) {
   return follow_pte(mmu_idmap, (uintptr_t)p);
 }
 
+static inline pmdval_t *litmus_tr_pmd(void *p) {
+  return follow_pmd(mmu_idmap, (uintptr_t)p);
```
murzinv: I assume implementation of `follow_pmd` left on `kvm-unit-test` side?

herd/tests/instructions/AArch64.kvm/TABT003.litmus.expected-failure
```diff
@@ -0,0 +1 @@
+Warning: File "./herd/tests/instructions/AArch64.kvm/TABT003.litmus": access_of_location_std on non-standard symbol 'TTD(s1, lv2, y)'
```
murzinv: I assume we do not expect such warning, right?

lib/AArch64Op.ml
```diff
@@ -22,6 +22,8 @@ type 'op1 unop =
   | DBM (* get DBM from PTE entry *)
   | Valid (* get Valid bit from PTE entry *)
   | EL0 (* get EL0 bit from PTE entry *)
+  | Contig (* get contiguous bit from PTE *)
```
murzinv: Would it make sense to postpone addition of contiguous bits until we get use-case for it?

litmus/libdir/_aarch64/kvm-headers.h
```diff
@@ -24,10 +24,19 @@
 
 #define LITMUS_PAGE_SIZE PAGE_SIZE
 
+// Assumes blocks are at level 2
```
murzinv: `kvm-unit-test` can be configured with different page size 
```
    --page-size=PAGE_SIZE
                           Specify the page size (translation granule). PAGE_SIZE can be
                           4k [default], 16k, 64k for arm64.
...
```
So we need a way to check that our assumption hold. Alternatively, `kvm_unit_test` offer handy macros `PMD_{SHIFT,SIZE,MASK}` we can use which automatically set to match `PAGE_SIZE`

litmus/libdir/_aarch64/kvm-headers.h
```diff
@@ -99,27 +108,92 @@ static inline int litmus_same_oa(pteval_t p,pteval_t q) {
...
+  pteval_t old = *p;
+  // Extract OA from pte: bits [47:12]
+  uint64_t oa_l3 = old & 0x0000FFFFFFFFF000ULL;
+  // Align down to 2 MiB (L2 block): clear bits [20:0]
+  uint64_t oa_l2 = oa_l3 & ~((1ULL << 21) - 1);
+  desc |= oa_l2;
+
+  // Extract attributes from PTE: everything except OA and type
+  uint64_t attrs = old & ~(0x0000FFFFFFFFF000ULL | 0x3ULL);
+  desc |= attrs;
```
murzinv: Maybe `GENMASK_ULL`  (provided by `bitops.h`) can make it more readable? Something like
```
        pteval_t old = *p
        pmdval_t desc = 0;

        // Block
        desc |= 0b01;

        // Lower attributes
        desc |= old & GENMASK_ULL(11,2);

        // Output address
        desc |= old & GENMASK_ULL(47,PMD_SHIFT);

        // Upper attributes
        desc |= old & GENMASK_ULL(63,50);
```

litmus/preSi.ml
```diff
@@ -564,6 +564,7 @@ module Make
         | Alloc.Static|Alloc.Before -> not Cfg.is_kvm
 
       let dbg = false
+      let is_mlpt = true
```
murzinv: I assume it would be turned into some conditional logic, so we won't collect pmds for all locations and won't allocate them within the same block...

litmus/preSi.ml
```diff
@@ -1379,19 +1404,32 @@ module Make
           if Misc.consp test.T.globals then begin
               O.oi "const size_t _sz = LINE/sizeof(intmax_t);";
               O.oi "pteval_t *_p;" ;
+              if is_mlpt then begin
+                O.oi "pmdval_t *_q;";
+                O.o "" ;
+                O.oi "_mem = block_align_up(_mem);";
```
murzinv: Can we please check that we have enough memory to align?
IIUC we get memory from
```
RUN
-> init_global
     -> instance_init
          -> vars_init
```
In `RUN` we allocate memory
```
glo_ptr->mem = malloc_check(MEMSZ*sizeof(*glo_ptr->mem));
```
where
```
sizeof(*glo_ptr->mem)
```
is `sizeof(intmax_t)` and `MEMSZ` is
```
MEMSZ ((NVARS*NEXE+1)*LINESZ)
```
here `LINESZ` is 
```
#define LINE LITMUS_PAGE_SIZE
LINESZ (LINE/sizeof(intmax_t))
```
and `NVARS` and `NEXE` are test dependent.

So assume `NVARS` and `NEXE` are both 1. `LITMUS_PAGE_SIZE` is 4096 then
- `LINESZ` is 512
- `MEMSZ` is 1024
- `glo_ptr->mem` is 8192 which is 8Kb

Have I missed something?

litmus/libdir/_aarch64/kvm-headers.h
diaolo01: Yes, that is right.

litmus/libdir/_aarch64/kvm_fault_type.c
diaolo01: I did not pay attention to the trick used to offset into the fault type enum. Having looked into it, does it not categorise all Synchronous External abort as Tag Check Faults?

murzinv: Right, it won't be out of bound access due to guard check
```
    if (fault_type > FaultTagCheck)
      return FaultUnsupported;
    else
      return fault_type;
```
and TLB Conflict abort would be  reported as `Unsupported`, which is probably not what we are after...

litmus/libdir/_aarch64/kvm-headers.h
relokin: Can we implement follow_pmd here instead? All we need is to be able to:
* Tell whether a TTD at any level is valid.
* Extract the OA from the current TTD.
* To extract the next VA bits and add them to the OA. 
I suspect that all of these are functions we need in litmus7 regardless.

herd/access.ml
```diff
@@ -14,7 +14,7 @@
 (* "http://www.cecill.info". We also give a copy in LICENSE.txt.            *)
 (****************************************************************************)
 
-type t = REG | VIR | PHY | PTE | TLB | TAG | PHY_PTE
+type t = REG | VIR | PHY | PTE | TLB | TAG | PHY_PTE | TTD
```
relokin: Why do we need a new access type?

diaolo01: We don't need this, I have removed it.

litmus/preSi.ml
diaolo01: We agreed that for now it can be always true and can be extracted as litmus variant later on once we have the desired functionality in place.

herd/tests/instructions/AArch64.kvm/TABT003.litmus.expected-failure
diaolo01: Agreed. It didn't affect the litmus side so I did not investigate. It's tied to a comment below about new access types. By fixing that this also goes away.

litmus/preSi.ml
diaolo01: You are right. I did not get a segmentation fault or anything to indicate that memory was being used out of bounds of what was allocated. It is unclear to me what is the purpose of the extra cache line, do you know?

The simplest way would be to change the MEMSZ definition to be:
`PMD_SIZE/sizeof(intmax_t)*NEXE+LINESZ`
This would fail for more than 512 variables being used in one test. Do we cover tests with more than 512 variables, or do we plan to in the future?
Also, this will allocate a large amount of memory for tests with a high NEXE. Is it reasonable to implement MEMSZ always to be as suggested above or do we want to limit this definition to multi-level page table tests only?

murzinv: > You are right. I did not get a segmentation fault or anything to indicate that memory was being used out of bounds of what was allocated.

I suspect the reasons you do not see side-effects are:
- `kvm-unit-test` setup 1:1 mappings upfront, IIUC
-  memory was only read or writes to the memory did not affect viable data 

> It is unclear to me what is the purpose of the extra cache line, do you know?

I have no idea :disappointed: 

> The simplest way would be to change the MEMSZ definition to be:
PMD_SIZE/sizeof(intmax_t)*NEXE+LINESZ
This would fail for more than 512 variables being used in one test. Do we cover tests with more than 512 variables, or do we plan to in the future?
Also, this will allocate a large amount of memory for tests with a high NEXE. Is it reasonable to implement MEMSZ always to be as suggested above or do we want to limit this definition to multi-level page table tests only?

I have no strong opinion on that, tbh. Maybe we should try to get requirements to the memory and see what would be the best and cleaner way to satisfy them:
- each `VAR` need to be `PAGE_SIZE`
- we assume that `VAR` never exceed `PAGE_SIZE`
- we need `NVARS`
- we need `VAR`s to be aligned to `PAGE_SIZE` or `PMD_SIZE`
- all allocated memory need to be multiplied by `NEXE`

so we can try employ kvm-unit-test's `memalign_pages` and formula could look like
```
mem = memalign_pages(PAGE_SIZE, NVARS*PAGE_SIZE*NEXE)
// or for PMD case
mem = memalign_pages(PMD_SIZE, ALIGN(NVARS*PAGE_SIZE, PMD_SIZE)*NEXE)
```
I do not insist on that since reworking memory allocation logic might be slightly out of scope (and deserves it's own patch), it is up to you which approach to take given how much bandwidth you have.

lib/AArch64Op.ml
diaolo01: Yes. I am going to make a separate PR for the contiguous bit to allow progress of this.

litmus/libdir/_aarch64/kvm-headers.h
diaolo01: The functionalities you mention, already exist in kvm-unit-tests. I did an import of the relevant file and transitioned the logic as suggested.

litmus/libdir/_aarch64/kvm-headers.h
diaolo01: Thank you - it's a lot cleaner.

lib/descriptorType.ml
ShaleXIONG, line 1: Question of coding style: since the `lib/descriptorType.mli` exposes everything in `lib/descriptorType.ml`. I do not think you need the `*.mli` file?

PR #1447, opened 2025-09-18
[ASL][ACL2] Various updates:

solswords:
- Do stdlib proofs on versions both with and without primitives
 - Switch record/storage structures to canonically ordered maps
 - Add a tracing version of the ACL2 ASL interpreter
 - Add file to build documentation for the interpreter
 - A few proofs about the interpreter: independence of static env, readonly sufficient condition, correctness of tracing interpreter


solswords:
Need to revise this; will try again.

PR #1448, opened 2025-09-19
Fix register identifier which causes warning on litmus7

diaolo01:
Before the change the following warning was reported: "Register x2 has different types: <int> and <int64_t>". We agreed such warnings must be fixed.


relokin:
When we modify a litmus test then their hash changes as well. So in most cases this is undesirable as it has the potential to create problems for hardware logs. In this case, this is a new test and I doubt there are any logs with it (@maranget can you confirm?). So we can make this change.

But before merging you will have to adjust the hash signature in the .expected file as well.

PR #1449, opened 2025-09-19
[ci] Improve make test

HadrienRenaud:
This PR moves the formatting check from the job "Make test latest lts" to its own job. The rationale is that running opam install is poorly cached by setup-ocaml, so what I've done is that I've put it in its own job.

Now, with caches warmed, the new job takes less than a minute (more 40s), and make-test is faster by around a minute.

### DIFF
--- a/.github/workflows/make-test.yml
+++ b/.github/workflows/make-test.yml
@@ -72,7 +72,7 @@ jobs:
       - name: Set-up OCaml
         uses: ocaml/setup-ocaml@v3
         with:
-          ocaml-compiler: ocamlformat.0.26.2
+          ocaml-compiler: ocamlformat.0.27.0
           dune-cache: true
 
       - run: opam exec -- dune fmt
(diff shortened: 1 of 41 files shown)


.github/workflows/make-test.yml
```diff
@@ -57,8 +57,23 @@ jobs:
 
       - run: opam exec -- make install
 
-      - run: opam install ocamlformat.0.26.2
-        if: ${{ matrix.cfg.with-fmt }}
+  check-fmt:
```
hrutvik, line 60: Just to check - why do we need an extra "step" here?

HadrienRenaud: So here what I'm doing is I'm removing the formatting check from the job "Make test latest lts" to its own job. The rationale is that running `opam install` is poorly cached by `setup-ocaml`, so what I've done is that I've put it in its own job.

Now, with caches warmed, the new job takes less than a minute (more 40s), and `make-test` is faster by around a minute.

PR #1451, opened 2025-09-20
[asl] A new DSL for specifying semantics fo ASL itself

Roman-Manevich:
This PR introduces a new domain-specific language called `aslspec` and a new tool of the same name.
- The tools is implemented in the sub-directory `asllib/aslspec`
- The file `asllib/doc/asl.spec` contains the definitions for ASL from which LaTeX content is generated in order to be included in the ASL Reference
- The tool generates two files:  `generated_macros`, which adds macros in addition to `ASLmacros.tex`, and `generated_elements.tex`, which is a temporary file containing the result of rendering the elements defined in `asllib/doc/asl.spec`
- The python scripts `doclint.py` and `extended_macros.py` help extract LaTeX blocks from `generated_elements.tex` into the LaTeX files in `asllib/doc`.
- The `Makefile` is also updated to automatically run `aslspec` and copy the generated LaTeX blocks into the relevant files in the ASL Reference.

In order to include the rendered LaTeX for an element (type, constant, relation) in a file, add `% BEGIN_GENERATED_ELEMENT(name) % END_GENERATED_ELEMENT` where `name` specifies the name of the element as it appears in `asl.spec`.


asllib/aslspec/AST.ml
```diff
@@ -0,0 +1,304 @@
...
+      (** No specific layout, appropriate for atomic terms and terms with singleton lists. *)
+  | Horizontal of layout list  (** Layout terms horizontally. *)
+  | Vertical of layout list  (** Layout terms vertically. *)
+
+(** A module for key-value attributes. *)
+module Attributes = struct
+  let prose_description = "prose_description"
+  let prose_application = "prose_application"
+  let math_macro = "math_macro"
+  let math_layout = "math_layout"
```
HadrienRenaud: Are those the 4 possible attribute keys? In which case, could you not have it as variants of a sum type?

asllib/doc/Bitfields.tex
```diff
@@ -108,13 +108,13 @@ \section{Nested Bitfields\label{sec:Nested Bitfields}}
...
-\bitfield \derives\ & \BitFieldSimple(\identifier, \slice^{*})
-  & \\
-  |\ & \BitFieldNested(\identifier, \slice^{*}, \bitfield^{*})
-  & \\
-  |\ & \BitFieldType(\identifier, \slice^{*}, \ty) &
+\bitfield \derives\ & \BitFieldSimple(\overtext{\Identifier}{name}, \overtext{\KleeneStar{\slice}}{slices})\\
+|\ & \BitFieldNested(\overtext{\Identifier}{name}, \overtext{\KleeneStar{\slice}}{slices}, \overtext{\KleeneStar{\bitfield}}{nested bitfields})\\
+|\ & \BitFieldType(\overtext{\Identifier}{name}, \overtext{\KleeneStar{\slice}}{slices}, \overtext{\ty}{type})
 \end{flalign*}
+% END_GENERATED_ELEMENT
```
HadrienRenaud: I'm wondering why you're committing the generated elements? Why not leave a placeholder and replace it?

asllib/doc/Expressions.tex
```diff
@@ -53,15 +53,30 @@ \chapter{Expressions\label{chap:Expressions}}
...
+ resulting in the following:
+ $\texttt{t}$ is the type inferred for $\texttt{e}$;
+ $\texttt{new\_e}$ is the \typedast{} for $\texttt{e}$, also known as the
+\emph{annotated expression};
+ and $\texttt{ses}$ is the \sideeffectsetterm{} inferred for $\texttt{e}$.
+\ProseOtherwiseTypeError
+% END_GENERATED_ELEMENT
 
-\hypertarget{def-annotateexpr}{}
+\BackupOriginalRelation{
```
HadrienRenaud, line 58: Why keeping the backup? You can probably get this with a diff on master right?

asllib/doc/Expressions.tex
```diff
@@ -1006,19 +1022,16 @@ \subsection{Abstract Syntax}
...
 
+% BEGIN_GENERATED_ELEMENT(call)[remove_hypertargets]
 \begin{flalign*}
-\call \derives\ &
-{
-\left\{
-  \begin{array}{rcl}
- \callname &:& \Strings, \\
- \callparams &:& \expr, \\
- \callargs &:& \expr, \\
```
HadrienRenaud, line 1016: Strange that there was not a star already?

asllib/doc/Expressions.tex
Roman-Manevich: For two reasons:
1. I'm going to be automating the translation from LaTeX to the new DSL. It's important to be able to spot mistakes, improve the translation and re-run it.
2. It's useful as examples for training an AI.

asllib/doc/Expressions.tex
Roman-Manevich: Apparently no. Strange indeed.

asllib/doc/Expressions.tex
HadrienRenaud: Ok thanks

asllib/aslspec/AST.ml
```diff
@@ -0,0 +1,304 @@
...
+      (** A set containing all subsets of the given type. *)
+  | Option of type_term
+      (** Either the empty set of a set containing a single value of the given type. *)
+  | Tuple of opt_named_type_term list
+      (** A set containing all tuples formed by the given components.
+        A tuple containing a single term is a special case - its domain is the domain
+        of that term.
+    *)
+  | LabelledTuple of { label : string; components : opt_named_type_term list }
+      (** A set containing all labelled tuples formed by the given components. *)
```
hrutvik: nit: I wonder if it's worth condensing these into a single variant, just to streamline the AST? Similarly for `Record` and `LabelledRecord` below

asllib/doc/asl.spec
```diff
@@ -0,0 +1,954 @@
...
+// attach information needed to emit LaTeX mathematical descriptions and
+// LaTeX prose for the types and rules.
+// The document `../aslspec/aslspec.md` is a tutorial for aslspec.
+
+////////////////////////////////////////////////////////////////////////////////
+// Generic Types
+////////////////////////////////////////////////////////////////////////////////
+
+typedef Bool
+{  prose_description = "Boolean",
```
hrutvik: Suggestion for readability: could we shorten this to e.g. "description" instead of "prose_description" everywhere, just to streamline things a bit?

asllib/aslspec/tests.t/run.t
```diff
@@ -0,0 +1,17 @@
+# Basic tests
+  $ aslspec hello.spec
+  $ aslspec typedefs.spec
+  $ aslspec relations.spec
+
+  $ aslspec type_name.bad
+  Fatal error: exception Dune__exe__AST.SpecError("element-defining identifiers must not contain digits: t2")
+  [2]
+  $ aslspec type_instance.bad
```
hrutvik, line 9: Just to check - this is named `bad`, but does not generate an error - is this intended?

asllib/doc/PatternMatching.tex
```diff
@@ -599,7 +599,7 @@ \subsection{Syntax}
 
 \subsection{Abstract Syntax}
 \begin{flalign*}
-\pattern \derives\ & \PatternMask(\overtext{\{0,1,x\}^*}{mask constant}) &
+\pattern \derives\ & \PatternMask(\overtext{\{\texttt{0},\texttt{1},\texttt{x}\}^*}{mask constant}) &
```
hrutvik: Why not use `\zerobit` and friends here?

asllib/doc/Semantics.tex
```diff
@@ -345,87 +410,105 @@ \subsection{Kinds of Semantic Configurations\label{sec:KindsOfSemanticConfigurat
...
+\cup\ & \ResultExprSEF(\overtext{\nativevalue}{value}, \overtext{\XGraphs}{graph})\hypertarget{type-ResultLexpr}{}\\
+\cup\ & \ResultLexpr(\overtext{\XGraphs}{graph}, \overtext{\envs}{environment})\hypertarget{type-ResultLDI}{}\\
+\cup\ & \ResultLDI(\overtext{\XGraphs}{graph}, \overtext{\envs}{environment})\hypertarget{type-ResultSlices}{}\\
+\cup\ & \ResultSlices(\overtext{(\KleeneStar{(\nativevalue, \nativevalue)}, \XGraphs)}{slices and graph}, \overtext{\envs}{environment})\hypertarget{type-ResultExprList}{}\\
+\cup\ & \ResultExprList(\overtext{(\KleeneStar{\nativevalue}, \XGraphs)}{values and graph}, \overtext{\envs}{environment})\hypertarget{type-ResultExprListM}{}\\
+\cup\ & \ResultExprListM(\overtext{\KleeneStar{(\nativevalue, \XGraphs)}}{value graph pairs}, \overtext{\envs}{environment})\hypertarget{type-ResultPattern}{}\\
+\cup\ & \ResultPattern(\overtext{\tbool}{boolean value}, \overtext{\XGraphs}{graph})\hypertarget{type-ResultCall}{}\\
+\cup\ & \ResultCall(\overtext{\KleeneStar{(\nativevalue, \XGraphs)}}{value graph pairs}, \overtext{\envs}{environment})
```
hrutvik: These are new right?

asllib/doc/TypeDeclarations.tex
```diff
@@ -571,10 +576,10 @@ \chapter{Type Declarations\label{chap:TypeDeclarations}}
 The function
 \[
 \declareenumlabels(\overname{\staticenvs}{\tenv} \aslsep
-  \overname{\identifier}{\name} \aslsep
-  \overname{\identifier^+}{\ids} \aslsep
+  \overname{\Identifier}{\name} \aslsep
+  \overname{\Identifier^+}{\ids} \aslsep
```
hrutvik: `\KleenePlus`?

asllib/doc/asl.spec
hrutvik, line 1: If I understand correctly, `aslspec` is already able to take multiple files - is it worth having an intentionally structured set of files from the very start, rather than this single one?

asllib/aslspec/SpecLexer.mll
```diff
@@ -0,0 +1,79 @@
...
+    | "//" [^'\n']*            { token lexbuf }
+    | "/*"                     { c_comments lexbuf }
+
+    (* Strings *)
+    | '"' ([^'"']* as s) '"'   {  (* Count the number of newline characters in the string.
+                                     This doesn't account for escape sequences (TODO). *)
+                                  let new_lines =  count_new_lines s in
+                                  for _i = 0 to (new_lines - 1) do
+                                    Lexing.new_line lexbuf
+                                  done;
```
hrutvik, line 29: I'm not quite sure I understand what the purpose of this is

asllib/aslspec/SpecLexer.mll
```diff
@@ -0,0 +1,79 @@
...
+    | identifier as lxm { IDENTIFIER(lxm) }
+    | latex_macro as lxm { LATEX_MACRO(lxm) }
+
+    | eof            { EOF }
+    | _ {
+      let lxm = Lexing.lexeme lexbuf in
+      let msg = Format.sprintf "unrecognized token %s" lxm in raise (Error msg) }
+
+(* C-style comments *)
+and c_comments = parse
```
hrutvik, line 76: We could stick to e.g. `//` comments only (or something else) if that was convenient

asllib/aslspec/SpecParser.mly
```diff
@@ -0,0 +1,186 @@
...
+let opt_named_type_term ==
+    | name=IDENTIFIER; COLON; ~=type_term; { (Some name, type_term) }
+    | ~=type_term; { (None, type_term) }
+
+let math_layout :=
+    | IDENTIFIER; { Unspecified }
+    | LPAR; inner=clist0(math_layout); RPAR; { Horizontal inner }
+    | IDENTIFIER; LPAR; inner=clist0(math_layout); RPAR; { Horizontal inner }
+    | LBRACKET; inner=clist0(math_layout); RBRACKET; { Vertical inner }
+    | IDENTIFIER; LBRACKET; inner=clist0(math_layout); RBRACKET; { Vertical inner }
```
hrutvik, line 215: This is a nice idea!

asllib/doc/Bitfields.tex
Roman-Manevich: Done. This is now just `\RenderType[remove_hypertargets]{bitfield}`.

asllib/aslspec/AST.ml
```diff
@@ -0,0 +1,331 @@
...
+  let attributes self = self.att
+end
+
+(** A datatype for a type definition. *)
+module Type : sig
+  include HasAttributes
+
+  val kind : t -> type_kind
+  val name : t -> string
+  val variants : t -> TypeVariant.t list
```
hrutvik: nit: stylistically I wonder if it's worth just exposing the type definition directly (similarly through the file)

asllib/aslspec/AST.ml
```diff
@@ -0,0 +1,331 @@
...
+    match find_opt AttributeKey.Math_Macro self.att with
+    | Some (StringAttribute s) -> Some s
+    | _ -> None
+
+  let attributes_to_list self = Attributes.bindings self.att
+  let attributes self = self.att
+end
+
+type type_subset_pointer = { type_name : string; variant_names : string list }
+type render = { render_name : string; pointers : type_subset_pointer list }
```
hrutvik: consistency nit: why does this not have its own module? (though as above, I would suggest making the other types exposed like this one)

asllib/aslspec/spec.ml
```diff
@@ -0,0 +1,389 @@
...
+
+let math_macro_opt_for_node = function
+  | Node_Type def -> Type.math_macro def
+  | Node_Relation def -> Relation.math_macro def
+  | Node_TypeVariant def -> TypeVariant.math_macro def
+  | Node_Constant def -> Constant.math_macro def
+
+(** Utility functions for handling layouts. *)
+module Layout = struct
+  (** For a relation like [ r(a,b,c) -> A|B|C ] generates a term
```
hrutvik, line 38: Just to check - not sure I've seen the `A|B|C` syntax just yet - have I missed an example somewhere?

asllib/aslspec/spec.ml
```diff
@@ -0,0 +1,389 @@
...
+            acc_nodes (Type.variants def)
+      | Elem_Constant def -> Node_Constant def :: acc_nodes
+      (* Although a render defines an identifier, it does not carry semantic
+         meaning, and cannot be referenced elsewhere. *)
+      | Elem_Render _ -> acc_nodes)
+    [] ast
+
+(** Generates a map from identifiers to the nodes where they are defined. If two
+    nodes with the same identifier exist, a [SpecError] is raised. *)
+let gen_id_to_definition_node definition_nodes =
```
hrutvik: It feels a bit like this could be done as part of `list_definition_nodes`?

asllib/aslspec/spec.ml
```diff
@@ -0,0 +1,389 @@
...
+        raise (SpecError msg)
+      else StringMap.add name node acc_map)
+    StringMap.empty definition_nodes
+
+type t = {
+  ast : AST.t;  (** The original AST as parsed. *)
+  id_to_defining_node : definition_node StringMap.t;
+      (** Associate identifiers with the AST nodes where they are defined. *)
+  defined_ids : string list;
+      (** The list of identifiers defined in the spec in the order they appear.
```
hrutvik: I might be missing something - but why is order relevant here?

asllib/aslspec/spec.ml
```diff
@@ -0,0 +1,389 @@
...
+      | [] -> None
+      | str :: tail ->
+          if StringSet.mem str set then Some str
+          else find_with_set (StringSet.add str set) tail
+    in
+    find_with_set StringSet.empty strs
+
+  let check_definition_name name =
+    let () = assert (String.length name > 0) in
+    if Utils.string_exists (fun c -> '0' <= c && c <= '9') name then
```
hrutvik: Could this be a regular expression for permitted names instead? Or handled in lexing/parsing?

asllib/aslspec/spec.ml
```diff
@@ -0,0 +1,389 @@
...
+      let msg =
+        Format.sprintf
+          "element-defining identifiers must not contain digits: %s" name
+      in
+      raise (SpecError msg)
+
+  let check_defining_names defined_ids =
+    List.iter check_definition_name defined_ids
```
hrutvik: nit: stylistic suggestion to make `check_definition_name` local to `check_defining_names` (and similarly for other things below)

Or perhaps it would be quite clean to have small submodules inside `Check`, so that the checks end up looking like `Check.SpecificCheck.run` or something

asllib/aslspec/spec.ml
```diff
@@ -0,0 +1,389 @@
...
+            List.iter
+              (fun (_, type_term) ->
+                check_only_top_level_types id_to_defining_node type_term)
+              (Relation.input rel)
+        | _ -> ())
+      elements
+
+  (** TODO:
+      - Check that instantiated type terms for labelled type terms and labelled
+        records match their definition (see [tests/type_instance.bad]). *)
```
hrutvik, line 360: Ah I see now that this test is expected not to produce an error just yet

asllib/aslspec/render.ml
```diff
@@ -0,0 +1,676 @@
...
+  match math_macro_opt with
+  | Some str -> str
+  | None -> Text.elem_name_to_math_macro id
+
+let pp_hypertarget fmt target_str =
+  fprintf fmt {|\hypertarget{%s}{}|} target_str
+
+(** [hyperlink_target_for_id spec id] returns a string [target] that can be used
+    for the LaTeX [\hypertarget{target}{}] for [id]. This is not needed when the
+    [id] is a spec element with a [math_macro] attribute. *)
```
hrutvik: Sorry I think I'm missing something - why is this?

asllib/aslspec/render.ml
```diff
@@ -0,0 +1,676 @@
...
+  if large then
+    match parenthesis with
+    | Parens -> fprintf fmt {|\left(%a\right)|} pp_elem elem
+    | Braces -> fprintf fmt {|\left\{%a\right\}|} pp_elem elem
+    | Brackets -> fprintf fmt {|\left[%a\right]|} pp_elem elem
+  else
+    match parenthesis with
+    | Parens -> fprintf fmt {|(%a)|} pp_elem elem
+    | Braces -> fprintf fmt {|\{%a\}|} pp_elem elem
+    | Brackets -> fprintf fmt {|[%a]|} pp_elem elem
```
hrutvik: nit: how about something like the following
```suggestion
let pp_parenthesized parenthesis large pp_elem fmt elem =
  let left = if large then {|\left|} else "" in
  let right = if large then {|\right|} else "" in
  match parenthesis with
  | Parens -> fprintf fmt {|%s(%a%s)|} left pp_elem elem right
  | Braces -> fprintf fmt {|%s\{%a%s\}|} left pp_elem elem right
  | Brackets -> fprintf fmt {|%s[%a%s]|} left pp_elem elem right
```

asllib/aslspec/render.ml
```diff
@@ -0,0 +1,676 @@
...
+        PP.pp_type_term type_term PP.pp_math_shape layout
+  in
+  let layout_contains_vertical = Layout.contains_vertical layout in
+  match type_term with
+  | Label name -> pp_print_string fmt (get_or_gen_math_macro spec name)
+  | Powerset { term = sub_term; finite } ->
+      let powerset_macro =
+        if finite then
+          if layout_contains_vertical then {|\Powfin|} else {|\powfin|}
+        else if layout_contains_vertical then {|\Pow|}
```
hrutvik: I wonder (similarly for other macros):
- would there be any harm in always using the `\Pow` form with `\left` and `\right`?
- could we streamline this by deferring more to LaTeX? I.e. have a macro with an optional "vertical" argument, and then just pass it appropriately?

asllib/aslspec/render.ml
```diff
@@ -0,0 +1,676 @@
...
+    | Node_Constant _ -> Text.TextSF
+  in
+  if Option.is_some (Spec.math_macro_opt_for_node node) then ()
+  else
+    let typeset_macro = Text.spec_var_to_latex_var ~font_type name in
+    fprintf fmt {|\newcommand%s[0]{ \hyperlink{%s}{%s} } %% Generated from %s|}
+      (get_or_gen_math_macro spec name)
+      hyperlink_target typeset_macro name
+
+let generate_latex_macros spec =
```
hrutvik: Most of the functions in this file are parameterised by `spec` - I wonder if it's worth turning this file into a functor, with the input module defining a `spec` value, to better reflect the "ambient" spec?

asllib/aslspec/tests.t/run.t
hrutvik: I see later that it is

asllib/doc/PatternMatching.tex
Roman-Manevich: This bit of latex is in backup now. The generated latex indeed uses the correct macros.

asllib/doc/Semantics.tex
Roman-Manevich: They should have been there. They were used, but I apparently forgot to list them here, and working on the new DSL made me notice.

asllib/doc/asl.spec
Roman-Manevich: How about having just a string `"..."` to implicitly stand for `prose_description = "..."`?

asllib/aslspec/SpecLexer.mll
Roman-Manevich: It's a hack to count the number of new line characters in a string. This is used to update the line numbering in strings for reporting lexical errors. It's hack because the correct way is to parse strings by taking into account escape sequences. I guess I will fix it when I add position information to AST and errors.

asllib/aslspec/AST.ml
Roman-Manevich: done

asllib/aslspec/AST.ml
Roman-Manevich: done

asllib/aslspec/AST.ml
Roman-Manevich: I've made the types exposed but kept the modules around name to avoid clashes between field names (like `name`). Also, wrapped `render` in its own module for consistency.

asllib/aslspec/spec.ml
Roman-Manevich: `relation annotate_expr(tenv: static_envs, e: expr) -> (t: ty, new_e: expr, ses: powerset(TSideEffect)) | type_error`

asllib/aslspec/spec.ml
Roman-Manevich: Switched to a regular expression and moved to the parser.

asllib/aslspec/render.ml
Roman-Manevich: done

asllib/aslspec/render.ml
Roman-Manevich: I don't remember. I don't think it's correct so I removed the second sentence.

asllib/aslspec/spec.ml
Roman-Manevich: I went with the first option.

asllib/doc/asl.spec
hrutvik: As discussed offline, sticking to a single file to simplify searching for now

asllib/aslspec/SpecLexer.mll
hrutvik: As discussed offline, no need

asllib/aslspec/spec.ml
hrutvik: As discussed offline, chosen to keep the functionality separate

asllib/aslspec/spec.ml
hrutvik: As discussed offline, simplifies debugging etc.

asllib/aslspec/render.ml
Roman-Manevich: Let's always use `\left` and `\right` and see how it goes.

asllib/aslspec/AST.ml
```diff
@@ -0,0 +1,362 @@
...
+
+(** A module for totally ordered attribute keys. *)
+module AttributeKey = struct
+  type t =
+    | Prose_Description  (** A description of the element in prose. *)
+    | Prose_Application
+        (** A description of the element in prose describing its application in
+            an inference rule premise. *)
+    | Math_Macro  (** A LaTeX macro for the element. *)
+    | Math_Layout  (** The layout of the element in a mathematical context. *)
```
HadrienRenaud, line 93: Neat

asllib/aslspec/AST.ml
```diff
@@ -0,0 +1,362 @@
...
+  let compare a b =
+    let key_to_int = function
+      | Prose_Description -> 0
+      | Prose_Application -> 1
+      | Math_Macro -> 2
+      | Math_Layout -> 3
+    in
+    let a_int = key_to_int a in
+    let b_int = key_to_int b in
+    Stdlib.compare a_int b_int
```
HadrienRenaud, line 105: What about `Int.compare`?


```suggestion
    Int.compare a_int b_int
```

asllib/aslspec/PP.ml
```diff
@@ -0,0 +1,132 @@
+(** A module for pretty-printing the AST. *)
+
+open Format
+open AST
+open SpecParser
+
+let tok_str = TokenStrings.string_of_token
+let pp_comma fmt () = fprintf fmt ",@ "
+let pp_double_newline fmt () = fprintf fmt "\n@ "
```
HadrienRenaud, line 9: Is this even used? And this looks quite ugly, but I guess it can just be removed

asllib/aslspec/PP.ml
```diff
@@ -0,0 +1,132 @@
...
+      fprintf fmt "%s %a -> %a" keyword pp_type_term from_type pp_type_term
+        to_type
+
+and pp_named_type_term fmt (name, term) =
+  fprintf fmt "%s: %a" name pp_type_term term
+
+and pp_opt_named_type_term fmt (name_opt, term) =
+  match name_opt with
+  | Some name -> pp_named_type_term fmt (name, term)
+  | None -> fprintf fmt "%a" pp_type_term term
```
HadrienRenaud, line 72:
```suggestion
  | None -> pp_type_term fmt term
```

asllib/aslspec/PP.ml
```diff
@@ -0,0 +1,132 @@
...
+
+and pp_named_type_terms fmt named_terms =
+  (pp_sep_list ~sep:", " pp_named_type_term) fmt named_terms
```
HadrienRenaud, line 75: I'm pretty sure the parentheses are not needed, any reason why they are here?

asllib/aslspec/PP.ml
```diff
@@ -0,0 +1,132 @@
...
+
+let pp_variants_with_attributes fmt variants =
+  (pp_sep_list ~sep:" @,| " pp_type_term_with_attributes) fmt variants
+
+let pp_variants fmt variants =
+  (pp_sep_list ~sep:" | " pp_type_term) fmt variants
+
+let pp_relation_definition fmt ({ Relation.name; input; output } as relation) =
+  let module Rel = Relation in
+  fprintf fmt "%s %s(%a) -> %a@.%a;" (tok_str RELATION) name
```
HadrienRenaud, line 92:
```suggestion
  fprintf fmt "@[<v>%s %s(%a) -> %a@,%a@];" (tok_str RELATION) name
```

asllib/aslspec/PP.ml
```diff
@@ -0,0 +1,132 @@
...
+    pointers
+
+let pp_elem fmt = function
+  | Elem_Type def -> pp_type_definition fmt def
+  | Elem_Relation def -> pp_relation_definition fmt def
+  | Elem_Constant def -> pp_constant_definition fmt def
+  | Elem_Render def -> pp_render_definition fmt def
+
+let pp_spec fmt spec =
+  fprintf fmt "@[<v>%a@]" (pp_sep_list ~sep:"\n\n" pp_elem) spec
```
HadrienRenaud, line 132:
```suggestion
  fprintf fmt "@[<v>%a@]" (pp_sep_list ~sep:"@,@," pp_elem) spec
```

Strange to use both `\n` and a vertical formatting box

asllib/aslspec/AST.ml
```diff
@@ -0,0 +1,362 @@
...
+type attribute = StringAttribute of string | MathLayoutAttribute of layout
+
+(** A module for associating attributes with attribute keys. *)
+module Attributes = struct
+  include Map.Make (AttributeKey)
+
+  type 'a map = 'a t
+  (** Shadows [Map.t] with a string-to-string map type. *)
+
+  type t = attribute map
```
HadrienRenaud, line 126: Fun fact, you can also do:


```suggestion
  type nonrec t = attribute t
```

asllib/aslspec/render.ml
```diff
@@ -0,0 +1,652 @@
...
+
+(** * A signature for modules that provide a specification to the Make functor.
+*)
+module type SPEC_VALUE = sig
+  val spec : Spec.t
+end
+
+(** A functor module for rendering an ASL semantics-specification for a given
+    specification. *)
+module Make (S : SPEC_VALUE) = struct
```
HadrienRenaud, line 17: I'm not sure this would be the case here, but for `Asllib.Interpreter.ml`, using that kind of thing instead of passing `spec` as function argument made the interpreter much slower.

I would also find it easier to understand

asllib/aslspec/SpecParser.mly
```diff
@@ -0,0 +1,224 @@
...
+let check_definition_name name =
+let () = assert (String.length name > 0) in
+let id_regexp = Str.regexp "^[A-Za-z_']+$" in
+if not (Str.string_match id_regexp name 0) then
+    let msg =
+    Format.sprintf
+        "illegal element-defining identifier: %s"
+        name
+    in
+    raise (SpecError msg)
```
HadrienRenaud, line 14: Can we respect OCaml indentation here?

asllib/aslspec/render.ml
HadrienRenaud, line 1: I'm struggling a bit to understand what is in this file, especially when the symbols for Format.fprintf mix with the latex notation.
I'm also wondering what's the point of using Format here when we don't really care about indentation? Why not use Printf?

Also wondering why not have a latex ast, convert the spec to the latex ast, then care about printing this ast (with or without indentation)?

asllib/aslspec/render.ml
```diff
@@ -0,0 +1,652 @@
...
+          | Unspecified -> assert false
+        in
+        let input_layout = List.nth layout_list 0 in
+        let output_layout = List.nth layout_list 1 in
+        match layout with
+        | Horizontal _ ->
+            fprintf fmt {|%a %s %a|} pp_type_term (from_type, input_layout)
+              arrow_symbol pp_type_term (to_type, output_layout)
+        | Vertical _ ->
+            fprintf fmt {|\begin{array}{c}@.%a %s\\@.%a@.\end{array}@.|}
```
HadrienRenaud, line 207: Why not use `pp_latex_array`?

asllib/aslspec/render.ml
```diff
@@ -0,0 +1,652 @@
...
+
+  type parenthesis = Parens | Braces | Brackets
+
+  let pp_parenthesized parenthesis large pp_elem fmt elem =
+    let left = if large then {|\left|} else "" in
+    let right = if large then {|\right|} else "" in
+    match parenthesis with
+    | Parens -> fprintf fmt {|%s(%a%s)|} left pp_elem elem right
+    | Braces -> fprintf fmt {|%s\{%a%s\}|} left pp_elem elem right
+    | Brackets -> fprintf fmt {|%s[%a%s]|} left pp_elem elem right
```
HadrienRenaud, line 108: If possible I would use `"` over `{|` because that makes it quite confusing with latex notation

```suggestion
    | Brackets -> fprintf fmt "%s[%a%s]" left pp_elem elem right
```

asllib/aslspec/render.ml
```diff
@@ -0,0 +1,652 @@
...
+      if false then
+        fprintf Format.std_formatter "%a with original layout %a@."
+          PP.pp_type_term type_term PP.pp_math_shape layout
+    in
+    let layout_contains_vertical = Layout.contains_vertical layout in
+    match type_term with
+    | Label name -> pp_print_string fmt (get_or_gen_math_macro name)
+    | Powerset { term = sub_term; finite } ->
+        let powerset_macro = if finite then {|\powfin|} else {|\pow|} in
+        fprintf fmt {|%s{%a}|} powerset_macro pp_type_term (sub_term, layout)
```
HadrienRenaud, line 161: Could we have a `pp_print_command fmt name args` that would be just adding the relevant `\` at the start and making sure that no command is called without brackets after.
Something I'm fearing is that inadvertantly a macro without argument and without brackets ends up being called on something it shouldn't.

Something like:
```ocaml
let pp_print_command fmt name args_format = fprintf fmt ("\\%s{" ^^ args_format ^^ "}") name
```

Or even something simple like
```ocaml
let pp_print_command_name fmt name = fprintf fmt "\\%s" name
```

Here it would look like

```suggestion
        fprintf fmt "%a{%a}" pp_command_name (if finite then "powfin" else "pow")  pp_type_term (sub_term, layout)
```

asllib/aslspec/render.ml
```diff
@@ -0,0 +1,652 @@
...
+    let num_terms = List.length opt_type_terms in
+    match layout with
+    | Horizontal _ ->
+        List.iteri
+          (fun i (opt_named_term, layout) ->
+            pp_opt_named_type_term fmt (opt_named_term, layout);
+            if i < num_terms - 1 then fprintf fmt ", " else ())
+          opt_terms_with_layouts
+    | Vertical _ ->
+        fprintf fmt {|\begin{array}{c}@.|};
```
HadrienRenaud, line 238: Why not use `pp_print_array`?

asllib/aslspec/render.ml
```diff
@@ -0,0 +1,652 @@
...
+
+  let pp_type_term_union fmt (terms, layout) =
+    if Utils.is_singleton_list terms then
+      fprintf fmt {|%a|} pp_type_term (List.hd terms, layout)
+    else
+      let layout = Layout.horizontal_for_list layout terms in
+      match layout with
+      | Vertical layouts ->
+          let terms_with_layouts = List.combine terms layouts in
+          fprintf fmt
```
HadrienRenaud, line 296: Similarly, why not use `pp_print_array`?

asllib/aslspec/PP.ml
Roman-Manevich: I don't know. Removed them along with a few similar occurrences.

asllib/aslspec/PP.ml
Roman-Manevich: Removed `pp_comma` and `pp_double_newline`. I'm not sure if you meant to comment on `tok_str` as well - I thought it was nicer than using `TokenStrings.string_of_token` everywhere.

asllib/aslspec/render.ml
Roman-Manevich: Interesting. Thanks for sharing.
In this specific case, I don't expect performance to be an issue, but I'll keep an eye on it, in case it does become an issue.

asllib/aslspec/render.ml
Roman-Manevich: switched to `pp_latex_array`

asllib/aslspec/render.ml
Roman-Manevich: Done similar to your suggestion in `pp_operator`.

asllib/aslspec/render.ml
Roman-Manevich: I will try to simplify and add abstractions to help readability.
We do care about indentation to some degree, since it makes it easier to read and understand the generated latex.
LaTeX is rather flat. Other than arrays, there isn't much structure so I think a LaTeX AST would be mostly useless.

HadrienRenaud: Well my concern is maybe about commands and making sure that the correct thing is understood as an argument of the command.

PR #1453, opened 2025-09-24
[asl] Simplify slightly integer division

HadrienRenaud:
We now use matching ZArith primitives for DIVRM and MOD.

Suggested by Anthony Fox.


Roman-Manevich (approved):
Looks good to me.

PR #1461, opened 2025-09-29
[asl] Improve error printing in v0

HadrienRenaud:
This PR adds a command line argument to enable or disable AST splitting in v0.

AST splitting removes a combinatorial explosion in parsing error cases, but messes up with the position tracking in the lexer, and thus error become untrackable. More precisely, the field `pos_cnum` becomes irrelevant, which makes error printing difficult.
This PR hides this behind a command line flag.

This option is always activated in `herd`.

### DIFF
--- a/asllib/builder.ml
+++ b/asllib/builder.ml
@@ -39,6 +39,7 @@ type parser_config = {
   allow_single_arrows : bool;
   allow_empty_structured_type_declarations : bool;
   allow_function_like_statements : bool;
+  v0_use_split_chunks : bool;
 }
 
 type version_selector = [ `ASLv0 | `ASLv1 | `Any ]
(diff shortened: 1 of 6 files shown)


asllib/builder.ml
```diff
@@ -39,6 +39,7 @@ type parser_config = {
   allow_single_arrows : bool;
   allow_empty_structured_type_declarations : bool;
   allow_function_like_statements : bool;
+  v0_use_split_chunks : bool;
```
hrutvik, line 42: nit: consistency of `v0_use_chunks` vs. `v0_use_split_chunks`

hrutvik (approved):

asllib/builder.ml
HadrienRenaud: done thanks

PR #1468, opened 2025-09-30
[aslspec] Minor improvements and fixes to aslspec

Roman-Manevich:
- Added proper support for macro attributes in terms of pretty-printing.
- Added identation to attributes in pretty-printing.
- Allowed (optional) naming sub-terms for `option`, `list0`, `list1`, `powerset`, `fun`, and `partial`.
- Droped the check `check_only_top_level_types`, since it looks like we do want to use constants and types defined as variants in relations signatures.
- Removed extra spaces around `\BeginDefineConstant` and `\BeginDefineType`, since they create redundant extra spaces in the PDF.

### DIFF
--- a/asllib/aslspec/PP.ml
+++ b/asllib/aslspec/PP.ml
@@ -21,8 +21,11 @@ let rec pp_math_shape fmt layout =
 let pp_attribute_key fmt key = pp_print_string fmt (AttributeKey.to_str key)
 
 let pp_attribute fmt = function
-  | StringAttribute s -> fprintf fmt {|"%s"|} s
+  | StringAttribute s ->
+      let s = Str.global_replace (Str.regexp "\n") " " s in
+      fprintf fmt "\"@[<hov>%a@]\"" pp_print_text s
   | MathLayoutAttribute layout -> pp_math_shape fmt layout
+  | MathMacroAttribute macro -> pp_print_string fmt macro
 
 let pp_attribute_key_value fmt (key, value) =
   fprintf fmt {|%a = %a|} pp_attribute_key key pp_attribute value
@@ -30,10 +33,10 @@ let pp_attribute_key_value fmt (key, value) =
 let pp_attribute_key_values fmt attributes =
   if Utils.list_is_empty attributes then ()
   else
-    fprintf fmt "{@[<v>@.%a@.@]}"
+    fprintf fmt "{@[<v>@;<0 2>%a@,@]}"
       (fun fmt attrs ->
         pp_sep_list ~sep:",@,"
-          (fun fmt attr -> fprintf fmt "  %a" pp_attribute_key_value attr)
+          (fun fmt attr -> fprintf fmt "%a" pp_attribute_key_value attr)
           fmt attrs)
       attributes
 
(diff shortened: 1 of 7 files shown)


asllib/aslspec/render.ml
```diff
@@ -51,17 +51,25 @@ module Make (S : SPEC_VALUE) = struct
   (** [vars_of_type_term term] returns the list of term-naming variables that
       occur at any depth inside [term]. *)
   let rec vars_of_type_term term =
+    let option_to_list = function Some x -> [ x ] | None -> [] in
```
hrutvik: nit: perhaps this helper could handle the pair of name and term, to avoid duplication below?

asllib/aslspec/spec.ml
```diff
@@ -325,36 +328,6 @@ module Check = struct
     in
     List.iter (check_no_undefined_ids_in_elem id_to_defining_node) elements
 
-  (** Checks that [type_term] references only top-level types. That is, not
-      constants or type variants. This check assumes that all identifiers in
-      [type_term] are mapped to their defining nodes in [id_to_defining_node].
-  *)
-  let check_only_top_level_types id_to_defining_node type_term =
```
hrutvik, line 332: nit: perhaps a test to illustrate this new flexibility?

asllib/aslspec/tests.t/run.t
```diff
@@ -14,4 +14,32 @@
 
 # Test that --pp is idempotent
   $ aslspec typedefs.spec --pp > tmp1.spec; aslspec tmp1.spec --pp > tmp2.spec; diff tmp1.spec tmp2.spec
+  29,30c29,30
+  <   prose_description = "either the empty set or a singleton set with an Int
+  <                        value"
+  ---
+  >   prose_description = "either the empty set or a singleton set with an Int   
+  >                                            value"
```
HadrienRenaud: Is this meant to be?

asllib/aslspec/tests.t/run.t
```diff
@@ -14,4 +14,32 @@
...
+  37,38c37,38
+  <   prose_description = "infers the type {inferred_type} for the plus
+  <                        expression {input}",
+  ---
+  >   prose_description = "infers the type {inferred_type} for the plus          
+  >                                     expression {input}",
+  40c40
+  <                        {inferred_type}"
+  ---
+  >                                               {inferred_type}"
```
HadrienRenaud: Why are there diffs here?

asllib/aslspec/PP.ml
```diff
@@ -21,29 +21,32 @@ let rec pp_math_shape fmt layout =
...
+  | MathMacroAttribute macro -> pp_print_string fmt macro
 
 let pp_attribute_key_value fmt (key, value) =
   fprintf fmt {|%a = %a|} pp_attribute_key key pp_attribute value
 
 let pp_attribute_key_values fmt attributes =
   if Utils.list_is_empty attributes then ()
   else
-    fprintf fmt "{@[<v>@.%a@.@]}"
+    fprintf fmt "{@.  @[<v>%a@.@]}"
```
HadrienRenaud: I find it strange to use a `@.` in the middle of a formatting thing. Especially if you have `<v>`, you shouldn't need this.

The manual says:
> All open pretty-printing boxes are closed, all pending text is printed.

Why do you put this in the middle of a box?

Do you want to do something like:

```suggestion
    fprintf fmt "{@[<v>%a@ @]}"
```

Maybe with an extra separator at the start, like:

```suggestion
    fprintf fmt "{@[<v>@;<0 2>%a@,@]}"
```

Roman-Manevich: Inexperience with OCaml formatting. Thanks for the suggestion.

asllib/aslspec/tests.t/run.t
Roman-Manevich: Oops. I added `--ignore-all-space` to have more stable diffing. They are gone now.

asllib/aslspec/tests.t/run.t
Roman-Manevich: fixed

PR #1470, opened 2025-10-01
[herd] No more mandatory label  after SVC for ERET to operate

maranget:
More specifically,  the instruction `ERET` was not accepting code addresses in the numeric form as introduced by PR #482.  This is now fixed. 

As a consequence a  label is no longer required after the instruction `SVC` or any faulting instruction for explicit handlers to return with `ERET`.


artkhyzha (approved):
Super -- thank you, Luc!

maranget:
Merged, thanks @artkhyzha an @relokin.

PR #1472, opened 2025-10-02
[asl] Revert optimisation not needed anymore

HadrienRenaud:
This optimisation was there to handle recursive calls, but this is not
needed anymore because of PR #1422.

See `make test-all` workflow running [here](https://github.com/HadrienRenaud/herdtools7/actions/runs/18189177961) (succeeded).


maranget (approved):
I have compared this PR with master on various tests. The PR is slightly faster.

PR #1473, opened 2025-10-02
[Menhir2bnfc build

hrutvik:


asllib/menhir2bnfc/BNFC.ml
HadrienRenaud, line 1: Those are expected because of the bump to `ocamlformat` version, which introduced formatting for documentation comments

dune-install.sh
```diff
@@ -33,7 +33,7 @@ cpdir () {
 }
 
 # Copy binaries
-dune install --only-packages herdtools7 --prefix "${prefix}"
+dune install herdtools7 --prefix "${prefix}"
```
HadrienRenaud, line 36: Why do we need this?
I think `menhir2bnfc` should be installable with `dune install --only-packages menhir2bnfc --prefix "$prefix"`. In which case we could add is as a separate entrypoint in the `Makefile` and not change `herdtools7` installation?

hrutvik: This was because leaving the `--only-packages` flag produces the following error when using an `opam` switch that does not fulfil the requirements to build `menhir2bnfc`:
```
Error: The following <package>.install are missing:
- _build/default/asllib/menhir2bnfc/menhir2bnfc.install
Hint: try running: dune build [-p <pkg>] @install
```

HadrienRenaud: Ah yes that's true, thanks

PR #1474, opened 2025-10-02
[herd] Fix early page table access faults

maranget:
This PR builds on top of PR #1470 and is draft until this previous PR is merged.

In standard VMSA mode,  checking the privilege level for  accessing tho the page table is performed very early and the corresponding potential fault was emitted in an obsolete way. Fix this by emitting the same effects (fault effect + link register write) and control results (B.Fault) as other VMSA faults.


relokin (approved):
Looks good to me!

PR #1475, opened 2025-10-03
[build] Ignore some menhir diagnostic.

maranget:
"Trop de warning tue le warning"

### DIFF
--- a/asllib/dune
+++ b/asllib/dune
@@ -38,18 +38,19 @@
   (:parser Parser.mly)
   (:msg parser_errors.messages))
  (action
-  (with-stdout-to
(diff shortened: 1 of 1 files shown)


hrutvik:
I notice in the `menhir` documentation (section §11.2):

> The correctness and irredundancy of a .messages file are checked by supplying `--compile-errors` *filename* ... This command fails if a sentence does not cause an error at all, or causes an error too early. It
also fails if two sentences lead to the same error state. 

I wonder if the error cases above cover all of the warnings/errors that we might care about for this rule, in which case we might be OK to discard the `stderr` output?

PR #1476, opened 2025-10-03
Aling the definition of same-loc in aarch64loc.cat with the architectural intent

relokin:
The definition of the Same Location is complicated by the fact that we need to define it differently when translation is enabled (-variant vmsa) and from the fact that we want to handle faults. MMU:Translation Faults don't have a valid PA and therefore we resort to comparing the low-order bits whereas all other faults have a known PA which is used to determine if two Effects are to the Same Location or not.

### DIFF
--- a/herd/libdir/aarch64loc.cat
+++ b/herd/libdir/aarch64loc.cat
@@ -4,10 +4,25 @@ let tr-ib =
 
 (* Notions of Same Location - PA, VA, and including Fault Effects *)
 let TTD-same-oa =  same-oa(TTD*TTD) 
-let same-loc = [M]; loc; [M]
-  | [MMU & Translation & FAULT]; same-low-order-bits
-  | same-low-order-bits; [MMU & Translation & FAULT] 
-  | (tr-ib^-1; TTD-same-oa; tr-ib) & same-low-order-bits
+let MMU-Fault-valid-PA = MMU & AccessFlag & FAULT | MMU & Permission & FAULT
+let MMU-Fault-invalid-PA = MMU & Translation & FAULT
+let TTD-valid-OA = TTDV
+
+(* same Physical Address *)
+let same-pa =
+  loc
+  | [MMU-Fault-valid-PA]; (tr-ib^-1; TTD-same-oa; tr-ib) & same-low-order-bits
+  | (tr-ib^-1; TTD-same-oa; tr-ib) & same-low-order-bits; [MMU-Fault-valid-PA]
+
+(* Effect with valid PA *)
+let E-valid-PA = if "vmsa" then (range([TTD-valid-OA]; tr-ib) | TTD & M) else (M | DC.CVAU | IC)
+
+(* Same Location *)
+let same-loc =
+  [E-valid-PA & ~(Tag & M)]; same-pa; [E-valid-PA & ~(Tag & M)]
+  | [(Tag & M) | MMU-Fault-valid-PA]; same-pa; [(Tag & M) | MMU-Fault-valid-PA]
+  | [MMU-Fault-invalid-PA]; same-low-order-bits
+  | same-low-order-bits; [MMU-Fault-invalid-PA]
 let po-loc = po & same-loc
 
 (* Same-cache-line relation *)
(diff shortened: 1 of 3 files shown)


herd/libdir/aarch64loc.cat
```diff
@@ -4,10 +4,25 @@ let tr-ib =
...
+let TTD-valid-OA = TTDV
+
+(* same Physical Address *)
+let same-pa =
+  loc
+  | [MMU-Fault-valid-PA]; (tr-ib^-1; TTD-same-oa; tr-ib) & same-low-order-bits
+  | (tr-ib^-1; TTD-same-oa; tr-ib) & same-low-order-bits; [MMU-Fault-valid-PA]
+
+(* Effect with valid PA *)
+let E-valid-PA = if "vmsa" then (range([TTD-valid-OA]; tr-ib) | TTD & M) else (M | DC.CVAU | IC)
```
artkhyzha, line 18: Should it be `Exp & TTD & M` in this case?

herd/libdir/aarch64loc.cat
```diff
@@ -4,10 +4,25 @@ let tr-ib =
...
+
+(* Same Location *)
+let same-loc =
+  [E-valid-PA & ~(Tag & M)]; same-pa; [E-valid-PA & ~(Tag & M)]
+  | [(Tag & M) | MMU-Fault-valid-PA]; same-pa; [(Tag & M) | MMU-Fault-valid-PA]
```
artkhyzha, line 23: I am curious why this is not simply: `[E-valid-PA]; same-pa; [E-valid-PA]`. Can't seem to spot the interaction that would be nuanced.

herd/libdir/aarch64loc.cat
```diff
@@ -4,10 +4,25 @@ let tr-ib =
...
 let TTD-same-oa =  same-oa(TTD*TTD) 
-let same-loc = [M]; loc; [M]
-  | [MMU & Translation & FAULT]; same-low-order-bits
-  | same-low-order-bits; [MMU & Translation & FAULT] 
-  | (tr-ib^-1; TTD-same-oa; tr-ib) & same-low-order-bits
+let MMU-Fault-valid-PA = MMU & AccessFlag & FAULT | MMU & Permission & FAULT
+let MMU-Fault-invalid-PA = MMU & Translation & FAULT
+let TTD-valid-OA = TTDV
+
+(* same Physical Address *)
```
artkhyzha, line 11: Capitalise "(* Same Physical Address *)"?

herd/libdir/aarch64hwreqs.cat
```diff
@@ -161,8 +161,9 @@ let rec lob =
  | po-scl-ob
  | DSB-ob
  | IFB-ob
- | lwfs
- | lwfs; sca-class
+ | lws
+ | lws; sca-class
```
artkhyzha, line 165: Elsewhere we used `sca-class?` -- is there a miaou7-related reason to go with one over the other?

herd/libdir/aarch64loc.cat
relokin: I think you are referring to the `TTD & M` in `if "vmsa" then (range([TTD-valid-OA]; tr-ib) | TTD & M) else ...`?

Architecturally here we need to include `Imp & TTD & M`. The Effects that have a valid PA are either Effects that are generated after a successful address translation or Effects due to hardware TTD walks (the PA for TTDs is determined without translation). 

Unfortunately, we can't just use `Imp & TTD & R` because in herd7 loads and stores to TTDs are not translated (unless we enabled `-variant PTE2`). So here we have to also include them as a special case.

herd/libdir/aarch64loc.cat
relokin: Assuming `0:X0=x`, take for example:
```
P0          ;
LDR X1,[X0] ;
LDG X2,[X0] ;
```
Architecturally, both instructions operate on the same Location x. In herd7 we distinguish between the two by creating two symbolic locations `x` and `tag(x)` but architecturally we don't define `tag(x)`.

herd/libdir/aarch64hwreqs.cat
relokin: Sorry this shouldn't have been here. I only wanted to go through the changes to Same Location.

In any case, the answer is that I split `lwfs` to `lws` and `lfs` and otherwise kept the logic we had before.

PR #1478, opened 2025-10-05
[aslspec] imrovements and bug fixes

Roman-Manevich:


asllib/aslspec/SpecParser.mly
```diff
@@ -179,15 +179,15 @@ let type_term_with_attributes := ~=type_term; ~=type_attributes;
 
 let type_term :=
     | name=IDENTIFIER; { check_definition_name name; Label name }
-    | POWERSET; LPAR; ~=opt_named_type_term; RPAR; { Powerset {term=opt_named_type_term; finite=false} }
-    | POWERSET_FINITE; LPAR; ~=opt_named_type_term; RPAR; { Powerset {term=opt_named_type_term; finite=true} }
-    | OPTION; LPAR; ~=opt_named_type_term; RPAR; { Option opt_named_type_term }
+    | POWERSET; LPAR; ~=opt_named_type_term; RPAR; { make_operator Powerset opt_named_type_term }
+    | POWERSET_FINITE; LPAR; ~=opt_named_type_term; RPAR; { make_operator Powerset_Finite opt_named_type_term }
+    | LIST0; LPAR; member_type=opt_named_type_term; RPAR; { make_operator List0 member_type }
+    | LIST1; LPAR; member_type=opt_named_type_term; RPAR; { make_operator List1 member_type }
```
hrutvik: nit: consistency
```suggestion
    | LIST0; LPAR; ~=opt_named_type_term; RPAR; { make_operator List0 opt_named_type_term }
    | LIST1; LPAR; ~=opt_named_type_term; RPAR; { make_operator List1 opt_named_type_term }
```

asllib/aslspec/main.ml
```diff
@@ -129,3 +126,17 @@ let () =
...
+  | CLI.CLIError msg ->
+      Format.eprintf "%sUsage Error: %s%s\n" Text.red msg Text.reset_color;
+      exit 1
+  | ParseError msg ->
+      Format.eprintf "%sSyntax Error: %s%s\n" Text.red msg Text.reset_color;
+      exit 1
+  | AST.SpecError msg ->
+      Format.eprintf "%sSpecification Error: %s%s\n" Text.red msg
+        Text.reset_color;
+      exit 1
```
hrutvik: nit: could this be de-duplicated a bit?

asllib/aslspec/render.ml
```diff
@@ -156,11 +114,12 @@ module Make (S : SPEC_VALUE) = struct
         (fun cell_index pp_fun ->
           let () = pp_fun fmt in
           let () =
-            if cell_index mod num_columns <> 0 then fprintf fmt {| & |} else ()
+            if (cell_index + 1) mod num_columns <> 0 then fprintf fmt {| & |}
+            else ()
```
hrutvik: I'm wondering if this could be done with `pp_print_list` with an appropriate `~pp_sep`

asllib/aslspec/render.ml
```diff
@@ -330,10 +297,19 @@ module Make (S : SPEC_VALUE) = struct
...
+                [
+                  (fun fmt -> pp_type_term fmt (term, layout));
+                  (if term_counter < List.length terms - 1 then fun fmt ->
+                     pp_print_string fmt {|\cup|}
+                   else fun _fmt -> ());
+                ])
+              terms_with_layouts
+            |> List.concat
+          in
+          pp_latex_array "ll" fmt term_pp_funs
```
hrutvik, line 310: Similarly, could this use `pp_print_list` with an appropriate `~pp_sep`?

asllib/aslspec/render.ml
```diff
@@ -296,21 +263,21 @@ module Make (S : SPEC_VALUE) = struct
...
+                (fun fmt -> pp_print_string fmt " : ");
                 (fun fmt -> pp_type_term fmt (field_term, layout));
               ])
             fields_with_layouts
         in
         fprintf fmt {|%a|}
-          (pp_parenthesized Brackets true pp_latex_array_rows)
-          row_funs
+          (pp_parenthesized Brackets true (pp_latex_array "lcl"))
+          field_pp_funs
```
hrutvik, line 279: Similarly, could this use `pp_print_list`?

asllib/aslspec/render.ml
hrutvik, line 1: In general: it's tricky to understand the impact of some of the LaTeX output changes. I wonder: should we have some tests with expected LaTeX output checked in? That way, changes like this are accompanied with a `dune promote` that makes it easy to understand how the generated LaTeX is affected.

asllib/aslspec/spec.ml
```diff
@@ -4,6 +4,36 @@
...
+        opt_named_term_to_var_list from_type
+        @ opt_named_term_to_var_list to_type
+  in
+  List.sort_uniq String.compare listed_vars
+
+(** [opt_named_term_to_var_list (var, t)] returns the list of term-naming
+    variables that occur at any depth inside [t], plus [var] if it is [Some x].
+*)
+and opt_named_term_to_var_list (var, t) =
+  (match var with Some x -> [ x ] | None -> []) @ vars_of_type_term t
```
```suggestion
  Option.to_list var @ vars_of_type_term t
```

asllib/aslspec/spec.ml
```diff
@@ -328,6 +379,79 @@ module Check = struct
...
+        does not contain a [{var}] where [var] is not in [vars]. Otherwise,
+        raises a [SpecError] detailing the unmatched variables. *)
+    let check_prose_template_for_vars template vars =
+      let open Text in
+      (* Associate each [var] with [{var}] *)
+      let substitutions_map =
+        List.fold_left
+          (fun acc_map var_str ->
+            let template_var = spec_var_to_template_var var_str in
+            StringMap.add template_var "" acc_map)
```
hrutvik: It looks a bit like this `StringMap` is being used like a `StringSet`?

asllib/aslspec/spec.ml
```diff
@@ -328,6 +379,79 @@ module Check = struct
...
+        in
+        raise (SpecError msg)
+
+    let check_prose_template_for_definition_node defining_node =
+      let prose_description = prose_description_for_node defining_node in
+      let vars = vars_of_node defining_node in
+      let () = check_prose_template_for_vars prose_description vars in
+      match defining_node with
+      | Node_Type _ | Node_TypeVariant _ | Node_Constant _ ->
+          check_prose_template_for_vars prose_description vars
```
hrutvik: This looks like a repeat of line 440

asllib/aslspec/spec.ml
```diff
@@ -328,6 +379,79 @@ module Check = struct
...
+      | Node_Relation def ->
+          let prose_description = Relation.prose_description def in
+          let () = check_prose_template_for_vars prose_description vars in
```
hrutvik: This also seems unnecessary after line 440

asllib/aslspec/render.ml
Roman-Manevich: The problem is that the separator works off of the index, which `pp_print_list` doesn't pass. I've simplified the code a little.

asllib/aslspec/render.ml
Roman-Manevich: Done. Take a look now.

asllib/aslspec/tests.t/run.t
```diff
@@ -1,11 +1,142 @@
 # Basic tests
-  $ aslspec hello.spec
-  $ aslspec typedefs.spec
-  $ aslspec relations.spec
+  $ aslspec hello.spec --render; cat generated_macros.tex; rm -f generated_macros.tex
```
HadrienRenaud: Thank you for this, this is very helpful for reviewing.

asllib/aslspec/tests.t/run.t
```diff
@@ -1,11 +1,142 @@
...
+  Generated LaTeX macros into generated_macros.tex
+  % ==================================================
+  % AUTO-GENERATED - DO NOT EDIT
+  % ==================================================
+  
+  % ------------------
+  % Macros for symbols
+  % ------------------
+  
+  \newcommand\t[0]{ \hyperlink{type-t}{\textsf{t}} } % Generated from t
```
HadrienRenaud: Could there be a clash between different names? say I also have a constant called `t`, or a math variable called `t`?

asllib/aslspec/tests.t/run.t
```diff
@@ -1,11 +1,142 @@
...
+  \begin{flalign*}
+  \s \triangleq\ & \pow{\t}\hypertarget{type-A}{}\\
+  \cup\ & \A\hypertarget{type-B}{}\\
+  \cup\ & \B(\s, \s)\end{flalign*}} % EndDefineType
+  
+  
+
+
+  $ aslspec typedefs.spec --render; cat generated_macros.tex; rm -f generated_macros.tex
+  Fatal error: exception File "asllib/aslspec/render.ml", line 250, characters 23-29: Assertion failed
```
HadrienRenaud: That doesn't sound good.

asllib/aslspec/tests.t/run.t
```diff
@@ -1,11 +1,142 @@
...
+  \newcommand\t[0]{ \hyperlink{type-t}{\textsf{t}} } % Generated from t
+  \newcommand\s[0]{ \hyperlink{type-s}{\textsf{s}} } % Generated from s
+  \newcommand\A[0]{ \hyperlink{type-A}{\textsf{A}} } % Generated from A
+  \newcommand\B[0]{ \hyperlink{type-B}{\textsf{B}} } % Generated from B
+  
+  % -------------------
+  % Macros for elements
+  % -------------------
+  
+  \BeginDefineType{t}{\hypertarget{type-t}{}$\t$} % EndDefineType
```
HadrienRenaud: I personally find it strange that the macros are called `\BeginSomething`: why would we need `Begin` if there's no need for a `\EndSomething`? Why not call it `\DefineType`?

asllib/aslspec/render.ml
HadrienRenaud: I think `pp_latex_array` is `pp_print_list` with the correct `pp_sep` under the hood.

(well in practice it is not, but it is the same implementation as `pp_print_list`)

asllib/aslspec/SpecParser.mly
```diff
@@ -179,15 +179,15 @@ let type_term_with_attributes := ~=type_term; ~=type_attributes;
...
 let type_term :=
     | name=IDENTIFIER; { check_definition_name name; Label name }
-    | POWERSET; LPAR; ~=opt_named_type_term; RPAR; { Powerset {term=opt_named_type_term; finite=false} }
-    | POWERSET_FINITE; LPAR; ~=opt_named_type_term; RPAR; { Powerset {term=opt_named_type_term; finite=true} }
-    | OPTION; LPAR; ~=opt_named_type_term; RPAR; { Option opt_named_type_term }
+    | POWERSET; LPAR; ~=opt_named_type_term; RPAR; { make_operator Powerset opt_named_type_term }
+    | POWERSET_FINITE; LPAR; ~=opt_named_type_term; RPAR; { make_operator Powerset_Finite opt_named_type_term }
+    | LIST0; LPAR; ~=opt_named_type_term; RPAR; { make_operator List0 opt_named_type_term }
+    | LIST1; LPAR; ~=opt_named_type_term; RPAR; { make_operator List1 opt_named_type_term }
+    | OPTION; LPAR; ~=opt_named_type_term; RPAR; { make_operator Option opt_named_type_term }
```
HadrienRenaud: Maybe it would be more intuitive to put the operators in a declaration to be inlined?

```ocaml
let operator ==
  | POWERSET; { Powerset }
  (* ... *)
  | OPTION; { Option }

let type_term :=
  (* ... *)
  | o=operator; LPAR; t=opt_name_type_term; RPAR; { make_operator o t }
  (* ... *)
```

asllib/aslspec/render.ml
```diff
@@ -155,22 +113,26 @@ module Make (S : SPEC_VALUE) = struct
...
-            else ()
-          in
-          ())
+          if (cell_index + 1) mod num_columns <> 0 then fprintf fmt {| & |}
+          else if cell_index < List.length pp_funs - 1 then fprintf fmt {|\\@.|}
+          else ())
         pp_funs
     in
     let () = fprintf fmt {|\end{array}|} in
     ()
```
HadrienRenaud: Can I suggest an implementation based on `pp_print_list` instead of re-implementing it?

```suggestion
(* Transform a list of (n*m) elements into a list with n list of m elements. *)
let _list_divide n =
  let rec _list_take n l acc =
    (* Loosely inspired from stdlib *)
    match n, l with
    | 0, _ | _, [] -> List.rev acc, l
    | n, x :: l -> _list_take (n - 1) l (x :: acc)
  in
  let rec _loop l acc =
    if List.is_empty l then List.rev acc
    else
      let heads, tail = _list_take n l [] in
      _loop (heads :: acc) tail
  in
  fun l -> _loop l []

let pp_latex_array =
  let pp_cell fmt pp_fun = pp_fun fmt in
  let pp_inside_row fmt pp_funs =
    pp_print_list ~pp_sep:(fun fmt () -> fprint fmt "@ & ") pp_cell pp_funs
  in
  let pp_row fmt pp_funs =
    fprintf fmt "@[%a@]\\\\@ " pp_inside_row pp_funs
  in
  let pp_inside num_columns fmt pp_funs =
    pp_print_list ~pp_sep:pp_print_space pp_print_row (_list_divide num_columns pp_funs)
  in
  fun alignment fmt ->
    let num_columns = String.length alignment in
    fprintf fmt "@[<hov 2>\\begin{array}{%s}@ %a@ \\end{array}@]" alignment (pp_inside num_columns)
```

Ideally `pp_funs` would already be a list of list so you wouldn't need to do the whole `_list_divide` manipulation.

asllib/aslspec/tests.t/run.t
Roman-Manevich: The tool checks for name clashes, but it doesn't know about the extrenal set of macros so that can possibly lead to clashes.

asllib/aslspec/tests.t/run.t
Roman-Manevich: fixed

asllib/aslspec/tests.t/run.t
Roman-Manevich: done

asllib/aslspec/render.ml
```diff
@@ -120,56 +97,40 @@ module Make (S : SPEC_VALUE) = struct
...
+    let i_ref = ref 0 in
+    let array_sep fmt () =
+      if (!i_ref + 1) mod num_columns <> 0 then fprintf fmt " & "
+      else fprintf fmt {|\\@ |};
+      i_ref := !i_ref + 1
+    in
+    let pp_elt = fun fmt pp_fun -> pp_fun fmt in
+    fprintf fmt "@[<v>\\begin{array}{%s}@ %a@ \\end{array}@]" alignment
+      (pp_print_list ~pp_sep:array_sep pp_elt)
+      pp_funs
```
hrutvik: This still seems a bit non-ideal - I would echo @HadrienRenaud's [comment](https://github.com/herd/herdtools7/pull/1478#discussion_r2406049719), specifically the part about converting `pp_latex_array` into a function that accepts a list of lists for `pp_funs` (top-level list are the rows, internal lists are the columns of each row). Then this could perhaps use a nested `pp_print_list` - what do you think? I'm aware you have a few stacked PRs, so feel free to file this as something to improve in one of the later PRs.

asllib/aslspec/render.ml
```diff
@@ -296,21 +257,21 @@ module Make (S : SPEC_VALUE) = struct
     in
     match layout with
     | Vertical _ ->
-        let row_funs =
-          List.map
+        let field_pp_funs =
+          Utils.list_concat_map
```
hrutvik: I think a modified `pp_latex_array` could then convert this to a simple `map` (i.e. no `concat`)

asllib/aslspec/render.ml
```diff
@@ -330,10 +291,19 @@ module Make (S : SPEC_VALUE) = struct
...
+            List.mapi
+              (fun term_counter (term, layout) ->
+                [
+                  (fun fmt -> pp_type_term fmt (term, layout));
+                  (if term_counter < List.length terms - 1 then fun fmt ->
+                     pp_print_string fmt {|\cup|}
+                   else fun _fmt -> ());
+                ])
+              terms_with_layouts
+            |> List.concat
```
hrutvik: Similarly, perhaps could delete `concat` here

asllib/aslspec/tests.t/run.t
```diff
@@ -1,11 +1,14 @@
 # Basic tests
-  $ aslspec hello.spec
-  $ aslspec typedefs.spec
-  $ aslspec relations.spec
+  $ aslspec hello.spec --render; diff generated_macros.tex hello.expected; rm -f generated_macros.tex
```
hrutvik, line 2: I think we should work towards having a separate expect-test directory for these first few, which would use `dune` rules to generate/diff/promote the `*.tex` files - but happy for this to be in a later PR

asllib/aslspec/render.ml
```diff
@@ -330,10 +291,19 @@ module Make (S : SPEC_VALUE) = struct
...
-            {|\left(\begin{array}{ll}@[<hv>%a@] & \\@.\end{array}\right)|}
-            (PP.pp_sep_list ~sep:{| & \cup \\@.|} pp_type_term)
-            terms_with_layouts
+          let term_pp_funs =
+            List.mapi
+              (fun term_counter (term, layout) ->
+                [
+                  (fun fmt -> pp_type_term fmt (term, layout));
+                  (if term_counter < List.length terms - 1 then fun fmt ->
+                     pp_print_string fmt {|\cup|}
```
hrutvik, line 305: Then I wonder if a modified `pp_latex_array` could be parametrised by optional additional separators to insert between row/columns, which might help tidy this up too

asllib/aslspec/render.ml
```diff
@@ -120,56 +97,37 @@ module Make (S : SPEC_VALUE) = struct
...
-          in
-          ())
-        pp_funs
-    in
-    let () = fprintf fmt {|\end{array}|} in
-    ()
+    let pp_elt = fun fmt pp_fun -> pp_fun fmt in
+    let pp_one_row fmt pp_funs =
+      let () = assert (List.compare_length_with pp_funs num_columns = 0) in
+      pp_print_list ~pp_sep:(fun fmt () -> fprintf fmt " & ") pp_elt fmt pp_funs
```
HadrienRenaud, line 114: I'm wondering why you're not putting the row in a box, and putting some breaking spaces in columns?
(This is what I had suggested in the previous PR's review)

```suggestion
      fprintf fmt "@[%a@ \\\\@]" (pp_print_list ~pp_sep:(fun fmt () -> fprintf fmt "@ & ") pp_elt) pp_funs
```

(I've also moved the latex line break to the row, in my opinion this is where it belongs but feel free to ignore this one)

asllib/aslspec/render.ml
```diff
@@ -274,15 +236,19 @@ module Make (S : SPEC_VALUE) = struct
         let term_pp_funs =
           List.mapi
             (fun term_counter term_and_layout ->
-              if term_counter < num_terms - 1 then fun fmt ->
-                fprintf fmt {|%a,@.|} pp_opt_named_type_term term_and_layout
-              else fun fmt -> pp_opt_named_type_term fmt term_and_layout)
+              [
+                (fun fmt ->
+                  if term_counter < num_terms - 1 then
+                    fprintf fmt {|%a,@.|} pp_opt_named_type_term term_and_layout
```
HadrienRenaud: If you're using `@.` you're breaking the array alignment. Why not simply put the content in a box?


```suggestion
                    fprintf fmt "@[%a,@]" pp_opt_named_type_term term_and_layout
```

asllib/aslspec/render.ml
Roman-Manevich: I tried it, but the usage becomes quite cumbersome.

asllib/aslspec/render.ml
Roman-Manevich: This created an extra break line, which fails LaTeX.

HadrienRenaud: Ah this is sad

PR #1482, opened 2025-10-07
[aslspec] inference rules

Roman-Manevich:
This PR introduces inference rules to relations with two examples for `annotate_literal` and `annotate_stmt`.
Here is a snippet from `annotate_literal`:
```
typing function annotate_literal(tenv: static_envs, l: literal) -> (t: ty) {...} =
  case Int {
    l = L_Int(n);
    cs := LIST(Constraint_Exact(E_Literal(L_Int(n))));
    --
    T_Int(WellConstrained(cs));
  }

  case Bool {
    l = L_Bool(_);
    --
    T_Bool;
  }
...
}
```
The notation tries to be close to OCaml by allowing (nested) case splitting and `-- result` to stand for the output of a conclusion transition.

- The language of expressions that is allowed in judgments is a bit limited right now, but it will grow in subsequent PRs.
The rendering currently only generates inference rules, not prose.
- short-circuiting is automated, but slightly broken. That is, if no short-circuit terms are specified for a given transition, aslspec inserts the default terms based on the signature of the relation. This is good enough for 99% of the cases. One can override the list of short-circuit alternative for a transition by terms, which is fine, except that it will not generate the `#TE`-style macros. I will later extend aslspec to support this.
- Currently, the rendering is almost identical to the manual rules, except that the output configuration of the conclusion transition does not name its sub-expressions. I will add support for this in a later PR.
To generate a LaTeX macro for a rule two options are given:
- All cases of a relation. The following example shows how to render all cases for `annotate_literal`: `render rule annotate_literal;`
- A subset of the cases. This is done by specifying the absolute name of a case, which generates all cases nested in it. For example: `render rule annotate_stmt_SDecl = annotate_stmt(SDecl);` generates the macro `annotate_stmt_SDecl`, which renders both cases for the `SDecl` case of `annotate_stmt`.

The generated macros are then used in LaTeX as follows: `\RenderRule{annotate_stmt_SDecl}`.

### DIFF
--- a/asllib/aslspec/AST.ml
+++ b/asllib/aslspec/AST.ml
@@ -411,7 +411,7 @@ module Rule = struct
         (** The first identifier is a variable and the rest are field names. *)
     | ListIndex of { var : string; index : string }
     | Record of { label : string; fields : (string * expr) list }
-        (** A record construction expression, optionally labelled. *)
+        (** A record construction expression. *)
     | Application of { applicator : applicator; args : expr list }
         (** An application of [applicator] to the list of argument expressions
             [args]. *)
--- a/asllib/aslspec/PP.ml
+++ b/asllib/aslspec/PP.ml
@@ -127,7 +127,7 @@ let rec pp_expr fmt =
   | Application { applicator; args } ->
       fprintf fmt "%a(%a)" pp_application_lhs applicator (pp_comma_list pp_expr)
         args
-  | FieldAccess path -> fprintf fmt "%s" (String.concat "." path)
+  | FieldAccess path -> pp_print_string fmt (String.concat "." path)
   | ListIndex { var; index } -> fprintf fmt "%s[%s]" var index
   | Record { label; fields } ->
       fprintf fmt "%s[%a]" label
@@ -147,7 +147,7 @@ and pp_application_lhs fmt =
   | Relation name -> pp_print_string fmt name
   | TupleLabel name -> pp_print_string fmt name
   | ExprOperator op -> pp_print_string fmt (expr_operator_to_string op)
-  | Fields path -> fprintf fmt "%s" (String.concat "." path)
+  | Fields path -> pp_print_string fmt (String.concat "." path)
   | Unresolved expr -> pp_expr fmt expr
 
 and pp_short_circuit fmt short_circuit =
@@ -221,9 +221,9 @@ let pp_constant_definition fmt ({ Constant.name } as def) =
     (Constant.attributes_to_list def)
 
 let type_subset_pointer fmt { TypesRender.type_name; variant_names } =
-  if Utils.list_is_empty variant_names then fprintf fmt {|%s(-)|} type_name
+  if Utils.list_is_empty variant_names then fprintf fmt "%s(-)" type_name
   else
-    fprintf fmt {|%s(%a)|} type_name
+    fprintf fmt "%s(%a)" type_name
       (pp_sep_list ~sep:", " pp_print_string)
       variant_names
 
--- a/asllib/aslspec/SpecParser.mly
+++ b/asllib/aslspec/SpecParser.mly
@@ -159,7 +159,7 @@ let elem :=
     | relation_definition
     | constant_definition
     | render_types
-    |render_rule
+    | render_rule
 
 let type_kind := TYPEDEF; { TypeKind_Generic }
     | AST; { TypeKind_AST }
@@ -337,8 +337,8 @@ let expr :=
       { Rule.make_infix_operator_application infix_expr_operator lhs rhs }
     | ~=field_path;
       { Rule.FieldAccess field_path }
-    | list=IDENTIFIER; LBRACKET; index=IDENTIFIER; RBRACKET;
-      { Rule.make_list_index list index }
+    | list_var=IDENTIFIER; LBRACKET; index=IDENTIFIER; RBRACKET;
+      { Rule.make_list_index list_var index }
     | name=IDENTIFIER; LBRACKET; fields=tclist1(field_and_value); RBRACKET;
       { Rule.make_record name fields }
     | lhs=expr; ARROW; rhs=expr; ~=short_circuit;
--- a/asllib/aslspec/latex.ml
+++ b/asllib/aslspec/latex.ml
@@ -182,11 +182,7 @@ let pp_elements_and_separator ~pp_sep ~alignment ~same_column pp_elem layout fmt
                   pp_elem fmt elem_with_layout;
                   pp_cond_sep fmt);
               ]
-            else
-              [
-                (fun fmt -> pp_elem fmt elem_with_layout);
-                (fun fmt -> pp_cond_sep fmt);
-              ])
+            else [ (fun fmt -> pp_elem fmt elem_with_layout); pp_cond_sep ])
           elements_with_layouts
       in
       let alignment = if same_column then alignment else alignment ^ "c" in
@@ -229,7 +225,7 @@ let pp_connect_pair ~alignment fmt pp_lhs_with_layout lhs connector_macro_name
   in
   match layout with
   | Horizontal _ ->
-      fprintf fmt {|%a %a %a|} pp_lhs_with_layout lhs_with_layout pp_macro
+      fprintf fmt "%a %a %a" pp_lhs_with_layout lhs_with_layout pp_macro
         connector_macro_name pp_rhs_with_layout rhs_with_layout
   | Vertical _ ->
       fprintf fmt {|\begin{array}{%s}@.  %a %a\\@.  %a@.\end{array}|} alignment
@@ -268,15 +264,13 @@ let pp_fields pp_field_name pp_field_value fmt (fields, layout) =
             ])
           fields_values_with_layouts
       in
-      fprintf fmt {|%a|}
-        (pp_parenthesized Braces true (pp_latex_array "lcl"))
-        field_pp_funs
+      pp_parenthesized Braces true (pp_latex_array "lcl") fmt field_pp_funs
   | Horizontal _ ->
       let pp_field fmt (field_name, (fields_value, layout)) =
-        fprintf fmt {|%a : %a|} pp_field_name field_name pp_field_value
+        fprintf fmt "%a : %a" pp_field_name field_name pp_field_value
           (fields_value, layout)
       in
-      fprintf fmt {|%a|}
-        (pp_parenthesized Braces true (PP.pp_sep_list ~sep:", " pp_field))
-        fields_values_with_layouts
+      pp_parenthesized Braces true
+        (PP.pp_sep_list ~sep:", " pp_field)
+        fmt fields_values_with_layouts
   | Unspecified -> assert false
--- a/asllib/aslspec/main.ml
+++ b/asllib/aslspec/main.ml
@@ -147,9 +147,12 @@ let () =
   let config = CLI.parse_args () in
   if !CLI.arg_debug then
     (* Allow the exception stack trace to be printed for debugging. *)
-    execute config |> fun () -> exit 0
+    let () = execute config in
+    exit 0
   else
-    try execute config |> fun () -> exit 0
+    try
+      let () = execute config in
+      exit 0
     with error ->
       let error_type, msg =
         match error with
--- a/asllib/aslspec/render.ml
+++ b/asllib/aslspec/render.ml
@@ -5,6 +5,7 @@ open AST
 open Format
 open Latex
 open LayoutUtils
+open Macros
 open Spec
 
 (** * A signature for modules that provide a specification to the Make functor.
@@ -20,6 +21,7 @@ module Make (S : SPEC_VALUE) = struct
       is given in which case it generates a math macro name and returns it. *)
   let get_or_gen_math_macro id =
     assert (not (String.equal id ""));
+    (* We add a prefix to the macro name to avoid potential clashes. *)
     let macro_prefix = function
       | Node_Relation _ | Node_Type _ | Node_TypeVariant _ | Node_Constant _ ->
           ""
@@ -69,7 +71,7 @@ module Make (S : SPEC_VALUE) = struct
   (** [pp_constant_definition fmt def] renders the definition of the constant
       given by [def] with the formatter [fmt]. *)
   let pp_constant_definition fmt { Constant.name } =
-    fprintf fmt {|$%a$|} pp_id_as_macro name
+    fprintf fmt "$%a$" pp_id_as_macro name
 
   (** [pp_constant_definition_macro fmt def] renders the LaTeX wrapper macro
       [\DefineConstant{name}{...}] around the rendering of a constant definition
@@ -83,11 +85,11 @@ module Make (S : SPEC_VALUE) = struct
       [arg] is rendered by [pp_arg fmt arg]. *)
   let pp_type_operator fmt op pp_arg arg =
     let operator_to_macro_name = function
-      | Powerset -> "pow"
-      | Powerset_Finite -> "powfin"
-      | List0 -> "KleeneStar"
-      | List1 -> "KleenePlus"
-      | Option -> "Option"
+      | Powerset -> pow_macro_name
+      | Powerset_Finite -> powfin_macro_name
+      | List0 -> kleene_star_macro_name
+      | List1 -> kleene_plus_macro_name
+      | Option -> option_macro_name
     in
     pp_one_arg_macro (operator_to_macro_name op) pp_arg fmt arg
 
@@ -124,18 +126,17 @@ module Make (S : SPEC_VALUE) = struct
               (field_name, field))
             fields
         in
-        fprintf fmt {|%a%a|} pp_id_opt_as_macro label_opt
+        fprintf fmt "%a%a" pp_id_opt_as_macro label_opt
           (pp_fields pp_field_name pp_record_field_as_pair)
           (pp_record_fields_as_pairs, layout)
     | ConstantsSet constant_names ->
         let layout_contains_vertical = LayoutUtils.contains_vertical layout in
-        fprintf fmt {|%a|}
-          (pp_parenthesized Braces layout_contains_vertical
-             (PP.pp_sep_list ~sep:", " pp_id_as_macro))
-          constant_names
+        pp_parenthesized Braces layout_contains_vertical
+          (PP.pp_sep_list ~sep:", " pp_id_as_macro)
+          fmt constant_names
     | Function { from_type; to_type; total } ->
         let arrow_macro_name =
-          if total then {|rightarrow|} else {|partialto|}
+          if total then rightarrow_macro_name else partialto_macro_name
         in
         pp_connect_pair ~alignment:"r" fmt pp_opt_named_type_term from_type
           arrow_macro_name pp_opt_named_type_term to_type layout
@@ -145,7 +146,7 @@ module Make (S : SPEC_VALUE) = struct
       laid out according to [layout]. *)
   and pp_opt_named_type_term fmt ((name_opt, term), layout) =
     match name_opt with
-    | None -> fprintf fmt {|%a|} pp_type_term (term, layout)
+    | None -> pp_type_term fmt (term, layout)
     | Some name -> pp_overtext fmt pp_type_term (term, layout) pp_var name
 
   (** [pp_opt_named_type_terms fmt (opt_type_terms, layout)] formats a list of
@@ -163,7 +164,7 @@ module Make (S : SPEC_VALUE) = struct
       in with the given [layout]. *)
   let pp_output_types fmt (terms, layout) =
     if Utils.is_singleton_list terms then
-      fprintf fmt {|%a|} pp_type_term (List.hd terms, layout)
+      pp_type_term fmt (List.hd terms, layout)
     else
       let pp_multiple_terms fmt (terms, layout) =
         pp_aligned_elements_and_operators
@@ -181,8 +182,8 @@ module Make (S : SPEC_VALUE) = struct
     in
     let property_macro_name =
       match property with
-      | RelationProperty_Relation -> "bigtimes"
-      | RelationProperty_Function -> "longrightarrow"
+      | RelationProperty_Relation -> bigtimes_macro_name
+      | RelationProperty_Function -> longrightarrow_macro_name
     in
     pp_connect_pair ~alignment:"c" fmt pp_type_term input_as_labelled_tuple
       property_macro_name pp_output_types output layout
@@ -222,7 +223,7 @@ module Make (S : SPEC_VALUE) = struct
       | Some layout -> layout
       | None -> Layout.for_type_term term
     in
-    fprintf fmt "%a" pp_type_term (term, layout)
+    pp_type_term fmt (term, layout)
 
   (** [pp_typename_with_hypertarget fmt type_name] renders the type name
       [type_name] along with its hypertarget. *)
@@ -230,7 +231,7 @@ module Make (S : SPEC_VALUE) = struct
     let hyperlink_target = hypertarget_for_id type_name in
     (* The hypertarget macro must come after the type name macro.
        Otherwise LaTeX compilation fails. *)
-    fprintf fmt {|%a%a|} pp_id_as_macro type_name pp_mathhypertarget
+    fprintf fmt "%a%a" pp_id_as_macro type_name pp_mathhypertarget
       hyperlink_target
 
   (** [pp_variant_with_hypertarget fmt variant] renders the variant [variant]
@@ -247,13 +248,13 @@ module Make (S : SPEC_VALUE) = struct
               hypertarget_for_id field_name)
             fields
         in
-        fprintf fmt {|%a%a%a|}
+        fprintf fmt "%a%a%a"
           (Format.pp_print_option pp_mathhypertarget)
           hyperlink_target_opt
           (PP.pp_sep_list ~sep:"" pp_mathhypertarget)
           field_hyperlink_targets pp_variant variant
     | _ ->
-        fprintf fmt {|%a%a|} pp_variant variant
+        fprintf fmt "%a%a" pp_variant variant
           (Format.pp_print_option pp_mathhypertarget)
           hyperlink_target_opt
 
@@ -266,8 +267,8 @@ module Make (S : SPEC_VALUE) = struct
       ?(is_last = true) fmt ({ Type.type_kind; Type.name }, variants) =
     let equality_symbol, join_symbol =
       match type_kind with
-      | TypeKind_AST -> ({|\derives|}, "|")
-      | TypeKind_Generic -> ({|\triangleq|}, {|\cup|})
+      | TypeKind_AST -> (derivation_macro_name, pipe_macro_name)
+      | TypeKind_Generic -> (triangleq_macro_name, union_macro_name)
     in
     let first_variant, variants_tail =
       match variants with
@@ -282,24 +283,23 @@ module Make (S : SPEC_VALUE) = struct
       if lhs_hypertargets then pp_typename_with_hypertarget else pp_id_as_macro
     in
     let _first_line =
-      fprintf fmt {|@.%a %s\ & %a|} pp_typename name equality_symbol
+      fprintf fmt {|@.%a %a\ & %a|} pp_typename name pp_macro equality_symbol
         pp_variant_with_hypertarget first_variant
     in
     let _add_latex_line_break_only_if_more_variants =
       if List.length variants > 1 then fprintf fmt {|\\@.|}
-      else fprintf fmt {|@.|}
+      else fprintf fmt "@."
     in
     let num_variants_tail = List.length variants_tail in
     let _render_variants_tail =
       List.iteri
         (fun counter variant ->
           let () =
-            fprintf fmt {|@[<h>%s\ & %a@]|} join_symbol
+            fprintf fmt {|@[<h>%a\ & %a@]|} pp_macro join_symbol
               pp_variant_with_hypertarget variant
           in
           let _add_latex_line_break_except_on_last_line =
             if counter < num_variants_tail - 1 then fprintf fmt {|\\@.|}
-            else fprintf fmt {||}
           in
           ())
         variants_tail
@@ -311,8 +311,7 @@ module Make (S : SPEC_VALUE) = struct
       variants) [basic_type] with the formatter [fmt]. *)
   let pp_basic_type fmt { Type.name } =
     let hyperlink_target = hypertarget_for_id name in
-    fprintf fmt {|%a$%a$|} pp_texthypertarget hyperlink_target pp_id_as_macro
-      name
+    fprintf fmt "%a$%a$" pp_texthypertarget hyperlink_target pp_id_as_macro name
 
   (** [pp_type_definition fmt def] renders the type definition [def] with the
       formatter [fmt]. *)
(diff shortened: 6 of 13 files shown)


asllib/aslspec/AST.ml
```diff
@@ -272,20 +300,71 @@ end = struct
...
+  (** A term that can be used to form a judgement. *)
+  type expr =
+    | Var of string
+    | Operator of operator
+    | Tuple of expr list
+    | Application of { func : expr; components : expr list }
+    | Field of { record : expr; field : string }
+    | ListIndex of { list : expr; index : string }
+    | Record of { name : expr; fields : (string * expr) list }
+
```
hrutvik, line 434: This is quite flexible - do we expect e.g. arbitrary expressions as functions or record names?

asllib/aslspec/AST.ml
```diff
@@ -272,20 +300,71 @@ end = struct
...
+  type judgment_form =
+    | Expr of expr  (** a Boolean-valued expression *)
+    | Output of expr  (** The output configuration of a conclusion judgement. *)
+    | Transition of { lhs : expr; rhs : expr }
+        (** A transition from the [lhs] configuration to the [rhs]
+            configuration. *)
+    | Indexed of { index : string; list : string; body : judgment_form }
+    | LaTeX of { uses : string list; defines : string list; latex : string }
```
hrutvik: I suppose this is to permit flexibility in case the normal grammar is not enough?

asllib/aslspec/SpecParser.mly
```diff
@@ -218,3 +237,53 @@ let render_definition :=
 let type_subset_pointer :=
     | type_name=IDENTIFIER; LPAR; MINUS; RPAR; { (type_name, []) }
     | type_name=IDENTIFIER; LPAR; variant_names=tclist1(IDENTIFIER); RPAR; { (type_name, variant_names) }
+
+let opt_relation_rule := { None }
+    | EQ; rule=rule; { Some rule }
+
+let rule := elements=list1(rule_element); { Rule.Case { name = "_"; elements } }
```
hrutvik: Any significance to the underscore `"_"` here?

asllib/aslspec/SpecParser.mly
```diff
@@ -218,3 +237,53 @@ let render_definition :=
...
+
+let rule_element :=
+    | ~=judgment; SEMI; { Rule.Judgement judgment }
+    | CASE; name=IDENTIFIER; LBRACE; elements=terminated(list(rule_element), RBRACE);
+      { Rule.Case { name; elements } }
+
+let judgment :=
+    | ~=judgment_form; ~=judgment_attributes; { Rule.make_judgement judgment_form judgment_attributes  }
+    | ~=judgment_form; { Rule.make_judgement judgment_form [] }
```
hrutvik: nit: could these cases be combined using `loption(judgment_atrributes)`?

asllib/aslspec/SpecParser.mly
```diff
@@ -218,3 +237,53 @@ let render_definition :=
...
+
+let judgment_form :=
+    | ~=expr;
+      { Rule.Expr expr }
+    | expr1=expr; COLON_EQ; expr2=expr;
+      { Rule.Expr (Rule.Application
+       { func = Rule.Operator Rule.Assign; components = [expr1; expr2] }) }
+    | MINUS_MINUS; ~=expr;
```
hrutvik: This seems a little odd - in that I assume we can expect at most one output, and it will be at the bottom? But this allows a lot of flexibility in e.g. multiple output lines. I wonder if the output part should be handled at a higher-level production (e.g. `judgment`)?

asllib/doc/asl.spec
```diff
@@ -1205,19 +1230,3449 @@ relation annotate_literal(tenv: static_envs, l: literal) -> (t: ty)
...
+    --
+    (S_Assign(le1, re1), tenv, ses);
+  }
+
+  case SDecl {
+    case Some {
+      s = S_Decl(ldk, ldi, ty_opt, Some(e));
+      annotate_expr(tenv, e) -> (t_e, e', ses_e);
+      annotate_local_decl_type_annot(tenv, ty_opt, t_e, ldk, e', ldi) -> (tenv1, ty_opt', ses_ldi)
+      { [_,_] };
```
hrutvik: suggestion: this is kind of "hidden" or easy to miss - I wonder if having more visible delimiters for the layout could be worthwhile?

asllib/aslspec/AST.ml
Roman-Manevich: No, but the flexibility is necessary to avoid shift/reduce conflicts. The spec will have more checks to reject unwanted constructs.

asllib/aslspec/AST.ml
Roman-Manevich: Right. We need an escaped hatch.

Roman-Manevich: I actually removed this temporarily.

asllib/aslspec/SpecParser.mly
Roman-Manevich: This has been refactored away.

asllib/aslspec/AST.ml
Roman-Manevich: This has been refactored and simplified.

asllib/aslspec/SpecParser.mly
Roman-Manevich: Rules can have arbitrarily nested cases. When we expand them out, we require exactly one output, which should appear last. However, hat condition is hard to check during parser so we defer it to after the rule expansion ste in `spec.ml`.

asllib/aslspec/SpecParser.mly
Roman-Manevich: This has been refactored by now.

asllib/doc/asl.spec
Roman-Manevich: We do have more visible delimiters - you can use the full attribute notation: `math_layout = [_,_]`. I'm applying it now.

asllib/doc/ASLmacros.tex
```diff
@@ -275,12 +276,14 @@
 \renewcommand\triangleq[0]{\hyperlink{def-triangleq}{\OldTriangleq}}
 \newcommand\cartimes[0]{\hyperlink{def-cartimes}{\times}}
 
+\newcommand\equal[0]{=}
```
hrutvik, line 279: Perhaps use `\mathrel`?

asllib/doc/ASLmacros.tex
```diff
@@ -275,12 +276,14 @@
...
 \newcommand\eqname[0]{\hyperlink{def-deconstruction}{\stackrel{\mathsmaller{\mathsf{is}}}{=}}}
 \newcommand\eqdef[0]{\hyperlink{def-eqdef}{:=}}
 \newcommand\overname[2]{\overbrace{#1}^{#2}}
 \newcommand\overtext[2]{\overbracket{#1}^{#2}}
 \newcommand\emptyfunc[0]{\hyperlink{def-emptyfunc}{{\emptyset}_\lambda}}
 \newcommand\restrictfunc[2]{{#1}\hyperlink{def-restrictfunc}{|}_{#2}}
+\newcommand\COLON[0]{:}
```
hrutvik, line 286: Perhaps use `\mathbin`?

asllib/aslspec/AST.ml
```diff
@@ -364,6 +364,123 @@ end = struct
...
+    | Unresolved of expr
+
+  (** A term that can be used to form a rule judgment. *)
+  and expr =
+    | Var of string
+    | FieldAccess of string list
+        (** The first identifier is a variable and the rest are field names. *)
+    | ListIndex of { var : string; index : string }
+    | Record of { label : string; fields : (string * expr) list }
+        (** A record construction expression, optionally labelled. *)
```
hrutvik: Is the "optionally labelled" still accurate? It doesn't seem optional from the type alone at least

asllib/aslspec/latex.ml
```diff
@@ -199,15 +199,15 @@ let pp_elements_and_separator ~pp_sep ~alignment ~same_column pp_elem layout fmt
     separators are rendered in the same column as the elements, and the elements
     are aligned according to [alignment]. *)
 let pp_aligned_elements ~pp_sep ~alignment pp_elem layout fmt elements =
-  pp_elements_and_separator ~pp_sep ~alignment ~same_column:false pp_elem layout
+  pp_elements_and_separator ~pp_sep ~alignment ~same_column:true pp_elem layout
```
hrutvik, line 198: Just to check - why have these two `same_column` arguments flipped?

asllib/aslspec/main.ml
```diff
@@ -139,15 +144,20 @@ let parse_command_line_args_and_execute () =
...
-      | AST.SpecError msg -> ("Specification Error", msg)
-      | _ -> raise error
-    in
-    Format.eprintf "%s%s: %s%s\n" TextColor.red error_type msg
-      TextColor.reset_color;
-    exit 1
+  let config = CLI.parse_args () in
+  if !CLI.arg_debug then
+    (* Allow the exception stack trace to be printed for debugging. *)
+    execute config |> fun () -> exit 0
```
hrutvik: Could this be:
```suggestion
    execute config;
    exit 0
```
?

asllib/aslspec/main.ml
```diff
@@ -139,15 +144,20 @@ let parse_command_line_args_and_execute () =
...
+  else
+    try execute config |> fun () -> exit 0
```
hrutvik: Similarly

asllib/aslspec/PP.ml
```diff
@@ -100,6 +100,87 @@ let pp_variants_with_attributes fmt variants =
...
+  | Var name -> pp_print_string fmt name
+  | Application { applicator = ExprOperator op; args = [ lhs; rhs ] }
+    when is_infix_operator op ->
+      fprintf fmt "%a %s %a" pp_expr lhs
+        (expr_operator_to_string op)
+        pp_expr rhs
+  | Application { applicator; args } ->
+      fprintf fmt "%a(%a)" pp_application_lhs applicator (pp_comma_list pp_expr)
+        args
+  | FieldAccess path -> fprintf fmt "%s" (String.concat "." path)
```
hrutvik: nit: could this be the following?

```suggestion
  | FieldAccess path -> pp_print_string fmt (String.concat "." path)
```

asllib/aslspec/PP.ml
```diff
@@ -100,6 +100,87 @@ let pp_variants_with_attributes fmt variants =
...
+      fprintf fmt "%s(%s, %s: %a)" (tok_str INDEX) index list pp_expr body
+
+and pp_application_lhs fmt =
+  let open Rule in
+  function
+  | EmptyApplicator -> ()
+  | Relation name -> pp_print_string fmt name
+  | TupleLabel name -> pp_print_string fmt name
+  | ExprOperator op -> pp_print_string fmt (expr_operator_to_string op)
+  | Fields path -> fprintf fmt "%s" (String.concat "." path)
```
hrutvik: Similarly

asllib/aslspec/PP.ml
```diff
@@ -100,6 +100,87 @@ let pp_variants_with_attributes fmt variants =
...
+  | Unresolved expr -> pp_expr fmt expr
```
hrutvik, line 151: Do I understand correctly that `Unresolved` only appears before resolving/checking of rules? And then they are all removed?

asllib/aslspec/render.ml
```diff
@@ -387,12 +387,256 @@ module Make (S : SPEC_VALUE) = struct
...
+
+    (** Renders the field path [path] with [fmt]. *)
+    let pp_field_path fmt path =
+      let var, fields = (List.hd path, List.tl path) in
+      let pp_dot_field fmt field_name =
+        fprintf fmt ".%a" pp_field_name field_name
+      in
+      fprintf fmt "%a%a" pp_var var
+        (pp_print_list ~pp_sep:(fun _ _ -> ()) pp_dot_field)
+        fields
```
hrutvik: Is this the same as something like:
```suggestion
    let pp_field_path fmt path =
      let pp_sep fmt () = fprintf f "." in
      fprintf fmt "%a% (pp_print_list ~pp_sep pp_field_name) path
```
?

asllib/aslspec/render.ml
```diff
@@ -387,12 +387,256 @@ module Make (S : SPEC_VALUE) = struct
...
+      | Application { applicator; args } ->
+          let pp_lhs fmt lhs =
+            match lhs with
+            | EmptyApplicator ->
+                () (* No left-hand side so nothing to render. *)
+            | Relation name | TupleLabel name -> pp_id_as_macro fmt name
+            | Fields path -> pp_field_path fmt path
+            | ExprOperator _ | Unresolved _ -> assert false
+          in
+          fprintf fmt {|%a%a|} pp_lhs applicator
```
hrutvik: nit:
```suggestion
          fprintf fmt "%a%a" pp_lhs applicator
```
(similarly below)

asllib/aslspec/render.ml
```diff
@@ -387,12 +387,256 @@ module Make (S : SPEC_VALUE) = struct
...
+      | FieldAccess path -> pp_field_path fmt path
+      | Indexed { index; list; body } ->
+          let list_of_indices_macro_name = "listrange" in
+          let pp_indexed_lhs fmt ((index, list_var), _layout) =
+            fprintf fmt {|%a \in %a(%a)|} pp_var index pp_macro
+              list_of_indices_macro_name pp_var list_var
+          in
+          pp_connect_pair ~alignment:"ll" fmt pp_indexed_lhs (index, list)
+            "COLON" pp_expr body layout
+          (* fprintf fmt {|%a \;:\; %a|} pp_indexed_lhs (index, list) pp_expr
```
hrutvik: Should this still be here?

asllib/aslspec/render.ml
```diff
@@ -387,12 +387,256 @@ module Make (S : SPEC_VALUE) = struct
...
+            args
+      | Record { label; fields } ->
+          fprintf fmt {|%a%a|} pp_id_as_macro label
+            (pp_fields pp_field_name pp_expr)
+            (fields, layout)
+      | ListIndex { var; index } ->
+          fprintf fmt {|%a[%a]|} pp_var var pp_var index
+      | FieldAccess path -> pp_field_path fmt path
+      | Indexed { index; list; body } ->
+          let list_of_indices_macro_name = "listrange" in
```
hrutvik: I'm a little wary of spreading these macro names through the file (e.g. `COLON` below) - is it worth grouping all such names into a small module somewhere?

asllib/aslspec/render.ml
```diff
@@ -387,12 +387,256 @@ module Make (S : SPEC_VALUE) = struct
...
+
+    let pp_math_expanded_rule fmt { ExpandRules.name_opt; judgments } =
+      let premises, conclusion = Utils.split_last judgments in
+      let pp_premise fmt premise = pp_judgment fmt premise in
+      let pp_conclusion fmt conclusion = pp_judgment fmt conclusion in
+      fprintf fmt {|\begin{mathpar}@.\inferrule%a{%a}{@.%a@.}@.\end{mathpar}|}
+        pp_case_name_opt name_opt
+        (pp_print_list
+         (* The quadruple backslash means the next premise definitely starts on a new line. *)
+           ~pp_sep:(fun fmt () -> fprintf fmt "\\\\\\\\@.")
```
hrutvik:
```suggestion
           ~pp_sep:(fun fmt () -> fprintf fmt {|\\\\@.|})
```

asllib/aslspec/render.ml
```diff
@@ -387,12 +387,256 @@ module Make (S : SPEC_VALUE) = struct
...
+      let { Relation.rule_opt; category } =
+        Spec.relation_for_id S.spec relation_name
+      in
+      let rule = Option.get rule_opt in
+      let expanded_rules = ExpandRules.expand category rule in
+      let expanded_subset =
+        List.filter
+          (fun { ExpandRules.name_opt; _ } ->
+            match name_opt with
+            | None -> true
```
hrutvik, line 606: Why include unnamed rules here?

asllib/aslspec/SpecLexer.mll
```diff
@@ -49,16 +52,26 @@ rule token = parse
     | "prose_description"   { PROSE_DESCRIPTION }
     | "relation"            { RELATION }
     | "render"              { RENDER }
+    | "rule"                { RULE }
     | "semantics"           { SEMANTICS }
     | "short_circuit_macro" { SHORT_CIRCUIT_MACRO }
     | "typedef"             { TYPEDEF }
     | "typing"              { TYPING }
+    | "UNION"             { UNION }
```
hrutvik, line 60: nit: alignment

asllib/aslspec/SpecParser.mly
```diff
@@ -134,6 +159,7 @@ let elem :=
     | relation_definition
     | constant_definition
     | render_types
+    |render_rule
```
hrutvik: nit: 
```suggestion
    | render_rule
```

asllib/aslspec/SpecParser.mly
```diff
@@ -273,3 +299,90 @@ let render_types_attribute :=
...
+    | ~=prefix_expr_operator; args=plist0(expr);
+      { Rule.make_prefix_operator_application prefix_expr_operator args }
+    | lhs=expr; args=plist0(expr);
+      { Rule.make_application lhs args }
+    | lhs=expr; ~=infix_expr_operator; rhs=expr;
+      { Rule.make_infix_operator_application infix_expr_operator lhs rhs }
+    | ~=field_path;
+      { Rule.FieldAccess field_path }
+    | list=IDENTIFIER; LBRACKET; index=IDENTIFIER; RBRACKET;
+      { Rule.make_list_index list index }
```
hrutvik: nit: `list` makes syntax highlighting confused (similarly below)

asllib/aslspec/SpecParser.mly
```diff
@@ -273,3 +299,90 @@ let render_types_attribute :=
...
+    | name=IDENTIFIER; LBRACKET; fields=tclist1(field_and_value); RBRACKET;
+      { Rule.make_record name fields }
+    | lhs=expr; ARROW; rhs=expr; ~=short_circuit;
+      { Rule.Transition { lhs; rhs; short_circuit } }
+    | INDEX; LPAR; index=IDENTIFIER; COMMA; list=IDENTIFIER; COLON; body=expr; RPAR;
+      { Rule.Indexed { index; list; body } }
+
+(** We currently do not need field paths longer than a variable with two fields. *)
```
hrutvik, line 349: Is it worth generalising this to more than two fields from the start?

asllib/aslspec/spec.ml
```diff
@@ -496,6 +846,44 @@ module Check = struct
...
 
+  (** A module for checking that all rules are well-formed. That is, that they
+      end with an output judgment and that all preceding judgments are
+      non-output judgments. *)
+  module CheckRules = struct
+    let check_well_formed_expanded relation_name expanded_rule =
+      let open ExpandRules in
+      match List.rev expanded_rule.judgments with
+      | [] -> Error.empty_rule relation_name
+      | { expr = Transition _; is_output = true } :: prefix_fules ->
```
hrutvik: typo?
```suggestion
      | { expr = Transition _; is_output = true } :: prefix_rules ->
```

asllib/aslspec/spec.ml
```diff
@@ -291,6 +323,323 @@ type t = {
...
+        let resolved_fields =
+          List.map
+            (fun (field_name, field_expr) ->
+              (field_name, resolve_expr id_to_defining_node field_expr))
+            fields
+        in
+        Record { label; fields = resolved_fields }
+    | Application ({ applicator; args } as app) -> (
+        let resolved_args =
+          List.map (fun e -> resolve_expr id_to_defining_node e) args
```
hrutvik: nit:
```suggestion
          List.map (resolve_expr id_to_defining_node) args
```

asllib/aslspec/spec.ml
```diff
@@ -291,6 +323,323 @@ type t = {
...
+            match StringMap.find_opt id id_to_defining_node with
+            | Some (Node_Relation _) ->
+                Application { applicator = Relation id; args = resolved_args }
+            | Some (Node_TypeVariant _) ->
+                Application { applicator = TupleLabel id; args = resolved_args }
+            | Some (Node_Constant { Constant.name })
+            | Some (Node_Type { Type.name }) ->
+                Error.invalid_application_of_symbol_in_expr name expr
+            | _ ->
+                Application { applicator = Fields [ id ]; args = resolved_args }
```
hrutvik, line 362: Is this effectively a way to handle an `Unresolved` that has failed to resolve?

asllib/aslspec/spec.ml
```diff
@@ -291,6 +323,323 @@ type t = {
...
+        | Unresolved (FieldAccess path) ->
+            Application { applicator = Fields path; args = resolved_args }
+        | Unresolved _ ->
+            Error.invalid_application_of_expression_to_arguments expr)
+    | Transition { lhs; rhs; short_circuit } ->
+        let resolved_lhs = resolve_expr id_to_defining_node lhs in
+        let resolved_rhs = resolve_expr id_to_defining_node rhs in
+        let resolved_short_circuit =
+          Option.map
+            (List.map (fun e -> resolve_expr id_to_defining_node e))
```
hrutvik: nit:
```suggestion
            (List.map (resolve_expr id_to_defining_node))
```

asllib/aslspec/spec.ml
```diff
@@ -291,6 +323,323 @@ type t = {
...
+    match rule_element with
+    | Judgment ({ expr; is_output } as judgment) ->
+        let resolved_expr = resolve_expr id_to_defining_node expr in
+        let resolved_expr =
+          if is_output then
+            Transition
+              {
+                lhs = conclusion_lhs;
+                rhs = resolved_expr;
+                short_circuit = Some [];
```
hrutvik, line 398: So the use of `Some []` rather than `None` ensures that the alternatives are not automatically inserted?

asllib/aslspec/spec.ml
```diff
@@ -291,6 +323,323 @@ type t = {
...
+    (* Second, aggregate consecutive Cases elements. *)
+    List.fold_right
+      (fun rule_elem suffix_elements ->
+        match (rule_elem, suffix_elements) with
+        | Judgment _, _ | Rule.Cases _, Judgment _ :: _ ->
+            rule_elem :: suffix_elements
+        | Rule.Cases cases, [] -> [ Rule.Cases cases ]
+        | Rule.Cases cases, Cases next_cases :: tail ->
+            Rule.Cases (cases @ next_cases) :: tail)
+      rule_elements []
```
hrutvik: Is this more clear without the `fold_right`? I.e. something like:
```suggestion
    let aggregate = function
      | [] -> []
      | Rule.Cases cases1 :: Rule.Cases cases2 :: rest ->
          aggregate (Rule.Cases (cases1 @ cases2) :: rest)
      | rule_element :: rest -> rule_element :: aggregate rest
    in
    aggregate rule_elements
```

asllib/aslspec/spec.ml
```diff
@@ -291,6 +323,323 @@ type t = {
...
+                | None -> def
+                | Some elements ->
+                    let elements = aggregate_case_blocks elements in
+                    let conclusion_lhs = lhs_of_conclusion def in
+                    let resolved_elements =
+                      List.map
+                        (fun rule_elem ->
+                          resolve_rule_element id_to_defining_node
+                            conclusion_lhs rule_elem)
+                        elements
```
hrutvik, line 525: nit:
```suggestion
                      List.map
                        (resolve_rule_element id_to_defining_node
                            conclusion_lhs)
                        elements
```

asllib/aslspec/spec.ml
```diff
@@ -291,6 +323,323 @@ type t = {
...
+(** A module for expanding rules with cases into multiple rules without cases.
+*)
+module ExpandRules = struct
+  open Rule
+
+  type expanded_rule = {
+    name_opt : string option;
+        (** The optional name of the expanded rule, if it originated from a
+            case. *)
+    category_opt : Relation.relation_category option;
```
hrutvik: I'm a bit confused by `category_opt` - in the code below, it seems to be effectively "constant", and just passed around as-is. Does it need to be in this type, or could it be factored out?

asllib/doc/ASLmacros.tex
Roman-Manevich: Sure. I'll do it here though: https://github.com/herd/herdtools7/pull/1587

asllib/doc/ASLmacros.tex
Roman-Manevich: Sure, I'll do it here though: https://github.com/herd/herdtools7/pull/1587

asllib/aslspec/AST.ml
Roman-Manevich: right. fixed.

asllib/aslspec/latex.ml
Roman-Manevich: The `false` was a mistake, which I only found while working on this PR.

asllib/aslspec/main.ml
Roman-Manevich: I changed to
```
let () = execute config in
    exit 0
```

asllib/aslspec/PP.ml
Roman-Manevich: right

asllib/aslspec/render.ml
Roman-Manevich: No, the logic for rendering the first variable differs from the logic for rendering the fields. For example, in `name.name`, the first `name` is rendered plainly as `\texttt{name}`, whereas the second `name` is rendered by using the right macro for the field.

asllib/aslspec/render.ml
Roman-Manevich: removed

asllib/aslspec/render.ml
Roman-Manevich: Some simple relations are defined by a single rule, which does not require any `case`. This catches these cases. I'll add a comment.

asllib/aslspec/SpecLexer.mll
Roman-Manevich: Handled in https://github.com/herd/herdtools7/pull/1587.

asllib/aslspec/SpecParser.mly
Roman-Manevich: changed to `list_var`.

asllib/aslspec/SpecParser.mly
Roman-Manevich: It would most likely be an error. Let's see if we come across that.

asllib/aslspec/spec.ml
Roman-Manevich: yes

asllib/aslspec/spec.ml
Roman-Manevich: Right. I edited the comment to explain this.

asllib/aslspec/spec.ml
Roman-Manevich: hmm... I'm usually in favor of eta reductions when the function has just one argument. When it has more arguments, I find that it makes the code harder to read - I usually have to look the function up, which wastes time.
If you feel strongly about this, I'll fix it though.

asllib/aslspec/spec.ml
Roman-Manevich: Turns out it was a temporary hack that I forgot to remove.

asllib/aslspec/render.ml
```diff
@@ -124,18 +126,18 @@ module Make (S : SPEC_VALUE) = struct
...
+        fprintf fmt "%a%a" pp_id_opt_as_macro label_opt
           (pp_fields pp_field_name pp_record_field_as_pair)
           (pp_record_fields_as_pairs, layout)
     | ConstantsSet constant_names ->
         let layout_contains_vertical = LayoutUtils.contains_vertical layout in
-        fprintf fmt {|%a|}
+        fprintf fmt "%a"
           (pp_parenthesized Braces layout_contains_vertical
              (PP.pp_sep_list ~sep:", " pp_id_as_macro))
           constant_names
```
```suggestion
        pp_parenthesized Braces layout_contains_vertical
         (PP.pp_sep_list ~sep:", " pp_id_as_macro)
         fmt constant_names
```

asllib/aslspec/render.ml
```diff
@@ -145,7 +147,7 @@ module Make (S : SPEC_VALUE) = struct
       laid out according to [layout]. *)
   and pp_opt_named_type_term fmt ((name_opt, term), layout) =
     match name_opt with
-    | None -> fprintf fmt {|%a|} pp_type_term (term, layout)
+    | None -> fprintf fmt "%a" pp_type_term (term, layout)
```
```suggestion
    | None -> pp_type_term fmt (term, layout)
```

asllib/aslspec/render.ml
```diff
@@ -163,7 +165,7 @@ module Make (S : SPEC_VALUE) = struct
       in with the given [layout]. *)
   let pp_output_types fmt (terms, layout) =
     if Utils.is_singleton_list terms then
-      fprintf fmt {|%a|} pp_type_term (List.hd terms, layout)
+      fprintf fmt "%a" pp_type_term (List.hd terms, layout)
```
```suggestion
      pp_type_term fmt (List.hd terms, layout)
```

asllib/aslspec/render.ml
```diff
@@ -282,24 +284,24 @@ module Make (S : SPEC_VALUE) = struct
...
         (fun counter variant ->
           let () =
-            fprintf fmt {|@[<h>%s\ & %a@]|} join_symbol
+            fprintf fmt {|@[<h>%a\ & %a@]|} pp_macro join_symbol
               pp_variant_with_hypertarget variant
           in
           let _add_latex_line_break_except_on_last_line =
             if counter < num_variants_tail - 1 then fprintf fmt {|\\@.|}
-            else fprintf fmt {||}
+            else fprintf fmt ""
```
```suggestion
            if counter < num_variants_tail - 1 then fprintf fmt {|\\@.|}
```

asllib/aslspec/render.ml
HadrienRenaud:
```suggestion
    let pp_field_path fmt = function
    | [] -> assert false
    | var :: [] -> pp_var fmt var
    | var :: path ->
        let pp_sep fmt () = fprintf f "." in
        fprintf fmt "%a.%a%" pp_var var (pp_print_list ~pp_sep pp_field_name) path
```

asllib/aslspec/render.ml
```diff
@@ -387,12 +388,249 @@ module Make (S : SPEC_VALUE) = struct
...
+              short_circuit
+          in
+          pp_connect_pair ~alignment:"r" fmt pp_expr expr_lhs arrow_macro_name
+            pp_rhs_with_short_circuit rhs layout
+      | Transition _ ->
+          let msg =
+            Format.asprintf "Unexpected LHS in transition judgment: %a"
+              PP.pp_expr expr
+          in
+          raise (SpecError msg)
```
```suggestion
      | Transition { lhs; rhs; short_circuit } -> (
          let relation_name =
            match lhs with
            | Application { applicator = Relation relation_name } -> relation_name
            | _ ->
                let msg =
                  Format.asprintf "Unexpected LHS in transition judgment: %a"
                    PP.pp_expr expr
                in
                raise (SpecError msg)
          in
          let { Relation.category } =
            Spec.relation_for_id S.spec relation_name
          in
          let arrow_macro_name = arrow_macro_name_for_category_opt category in
          let pp_rhs_with_short_circuit fmt rhs_with_layout =
            fprintf fmt "%a%a" pp_expr rhs_with_layout
              (pp_short_circuit relation_name)
              short_circuit
          in
          pp_connect_pair ~alignment:"r" fmt pp_expr lhs arrow_macro_name
            pp_rhs_with_short_circuit rhs layout
```

asllib/aslspec/render.ml
```diff
@@ -387,12 +388,249 @@ module Make (S : SPEC_VALUE) = struct
...
+          in
+          match alternative_macros with
+          | [] -> ()
+          | _ ->
+              fprintf fmt {|\;%a\;%a|} pp_macro terminateas_macro_name
+                (PP.pp_sep_list ~sep:", " pp_print_string)
+                alternative_macros)
+      | Some alternatives ->
+          if Utils.list_is_empty alternatives then ()
+          else
```
```suggestion
      | Some [] -> ()
      | Some alternatives ->
```

HadrienRenaud (approved):
This looks good. I've left a couple of comments in #1586 that are probably more relevant on here apologies.

PR #1487, opened 2025-10-09
[asl] Clean-up: remove empty accessors

HadrienRenaud:
After removal of empty accessors in ASL v1, there is a problem with interoperability between ASLv0 and ASLv1: as there is no way of writing an empty-caller in ASLv1, there is no way to express a v1 callee that would match a v0 call site.
We thus suggest removing the difference between empty and non-empty accessors in ASLv0. Note that nothing would prevent a ASLv0 user to maintain the discipline of differentiating between empty and non-empty accessors.

---

**Documentation changes**

The new version should be equivalent to the previous one, just a slight refacto to match the version in the code. New documentation:

Screenshot 2025-10-11 at 12 20 52
Screenshot 2025-10-11 at 12 20 39
Screenshot 2025-10-11 at 12 20 03

---

Note: there is a slight change in `herd/libdir/asl-pseudocode/implementations.asl` that removes a declaration for an accessor for `SCTLR_EL1`. It is important to note that there is also a ASLv1 variable called `SCTLR_EL1` in the same file. 
Previously, the accessor was not used, because it is a non-empty ASLv1 accessor that couldn't get called from the shared pseudocode in ASLv0. With the change in this PR, the ASLv0 code would actually call it instead of using the correctly instantiated variable.

### DIFF
--- a/asllib/doc/SubprogramCalls.tex
+++ b/asllib/doc/SubprogramCalls.tex
@@ -1502,12 +1502,6 @@ The table below shows keywords used to define functions, and their possible comp
 \ProseParagraph
 \OneApplies
 \begin{itemize}
-  \item \AllApplyCase{equal}
-  \begin{itemize}
-    \item $\funcdef.\funcsubprogramtype$ is equal to $\calltype$;
-    \item $\vb$ is $\True$.
-  \end{itemize}
-
   \item \AllApplyCase{getter\_func}
   \begin{itemize}
     \item $\funcdef.\funcsubprogramtype$ is $\STGetter$;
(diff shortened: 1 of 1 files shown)


asllib/Typing.ml
```diff
@@ -297,16 +292,11 @@ module FunctionRenaming (C : ANNOTATE_CONFIG) = struct
 
   (* Begin CallTypeMatches *)
   let call_type_matches func call_type =
-    func.subprogram_type = call_type
-    ||
-    match (func_version func, func.subprogram_type, call_type) with
+    match (func.subprogram_type, call_type) with
```
Roman-Manevich, line 295: Please reflect this in the inference rules for`TypingRule.CallTypeMatches`

herd/libdir/asl-pseudocode/implementations.asl
```diff
@@ -421,19 +421,6 @@ end;
 
 // =============================================================================
 
-accessor SCTLR_EL1() <=> v: SCTLRType
```
Roman-Manevich, line 424: Why is this not needed now?

HadrienRenaud: Answered in the PR description!

asllib/Typing.ml
HadrienRenaud: Done thanks, see PR description

PR #1495, opened 2025-10-11
[aslspec] added a check that all type terms are well-formed

Roman-Manevich:
Aslspec allows defining types and also constructing types out of those definitions. This check ensures that every constructed type is subsumed in the corresponding defined type constructor. The check is conservative - if the check cannot prove subsumption the result is an error.

The tests `instantiation_xxx` test various cases where instantiation should fail.

### DIFF
--- a/asllib/aslspec/spec.ml
+++ b/asllib/aslspec/spec.ml
@@ -448,9 +448,41 @@ module Check = struct
       List.iter check_prose_template_for_definition_node defining_nodes
   end
 
-  (** * A module for conservatively checking that all type terms are
-      well-formed. That is each type term that instantiates another defined type
-      term is subsumed by it. *)
+  (** * A module for conservatively checking that all type terms are well-formed
+      and that all type instantiations are valid. That is each type term that
+      instantiates another defined type term is subsumed by it.
+
+      For example, given the following definitions:
+      {[
+        typedef Z;
+        typedef Bool = True | False;
+        ast literal = L_Int(whole_number: Z) | L_Bool(value: Bool);
+        typedef native_value = NV_Literal(l: literal);
+        typedef tint = ( NV_Literal(L_Int(v: Z)) );
+      ]}
+      The type [native_value] defines the labelled tuple type [L_Int] with a
+      single component of type [literal].
+
+      The type term for [tint] instantiates the type term for
+      [NV_Literal(L_Int(v: Z))] rather than defining it, which can be seen since
+      it parenthesized. The check needs to ensure that [NV_Literal(L_Int(v: Z))]
+      is valid.
+
+      This is done by: 1. checking that [NV_Literal(L_Int(v: Z))] is well-formed
+      with respect to [NV_Literal(l: literal)], and
+
+      2. checking that all values of [NV_Literal(L_Int(v: Z))] are also value of
+      [NV_Literal(l: literal)]. In other words, [NV_Literal(L_Int(v: Z))] is
+      subsumed by [NV_Literal(l: literal)].
+
+      To check (1), we verify that the labelled tuple [NV_Literal(L_Int(v: Z))]
+      has a single component.
+
+      To check (2), we use structural induction, which means that we need to
+      verify that [L_Int(v: Z)] is subsumed by [l: literal]. This is done by
+      checking that [L_Int(v: Z)] is subsumed by at least one of the type
+      variants of [literal], which is indeed the case since [L_Int(v: Z)] is a
+      type variant of [literal]. *)
   module CheckTypeInstantiations : sig
     val check : definition_node StringMap.t -> elem list -> unit
     (** [check id_to_defining_node elems] conservatively checks that all type
@@ -479,9 +511,28 @@ module Check = struct
         of [super]. Labels that represent types may be expanded, that is,
         replaced by their list of type variants, using [id_to_defining_node]. To
         ensure termination on recursive types, this expansion is done at most
-        once by tracking the set of expanded labels in [expanded_types]. *)
+        once by tracking the set of expanded labels in [expanded_types].
+
+        In the example, [M(B, Num)] is not subsumed by [M(A, Num)]. The
+        algorithm checks whether [B] is subsumed by [A], which requires checking
+        whether [B] is subsumed by [L] and whether [B] is subsumed by
+        [M(A, Num)]. The first check fails, but the second check requires
+        expanding [B] to check whether all of its variants are subsumed by
+        [M(A, Num)], namely whether [M(B, Num)] is subsumed by [M(A, Num)],
+        which is the original subsumption test.
+
+        To avoid infinite recursion, the algorithm tracks which types have
+        already been expanded, and do not expand it again when checking
+        subsumption for [B]. Thus, [B] is not expanded again, and the
+        subsumption test returns [false].
+        {[
+          typedef;
+          typedef A = L | M(A, Num);
+          typedef B = ( M(B, Num) );
+        ]} *)
     let rec subsumed id_to_defining_node expanded_types sub super =
-      (* The domain of an unlabelled singleton tuple is the domain of its single type term. *)
+      (* The domain of an unlabelled singleton tuple is the domain
+         of its single type term. *)
       let equiv_singleton_tuple term =
         match term with
         | LabelledTuple
@@ -489,6 +540,7 @@ module Check = struct
             referenced_term
         | _ -> term
       in
+      (* In the example above [( M(B, Num) )] is equivalent to [M(B, Num)]. *)
       let sub = equiv_singleton_tuple sub in
       let super = equiv_singleton_tuple super in
       match (sub, super) with
@@ -498,15 +550,19 @@ module Check = struct
             | Label sub_label ->
                 String.equal sub_label super_label
                 ||
-                (* The case where [sub_label] is a type name. *)
+                (* The case where [sub_label] is a type name,
+                  like [B] of [M(B, Num)] in the example. *)
                 subsumed_typename_term_type id_to_defining_node expanded_types
                   sub_label super
             | _ -> false
           in
           sub_is_label_case
-          (* The case where [super_label] is a type name. *)
+          (* The case where [super_label] is a type name,
+              like [A] of [M(A, Num)] in the example.
+          *)
           || subsumed_term_type_typename id_to_defining_node expanded_types sub
                super_label
+      (* From here on the test operates via structural induction. *)
       | ( Operator { op = sub_op; term = _, sub_term },
           Operator { op = super_op; term = _, super_term } ) ->
           operator_subsumed sub_op super_op
@@ -531,6 +587,10 @@ module Check = struct
       | ( Function { from_type = _, sub_from_term; to_type = _, sub_to_term },
           Function
             { from_type = _, super_from_term; to_type = _, super_to_term } ) ->
+          (* Functions can be partial or total, which require different subsumption tests.
+             To make this simple, we require equivalence of the from-terms and to-terms,
+             which is sufficient for our needs.
+          *)
           let equivalence_test term term' =
             subsumed id_to_defining_node expanded_types term term'
             && subsumed id_to_defining_node expanded_types term' term
@@ -555,6 +615,11 @@ module Check = struct
         let expanded_types = StringSet.add typename expanded_types in
         match StringMap.find_opt typename id_to_defining_node with
         | Some (Node_Type { Type.variants; _ }) ->
+            (* [sub_term] is subsumed by the type [typename] if it is subsumed by at least
+              one of its variants.
+              For example [L_Int(v: Z)] is subsumed by [literal] since it is subsumed by the
+              variant [L_Int(whole_number: Z)] of [literal].
+            *)
             (not (Utils.list_is_empty variants))
             && List.exists
                  (fun { TypeVariant.term = super_term } ->
@@ -589,8 +654,7 @@ module Check = struct
         labelled tuple/labelled record [term] with label [label]. If the lists
         have different lengths, or if any term in [sub_terms] is not subsumed by
         the corresponding term in [super_terms], a [SpecError] is raised. *)
-    let check_subsumed_terms_lists id_to_defining_node term label sub_terms
-        super_terms =
+    let check_subsumed_terms_lists id_to_defining_node sub_terms super_terms =
       let check_subsumed id_to_defining_node sub super =
         if subsumed id_to_defining_node StringSet.empty sub super then ()
         else
@@ -600,20 +664,17 @@ module Check = struct
           in
           raise (SpecError msg)
       in
-      if List.compare_lengths sub_terms super_terms = 0 then
-        List.iter2 (check_subsumed id_to_defining_node) sub_terms super_terms
-      else
-        let msg =
-          Format.asprintf
-            "The type term `%a` cannot be instantiated since it has %i type \
-             terms and `%s` requires %i type terms"
-            PP.pp_type_term term (List.length sub_terms) label
-            (List.length super_terms)
-        in
-        raise (SpecError msg)
+      List.iter2 (check_subsumed id_to_defining_node) sub_terms super_terms
 
     (** [is_constant id_to_defining_node id] checks if [id] is either defined as
-        a constant directly or as a type variant with a label. *)
+        a constant directly or as a type variant with a label.
+
+        For example:
+        {[
+          typedef A = L;
+          constant C
+        ]}
+        defines the constant [L] and the constant [C]. *)
     let check_is_constant id_to_defining_node id =
       match StringMap.find_opt id id_to_defining_node with
       | Some (Node_Constant _) | Some (Node_TypeVariant { term = Label _ }) ->
@@ -625,16 +686,24 @@ module Check = struct
           in
           raise (SpecError msg)
 
-    (** [check_type_term id_to_defining_node term] checks that [term] is
-        well-formed with respect to the type definitions in the range of
-        [id_to_defining_node]. *)
-    let rec check_type_term id_to_defining_node term =
+    (** [check_well_typed id_to_defining_node term] checks that every type
+        referenced by [term] correctly instantiates its defining type with
+        respect to the type definitions in the range of [id_to_defining_node].
+        The check assumes that both [check_no_undefined_ids] and that
+        [check_well_formed] have already been run.
+
+        The check operates by structural induction on [term], except when a
+        label term references another type in which case the type variants of
+        the referenced type are considered. *)
+    let rec check_well_instantiated id_to_defining_node term =
       match term with
       | Operator { term = _, operator_term } ->
-          check_type_term id_to_defining_node operator_term
+          check_well_instantiated id_to_defining_node operator_term
       | LabelledTuple { label_opt; components } -> (
           let terms = List.map snd components in
-          let () = List.iter (check_type_term id_to_defining_node) terms in
+          let () =
+            List.iter (check_well_instantiated id_to_defining_node) terms
+          in
           match label_opt with
           | None -> ()
           | Some label -> (
@@ -646,19 +715,13 @@ module Check = struct
                       LabelledTuple { components = def_opt_named_components };
                   } ->
                   let def_terms = List.map snd def_opt_named_components in
-                  check_subsumed_terms_lists id_to_defining_node term label
-                    terms def_terms
-              | _ ->
-                  let msg =
-                    Format.asprintf
-                      "The type term `%a` cannot be instantiated since '%s' is \
-                       not a labelled tuple type"
-                      PP.pp_type_term term label
-                  in
-                  raise (SpecError msg)))
+                  check_subsumed_terms_lists id_to_defining_node terms def_terms
+              | _ -> assert false))
       | LabelledRecord { label_opt; fields } -> (
           let terms = List.map snd fields in
-          let () = List.iter (check_type_term id_to_defining_node) terms in
+          let () =
+            List.iter (check_well_instantiated id_to_defining_node) terms
+          in
           match label_opt with
           | None -> ()
           | Some label -> (
@@ -669,19 +732,11 @@ module Check = struct
                     TypeVariant.term = LabelledRecord { fields = def_fields; _ };
                   } ->
                   let def_terms = List.map snd def_fields in
-                  check_subsumed_terms_lists id_to_defining_node term label
-                    terms def_terms
-              | _ ->
-                  let msg =
-                    Format.asprintf
-                      "The type term `%a` cannot be instantiated since '%s' is \
-                       not a labelled record type"
-                      PP.pp_type_term term label
-                  in
-                  raise (SpecError msg)))
+                  check_subsumed_terms_lists id_to_defining_node terms def_terms
+              | _ -> assert false))
       | Function { from_type = _, from_term; to_type = _, to_term } ->
-          check_type_term id_to_defining_node from_term;
-          check_type_term id_to_defining_node to_term
+          check_well_instantiated id_to_defining_node from_term;
+          check_well_instantiated id_to_defining_node to_term
       | ConstantsSet labels ->
           List.iter (check_is_constant id_to_defining_node) labels
       | Label label -> (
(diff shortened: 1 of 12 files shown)


asllib/aslspec/spec.ml
```diff
@@ -448,16 +448,221 @@ module Check = struct
...
+    let operator_subsumed sub_op super_op =
+      match (sub_op, super_op) with
+      | Powerset, Powerset
+      | Powerset_Finite, Powerset_Finite
+      | Powerset_Finite, Powerset
+      | List1, List1
+      | List0, List0
+      | List1, List0
+      | Option, Option
+      | Option, Powerset_Finite
```
HadrienRenaud: Is this because option is a shortcut for the singleton sets? If so can we put a comment to justify this here?

asllib/aslspec/spec.ml
```diff
@@ -448,16 +448,221 @@ module Check = struct
...
+        ) ->
+          Option.equal String.equal sub_label_opt super_label_opt
+          && List.for_all2
+               (fun (_, sub_term) (_, super_term) ->
+                 subsumed id_to_defining_node sub_term super_term)
+               sub_fields super_fields
+      | ( Function { from_type = _, sub_from_term; to_type = _, sub_to_term },
+          Function
+            { from_type = _, super_from_term; to_type = _, super_to_term } ) ->
+          subsumed id_to_defining_node sub_from_term super_from_term
```
HadrienRenaud: Isn't it the opposite for function arguments?
Like you want the function to accept more arguments than what was in the contract?

asllib/aslspec/spec.ml
```diff
@@ -448,16 +448,221 @@ module Check = struct
...
+          operator_subsumed sub_op super_op
+          && subsumed id_to_defining_node sub_term super_term
+      | ( LabelledTuple
+            { label_opt = sub_label_opt; components = sub_components },
+          LabelledTuple
+            { label_opt = super_label_opt; components = super_components } ) ->
+          Option.equal String.equal sub_label_opt super_label_opt
+          && List.for_all2
+               (fun (_, sub_term) (_, super_term) ->
+                 subsumed id_to_defining_node sub_term super_term)
```
HadrienRenaud: Why do we ignore the labels of the components?

asllib/aslspec/spec.ml
```diff
@@ -448,16 +448,221 @@ module Check = struct
...
+      | ( Function { from_type = _, sub_from_term; to_type = _, sub_to_term },
+          Function
+            { from_type = _, super_from_term; to_type = _, super_to_term } ) ->
+          subsumed id_to_defining_node sub_from_term super_from_term
+          && subsumed id_to_defining_node sub_to_term super_to_term
+      | ConstantsSet sub_names, ConstantsSet super_names ->
+          List.for_all (fun name -> List.mem name super_names) sub_names
+      | _, Label super_label ->
+          (* The case where [super_label] references a type. *)
+          subsumed_term_type id_to_defining_node sub super_label
```
HadrienRenaud: Could we move this case next to the other `Label` case?

asllib/aslspec/spec.ml
```diff
@@ -448,16 +448,221 @@ module Check = struct
...
+      match StringMap.find_opt typename id_to_defining_node with
+      | None -> false
+      | Some defining_node -> (
+          match defining_node with
+          | Node_Type { Type.variants; _ } ->
+              List.exists
+                (fun { TypeVariant.term = super_term } ->
+                  subsumed id_to_defining_node sub_term super_term)
+                variants
+          | _ -> false)
```
HadrienRenaud: Why not merge the 2 `match`?

```suggestion
      match StringMap.find_opt typename id_to_defining_node with
      | Some (Node_Type { Type.variants; _ }) ->
          List.exists
            (fun { TypeVariant.term = super_term } ->
              subsumed id_to_defining_node sub_term super_term)
            variants
      | _ -> false
```

asllib/aslspec/spec.ml
```diff
@@ -448,16 +448,221 @@ module Check = struct
...
+                  in
+                  raise (SpecError msg)))
+      | LabelledRecord { label_opt; fields } -> (
+          let terms = List.map snd fields in
+          let () = List.iter (check_type_term id_to_defining_node) terms in
+          match label_opt with
+          | None ->
+              List.iter
+                (fun (_, term) -> check_type_term id_to_defining_node term)
+                fields
```
HadrienRenaud: Hasn't this been done just a few lines above?

asllib/aslspec/spec.ml
```diff
@@ -448,16 +448,221 @@ module Check = struct
...
+                    terms def_terms
+              | _ ->
+                  let msg =
+                    Format.asprintf
+                      "The type term `%a` cannot be instantiated since '%s' is \
+                       not a labelled tuple type"
+                      PP.pp_type_term term label
+                  in
+                  raise (SpecError msg)))
+      | LabelledRecord { label_opt; fields } -> (
```
HadrienRenaud, line 803: Is the difference between `LabelledRecord` and `LabelledTuple` just the recursive case with the label? If so wouldn't there be a way to have those 2 cases in one case?

asllib/aslspec/spec.ml
```diff
@@ -448,16 +448,221 @@ module Check = struct
...
+      | Operator { term = _, operator_term } ->
+          check_type_term id_to_defining_node operator_term
+      | LabelledTuple { label_opt; components } -> (
+          let terms = List.map snd components in
+          let () = List.iter (check_type_term id_to_defining_node) terms in
+          match label_opt with
+          | None ->
+              List.iter
+                (fun (_, term) -> check_type_term id_to_defining_node term)
+                components
```
HadrienRenaud: Same as 2 lines above?

asllib/aslspec/spec.ml
```diff
@@ -448,16 +448,221 @@ module Check = struct
...
+                       not a labelled record type"
+                      PP.pp_type_term term label
+                  in
+                  raise (SpecError msg)))
+      | Function { from_type = _, from_term; to_type = _, to_term } ->
+          check_type_term id_to_defining_node from_term;
+          check_type_term id_to_defining_node to_term
+      | ConstantsSet labels ->
+          List.iter (check_is_constant id_to_defining_node) labels
+      | Label _ -> ()
```
HadrienRenaud: You don't need to check that in `Label l`, `l` is well defined earlier?

asllib/aslspec/spec.ml
Roman-Manevich: This is true in type theory where types have the concept of subtyping and substitutability, but here functions are simply relations over values with no concept of subtyping.

asllib/aslspec/spec.ml
Roman-Manevich: The labels do not affect the domain of a type. They are only used as a way of referencing components in prose and in rules.

asllib/aslspec/spec.ml
Roman-Manevich: I already factored out the common stuff into `check_subsumed_terms_lists id_to_defining_node term label`. I don't see how to naturally combine the cases. The differences are: the term matched for `variant_def` is either `LabelledTuple` or `LabelledRecord`, the `def_terms` have to be pulled out differently, and the error message is a bit different.

asllib/aslspec/spec.ml
Roman-Manevich: Right. Done.

asllib/aslspec/spec.ml
```diff
@@ -448,16 +448,225 @@ module Check = struct
...
+    (** [check_subsumed id_to_defining_node sub super] checks that [sub] is
+        subsumed by [super]. If not, a [SpecError] is raised. *)
+    let check_subsumed id_to_defining_node sub super =
+      if not (subsumed id_to_defining_node sub super) then
+        let msg =
+          Format.asprintf "The type term `%a` is not subsumed by `%a`"
+            PP.pp_type_term sub PP.pp_type_term super
+        in
+        raise (SpecError msg)
+      else ()
```
HadrienRenaud: nit:

```suggestion
      if subsumed id_to_defining_node sub super then ()
      else
        let msg =
          Format.asprintf "The type term `%a` is not subsumed by `%a`"
            PP.pp_type_term sub PP.pp_type_term super
        in
        raise (SpecError msg)
```

asllib/aslspec/spec.ml
```diff
@@ -448,16 +448,225 @@ module Check = struct
...
+      if List.compare_lengths sub_terms super_terms <> 0 then
+        let msg =
+          Format.asprintf
+            "The type term `%a` cannot be instantiated since it has %i type \
+             terms and `%s` requires %i type terms"
+            PP.pp_type_term term (List.length sub_terms) label
+            (List.length super_terms)
+        in
+        raise (SpecError msg)
+      else List.iter2 (check_subsumed id_to_defining_node) sub_terms super_terms
```
HadrienRenaud: nit:

```suggestion
      if List.compare_lengths sub_terms super_terms = 0 then
        List.iter2 (check_subsumed id_to_defining_node) sub_terms super_terms
      else
        let msg =
          Format.asprintf
            "The type term `%a` cannot be instantiated since it has %i type \
             terms and `%s` requires %i type terms"
            PP.pp_type_term term (List.length sub_terms) label
            (List.length super_terms)
        in
        raise (SpecError msg)
```

asllib/aslspec/spec.ml
```diff
@@ -448,16 +448,225 @@ module Check = struct
...
+      (* Option represents sets with 0 or 1 values while Powerset represents sets of _any_ number of values. *)
+        ->
+          true
+      | _ -> false
+
+    (** [subsumed id_to_defining_node sub super] conservatively tests whether
+        all values in the domain of [sub] are also in the domain of [super]. *)
+    let rec subsumed id_to_defining_node sub super =
+      match (sub, super) with
+      | Label sub_label, Label super_label -> sub_label = super_label
```
HadrienRenaud: Can we avoid using polymorphic equality unless types are obviously simple?

```suggestion
      | Label sub_label, Label super_label -> String.equal sub_label super_label
```

asllib/aslspec/spec.ml
HadrienRenaud: I'm not sure I understand your comment:

```
typedef O;
typedef O2;
typedef P = O | O2;
typedef A = L(O -> O);
typedef B = L(P -> O);
```

Here you would instantiate `B` with a badly typed function argument

asllib/aslspec/tests.t/labelled_tuple_instantiation.bad
```diff
@@ -0,0 +1,8 @@
+typedef Int { prose_description = "integer" };
+
+typedef A { prose_description = "A"} =
+//    | (A(Int)) { prose_description = "Bad instantiation"}
```
hrutvik: nit: why is this comment here?

asllib/aslspec/tests.t/typedefs.spec
```diff
@@ -31,7 +31,7 @@ typedef B { prose_description = "B"} =
     { prose_description = "a labelled ordered pair of Int values" }
     | [f: Int, g: fun (Int, Int) -> powerset(Int)]
     { prose_description = "a record with two fields" }
-    | LRec[f: Rec(Int, Int, Int)]
+    | LRec[f: Rec(Int, Int), g: fun (Int, Int) -> powerset(Int)]
```
hrutvik, line 34: nit: why has this test changed?

asllib/aslspec/spec.ml
```diff
@@ -448,16 +448,225 @@ module Check = struct
...
+      match (sub_op, super_op) with
+      | Powerset, Powerset
+      | Powerset_Finite, Powerset_Finite
+      | Powerset_Finite, Powerset
+      | List1, List1
+      | List0, List0
+      | List1, List0
+      | Option, Option
+      | Option, Powerset_Finite
+      | Option, Powerset
```
hrutvik: nit: purely stylistic, but I find it more readable/intuitive to write as follows
```suggestion
      | Powerset, Powerset
      | Powerset_Finite, (Powerset_Finite | Powerset)
      | List1, (List1 | List0)
      | List0, List0
      | Option, (Option | Powerset_Finite | Powerset)
```

asllib/aslspec/spec.ml
```diff
@@ -448,16 +448,225 @@ module Check = struct
...
+              subsumed id_to_defining_node sub_term super_term)
+            variants
+      | _ -> false
+
+    (** [check_subsumed id_to_defining_node sub super] checks that [sub] is
+        subsumed by [super]. If not, a [SpecError] is raised. *)
+    let check_subsumed id_to_defining_node sub super =
+      if not (subsumed id_to_defining_node sub super) then
+        let msg =
+          Format.asprintf "The type term `%a` is not subsumed by `%a`"
```
hrutvik: Suggested to add a test showcasing this

asllib/aslspec/spec.ml
```diff
@@ -448,16 +448,225 @@ module Check = struct
...
+    (** [is_constant id_to_defining_node id] checks if [id] is either defined as
+        a constant directly or as a type variant with a label. *)
+    let check_is_constant id_to_defining_node id =
+      match StringMap.find_opt id id_to_defining_node with
+      | Some (Node_Constant _) | Some (Node_TypeVariant { term = Label _ }) ->
+          ()
+      | _ ->
+          let msg =
+            Format.sprintf
+              "%s is used as a constant even though it is not defined as one" id
```
hrutvik, line 685: Suggested to add a test showcasing this

asllib/aslspec/spec.ml
```diff
@@ -448,16 +448,225 @@ module Check = struct
...
+              | Node_TypeVariant
+                  {
+                    TypeVariant.term = LabelledRecord { fields = def_fields; _ };
+                  } ->
+                  let def_terms = List.map snd def_fields in
+                  check_subsumed_terms_lists id_to_defining_node term label
+                    terms def_terms
+              | _ ->
+                  let msg =
+                    Format.asprintf
```
hrutvik, line 833: Suggested to add a test showcasing this

asllib/aslspec/spec.ml
```diff
@@ -448,16 +448,225 @@ module Check = struct
...
+      | ConstantsSet labels ->
+          List.iter (check_is_constant id_to_defining_node) labels
+      | Label label -> (
+          let variant_def = StringMap.find label id_to_defining_node in
+          match variant_def with
+          | Node_TypeVariant _ | Node_Type _ -> ()
+          | _ ->
+              let msg =
+                Format.asprintf
+                  "The type term `%a` cannot be instantiated since '%s' is not \
```
hrutvik, line 854: Suggested to add a test showcasing this

asllib/aslspec/spec.ml
Roman-Manevich: No, there would be an error, since `P` is not subsumed by `O` (it is vice versa).
I'm adding a test to show this (including a bug fix).

HadrienRenaud: Can you also document the opposite case, for example would this also be accepted?

```
typedef O;
typedef O2;
typedef P = O | O2;
typedef A = L(P -> O);
typedef B = L(O -> O);
```

HadrienRenaud: Sorry I got confused in my previous example, I think this one is the one I wanted to ask about

Roman-Manevich: Right. I made the test stricter by requiring equivalence on both domain and range. This works both for partial and total functions and is probably a better way of catching bugs in a spec.

asllib/aslspec/tests.t/labelled_tuple_instantiation.bad
Roman-Manevich: A leftover from experiments. I removed it.

asllib/aslspec/tests.t/typedefs.spec
Roman-Manevich: It was a bug found by the subsumption test - `Rec` only takes two arguments, not three.
The `g: fun (Int, Int) -> powerset(Int)` is just to add another field to match the existing description, but not strictly needed.

asllib/aslspec/spec.ml
HadrienRenaud: Sounds good - we can always relax later!

asllib/aslspec/tests.t/instantiation_expansion.spec
hrutvik, line 1: What is the intent of this test? (It does not produce an error in `run.t`)

asllib/aslspec/spec.ml
```diff
@@ -448,16 +448,284 @@ module Check = struct
...
+          Option.equal String.equal sub_label_opt super_label_opt
+          && List.for_all2
+               (fun (_, sub_term) (_, super_term) ->
+                 subsumed id_to_defining_node expanded_types sub_term super_term)
+               sub_components super_components
+      | ( LabelledRecord { label_opt = sub_label_opt; fields = sub_fields },
+          LabelledRecord { label_opt = super_label_opt; fields = super_fields }
+        ) ->
+          Option.equal String.equal sub_label_opt super_label_opt
+          && List.for_all2
```
HadrienRenaud, line 583: Won't this raise a `Invalid_argument` if the lists are not of the same size? Similarly for `LabelledTuple` above

asllib/aslspec/spec.ml
```diff
@@ -448,16 +448,284 @@ module Check = struct
...
+               (fun (_, sub_term) (_, super_term) ->
+                 subsumed id_to_defining_node expanded_types sub_term super_term)
+               sub_fields super_fields
+      | ( Function { from_type = _, sub_from_term; to_type = _, sub_to_term },
+          Function
+            { from_type = _, super_from_term; to_type = _, super_to_term } ) ->
+          let equivalence_test term term' =
+            subsumed id_to_defining_node expanded_types term term'
+            && subsumed id_to_defining_node expanded_types term' term
+          in
```
HadrienRenaud, line 597: nit: Maybe this should be moved to be its own co-recursive function with `subsumed`?

asllib/aslspec/spec.ml
```diff
@@ -448,16 +448,284 @@ module Check = struct
...
+                ||
+                (* The case where [sub_label] is a type name. *)
+                subsumed_typename_term_type id_to_defining_node expanded_types
+                  sub_label super
+            | _ -> false
+          in
+          sub_is_label_case
+          (* The case where [super_label] is a type name. *)
+          || subsumed_term_type_typename id_to_defining_node expanded_types sub
+               super_label
```
HadrienRenaud: I find this whole thing of going down into the list of labels quite difficult to understand and quite error prone. Why not just going to the "structure" (to take an ASL name) and looking at this? If the labels are equal then the structure is going to be equal

asllib/aslspec/spec.ml
```diff
@@ -448,16 +448,284 @@ module Check = struct
...
+
+    (** [subsumed_term_type_typename id_to_defining_node sub_term typename]
+        checks if [sub_term] is subsumed by any of the type variants defined by
+        [typename] in [id_to_defining_node]. If [typename] is not a type with
+        type variants, returns false. *)
+    and subsumed_term_type_typename id_to_defining_node expanded_types sub_term
+        typename =
+      if StringSet.mem typename expanded_types then false
+      else
+        (* Prevent infinite recursion on recursive types by expanding a type name at most once. *)
```
HadrienRenaud, line 614: Could you give an example of infinite recursion that could happen please? I'm a bit struggling to picture what this would look like

asllib/aslspec/spec.ml
HadrienRenaud: Could we document a case where recursing on super (resp sub) before sub (resp super) makes type-checking fail?

asllib/aslspec/spec.ml
```diff
@@ -448,16 +448,284 @@ module Check = struct
...
+      | _ -> false
+
+    (** [subsumed id_to_defining_node expanded_types sub super] conservatively
+        tests whether all values in the domain of [sub] are also in the domain
+        of [super]. Labels that represent types may be expanded, that is,
+        replaced by their list of type variants, using [id_to_defining_node]. To
+        ensure termination on recursive types, this expansion is done at most
+        once by tracking the set of expanded labels in [expanded_types]. *)
+    let rec subsumed id_to_defining_node expanded_types sub super =
+      (* The domain of an unlabelled singleton tuple is the domain of its single type term. *)
```
HadrienRenaud: Why are unlabelled singleton tuple existing? Could we remove them?

asllib/aslspec/spec.ml
Roman-Manevich: RIght. I created `check_well_formed` and moved all "structural" kinds of checks there.

asllib/aslspec/spec.ml
Roman-Manevich: They allow you to reference types.
For example
```
typedef A = L;
typedef B = L;
```
Is illegal, since both occurrences of `L` define a label, whereas in
```
typedef A = L;
typedef B = ( L );
```
The label `L` is defined in `A` and referenced in `B`.

HadrienRenaud: Ah ok, maybe good to have this example documented somewhere?

Roman-Manevich: It is documented here https://github.com/herd/herdtools7/blob/master/asllib/aslspec/aslspec.md

HadrienRenaud: ok fair enough

asllib/aslspec/spec.ml
Roman-Manevich: The test `instantiation_recursion.bad` shows an example of a possible infinite recursion on the super term. An infinite recursion on the sub term may not be possible (I couldn't create one), but it's better to safe guard against anyway.

asllib/aslspec/tests.t/instantiation_constant.bad
```diff
@@ -0,0 +1,3 @@
+constant A {""};
+typedef Int {""};
+typedef B {""} = constants_set(A, Int);
```
HadrienRenaud, line 3: Probably not for this PR, but I think it could be good to allow for no description, maybe behind a command line flag. This way we wouldn't have the `{""}` that are getting in the way.

asllib/aslspec/spec.ml
```diff
@@ -448,16 +448,364 @@ module Check = struct
...
+          match variant_def with
+          | Node_TypeVariant _ | Node_Type _ -> ()
+          | _ ->
+              let msg =
+                Format.asprintf
+                  "The type term `%a` cannot be instantiated since '%s' is not \
+                   a type or type variant"
+                  PP.pp_type_term term label
+              in
+              raise (SpecError msg))
```
HadrienRenaud, line 753: I think all of this is caught by `check_well_formed` right?

asllib/aslspec/spec.ml
```diff
@@ -448,16 +448,364 @@ module Check = struct
...
+                    TypeVariant.term = LabelledRecord { fields = def_fields; _ };
+                  } ->
+                  let def_terms = List.map snd def_fields in
+                  check_subsumed_terms_lists id_to_defining_node terms def_terms
+              | _ -> assert false))
+      | Function { from_type = _, from_term; to_type = _, to_term } ->
+          check_well_typed id_to_defining_node from_term;
+          check_well_typed id_to_defining_node to_term
+      | ConstantsSet labels ->
+          List.iter (check_is_constant id_to_defining_node) labels
```
HadrienRenaud, line 741: I think this is caught by `check_well_formed`?

asllib/aslspec/tests.t/instantiation_expansion.spec
Roman-Manevich: I added a comment at the beginning of the file that explain it.

HadrienRenaud (commented):
I think this is good, thanks for all the comments.

I only have a (quite general I'm afraid) comment about extracting some basic operations that are used in various functions.

asllib/aslspec/tests.t/instantiation_constant.bad
Roman-Manevich: Right. I already have that in my list.

PR #1496, opened 2025-10-13
[asl,herd] Improve ASL interface by not overriding FetchDescriptor

HadrienRenaud:
This PR suggests an implementation of `PhysMemRead` which differentiate the
different access types that are passed to it. This makes is possible to use the
`FetchDescriptor` function from the manual.

---

Commit list:

- **[asl,herd] Use PhysMemRead for FetchDescriptor**


maranget:
LGTM. Thanks @HadrienRenaud (tried after rebasing on master that changed much recently).

PR #1499, opened 2025-10-14
[ASL][Reference] fixes

Roman-Manevich:


asllib/doc/SymbolicSubsumptionTesting.tex
```diff
@@ -599,7 +599,7 @@ \section{Symbolic Reasoning\label{sec:Symbolic Reasoning}}
 \hypertarget{def-approximationdirectionterm}{}
 In the following rules, we use the symbol \RenderConstant{Over} to stand for \emph{overapproximation},
 the symbol \RenderConstant{Under} to stand for \emph{underapproximation},
-and the type \RenderType{approximation_direction}.
+and the type \RenderType{approximation_direction}
```
HadrienRenaud, line 602: Why this change? you don't want to end the sentence?

Roman-Manevich: It renders badly. The period goes on the next line, since the type is in a math environment.

HadrienRenaud: For reference `\RenderTypes` creates an equation in a math paragraph, see screenshot:

Screenshot 2025-10-13 at 12 48 51

I think in that case maybe you should reference it, end your sentence and add the definition underneath but I think this is very ok.

asllib/doc/doclint.py
```diff
@@ -643,6 +651,61 @@ def spellcheck(reference_dictionary_path: str, latex_files: list[str]) -> int:
...
+    zero_arg_pattern = re.compile(r"\\newcommand\\([a-zA-Z]+)\[0\]")
+    for line in lines:
+        if is_skipped_line(line):
+            continue
+        matches = re.findall(zero_arg_pattern, line)
+        for match in matches:
+            # Exclude macros that end with "term" as they can be used with {} for styling
+            if (
+                not match.endswith("term")
+                and not match.endswith("Treerm")
```
hrutvik: Typo?

asllib/doc/AbstractSyntax.tex
```diff
@@ -858,7 +858,7 @@ \section{Building Parameterized Productions\label{sec:BuildingParameterizedProdu
 \inferrule[some]{
   b(\vv) \astarrow \astversion{\vv}
 }{
-  \buildoption[b](\overname{\namednode{\vv}{N}}{\vsym}) \astarrow \overname{\Option{\astversion{\vv}}}{\vsymast}
+  \buildoption[b](\overname{\namednode{\vv}{N}}{\vsym}) \astarrow \overname{\Option(\astversion{\vv})}{\vsymast}
```
hrutvik, line 861: Am I right in thinking that `\some` takes an argument but `\Option` does not? If so, is there a particular reason for this?

hrutvik (approved):

asllib/doc/AbstractSyntax.tex
Roman-Manevich: Yes. `some` needs an argument to wrap it with `\left( \right)`, whereas `\Option` is rendered by aslspec, which adds the parenthesis. When we render rules, we may switch `some` to a 0-argument macro.

PR #1490, opened 2025-10-10
[herd] Implement noret for non atomic memory accesses

HadrienRenaud:
With added information in ASL `accessdescriptor`, we are able to implement `NoRet` annotations for non-atomic memory accesses.

The newly added test `MP+rel+LDRnoret-dmb.ld` discriminates against this modification: it is forbidden by `herd7 -variant vmsa` and allowed with `herd7 -variant asl,vmsa` (on this PR).
See CI workflow [here](https://github.com/herd/herdtools7/actions/runs/18529404492/job/52807974637?pr=1490).

---

Commit breakdown:
- First 2 commits add tests to document behaviours that are changed by this PR.
- The 3rd commit, `Rename NoRet into XNoRet` is preparatory work for the next commit, only reflects that what previously was named `NoRet` is now called `XNoRet`, i.e. `NoRet` annotations for atomic accesses.
- 4th commit, `Implement NoRet annotations for simple loads`, implement in herd `NoRet` for `LDR` instructions.
- 5th commit is the ASL implementation of the previous commit.
- Commits 6 and 7 implement `NoRet` annotations for pair instructions (7 being a simple refacto)

PR #1503, opened 2025-10-16
[test] Introduce partial validity checks in regression

maranget:
Some work on tests:
  1. Run basic aarch64 regression tests in ASL mode.
  2. Add some options to regression test applications:
      a- `-checkobs` limit test validity the text part (Always/Sometimes/Never) of the `Observation` filed in test output.
      b- `-checkstates`  limit test validity the "final states" part.
      b- `-verbose` to output some diagnostics once each litmus test has completed.
  4. Document some of Makefile test targets.


herd/tests/instructions/AArch64/asl.cfg
```diff
@@ -0,0 +1,19 @@
...
+nonames L022,L023,L011,L040,L039,L038,L036,L035
+nonames L037,A94,A93,A92,A90,A89,L021,A87
+nonames A88,L014,L007,A91,AK001,AK002,AK003,A50
+nonames A49,A48,A47,A42,A41,A40,A39,A27
+nonames A26,A259,A258,A257,A250,A25,A247,A248
+nonames A241,A24,A240,C,A256,A254,A23,A22
+nonames A255,A246,A244,A21,A245,A243,A242,A253
+nonames A251,A20,A19,A18,A17,A16,A15,A149
+nonames A14,A172,A13,A12,A11,A107,A100,A09
+nonames A07,A06,A05,A03,A02,A01,AK005,A137,A136
```
HadrienRenaud, line 19: Oh no 😱

Looks like I need to implement some instructions ...

internal/lib/args.ml
```diff
@@ -36,6 +36,10 @@ let nohash b =
   "-nohash",Arg.Unit (fun () -> b := true),
   "do not check hashes"
 
+let checkobs b =
+  "-checkobs",Arg.Unit (fun () -> b := true),
```
HadrienRenaud: Wondering if we don't want to keep the same behaviour by default as before, so set this to default to `false` and then put in `herd/tests/instructions/AArch64/asl.cfg`
```
checkobs true
```

maranget: 1. Why change the default behaviour?
2. `-checkobs` is an option of the regression test application, `.cfg` files are for configuring **herd7**.

HadrienRenaud: Sorry my bad, you've already done this I missed something

internal/lib/args.ml
```diff
@@ -36,6 +36,14 @@ let nohash b =
...
+  "limit test validity to observation: Never/Sometimes/Always"
+
+let verbose b =
+  "-verbose",Arg.Unit (fun () -> b := true),
+  "limit test validity to observation: Never/Sometimes/Always"
```
HadrienRenaud: I think the documentation is a bit off?

HadrienRenaud (approved):
Looks good, thanks @maranget 

I'm interested in this PR as there are some litmus tests in #1490 that have an extra final state that I'd like to ignore.

internal/lib/args.ml
maranget: You think right... Fixed. Thanks.

maranget:
Hi @HadrienRenaud. As I understand it, option `-checkstates` of `herd_regression_test.exe` may be what you need for PR #1490.

maranget:
As there are other tests, where the differences are in count of executions, option `-checkstates` may be the best choice for `test.kvm.asl`. This is not finished though...

HadrienRenaud:
> As there are other tests, where the differences are in count of executions, option -checkstates may be the best choice for test.kvm.asl. This is not finished though...

Indeed it was not finished, but I've worked on it. Now the only difference is in the number of executions ...

maranget:
I meant this PR #1503 is not finished. I still have to work a bit on it. I am happy to see that PR #1490 is also progressing!

HadrienRenaud:
> Would you please review and decide if option -checkstates is adequate for PR #1490?

Hi @maranget, I have tried to rebase #1490 on top of this PR, and use `-checkstate` in the `Makefile`. There seems to be some problems.

Where before the logs were not matching we have:
```
Failed ./herd/tests/instructions/AArch64.kvm/MP+rel+LDNPnoret-dmb.ld.litmus : Missing file './herd/tests/instructions/AArch64.kvm/MP+rel+LDNPnoret-dmb.ld.litmus.expected-failure'
```
Maybe it's my fault on this one, I haven't checked.

When I tried to do `promote`, I have the following error:
```
❯ make test.kvm.asl REGRESSION_TEST_MODE=promote

Fatal error: exception Unix.Unix_error(Unix.ENOENT, "create_process", "-checkstates")
Fatal error: exception Unix.Unix_error(Unix.ENOENT, "create_process", "-checkstates")
...
```

I don't really know what's happening, I won't have time to investigate before tomorrow.

maranget:
This failing test, may suggest that we should abstract away the contents of error messages in mode `-checkstates` and `-checkobs` . In other words, consider that a failing test satisfies the regression test because it fails regardless of why.

Makefile
```diff
@@ -331,7 +338,7 @@ test-aarch64-noasl:
 		-libdir-path ./herd/libdir \
 		-litmus-dir ./herd/tests/instructions/AArch64.ASL \
 		-conf ./herd/tests/instructions/AArch64.ASL/noasl.cfg \
-		$(REGRESSION_TEST_MODE)
+		$(NOPROMOTE)
```
HadrienRenaud: Shouldn't this one be the one promoting and not the ASL one?

Makefile
```diff
@@ -344,7 +351,7 @@ test-aarch64-noasl-mixed:
 		-libdir-path ./herd/libdir \
 		-litmus-dir ./herd/tests/instructions/AArch64.ASL \
 		-conf ./herd/tests/instructions/AArch64.ASL/noasl-mixed.cfg \
-		$(REGRESSION_TEST_MODE)
+		$(NOPROMOTE)
```
HadrienRenaud: Same for this one

internal/herd_catalogue_regression_test.ml
```diff
@@ -266,7 +266,16 @@ let () =
     | "show" :: [] -> show_tests ?j ?timeout flags
     | "test" :: [] -> run_tests ?j ?timeout flags
     | "promote" :: [] -> promote_tests ?j flags
```
HadrienRenaud, line 268: Am I mistaken or all calls to `herd_catalogue_regression_test` are with `nopromote`? In that case, should we just put a warning or a failure with `promote`?

Makefile
maranget: Here the ASL test is the reference. For instance there were nonames lines in the configuration file `./herd/tests/instructions/AArch64.ASL/noasl.cfg`. This is no longer the case (cf. commit 6f1f9e54ad542b63376d128be33c8ef3e1680ebf), as ASL and non-ASL output are now identical for all litmus tests. Should this change in the future, I think the tests that make a difference should again be executed in ASL only.

Makefile
maranget: I guess there should be exactly one reference per regression test directory.

Makefile
HadrienRenaud: That makes sense, thanks

internal/herd_catalogue_regression_test.ml
maranget: I'd rather have `herd_catalogue_regression` to go on with accepting the `promote` command:
 1. Some makefile catalogue test targets do not turn "promote" into "nopromote" (for instance `make cata-aarch64-test`).
 2 It may be useful to directly invoke `herd_catalogue_regression.exe` directly with command `promote`. One may do so to check which tests do not terminate in the allocated amount of time, or for any other reason. The intent of "nopromote" is avoiding involuntary clobber of reference files when using `make .... REGRESSION_TEST_MODE=promote`.

HadrienRenaud: That makes sense, thanks

PR #1507, opened 2025-10-20
[herd] Introduce basic support for AArch64 Guarded Control Stack 

murzinv:
Guarded Control Stack (GCS), an area of memory in which procedure return addresses and exception return addresses are stored and protected from modification.
When the processor executes a Branch with Link instruction, such as BL, the return address is pushed onto the GCS as well as being written into the Link Register (LR). On a procedure return, the latest stored return address is popped from the GCS. The processor either compares the popped value with the LR, or uses the popped value directly.
There are times when the software needs to make manual adjustments to the control stack to enable this, the architecture provides instructions for maintaining the GCS.

### DIFF
--- a/catalogue/aarch64-GCS/tests/Co+gcsss1+gcsss1.litmus
+++ b/catalogue/aarch64-GCS/tests/Co+gcsss1+gcsss1.litmus
@@ -1,0 +1,18 @@
+AArch64 Co+gcsss1+gcsss1
+variant=shadowstack
+{
+  SS(x,2);
+  SS(y,2) = ssval_t: {0,SSCap(y,1)};
+
+  0:GCSPR_EL1=&x[1];
+  0:X1=&y[1];
+  0:X2=y;
+}
+
+P0             ;
+GCSSS1 X1      ; (* GCSSS1 Memory effect, GCSPR = y[1], y[1] = SSCap(x[1],5)   *)
+ADD X3,X2,#1   ; (* Synthesize "valid" cap                                     *)
+GCSSTR X3,[X1] ; (* GCS (Write) effect                                         *)
+GCSSS1 X1      ; (* GCSSS1 Memory effect, GCSPR = y[1], y[1] = SSCap(y[1],5) q *)
+
+forall ~fault(P0)
--- a/catalogue/aarch64-GCS/tests/coWR+poblret.litmus
+++ b/catalogue/aarch64-GCS/tests/coWR+poblret.litmus
@@ -1,0 +1,16 @@
+AArch64 coWR+poblret
+variant=shadowstack
+{
+
+  SS(x,1);
+
+  0:GCSPR_EL1=&x[1];
+}
+P0    ;
+BL L0 ; (* GCS (Write) effect *)
+B L1  ;
+L0:   ;
+RET   ; (* GCS (Read) effect  *)
+L1:   ;
+
+forall ~fault(P0,GCS:PRET)
--- a/catalogue/aarch64-GCS/tests/coWR+pogcspushmgcspopm.litmus
+++ b/catalogue/aarch64-GCS/tests/coWR+pogcspushmgcspopm.litmus
@@ -1,0 +1,14 @@
+AArch64 coWR+pogcspushmgcspopm
+variant=shadowstack
+{
+
+  SS(x,1);
+
+  0:GCSPR_EL1=&x[1];
+}
+P0         ;
+MOV X1,#4  ;
+GCSPUSHM X1; (* GCS (Write) effect *)
+GCSPOPM X2 ; (* GCS (Read) effect  *)
+
+forall 0:X2=4
--- a/catalogue/aarch64-GCS/tests/coWR+pogcsss1gcsss2.litmus
+++ b/catalogue/aarch64-GCS/tests/coWR+pogcsss1gcsss2.litmus
@@ -1,0 +1,14 @@
+AArch64 coWR+pogcsss1gcsss2
+variant=shadowstack
+{
+  SS(x,2);
+  SS(y,1) = ssval_t: {SSCap(y,1)};
+
+  0:GCSPR_EL1=&x[1];
+  0:X0=y;
+}
+P0        ;
+GCSSS1 X0 ; (* GCSSS1 Memory effect, GCSPR = y[0], y[0] = SSCap(x[1],5)  *)
+GCSSS2 X0 ; (* GCS effect                                                *)
+
+forall ~fault(P0,GCS:SS2)
--- a/herd/AArch64Arch_herd.ml
+++ b/herd/AArch64Arch_herd.ml
@@ -48,17 +48,17 @@ module Make (C:Arch_herd.Config)(V:Value.AArch64) =
 
     let is_ifetch_annot = function
       | NExp IFetch -> true
-      | NExp (AF|DB|AFDB|Other)|Exp -> false
+      | NExp (AF|DB|AFDB|Other|GCS)|Exp -> false
 
     let is_barrier b1 b2 = barrier_compare b1 b2 = 0
 
     let is_af = function (* Setting of access flag *)
       | NExp (AF|AFDB)-> true
-      | NExp (DB|IFetch|Other)|Exp -> false
+      | NExp (DB|IFetch|Other|GCS)|Exp -> false
 
     and is_db = function (* Setting of dirty bit flag *)
       | NExp (DB|AFDB) -> true
-      | NExp (AF|IFetch|Other)|Exp -> false
+      | NExp (AF|IFetch|Other|GCS)|Exp -> false
 
     module CMO = struct
       type t = | DC of AArch64Base.DC.op | IC of AArch64Base.IC.op
--- a/herd/AArch64Arch_herd.ml
+++ b/herd/AArch64Arch_herd.ml
@@ -115,6 +115,7 @@ module Make (C:Arch_herd.Config)(V:Value.AArch64) =
     | I_SMSTART _ | I_SMSTOP _ | I_LD1SPT _ | I_ST1SPT _
     | I_MOVA_TV _ | I_MOVA_VT _ | I_ADDA _
     | I_PAC _ | I_AUT _ | I_XPACI _ | I_XPACD _
+    | I_GCSPOPM _ | I_GCSPUSHM _ | I_GCSSTR _ | I_GCSSS1 _ | I_GCSSS2 _
       -> true
 
     let is_cmodx_restricted_value =
@@ -317,6 +318,7 @@ module Make (C:Arch_herd.Config)(V:Value.AArch64) =
       | I_RDVL _ | I_ADDVL _ | I_CNT_INC_SVE _ | I_CTERM _
       | I_SMSTART _ | I_SMSTOP _ | I_MOVA_TV _ | I_MOVA_VT _ | I_ADDA _
       | I_PAC _ | I_AUT _ | I_XPACI _ | I_XPACD _
+      | I_GCSPOPM _ | I_GCSPUSHM _ | I_GCSSTR _ | I_GCSSS1 _ | I_GCSSS2 _
           -> None
 
     let all_regs =
@@ -350,7 +352,7 @@ module Make (C:Arch_herd.Config)(V:Value.AArch64) =
       | I_FENCE _
       | I_IC _|I_DC _|I_TLBI _ | I_AT _
       | I_NOP|I_TBZ _|I_TBNZ _
-      | I_BL _ | I_BLR _ | I_RET _ | I_ERET | I_SVC _ | I_UDF _
+      | I_ERET | I_SVC _  | I_UDF _
       | I_ST1SP _ | I_ST2SP _ | I_ST3SP _ | I_ST4SP _
       | I_ST1SPT _ | I_CTERM _
         -> [] (* For -variant self only ? *)
@@ -400,11 +402,24 @@ module Make (C:Arch_herd.Config)(V:Value.AArch64) =
       | I_LD1SPT (_,r,_,_,_,_,_) | I_MOVA_TV (r,_,_,_,_) | I_MOVA_VT (r,_,_,_,_)
       | I_ADDA (_,r,_,_,_)
       | I_IRG (r,_,_)
+      | I_GCSSTR (r,_)
         -> [r]
       | I_LDAR (_,(XX|AX),r,_) |I_LDARBH (_,(XX|AX),r,_)
         -> [r;ResAddr;]
       | I_STXR (_,_,r,_,_) | I_STXP (_,_,r,_,_, _) | I_STXRBH (_,_,r,_,_)
         -> [r;ResAddr;]
+      | I_BL _ | I_BLR _  when C.variant Variant.ShadowStack
+        -> [Ireg R30; SysReg GCSPR_EL1]
+      | I_RET _ when C.variant Variant.ShadowStack
+        -> [SysReg GCSPR_EL1]
+      | I_BL _ | I_BLR _
+        -> [Ireg R30]
+      | I_RET _
+        -> []
+      | I_GCSPUSHM _  | I_GCSSS1 _
+        -> [SysReg GCSPR_EL1]
+      | I_GCSPOPM r | I_GCSSS2 r
+        -> [r; SysReg GCSPR_EL1]
       | I_MSR (SYS_NZCV,_)
         ->
           nzcv_regs
@@ -495,6 +510,7 @@ module Make (C:Arch_herd.Config)(V:Value.AArch64) =
       | I_LD1SPT _ | I_ST1SPT _
       | I_MOVA_TV _| I_MOVA_VT _ | I_ADDA _
       | I_PAC _ | I_AUT _ | I_XPACI _ | I_XPACD _
+      | I_GCSPOPM _ | I_GCSPUSHM _ | I_GCSSTR _ | I_GCSSS1 _ | I_GCSSS2 _
         -> MachSize.No
 
     let reg_defaults =
--- a/herd/AArch64Sem.ml
+++ b/herd/AArch64Sem.ml
@@ -44,6 +44,7 @@ module Make
     let pac = C.variant Variant.Pac
     let const_pac_field = C.variant Variant.ConstPacField
     let fpac = C.variant Variant.FPac
+    let gcs = C.variant Variant.ShadowStack
 
     let check_kvm ins =
       if not kvm then
@@ -79,6 +80,12 @@ module Make
       if not sme then
         Warn.user_error
           "SME instruction %s requires -variant sme"
+          (AArch64.dump_instruction inst)
+
+    let check_gcs inst =
+      if not gcs then
+        Warn.user_error
+          "GCS instruction %s requires -variant shadowstack"
           (AArch64.dump_instruction inst)
 
 (* Barrier pretty print *)
@@ -943,7 +950,7 @@ module Make
           (fun _ ->
              set_elr_el1 lbl_v ii
              >>|
-             mk_fault (Some a) dir an ii ft msg) None ii
+             mk_fault a dir an ii ft msg) None ii
         >>!  B.fault [AArch64Base.elr_el1, lbl_v]
 
       (* Specific fault when accessing PTE from EL0. *)
@@ -951,7 +958,7 @@ module Make
         let open FaultType.AArch64 in
         let ft = Some (MMU (domain, Permission))
         and msg = Some "EL0" in
-        emit_fault a ma dir an ft msg ii
+        emit_fault (Some a) ma dir an ft msg ii
 
       let an_xpte =
         let open Annot in
@@ -1216,7 +1223,6 @@ module Make
 (* Page tables and TLBs *)
       let do_inv op a ii = inv_loc op (A.Location_global a) ii
 
-
 (************************)
 (* Conditions and flags *)
 (************************)
@@ -1458,7 +1464,7 @@ module Make
         | _ -> true)
 
       let lift_kvm _tag dir updatedb mop ma an ii mphy branch domain =
-        let mfault ma a ft = emit_fault a ma dir an ft None ii in
+        let mfault ma a ft = emit_fault (Some a) ma dir an ft None ii in
         let maccess a ma =
           check_ptw ii.AArch64.proc dir updatedb false a ma an ii
             (mop (Access.PTE domain) ma |> branch)
@@ -2184,7 +2190,7 @@ Arguments:
           (rmw_to_read rmw)
           ii
 
-      let do_cas_fail do_wb sz an rn ma mv mop ii =
+      let do_cas_fail do_wb sz an rn ma mv mop tagcheck ii =
         let action checked ma =
           let do_action updatedb checked ma =
               (* Dir.W would force check for dbm bit:                  *)
@@ -2208,7 +2214,7 @@ Arguments:
           end
         in
 
-        if memtag && C.mte_store_only then
+        if tagcheck && C.mte_store_only then
           (* If FEAT_MTE_STORE_ONLY is implemented it is              *)
           (* CONSTRAINED UNPREDICTABLE whether the Tag Check          *)
           (* operation is performed.                                  *)
@@ -2227,18 +2233,18 @@ Arguments:
       let do_cas_fail_with_wb = do_cas_fail true
       let do_cas_fail_no_wb = do_cas_fail false
 
-      let do_cas sz an rn ma mv mop_success mop_fail_with_wb mop_fail_no_wb ii =
+      let do_cas sz an rn ma mv mop_success mop_fail_with_wb mop_fail_no_wb tagcheck ii =
         M.altT (
           (* CAS succeeds and generates an Explicit Write Effect *)
           (* there must be an update to the dirty bit of the TTD *)
-          lift_memop ~tag:"CAS" rn Dir.W true memtag mop_success (to_perms "rw" sz) ma mv an ii
+          lift_memop ~tag:"CAS" rn Dir.W true tagcheck mop_success (to_perms "rw" sz) ma mv an ii
         )( (* CAS fails *)
           M.altT (
             (* CAS generates an Explicit Write Effect              *)
-            do_cas_fail_with_wb sz an rn ma mv mop_fail_with_wb ii
+            do_cas_fail_with_wb sz an rn ma mv mop_fail_with_wb tagcheck ii
           )(
             (* CAS does not generate an Explicit Write Effect      *)
-            do_cas_fail_no_wb sz an rn ma mv mop_fail_no_wb ii
+            do_cas_fail_no_wb sz an rn ma mv mop_fail_no_wb tagcheck ii
           )
         )
 
@@ -2283,7 +2289,7 @@ Arguments:
         in
         let ma = read_reg_addr rn ii
         and mv = read_reg_data_sz sz rt ii in
-        do_cas sz an rn ma mv mop_success mop_fail_with_wb mop_fail_no_wb ii
+        do_cas sz an rn ma mv mop_success mop_fail_with_wb mop_fail_no_wb memtag ii
 
       let casp sz rmw rs1 rs2 rt1 rt2 rn ii =
         let an = rmw_to_read rmw in
@@ -2344,7 +2350,7 @@ Arguments:
         in
         let ma = read_reg_addr rn ii
         and mv = read_reg_data_sz sz rt1 ii >>> fun _ -> read_reg_data_sz sz rt2 ii in
-        do_cas sz an rn ma mv mop_success mop_fail_with_wb mop_fail_no_wb ii
+        do_cas sz an rn ma mv mop_success mop_fail_with_wb mop_fail_no_wb memtag ii
 
       (* Temporary morello variation of CAS *)
       let cas_morello sz rmw rs rt rn ii =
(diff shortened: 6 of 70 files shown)


herd/AArch64Annot.ml
```diff
@@ -18,6 +18,12 @@
 type t =
   A | XA | L | XL | X | N | Q | XQ | NoRet | S
   | NTA (* Non-Temporal, avoid clash with NT in AArch64Base *)
+  | GCS (* GCS Memory Effect *)
```
relokin: Can you also add a `GCSX` which will be the type of annotation that Memory Effects from a GCSSS1 will have? It should return true for `is_gcs` and true for `is_atomic`.

herd/AArch64Arch_herd.ml
relokin: This and the following suggestion along with the change in `lib/AArch64Base.ml` is to get GCSB.DSYNC Effects.
```suggestion
      fold_barrier false true
```

relokin, line 137:
```suggestion
        []
```

herd/AArch64Sem.ml
```diff
@@ -3553,6 +3575,419 @@ Arguments:
...
+              (* Branch depends on destination register (or LR) *)
+              M.short (is_this_reg r) (E.is_bcc) m)
+          ma
+          (read_reg_ord r ii)
+          an
+          ii)
+
+      let gcsss1 r ii =
+        let open AArch64Base in
+        let an = Annot.X
```
relokin, line 3858:
```suggestion
        let an = Annot.GCSX
```

herd/AArch64Sem.ml
```diff
@@ -1125,6 +1137,13 @@ module Make
       let do_inv op a ii = inv_loc op (A.Location_global a) ii
 
 
+(****************)
+(* Shadow Stack *)
+(****************)
+
+      let do_read_shadow_stack ac an a ii = do_read_mem_ret quad an aexp ac a ii
+      and do_write_shadow_stack ac an a v ii = do_write_mem quad an aexp ac a v ii
```
relokin:
```suggestion
      let do_read_shadow_stack ac an a ii = do_read_mem_ret quad an AArch64Explicit.NExp ac a ii
      and do_write_shadow_stack ac an a v ii = do_write_mem quad an AArch64Explicit.NExp ac a v ii
```

lib/AArch64Base.ml
```diff
@@ -686,6 +690,7 @@ let do_pp_barrier tag b = match b with
   | DMB (d,t) -> "DMB" ^ tag ^ pp_option d t
   | DSB (d,t) -> "DSB" ^ tag ^ pp_option d t
   | ISB -> "ISB"
+  | GCSB -> "GCSB DSYNC"
```
relokin:
```suggestion
  | GCSB -> "GCSB" ^ tag ^  "DSYNC"
```

lib/AArch64Base.ml
relokin, line 675:
```suggestion
  let k = f ISB k in
  let k = f GCSB k in
```

relokin (commented):
Some initial comments.

herd/AArch64Sem.ml
murzinv: `AArch64Explicit.NExp` expects one of  `AF|DB|AFDB|IFetch|Other` so is that Ok to use `Other` here or we need to add one for GCS?

relokin: Sorry I didn't think of that. I think you should add GCS here. And that probably means that you don't need the GCS Annot and therefore you won't need separate GCSX, GCSA, GCSL either.

murzinv: No problem, I've just added GCS access type and undo annotation.

herd/AArch64Sem.ml
```diff
@@ -1374,6 +1395,9 @@ module Make
             fun (r,_) -> M.unitT r
         else m
 
+      let set_elr_el1 v ii =
```
diaolo01: This already exists at line 1361.

herd/AArch64Sem.ml
```diff
@@ -3553,6 +3577,419 @@ Arguments:
...
+              M.op Op.Or index cap) >>= fun (v, offset) ->
+              (* Write it back *)
+              M.op Op.Add v offset
+
+      let reset_cap = set_cap V.zero
+      let make_valid = set_cap V.one (V.intToV 0xfff)
+      let make_inprogress = set_cap (V.intToV 0x5) (V.intToV 0x7)
+
+      let mk_gcs_fault ft ii=
+        let (>>!) = M.(>>!) in
```
diaolo01: This is already declared on line 128.

herd/AArch64Sem.ml
```diff
@@ -3553,6 +3577,419 @@ Arguments:
...
+        and off = MachSize.nbytes quad in
+        let m =
+        M.delay_kont "gcspopm"
+        (read_reg_addr rA ii)
+        (fun a_virt ma ->
+          let mop ac a =
+            let m = do_read_shadow_stack ac an a ii >>= fun v ->
+              let commit = commit_pred_txt (Some "PCAligned") ii in
+              let mok =
+                let(>>*=) = M.bind_control_set_data_input_first in
```
diaolo01, line 3723: There’s already a global `>>*=` with a different meaning; shadowing it here makes it easy to misread the code. Could we use a different operator name to highlight the difference?

herd/AArch64Sem.ml
```diff
@@ -3553,6 +3577,419 @@ Arguments:
...
+                commit >>*= fun () ->
+                  M.add a_virt (V.intToV off) >>= fun new_addr ->
+                    write_reg rd v ii >>|
+                    write_reg rA new_addr ii
+                    >>= M.ignore >>= B.next1T in
+              let mno =
+                let open FaultType.AArch64 in
+                let(>>*=) = M.bind_data_to_minimals in
```
diaolo01: Rebinding `>>*=` here to `bind_data_to_minimals` after it was just rebound to `bind_control_set_data_input_first` makes it more confusing to me.

herd/AArch64Sem.ml
```diff
@@ -3553,6 +3577,419 @@ Arguments:
...
+          mzero
+          an
+          ii) in
+        (* Value writen to GCSPR depends on previous read *)
+        let read e = (is_this_reg rA e) && (E.is_reg_load e ii.A.proc)
+        and write e = (is_this_reg rA e) && (E.is_reg_store e ii.A.proc) in
+        M.short read write m
+
+      (*
+       * Basically copy of lift_memop which allows mop to drive control flow
```
diaolo01, line 3758: Is it worth extracting the common parts with lift_memop into a helper function?

diaolo01 (commented):
I successfuly built using this PR and ran all the new GCS tests. Everything passes with the exception of tests with `SScap` in the post condition, since that's not yet supported. I was able to observe the GCS access in the produced graphs and the `GCSB` barrier. The barrier correctly becomes a NOP when `-variant shadowstack` is omitted. 
Most of my comments are clean-up. I have not performed a full review against the architecture, but I checked parts of it like the barrier and the implementation of the cap-related functions.

lib/symbValue.ml
```diff
@@ -594,8 +594,9 @@ module
         | Some o -> intToV o
         | None ->  illegal_offset v
       end
+  | Val (Concrete _) as v -> v
```
relokin, line 597: does this mean that a Concrete has a valid offset?

catalogue/aarch64-GCS/tests/Co+bl+gcsb+R.litmus
```diff
@@ -0,0 +1,17 @@
+AArch64 Co+bl+gcsb+R
```
relokin:
```suggestion
AArch64 coWR+gcsbblp
```

catalogue/aarch64-GCS/tests/Co+bl+ret.litmus
```diff
@@ -0,0 +1,17 @@
+AArch64 Co+bl+ret
```
relokin:
```suggestion
AArch64 coWR+poblret
```

catalogue/aarch64-GCS/tests/Co+gcspush+gcsb+R.litmus
```diff
@@ -0,0 +1,17 @@
+AArch64 Co+gcspush+gcsb+R
```
relokin:
```suggestion
AArch64 coWR+gcsbgcspushmp
```

catalogue/aarch64-GCS/tests/Co+gcspush+gcspop.litmus
```diff
@@ -0,0 +1,15 @@
+AArch64 Co+gcspush+gcspop
```
relokin:
```suggestion
AArch64 coWR+pogcspushmgcspopm
```

catalogue/aarch64-GCS/tests/Co+gcsss1+gcsss1.litmus
```diff
@@ -0,0 +1,18 @@
...
+  SS(y,2) = ssval_t: {0,SSCap(y,1)};
+
+  0:GCSPR_EL1=&x[1];
+  0:X1=&y[1];
+  0:X2=y;
+}
+
+P0             ;
+GCSSS1 X1      ; (* GCSSS1 Memory effect, GCSPR = y[1], y[1] = SSCap(x[1],5)   *)
+ADD X3,X2,#1   ; (* Syntesize "valid" cap                                      *)
```
relokin:
```suggestion
ADD X3,X2,#1   ; (* Synthesize "valid" cap                                     *)
```

catalogue/aarch64-GCS/tests/Co+gcsss1+gcsss1.litmus
```diff
@@ -0,0 +1,18 @@
+AArch64 Co+gcsss1+gcsss1
```
relokin, line 1: Is this test targeting a specific rule?

catalogue/aarch64-GCS/tests/Co+gcsss1+gcsss2.litmus
```diff
@@ -0,0 +1,14 @@
+AArch64 Co+gcsss1+gcsss2
```
relokin:
```suggestion
AArch64 coWR+pogcsss1gcsss2
```

catalogue/aarch64-GCS/tests/Co+gcsss1+R.litmus
```diff
@@ -0,0 +1,15 @@
+AArch64 Co+gcsss1+R
```
relokin:
```suggestion
AArch64 coWR+pogcsss1p

File: catalogue/aarch64-GCS/tests/Co+gcsstr+gcsss1.litmus [hunk 19]
```diff
@@ -0,0 +1,17 @@
+AArch64 Co+gcsstr+gcsss1
```
relokin:
```suggestion
AArch64 coWR+pogcsstrgcsss1
```

catalogue/aarch64-GCS/tests/Co+W+gcsb+gcspop.litmus
```diff
@@ -0,0 +1,19 @@
+AArch64 Co+W+gcsb+gcspop
```
relokin:
```suggestion
AArch64 coWR+gcsbpgcspopm
```

catalogue/aarch64-GCS/tests/Co+W+gcsb+ret.litmus
```diff
@@ -0,0 +1,20 @@
+AArch64 Co+W+gcsb+ret
```
relokin:
```suggestion
AArch64 coWR+gcsbpret
```

catalogue/aarch64-GCS/tests/Co+W+gcsss1.litmus
```diff
@@ -0,0 +1,20 @@
+AArch64 Co+W+gcsss1
```
relokin:
```suggestion
AArch64 coWR+popgcsss1
```

catalogue/aarch64-GCS/tests/LB+gcspop+gcsb+dmb.sy-dmb.ld+gcsb.litmus
```diff
@@ -0,0 +1,18 @@
+AArch64 LB+gcspop+gcsb+dmb.sy-dmb.ld+gcsb
```
relokin:
```suggestion
AArch64 LB+gcspop+gcsb-dmb.sy+dmb.ld-gcsb
```

catalogue/aarch64-GCS/tests/LB+gcsss2+dmb.sy-dmb.ld+gcsss1.litmus
```diff
@@ -0,0 +1,22 @@
+AArch64 LB+gcsss2+dmb.sy-dmb.ld+gcsss1
```
relokin:
```suggestion
AArch64 LB+gcsss2-dmb.sy+dmb.ld-gcsss1
```

catalogue/aarch64-GCS/tests/LB+ret+gcsb+dmb.sy-dmb.ld+gcsb.litmus
```diff
@@ -0,0 +1,21 @@
+AArch64 LB+ret+gcsb+dmb.sy-dmb.ld+gcsb
```
relokin:
```suggestion
AArch64 LB+ret-gcsb-dmb.sy+dmb.ld-gcsb
```

catalogue/aarch64-GCS/tests/MP+bl+gcsb+dmb.sy-dmb.ld.litmus
```diff
@@ -0,0 +1,23 @@
+AArch64 MP+bl+gcsb+dmb.sy-dmb.ld
```
relokin:
```suggestion
AArch64 MP+bl-gcsb-dmb.sy+dmb.ld
```
I won't comment on all the rest I think you get the point + to move the next thread and - within the thread.

herd/AArch64Arch_herd.ml
```diff
@@ -399,11 +401,17 @@ module Make (C:Arch_herd.Config)(V:Value.AArch64) =
       | I_RDVL (r,_) | I_ADDVL (r,_,_) | I_CNT_INC_SVE (_,r,_,_)
       | I_LD1SPT (_,r,_,_,_,_,_) | I_MOVA_TV (r,_,_,_,_) | I_MOVA_VT (r,_,_,_,_)
       | I_ADDA (_,r,_,_,_)
+      | I_GCSSTR (r,_)
         -> [r]
       | I_LDAR (_,(XX|AX),r,_) |I_LDARBH (_,(XX|AX),r,_)
         -> [r;ResAddr;]
       | I_STXR (_,_,r,_,_) | I_STXP (_,_,r,_,_, _) | I_STXRBH (_,_,r,_,_)
         -> [r;ResAddr;]
+      | I_BL _ | I_BLR _ | I_RET _
```
relokin: Is it easy to check if variant shadowstack has been enabled? It's not a huge problem to "kill" GCSPR_EL1 when not needed, so just checking.

lib/constant.mli
```diff
@@ -42,6 +42,7 @@ type symbolic_data =
    cap : cap ;
    offset : offset ;
    pac : PAC.t ;
+   fixup_offset: bool;
```
relokin: Sorry I didn't quite understand why you need this, can you please explain?

lib/faultType.ml
```diff
@@ -66,6 +98,14 @@ module AArch64 = struct
               MMU AccessFlag;
               MMU Permission;
               MMU Exclusive;];
+      "GCS", [GCS PRET;
```
relokin: This will create a "GCS" set in the cat file will correspond to a subset of Fault Effects, can we use something specific, perhaps "GCSDataCheck"?

herd/AArch64Sem.ml
```diff
@@ -568,7 +575,10 @@ module Make
       let commit_pred ii = commit_pred_txt None ii
 
 (* Fence *)
-      let create_barrier b ii = M.mk_singleton_es (Act.Barrier b) ii
+      let create_barrier b ii =
+        match b with
+        | AArch64Base.GCSB when not gcs -> M.mk_singleton_es (Act.NoAction) ii
```
relokin: Does this mean that you handle GCSB as a NOP when shadowstack is not a variant? Why doesn't that simply throw an error?

herd/AArch64Sem.ml
```diff
@@ -1153,6 +1163,17 @@ module Make
       let do_inv op a ii = inv_loc op (A.Location_global a) ii
 
 
+(****************)
```
relokin: I think you can group everything related to GCS into a separate module similarly to what I did for GICv5

herd/AArch64Sem.ml
```diff
@@ -3569,6 +3590,419 @@ Arguments:
...
+          let shortcut =
+            let read_reg e =
+              let is_reg = is_this_reg r e
+              and is_load = E.is_reg_load e ii.A.proc in
+              is_reg && is_load
+            in
+            M.short read_reg (E.is_pred_txt cond2) in
+         let mop_fail_no_wb ac ma _ =
+            (* GCSSS1 fails, there is no Explicit Write Effect *)
+            (* FIXME: GCSSS1 Memory Read effect *)
```
relokin: Can we remove this FIXME now? And the others in this function?

herd/AArch64Sem.ml
```diff
@@ -3569,6 +3590,419 @@ Arguments:
...
+              (* Make new offset *)
+              M.op Op.Or index cap) >>= fun (v, offset) ->
+              (* Write it back *)
+              M.op Op.Add v offset
+
+      let reset_cap = set_cap V.zero
+      let make_valid = set_cap V.one (V.intToV 0xfff)
+      let make_inprogress = set_cap (V.intToV 0x5) (V.intToV 0x7)
+
+      let mk_gcs_fault ft ii=
```
relokin: can't you can emit_fault?

herd/AArch64Sem.ml
```diff
@@ -3569,6 +3590,419 @@ Arguments:
...
+            (fun _ -> set_elr_el1 lbl_v ii >>| mk_fault (Some a) dir an ii ft None)
+            None ii >>! B.fault [AArch64Base.elr_el1, lbl_v] in
+          let maccess a ma =
+            check_ptw ii.AArch64.proc dir updatedb false a ma an ii
+            (fire_spurious_af dir a (mop Access.PTE ma))
+            mphy
+            mfault in
+          M.delay_kont "shadow_stack"
+          ma
+          (if pte2 then
```
relokin: pte2 is not relevant here, pte2 adds page walks for LDRs/STRs to PTEs.

herd/AArch64Sem.ml
```diff
@@ -3569,6 +3590,419 @@ Arguments:
...
+      (*
+       * Basically copy of lift_memop which allows mop to drive control flow
+       * (handy for BL{R}/RET insrucions with GCS enabled)
+       *)
+      let lift_shadow_stack dir updatedb mop ma mv an ii =
+        let mop = apply_mv mop mv in
+        if kvm then
+          let mphy ma a_virt =
+            let ma = get_oa a_virt ma in
+            if pte2 then
```
relokin: pte2 is not relevant here, pte2 adds page walks for LDRs/STRs to PTEs.

herd/AArch64Sem.ml
```diff
@@ -3569,6 +3590,419 @@ Arguments:
...
+              M.op1 Op.IsVirtual a_virt >>= fun c ->
+                M.choiceT c
+                (mop Access.PHY ma)
+                (fire_spurious_af dir a_virt (mop Access.PHY_PTE ma))
+            else
+              mop Access.PHY ma
+          in
+          (* lift_kvm dir updatedb mop ma an ii mphy in *)
+          let lbl_v = get_instr_label ii in
+          let mfault ma a ft =
```
relokin: isn't this another place where we can call emit_fault?

herd/AArch64Sem.ml
```diff
@@ -3569,6 +3590,419 @@ Arguments:
...
+              and write e = (is_this_reg rA e) && (E.is_reg_store e ii.A.proc) in
+              let m = M.short read write m in
+              (* Branch depends on destination register (or LR) *)
+              M.short (is_this_reg r) (E.is_bcc) m)
+          ma
+          (read_reg_ord r ii)
+          an
+          ii)
+
+      let gcsss1 r ii =
```
relokin, line 3856: Have you tried reusing the cas helper functions?

catalogue/aarch64-GCS/tests/Co+gcsss1+gcsss1.litmus
murzinv: I think this is for

```
Two effects E1 and E2 to the Same Location follow the same ordering requirements as two Explicit Memory effects to the Same Location if one of the following applies:
 - E1 and E2 are both GCSSS1 Memory effects.
```

herd/AArch64Arch_herd.ml
murzinv: It can be easily done. By the way it seems that existing code doesn't take into account that `BL` and `BLR` kill `X30`...

lib/faultType.ml
murzinv: Ack

herd/AArch64Sem.ml
murzinv: Ack. It is better to `check_gcs` in `I_FENCE` like we do for other instructions

herd/AArch64Sem.ml
murzinv: Done

herd/AArch64Sem.ml
murzinv: Ack

herd/AArch64Sem.ml
murzinv: Ack.

herd/AArch64Sem.ml
murzinv: No, that code predates `do_cas`. I've had a look just k now and it seems that it can be reused with minor change to `do_cas` to take parameter to check tags or not since GCS access is not TagChecked

lib/constant.mli
murzinv: That's for case where `SSCap`assigned to the register for instance `0:X0=SSCap(x,1) (* invalid cap *)`. Since shadow stack under the hood is basically array `herd7` would interpreter it as  `0:X0=x[1]` which would result in `.offset` initially set to `1`. However, `herd7` would scale up / fixup `.offset` once information about type of `x` get know so it will turn it into `.offset` `8`  (more details in #1371).

`fixup_offset` flag is attempt to control if we want such fixup or we want keep `.offset` as is. 

So far I have not experienced need for `0:X0=SSCap(x,1)` and instead synthesized SSCap in the litmus code and I think I can just drop `fixup_offset` from the PR..

lib/symbValue.ml
murzinv: Yes. One way to see it for `Concrete` would be treating value as offset from `0`. The reason why it is required is that shadow stack location by default initialized with `0`, so any attempt to use that value (`GCSPOPM`/`RET`/`GCSSS1`) would result in error. Possible alternative could be to disallow `Concrete` in shadow stack and allow only `Symbol` (`Data` and `Label`) yet it require more work... also having `0` in shadow stack can be result of Overshooting Memory Effect so we might run into the same problem once we want to support that

relokin: Sorry I am not sure I understand this. What do you think is the desired behaviour for this test:
```
AArch64 A
{
 SS(x,2);
}
 P0 ;
 GCSPOPM X0; 
```

catalogue/aarch64-GCS/tests/Co+gcsss1+gcsss1.litmus
relokin: In that case, why do you need the GCSSTR, now you have E1 and E2 generated by SS1 instructions but also an E3 which is to the same Location.

catalogue/aarch64-GCS/tests/coWR+poblret.litmus
```diff
@@ -0,0 +1,17 @@
+AArch64 coWR+poblret
+variant=shadowstack
+{
+
+  SS(x,1);
+
+  0:GCSPR_EL1=&x[1];
+  0:X0=x;
```
relokin:
```suggestion
```

catalogue/aarch64-GCS/tests/coWR+pogcspushmgcspopm.litmus
```diff
@@ -0,0 +1,15 @@
+AArch64 coWR+pogcspushmgcspopm
+variant=shadowstack
+{
+
+  SS(x,1);
+
+  0:GCSPR_EL1=&x[1];
+  0:X0=x;
```
relokin:
```suggestion
```

catalogue/aarch64-GCS/tests/coWR+pogcsss1gcsss2.litmus
```diff
@@ -0,0 +1,14 @@
+AArch64 coWR+pogcsss1gcsss2
+variant=shadowstack
+{
+  SS(x,2);
+  SS(y,1) = ssval_t: {SSCap(y,1)};
```
relokin, line 5: the semantics of this is that you initialise a shadow stack at the symbolic location y, which has one entry and that entry will have the value {SSCap(y,1)}? Isn't this value more like a list? Is there a reason why you didn't use [] rather than {}?

relokin (commented):
As we are getting closer to merging this do you also want to remove wip from the PR title and the commit titles?

catalogue/aarch64-GCS/tests/Co+gcsss1+gcsss1.litmus
murzinv: Good point about E3. However, I do not see how two SS1 instructions can access to the same location:
 - 1st SS1 would require location to hold valid cap and update location to in-progress cap on success
 - 2nd SS1 would again require location to hold valid cap, but it is in-progress due to 1st SS1, so it would fault
 :thinking:

lib/symbValue.ml
murzinv: SS(x,2) initialized to 0 by default, bit[1:0] are 0b00 thus , 0:X0=0 and no fault generated.

catalogue/aarch64-GCS/tests/coWR+pogcsss1gcsss2.litmus
murzinv: > the semantics of this is that you initialise a shadow stack at the symbolic location y, which has one entry and that entry will have the value {SSCap(y,1)}?

Correct

> Isn't this value more like a list? Is there a reason why you didn't use [] rather than {}?

Under the hood this is an array so I used array initialization syntax {} (even for single entry). I have no strong opinion of if to use [] or {}...

relokin: Thanks we can stick with {}!

lib/symbValue.ml
relokin: I think that's fine, thanks for elaborating on this.

catalogue/aarch64-GCS/tests/Co+gcsss1+gcsss1.litmus
relokin: Which begs the question why do we need this ordering rule. In any case, I think that's a good litmus test to have and as we clarify these ordering rules, we can see if this is still needed.

murzinv:
@relokin I’ve just rebased the PR, but I’m seeing a regression in `test.gcs` after the rebase. Reverting 6d26543b0f398f46cceb5de8409b9237c9829871 ("[herd] Improve the delaying monad operator") seems to restore the tests.

It may take some time to understand the root cause, unless @maranget has any insight into what I might be missing...

herd/AArch64Sem.ml
```diff
@@ -3624,6 +3630,376 @@ Arguments:
...
+            (fun ac ma mv ->
+              if is_branching && Access.is_physical ac then
+                M.bind_ctrldata_data ma mv (fun a v -> mop ac a v)
+              else
+                ma >>| mv >>= fun (a,v) -> mop ac a v)
+            (to_perms "w" quad)
+            ma
+            mzero
+            an
+            ii) in
```
relokin, line 3708: I think you don't need the M.delay here:
```suggestion
          let* a_virt = read_reg_addr rA ii in
          let* a_virt = M.add a_virt (V.intToV (-off)) in
          let mop ac a _v =
             write_reg rA a_virt ii >>|
             M.data_input_next
                (read_reg_data rs ii)
                (fun v -> GCSSem.write ac an a v ii)
             >>! B.nextSetT rA a_virt in
          lift_memop rA Dir.W true false
            (fun ac ma _mv ->
              if is_branching && Access.is_physical ac then
                M.bind_ctrldata ma (mop ac)
              else
                ma >>= mop ac)
            (to_perms "w" quad)
            (M.unitT a_virt)
            mzero
            an
            ii in
```

herd/AArch64Sem.ml
```diff
@@ -3624,6 +3630,376 @@ Arguments:
...
+          (fun ac ma _mv ->
+            if Access.is_physical ac then
+              M.bind_ctrldata ma (mop ac)
+            else
+              ma >>= mop ac)
+          (to_perms "r" quad)
+          ma
+          mzero
+          an
+          ii) in
```
relokin, line 3754: Similarly:
```suggestion
        let* a_virt = read_reg_addr rA ii in
        let mop ac a =
            let m = GCSSem.read ac an a ii >>= fun v ->
              let commit = commit_pred_txt (Some "PCAligned") ii in
              let mok =
                let(>>*=) = M.bind_control_set_data_input_first in
                commit >>*= fun () ->
                  M.add a_virt (V.intToV off) >>= fun new_addr ->
                    write_reg rd v ii >>|
                    write_reg rA new_addr ii
                    >>= M.ignore >>= B.next1T in
              let mask = V.intToV 0x3 in
               GCSSem.get_cap mask v >>= fun cap ->
               M.delay_kont "gcspopm(fault)"
               (M.op Op.Ne cap V.zero)
               (fun nonzero action ->
                 let open FaultType.AArch64 in
                 let mno = GCSSem.mk_fault action (GCSCheck POPM) ii in
                 M.choiceT nonzero mno mok)
            in
            (* Write to Rd depends on read from Shadow Stack *)
            M.short (E.is_mem_load) (is_this_reg rd) m in
          lift_memop rA Dir.R false false
          (fun ac ma _mv ->
            if Access.is_physical ac then
              M.bind_ctrldata ma (mop ac)
            else
              ma >>= mop ac)
          (to_perms "r" quad)
          (M.unitT a_virt)
          mzero
          an
          ii in
```

relokin (commented):
Sorry I am not entirely sure I understand what is happening with M.delay_kont, but I do think that in some cases you don't really need to use it.

murzinv:
Ok. Here is what happened.

To make debug easier I excluded VMSA bits and did it on top of `Introduce basic support for GCS`. Here `make test.gcs` fails for `tests/instructions/AArch64.gcs/G001.litmus` which now produces both faulting and non-faulting states
```
Test G001 Required
States 2
  ~Fault(P0:L0,GCS:POPM);
 Fault(P0:L0,GCS:POPM);
No
Witnesses
Positive: 1 Negative: 2
```

Since revering 6d26543b0f398f46cceb5de8409b9237c9829871 ("[herd] Improve the delaying monad operator") brings it back to expected state, I had a close look at use of `delay_kont` which is, in case of `G001.litmus`,  only present in
```
        | I_GCSPOPM rd ->
          check_gcs inst;
          let open AArch64Base in
          let rA = SysReg GCSPR_EL1
          and off = MachSize.nbytes quad in
          read_reg_addr rA ii >>= fun addr ->
            let m = GCSSem.read Annot.N addr ii
            >>= fun v ->
              let commit = 
                commit_pred_txt (Some "PCAligned") ii in
              let mok =
                (* Write to GCSPR depends on earlier read *)
                let(>>*=) = M.bind_control_set_data_input_first in
                commit >>*=
                fun () ->
                  !!(M.add addr (V.intToV off) >>= fun new_addr ->
                      write_reg rd v ii >>| 
                      write_reg rA new_addr ii) in
              let mask = V.intToV 0x3 in
              GCSSem.get_cap mask v >>= fun cap ->
              M.delay_kont "gcspop(fault)"
              (M.op Op.Ne cap V.zero)
              (fun nonzero action ->
                let open FaultType.AArch64 in
                let mno = GCSSem.mk_fault action (GCSCheck POPM) ii in
                M.choiceT nonzero mno mok)
              (* Write to Rd depends on read from Shadow Stack *)
              in M.short (E.is_mem_load) (is_this_reg rd) m
```
let's zoom it in
```
            (M.op Op.Ne cap V.zero)
              (fun nonzero action ->
                let open FaultType.AArch64 in
                let mno = GCSSem.mk_fault action (GCSCheck POPM) ii in
                M.choiceT nonzero mno mok)
```

It seems that prior 6d26543b0f398f46cceb5de8409b9237c9829871 ("[herd] Improve the delaying monad operator") `action` is implicitly (?) consumed (?) on both arms of `M.choiceT`, while after the change `action` seems to be expected to be consumed explicitly, so in code above it is done for `mno`, but not `mok`. Adding `action` to `mok` seems to make problem go away.

```
diff --git a/herd/AArch64Sem.ml b/herd/AArch64Sem.ml
index fe2d0938a..ce71c2262 100644
--- a/herd/AArch64Sem.ml
+++ b/herd/AArch64Sem.ml
@@ -4914,6 +4914,7 @@ Arguments:
               (fun nonzero action ->
                 let open FaultType.AArch64 in
                 let mno = GCSSem.mk_fault action (GCSCheck POPM) ii in
+                let mok = action >>= fun _ -> mok in
                 M.choiceT nonzero mno mok)
               (* Write to Rd depends on read from Shadow Stack *)
               in M.short (E.is_mem_load) (is_this_reg rd) m
```

If my analysis makes sense, I will update/redo the PR to match new semantic of `delay_kont`

relokin:
> It seems that prior [6d26543](https://github.com/herd/herdtools7/commit/6d26543b0f398f46cceb5de8409b9237c9829871) ("[herd] Improve the delaying monad operator") `action` is implicitly (?) consumed (?) on both arms of `M.choiceT`, while after the change `action` seems to be expected to be consumed explicitly, so in code above it is done for `mno`, but not `mok`. Adding `action` to `mok` seems to make problem go away.
>

That makes sense, thanks for looking into it.

I am a bit surprised that this was not flagged as an error by the solver, but indeed you use the result of `action` and therefore it need to be part of the final event_structure.
 
> ```
> diff --git a/herd/AArch64Sem.ml b/herd/AArch64Sem.ml
> index fe2d0938a..ce71c2262 100644
> --- a/herd/AArch64Sem.ml
> +++ b/herd/AArch64Sem.ml
> @@ -4914,6 +4914,7 @@ Arguments:
>                (fun nonzero action ->
>                  let open FaultType.AArch64 in
>                  let mno = GCSSem.mk_fault action (GCSCheck POPM) ii in
> +                let mok = action >>= fun _ -> mok in
>                  M.choiceT nonzero mno mok)
>                (* Write to Rd depends on read from Shadow Stack *)
>                in M.short (E.is_mem_load) (is_this_reg rd) m
> ```
> 
> If my analysis makes sense, I will update/redo the PR to match new semantic of `delay_kont`

Yes that makes sense, please go ahead.

maranget:
Hi @murzinv and @relokin. The change was introduced to permit not using `action` at all...  Here `action` contains the equation generated by `M.op Op.Ne cap V.zero`. Getting rid of this equation in the "else" branch of `M.choiceT` amounts to discard a constraint, leading to extra executions that this contraint would make contradictory...

We should probably add an warning in a comment before the signatire of `delay_kont` in monad.mli

murzinv:
> Hi @murzinv and @relokin. The change was introduce to permit not using `action` at all... Here `action` contains the equation generated by `M.op Op.Ne cap V.zero`. Getting rid of this equation in the "else" branch of `M.choiceT` amounts to discard a constraint, leading to extra executions that this contraint would make contradictory...

Thank for confirming @maranget !

> We should probably add an warning in a comment before the signatire of `delay_kont` in monad.mli

That would be handy, IMO

relokin:
@murzinv is this ready to be merged?

murzinv:
> @murzinv is this ready to be merged?

@relokin I do not see regressions anymore, so yes, it is ready, yet you might want glance over it once again in case there are any issues I did not notice.

relokin:
Great! Can you please rebase once more and we can merge.

murzinv:
> Great! Can you please rebase once more and we can merge.

@relokin done.

relokin:
Merged thanks Vladimir!

PR #1508, opened 2025-10-21
[tools] Add no-hash option to mprog

HadrienRenaud:
`mprog` is a great tool to automatically format litmus tests. However, it adds the hash of the litmus test to the information provided in the test, which is not always wanted, for example in regression tests.

This PR changes the default behaviour of `mprog7` **not** to add the hash of a test.
This PR also adds an option to `mprog7` to add the hash of a test.

For example, on a test that does not contain a hash before:
```
❯ mprog7 -mode text ./herd/tests/instructions/AArch64/A01.litmus
AArch64 A01

{
 uint64_t x=0;
 uint64_t 0:X0=0; 0:X1=x; 0:X2=0;
}
 P0                   ;
 LDRB W0,[X1,X2,UXTW] ;

exists (0:X0=0)

❯ mrpog7 -mode text -set-hash true ./herd/tests/instructions/AArch64/A01.litmus
AArch64 A01
Hash=fac9c0d41ef2dc619014a5f8545de0ad

{
 uint64_t x=0;
 uint64_t 0:X0=0; 0:X1=x; 0:X2=0;
}
 P0                   ;
 LDRB W0,[X1,X2,UXTW] ;

exists (0:X0=0)

```

This PR does not change the behaviour of `mprog7` when the hash is already in the litmus test (which is not to recompute nor change it).

### DIFF
--- a/lib/genParser.ml
+++ b/lib/genParser.ml
@@ -22,6 +22,7 @@
 module type Config = sig
   val debuglexer : bool
   val verbose : int
+  val set_hash : bool
   val check_kind : string -> ConstrGen.kind option
   val check_cond : string -> string option
 end
@@ -30,6 +31,7 @@ module DefaultConfig = struct
   let debuglexer = false let verbose = 0
   let check_kind _ = None
   let check_cond _ = None
+  let set_hash = true
 end
 
 (* input signature, a lexer and a parser for a given architecture *)
@@ -237,6 +239,7 @@ module Make
               MiscParser.condition =
               ConstrGen.set_kind k parsed.MiscParser.condition; } in
       let parsed =
+        if not O.set_hash then parsed else
         match MiscParser.get_hash parsed with
         | None ->
              let info = parsed.MiscParser.info in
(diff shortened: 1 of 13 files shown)


lib/genParser.ml
```diff
@@ -237,6 +239,7 @@ module Make
               MiscParser.condition =
               ConstrGen.set_kind k parsed.MiscParser.condition; } in
       let parsed =
+        if not O.set_hash then parsed else
```
maranget, line 242: Wouldn't be more simple to intervene at the printing step? This may involve less changes of configuration modules.

HadrienRenaud: No because that would change the behaviour when the test already contains a hash (and does not recompute it).

maranget: Understood,  so this PR will clash with mine. Stay tuned.

maranget:
LGTM, please merge so that I can build my `mshowhashes7 -rehash true...` on top of this.

PR #1513, opened 2025-10-22
[herd] Handle memory access on non-symbolic locations by failling

maranget:
This PR is an attempt to flag wrong memory accesses as fatal errors.

Consider for instance:
```
AArch64 A217
(* This test should fail on an attempt to dereference '2' *)
{
 [x]=2;
 0:X0=x;
}
 P0          ;
 LDR X1,[X0] ;
 LDR X2,[X1] ;

locations [0:X1;]
forall (true)
```
We had:
```
% herd7 A217.litmus
Test A217 Required
States 0
Ok
Witnesses
Positive: 0 Negative: 0
Condition forall (true)
Observation A217 Never 0 0
Time A217 0.00
```
The test is deterministic and one expects one execution that attempts to access to `[2]`. However, this execution disappeared.

With this PR we have:
```
% herd7 A217.litmus
Warning: File "A217.litmus": Non-symbolic memory access found on '[2]' (User error) 
```
Which matches more our expectations.

### DIFF
--- a/herd/tests/instructions/AArch64/A217.litmus
+++ b/herd/tests/instructions/AArch64/A217.litmus
@@ -1,0 +1,12 @@
+AArch64 A217
+(* This test should fail on an attempt to dereference '2' *)
+{
+ [x]=2;
+ 0:X0=x;
+}
+ P0          ;
+ LDR X1,[X0] ;
+ LDR X2,[X1] ;
+
+locations [0:X1;]
+forall (true)
--- a/herd/tests/instructions/AArch64/A217.litmus.expected-failure
+++ b/herd/tests/instructions/AArch64/A217.litmus.expected-failure
@@ -1,0 +1,1 @@
+Warning: File "./herd/tests/instructions/AArch64/A217.litmus": Non-symbolic memory access found on '[2]' (User error)
--- a/herd/tests/instructions/AArch64/A218.litmus
+++ b/herd/tests/instructions/AArch64/A218.litmus
@@ -1,0 +1,13 @@
+AArch64 A218
+(* This test should succeed *)
+{
+ int64_t *p = &x;
+ int64_t x = 2;
+ 0:X0=p;
+}
+ P0          ;
+ LDR X1,[X0] ;
+ LDR X2,[X1] ;
+
+locations [0:X1;0:X2;]
+forall (true)
--- a/herd/tests/instructions/AArch64/A218.litmus.expected
+++ b/herd/tests/instructions/AArch64/A218.litmus.expected
@@ -1,0 +1,10 @@
+Test A218 Required
+States 1
+0:X1=x; 0:X2=2;
+Ok
+Witnesses
+Positive: 1 Negative: 0
+Condition forall (true)
+Observation A218 Always 1 0
+Hash=7337f658da576c132ba6a403026111a6
+
--- a/herd/tests/instructions/AArch64/A219.litmus
+++ b/herd/tests/instructions/AArch64/A219.litmus
@@ -1,0 +1,17 @@
+AArch64 A219
+(* This test should fail, because some valid execution
+   attempts to dereference an integer *)
+{
+ int x=1;
+ int *p=x;
+ 0:X2=p;
+ 1:X2=p;
+ 2:X2=p;
+}
+ P0          | P1          | P2          ;
+ LDR X2,[X2] | STR X2,[X2] | LDR X2,[X2] ;
+ MOV X0,#2   |             | LDR X0,[X2] ;
+ STR X0,[X2] |             |             ;
+
+locations [0: X2; x; 2: X2; 2:X0; p]
+forall (true)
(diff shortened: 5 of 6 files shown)


herd/mem.ml
```diff
@@ -1103,26 +1116,42 @@ let match_reg_events es =
...
       let possible =
         List.map
           (fun (er,ws) ->
-            let ws = List.map (fun w -> Some w) ws in
+            let ws = List.map (fun w -> Write w) ws in
+            let ws =
+              let v = S.E.global_loc_of er in
+              match v with
+              | Some (V.Var _ as v) -> NoSymbol v::ws
+              | None|Some (V.Val _) -> ws in
```
HadrienRenaud, line 1152: Shouldn't we check that it's symbolic here?

```suggestion
              | None | Some (V.Val (Constant.Symbol _) -> ws
              | Some (V.Val _ as v) -> NoSymbol v :: ws in
```

lib/op.ml
```diff
@@ -135,6 +135,7 @@ type 'aop op1 =
   | IsInstr (* Check nature of constant *)
   | Promote (* Promote to higher precision *)
   | Demote  (* Demote to lower precision *)
+  | IsNotSymbolic  (* Runtime check *)
```
HadrienRenaud: nit: maybe we could invert it to `IsSymbolic`? I find having too many 'not' in a sentence make it hard to understand

nit: maybe we could name it with `Check[Not]Symbolic`? I would like to make it explicit that this will crash if it's not symbolic

nit: maybe say that it's for internal use in `mem` only?

herd/mem.ml
maranget: No really, if the location is a non-symbolic constant, this will be caught later by the general mechanism.

HadrienRenaud: Ok nice thanks

lib/op.ml
maranget: I chose "CheckSymbolic" and add the remark on internal use.

akiyks:
Hi @maranget,

After this was merged, I get:

```
Warning: File "/[...]/C-RomanPenyaev-list-rcu-rr.litmus": Non-symbolic memory access found on '[1]' (User error)
```

in running herd7 (with LKMM) against C-RomanPenyaev-list-rcu-rr.litmus available at:

https://github.com/paulmckrcu/litmus/blob/master/manual/kernel/C-RomanPenyaev-list-rcu-rr.litmus

Can you have a look into it?  Does the litmus test need any change?

As of commit e67da03607e8 ("Merge pull request #1515 from herd/aslref-parser-strict"), I got:

```
Test C-RomanPenyaev-list-rcu-rr Allowed
States 7
0:r1=w; 1:r1=w; 1:r2=y; 1:r3=y; 1:r4=z; [c]=z; [v]=w; [w]=y; [x]=0; [y]=z;
0:r1=x; 1:r1=w; 1:r2=x; 1:r3=w; 1:r4=y; [c]=y; [v]=w; [w]=y; [x]=0; [y]=z;
0:r1=x; 1:r1=w; 1:r2=x; 1:r3=x; 1:r4=y; [c]=0; [v]=w; [w]=y; [x]=0; [y]=z;
0:r1=x; 1:r1=w; 1:r2=x; 1:r3=x; 1:r4=y; [c]=y; [v]=w; [w]=y; [x]=0; [y]=z;
0:r1=y; 1:r1=w; 1:r2=x; 1:r3=x; 1:r4=y; [c]=y; [v]=w; [w]=y; [x]=0; [y]=z;
0:r1=y; 1:r1=w; 1:r2=y; 1:r3=y; 1:r4=z; [c]=z; [v]=w; [w]=y; [x]=0; [y]=z;
0:r1=z; 1:r1=w; 1:r2=y; 1:r3=y; 1:r4=z; [c]=z; [v]=w; [w]=y; [x]=0; [y]=z;
No
Witnesses
Positive: 0 Negative: 7
Condition exists (1:r1=0 \/ 1:r2=0 \/ 1:r3=0 \/ 1:r4=0)
Observation C-RomanPenyaev-list-rcu-rr Never 0 7
Time C-RomanPenyaev-list-rcu-rr 0.32
Hash=81ddf6a12ded1accf54318919160be87
```

akiyks:
Ahh, you mean something like:

```
diff --git a/manual/kernel/C-RomanPenyaev-list-rcu-rr.litmus b/manual/kernel/C-RomanPenyaev-list-rcu-rr.litmus
index 58bd2861..64bfc6ca 100644
--- a/manual/kernel/C-RomanPenyaev-list-rcu-rr.litmus
+++ b/manual/kernel/C-RomanPenyaev-list-rcu-rr.litmus
@@ -1,7 +1,7 @@
 C C-RomanPenyaev-list-rcu-rr
 
 {
-       int *z=1; (* List: v->w->x->y->z. Noncircular, but long enough. *)
+       int *z=t; (* List: v->w->x->y->z. Noncircular, but long enough. *)
        int *y=z;
        int *x=y;
        int *w=x;
```

Looks like this works.
I'll test more and see if there are any other tests in Paul's litmus repo who want similar tweaks.
Thanks you!

PR #1514, opened 2025-10-23
[aslspec] fixed parser error and made parser warnings into errors

Roman-Manevich:


asllib/aslspec/dune
```diff
@@ -10,7 +10,7 @@
 
 (menhir
  (modules SpecParser)
- (flags --unused-tokens --explain))
+ (flags --unused-tokens --explain --strict))
```
maranget, line 13: Fine, I was not aware of this `-strict` option. Perhaps we can use it everywhere?

Roman-Manevich: Let's do it in a different PR: https://github.com/herd/herdtools7/pull/1515

PR #1516, opened 2025-10-23
[gen] Introduce more wildcard syntax in `diy`

ShaleXIONG:
This pull request introduce some new wildcard syntax so it matches better to the cat file. For better usage, all wildcard syntax now are only accepted in `diycross7` and `diy7`.

For example, it is now allowed to use relaxation syntax like `[DpAddr Po**W]` which will be unfolded to `[DpAddrdW PodWW] [DpAddrdW PosWW] [DpAddrsW PodWW] [DpAddrsW PosWW] [DpAddrdR PodRW] [DpAddrdR PosRW] [DpAddrsR PodRW] [DpAddrsR PosRW]`. This provides closer syntax to those used in `cat` file.

The new wildcard syntax that automatically expanded as the following:
- wildcard `*` for different and same location, this can be used in `Po*WR` or `DMB.SY*W*` for example.
- `Amo` -> `Amo.Swp, Amo.Cas, Amo.LdAdd, Amo.LdSub .....`
- `Po` for (1) all possible location, different or same, and (2) all possible read and write combinations.
- `DpAddr`, `DpData`, `DpCtrl`, `DpAddrCsel`, `DpDataCsel`, `DpCtrlCsel`, for all possible location and all possible second memory event. However, `DpData` only have write hence it only unfold to `DpDatasW` and `DpDatadW`.
- Any fence for example `ISB***` (1) all possible location, different or same, and (2) all possible read and write combinations. Note that here it MUST explicit write `***` as `ISB` is already used for peudo-edge (i.e., no memory event edge)
- `Co` will be unfolded to `Coe` and `Coi`, similar to `Fr` and `Rf`.

We fix an inconsistent between the backward capability syntax of `Dp` and `Ctrl`. Now they all have `wildcard` syntax (while previously it only exists for `Ctrl`).

### DIFF
--- a/gen/alt.ml
+++ b/gen/alt.ml
@@ -70,6 +70,10 @@ module Make(C:Builder.S)
           (fun f -> List.exists (equal_fence f) fs)
 
     let choice_sc po_safe e1 e2 =
+      let seq_sd e1 e2 =
+        match Code.seq_sd e1 e2 with
+        | None -> Warn.user_error "Unexpected UnspecLoc"
+        | Some b -> b in
       let r = match e1.edge,e2.edge with
 (*
   Now accept internal with internal composition
@@ -564,7 +568,7 @@ module Make(C:Builder.S)
 
     let count_ext es = count_e 0 es
 
-    let change_loc e = not @@ Code.is_same_loc @@ loc_sd e
+    let change_loc e = Code.is_diff_loc @@ loc_sd e
 
     let count_p p =
       List.fold_left ( fun acc x -> if p x then acc + 1 else acc ) 0
--- a/gen/common/code.ml
+++ b/gen/common/code.ml
@@ -97,18 +97,21 @@ let pp_sd = function
 
 let is_same_loc = function
   | Same -> true
-  | Diff -> false
-  | UnspecLoc -> assert false
+  | _ -> false
 
-let is_both_loc = function
+let is_diff_loc = function
+  | Diff -> true
+  | _ -> false
+
+let is_unspec_loc = function
   | UnspecLoc -> true
   | _ -> false
 
 let seq_sd sd1 sd2 =
   match sd1,sd2 with
-  | UnspecLoc,_|_, UnspecLoc -> assert false
-  | Same,Same -> Same
-  | Diff,_|_,Diff -> Diff
+  | UnspecLoc,_|_, UnspecLoc -> None
+  | Same,Same -> Some Same
+  | Diff,_|_,Diff -> Some Diff
 
 let fold_ie wildcard f r = let r = if wildcard then (f UnspecCom r) else r in f Ext (f Int r)
 let fold_sd wildcard f r = let r = if wildcard then (f UnspecLoc r) else r in f Diff (f Same r)
--- a/gen/common/code.mli
+++ b/gen/common/code.mli
@@ -55,9 +55,10 @@ val pp_ie : ie -> string
 val pp_dir : dir -> string
 val pp_extr : extr -> string
 val pp_sd : sd -> string
-val seq_sd : sd -> sd -> sd
+val seq_sd : sd -> sd -> sd option
 val is_same_loc : sd -> bool
-val is_both_loc : sd -> bool
+val is_diff_loc : sd -> bool
+val is_unspec_loc : sd -> bool
 (* The first boolean indicates whether wildcard syntax is included in the fold *)
 val fold_ie : bool -> (ie -> 'a -> 'a) -> 'a -> 'a
 val fold_extr : bool -> (extr -> 'a -> 'a) -> 'a -> 'a
--- a/gen/common/edge.ml
+++ b/gen/common/edge.ml
@@ -388,7 +388,7 @@ let pp_dp_default tag sd e = sprintf "%s%s%s" tag (pp_sd sd) (pp_extr e)
   | Po (sd,_,_) | Fenced (_,sd,_,_) | Dp (_,sd,_) -> sd
   | Insert _|Store|Node _|Fr _|Ws _|Rf _|Hat|Rmw _|Id|Leave _|Back _ -> Same
 
-  let do_is_diff e = not @@ Code.is_same_loc @@ do_loc_sd e
+  let do_is_diff e = Code.is_diff_loc @@ do_loc_sd e
 
 let fold_tedges_compat f r =
   let r = fold_ie wildcard (fun ie -> f (Ws ie)) r in
@@ -445,7 +445,7 @@ let fold_tedges f r =
   let ok_non_rmw e a1 a2 =
     (* `do_is_diff` is safe to call when `e` is not
        wildcard `*`/UnspecLoc location. *)
-    Code.is_both_loc @@ do_loc_sd e ||
+    Code.is_unspec_loc @@ do_loc_sd e ||
     do_is_diff e || do_disjoint ||
     (overlap_atoms a1 a2 &&
      not (do_strict_overlap && same_access_atoms a1 a2))
@@ -1039,7 +1039,11 @@ let fold_tedges f r =
 
 (* compact *)
 
-  let seq_sd e1 e2 = Code.seq_sd (loc_sd e1) (loc_sd e2)
+  let seq_sd e1 e2 =
+    match Code.seq_sd (loc_sd e1) (loc_sd e2) with
+    | None -> Warn.user_error "Unexpected UnspecLoc"
+    | Some b -> b
+
 
   let fst_dp e1 e2 k = match e1.edge with
   | Dp (d,_,_) ->
--- a/gen/cycle.ml
+++ b/gen/cycle.ml
@@ -470,7 +470,7 @@ let next_loc e ((loc0,lab0),vs) = match E.is_fetch e with
 
 let same_loc e = Code.is_same_loc @@ E.loc_sd e
 
-let diff_loc e = not (same_loc e)
+let diff_loc e = Code.is_diff_loc @@ E.loc_sd e
 
 let same_proc e = E.get_ie e = Int
 let diff_proc e = E.get_ie e = Ext


gen/alt.ml
```diff
@@ -157,6 +157,7 @@ module Make(C:Builder.S)
                 | (Same,Same) | (Diff,Same) | (Same,Diff)
                   -> true
                 | Diff,Diff -> false
+                | _ -> assert false
```
relokin, line 168: This is not needed, is it? All cases are covered above. Otherwise the compiler would complain about it.

gen/common/code.ml
```diff
@@ -89,7 +89,7 @@ type dir = W | R
 type ie = Int|Ext
 
 (* Change of location across edge *)
-type sd = Same|Diff
+type sd = Same|Diff|Both
```
relokin: Would unspecified be a better name?

gen/common/edge.ml
```diff
@@ -312,15 +312,19 @@ and type rmw = F.rmw = struct
...
     | Fr ie -> sprintf "Fr%s" (pp_ie ie)
     | Ws ie -> if compat then sprintf "Ws%s" (pp_ie ie) else sprintf "Co%s" (pp_ie ie)
-    | Po (sd,e1,e2) -> sprintf "Po%s%s%s" (pp_sd sd) (pp_extr e1) (pp_extr e2)
-    | Fenced (f,sd,e1,e2) ->
-        sprintf "%s%s%s%s" (F.pp_fence f) (pp_sd sd) (pp_extr e1) (pp_extr e2)
-    | Dp (dp,sd,e) -> sprintf "Dp%s%s%s"
-          (F.pp_dp dp) (pp_sd sd) (pp_extr e)
+    | Po (sd,e1,e2) ->
+      if all_wildcard sd e1 e2 then sprintf "Po"
+      else sprintf "Po%s%s%s" (pp_sd sd) (pp_extr e1) (pp_extr e2)
```
relokin: Feel free to keep your version but you could also use the match construct for this
```suggestion
    | Po (Both,Irr,Irr) ->
      sprintf "Po"
    | Po (sd,e1,e2) ->
      sprintf "Po%s%s%s" (pp_sd sd) (pp_extr e1) (pp_extr e2)
```

gen/common/edge.ml
```diff
@@ -312,15 +312,19 @@ and type rmw = F.rmw = struct
...
+    | Fenced (f,sd,e1,e2) -> sprintf "%s%s%s%s" (F.pp_fence f) (pp_sd sd) (pp_extr e1) (pp_extr e2)
```
relokin: Why are you not changing this? Is it because a DMB can also be an Insert edge? If you're not changing this on purpose can you please revert the change?

gen/common/edge.ml
```diff
@@ -312,15 +312,19 @@ and type rmw = F.rmw = struct
...
+    | Dp (dp,sd,e) ->
```
relokin, line 305: as for Po

else if you want to keep the commits separate this goes into the next commit, doesn't it?

gen/edge.ml
```diff
@@ -440,6 +440,9 @@ let fold_tedges f r =
   let r =
     F.fold_dpw
       (fun dp -> fold_sd (fun sd -> f (Dp (dp,sd,Dir W)))) r in
+  let r =
+    F.fold_dpw
+      (fun dp -> fold_sd (fun sd -> f (Dp (dp,sd,Irr)))) r in
```
relokin: Sorry I am not sure I understand why you are changing this, can you please elaborate?

gen/common/AArch64Arch_gen.ml
```diff
@@ -797,7 +797,7 @@ let sequence_dp (d1,c1) (d2,c2) = match c1 with
 let expand_dp_dir (dir,_) = D.expand_dp_dir dir
 
 (* Read-Modify-Write *)
-type rmw =  LrSc | LdOp of atomic_op | StOp of atomic_op | Swp | Cas
+type rmw =  LrSc | LdOp of atomic_op | StOp of atomic_op | Swp | Cas | AllAmo
```
relokin, line 771: I am not sure I understand this. What's the point? `Amo` appears as a possible in diyone7 and when I tried I got:
```
$> diyone7 -arch AArch64 Amo Rfe PosRW Rfe -oneloc
Fatal error: exception File "gen/AArch64Arch_gen.ml", line 896, characters 16-22: Assertion failed
```
If I understand correctly the intent is for diy7 to go through all possible Amo, isn't it? Is there a way for this to not appear. as a valid relaxation in diyone7?

relokin: And by the way this seems to be the case for `Dp**` and I guess many others, some of these shouldn't appear as relaxation options for diyone7.

gen/libdir/forbidden.conf
```diff
@@ -45,11 +45,152 @@
...
+# `Hat` is an a read-read communication edge in `diy7`, namely
+# the end of a thread and the beginning of antoher thread are both read events
+# and they will read the same value. We swith them off by default.
+###
+#-safe [Hat Fre]
+#-safe [Hat DpAddr] [Hat DpData] [Hat DpCtrl]
+#-safe [Hat DMB.SY***] [Hat DMB.LD***]
+#-safe [Hat A PodR*] [Hat Q PodR*]
+#-safe [Hat Amo.Swp] [Hat Amo.Cas] [Hat Amo.LdAdd] [Hat Amo.StAdd]
+#-safe [Hat Amo]
```
relokin: Why have you disabled these?

ShaleXIONG:
> It looks like PR has side effect on `forbidden_ifetch.conf`
> 
> 
> 
> Commit ``[gen] Flag `allow_back` in diy7 by default, which is consistent with diyone7`` slightly increases number of generated tests from `269` to `272`, and it looks like new `3` tests are allowed.
> 
> 
> 
> Next, commit `[gen] Allow generate pseudo edge followed by communication by default in diy` increases number of generated tests further to `1506`, `11` of which are allowed and `3` seems have unreachable final clause (`herd7` reports `Observation Never` if run with `-through all`)

Let me pend these two commits to another pull. It seems I need to fix the ifetch configuration file first.

gen/common/AArch64Arch_gen.ml
ShaleXIONG: I think I can systematically add a flag to only introduce those wildcard syntax in diy7 and diycross7. Let me investigate a bit here.

[update] Done. Wildcard syntax is only accepted on `diycross7` and `diy7` now.

gen/common/edge.ml
ShaleXIONG: Oh this is better, I will try to avoid `if` as possible in general.

gen/libdir/forbidden.conf
ShaleXIONG: Sorry is a rebase mistake. I meant to enable them. Thanks for spotting it.

gen/common/edge.ml
ShaleXIONG: You are asking why not introduce `DMB` similar to `Po` which unfold into `PodRR`, `PodWR` ...? If so, yes there will be syntax collision as all fences such as `DMB` and `ISB` already are used for peudo edge. Before this commit, syntax like `ISBd**`, in this change we introduce wildcard `*` for `d`/`s` as well, so `ISB***` will become valid syntax in `diy7` and `diyone7`.

relokin: Thanks!

Doesn't that mean that we don't need `ISBd**`, we can achieve the same thing with `[Pod** ISB]`, can't we?

ShaleXIONG: Yes and no.
- Yes as if a test can be generated, for example, `PodWR ISB Fre PodWR Fre`, here `PodRW ISB` is equivalent to `ISBdRW`.
- No as `ISB` is a pseudo edges, which has restriction, namely pseudo edges (before this merge) must be next (left or right) to an external edge, hence the following cycles will be rejected:
    * `PosWR ISB Fri` 
    * `PodWR ISB PosRR Fre PodWR Fre`

However, the equivalent version using `ISBsWR` will be allowed. In fact, in this pull request I try to relax the first restriction in commit `55d4522` since it is just one line change. Yet it has a ripple effect on the configuration file for `ifetch` so I am planing to remove the commit from this pull for now and merge them in the future after revisiting the `ifetch` configuration file. Similarly, to relax the second restrict is certainly possible as well, which I guess will have some ripple effect too, so I shall be more careful here.

gen/edge.ml
ShaleXIONG: This is for introduce wildcard syntax for `dp` for example `dpaddr*`. Most the `fold` function in `diy` are used for built up the lookup table for parsing syntax. Hence we need to add the extra step in `fold` here for extra entry when the location is wildcard in dp.

gen/common/edge.ml
ShaleXIONG: Will check it !

gen/alt.ml
ShaleXIONG: We do need it as there is `Both` location now.

gen/common/code.ml
ShaleXIONG: Will change to `UnspecLoc`.

gen/common/code.mli
```diff
@@ -56,11 +56,14 @@ val pp_dir : dir -> string
...
 val pp_sd : sd -> string
 val seq_sd : sd -> sd -> sd
-val fold_ie : (ie -> 'a -> 'a) -> 'a -> 'a
-val fold_extr : (extr -> 'a -> 'a) -> 'a -> 'a
-val fold_sd : (sd -> 'a -> 'a) -> 'a -> 'a
-val fold_sd_extr : (sd -> extr -> 'a -> 'a) -> 'a -> 'a
-val fold_sd_extr_extr : (sd -> extr -> extr -> 'a -> 'a) -> 'a -> 'a
+val is_same_loc : sd -> bool
+val is_both_loc : sd -> bool
+(* The first boolean indicates on wildcard syntax *)
```
fsestini:
```suggestion
(* The first boolean indicates whether wildcard syntax is included in the fold *)
```

ShaleXIONG: Done.

gen/common/edge.ml
```diff
@@ -94,8 +93,8 @@ module type S = sig
   val compare_atomo : atom option -> atom option -> int
   val compare : edge -> edge -> int
 
-  val parse_atom : string -> atom
-  val parse_atoms : string list -> atom list
+  val parse_atom : string -> atom option
```
fsestini, line 96: I feel like the type signature of this function (and similar ones) is a bit confusing, since from the type alone one might think that this function returns `Some ...` on success and `None` on parsing failure, whereas here and throughout the module the `None` constructor of `atom option` is actually used to represent the "plain" atom `P`. Perhaps not in this PR, but in the future I think it would be great if we can:
- integrate `P` (which is really more similar to a wildcard) as a wildcard constructor of `atom`, similar to how  `sd`, `ie`, etc. have their own wildcard constructors
- add an additional constructor to `atom` to represent accesses that _must_ be plain. This is currently not possible since `P` behaves as a wildcard that defaults to plain access (but does not enforce it)

ShaleXIONG: You are right here, the change here is trying to match with the type `type edge = { edge: tedge; a1:atom option; a2: atom option; }`.  I agree with you the best way to progress here is to change `a1: atom option` to `a1: atom` and introduce `P` as a wildcard. We probably need to introduce a `M` for example so it will not be merged with other annotation.

Lets do it in a separate PR as you suggested.

gen/common/code.ml
```diff
@@ -92,18 +93,29 @@ let pp_extr = function
...
   | Same -> "s"
   | Diff -> "d"
+  | UnspecLoc -> "*"
+
+let is_same_loc = function
+  | Same -> true
+  | Diff -> false
+  | UnspecLoc -> assert false
+
+let is_both_loc = function
```
fsestini:
```suggestion
let is_unspec_loc = function
```

gen/common/code.mli
```diff
@@ -56,11 +56,14 @@ val pp_dir : dir -> string
 val pp_extr : extr -> string
 val pp_sd : sd -> string
 val seq_sd : sd -> sd -> sd
-val fold_ie : (ie -> 'a -> 'a) -> 'a -> 'a
-val fold_extr : (extr -> 'a -> 'a) -> 'a -> 'a
-val fold_sd : (sd -> 'a -> 'a) -> 'a -> 'a
-val fold_sd_extr : (sd -> extr -> 'a -> 'a) -> 'a -> 'a
-val fold_sd_extr_extr : (sd -> extr -> extr -> 'a -> 'a) -> 'a -> 'a
+val is_same_loc : sd -> bool
+val is_both_loc : sd -> bool
```
fsestini:
```suggestion
val is_unspec_loc : sd -> bool
```

gen/common/code.ml
```diff
@@ -92,18 +93,29 @@ let pp_extr = function
 let pp_sd = function
   | Same -> "s"
   | Diff -> "d"
+  | UnspecLoc -> "*"
+
+let is_same_loc = function
+  | Same -> true
+  | Diff -> false
+  | UnspecLoc -> assert false
```
fsestini: I'm a bit uneasy about throwing an exception on `UnspecLoc` here. As a consumer of the `Code` module, I would expect functions of type `sd -> bool` exposed by `Code` to be well-defined on every value of `sd`, especially considering `sd` is defined in that same `Code` module.

I understand that the question "is this the same location?", which `is_same_loc` is supposed to answer, is not very well defined now that `sd` includes wildcard constructor (as it's perhaps not definitively clear whether the wildcard should count as "same location" or not). But in that case, maybe we should update the interface exposed by `Code` to reflect this change to `sd`, and replace `is_same_loc` with a function that admits a non-partial definition. For instance:

```suggestion
let is_same_loc_strict = function
  | Same -> true
  | _ -> false
```

gen/common/code.ml
```diff
@@ -92,18 +93,29 @@ let pp_extr = function
...
+
+let is_both_loc = function
+  | UnspecLoc -> true
+  | _ -> false
 
 let seq_sd sd1 sd2 =
   match sd1,sd2 with
+  | UnspecLoc,_|_, UnspecLoc -> assert false
```
fsestini: Similarly to `is_same_loc`: I think this function should be changed (and possibly renamed) to accommodate the case where `sd1` or `sd2` is a wildcard. I'm not quite sure that kind of operation `seq_sd` implements, but in case such operation cannot be performed on wildcard constructors, perhaps a better way to handle this case is to have the function return `None` (so that the operation's partiality is clear from its type), and let the callers of this function handle the `None` case however they want (including possibly throwing an exception).

gen/common/code.ml
ShaleXIONG: I think the best way is I will introduce another function `is_diff_loc` too. I created the third case assert false, as previously other part of the code expects `not @@ is_same_loc` means `Diff`. However it is no longer the case. 

I will probably keep the function name as `is_same_loc` but do `| same -> true | _ -> false`

gen/common/code.mli
ShaleXIONG: Done.

gen/common/code.ml
ShaleXIONG: Done.

gen/common/code.ml
ShaleXIONG: I think `option` is a better idea.

ShaleXIONG: I just updated this part in a new commit.

fsestini (approved):
LGTM

PR #1517, opened 2025-10-23
ACL2 reorg and CAT interpreter

solswords:
Moves ACL2 ASL implementation from asllib/acl2 to acl2/asl
Adds ACL2 CAT interpreter in acl2/cat
Catches up on improvements to reasoning about ASL interpreter


slobodova (approved):
Changes affecting only ACL2 directory

PR #1518, opened 2025-10-24
[herd,litmus] Initial support for translating addresses of instructions

artkhyzha:
This PR adds basic support for address translation for instruction fetch into `herd7` and `litmus7`.

The syntax of litmus tests is extended with:
* a new annotation `.pagealign` that aligns the next instruction to the new page and inserts an immediate branch to that instruction;
* ability to specify a TTD of a label, so that we could write `TTD(P0:Page0)=...` where `Page0` is a label on P0;
* ability to specify an MMU fault as either coming from the data side (D-MMU) or the instruction side (I-MMU).

Functionality-wise, semantics of instruction fetch is reworked to benefit from the pre-existing support for address translation. Prior to the instruction fetch, the address of the instruction is translated, and a possibility of the address having been remapped onto a different page is considered.

---

**Ideas for future work.**

UI-wise, recommended, but not urgent:
* Currently herd hides register effects from graphs by default. This at times may result in "iico_ctrl" edges seemingly disappearing. With `-variant ifetch`, `iico_ctrl` edges may need to be displayed without '-showevents all'.
*  Printing of the instruction should be omitted from boxes until after the Branching effect for ifetch

Refactoring-wise, optional and just for consideration:
* `loader.ml` now makes use of a pseudoinstruction `Skip <imm>` denoting the offset of a padding until the next page. To make use of it, the Loader module does yet another transformation of the internal data structure for a litmus test. Wasted transformations aside, it would be an improvement to maintainability if the organisation of the transformations was clearer.
* `semExtra.ml` implements checks for the purposes of mixed-size mode, which are then used by `mem.ml` during the `rf` construction. One of such checks interacted with the way offsets are created for instruction addresses. That interaction received "duct-tape" treatment (a special case got added) and is marked as "FIXME". Ideally, offsets for instruction addresses would pass the usual mixed-size checks.
* support for labels as VAs resulted into the following: the name of a symbolic constant, when it corresponds to a VA, is of a type "Constant.Symbol", while the name of a symbolic constant, when it corresponds to a PA, is of a type "string". This mismatch results in odd-looking pattern matches. See [this comment](https://github.com/herd/herdtools7/pull/1518#discussion_r2727306930).

### DIFF
--- a/herd/tests/instructions/AArch64.PAC/A13.litmus.expected
+++ b/herd/tests/instructions/AArch64.PAC/A13.litmus.expected
@@ -1,6 +1,6 @@
 Test A13 Required
 States 1
- Fault(P0,x,MMU:Translation);
+ Fault(P0,x,D-MMU:Translation);
 Ok
 Witnesses
 Positive: 1 Negative: 0
--- a/herd/tests/instructions/AArch64.PAC/A14.litmus.expected
+++ b/herd/tests/instructions/AArch64.PAC/A14.litmus.expected
@@ -1,6 +1,6 @@
 Test A14 Required
 States 1
- Fault(P0,x,MMU:Translation);
+ Fault(P0,x,D-MMU:Translation);
 Ok
 Witnesses
 Positive: 1 Negative: 0
--- a/herd/tests/instructions/AArch64.kvm/A004.litmus.expected
+++ b/herd/tests/instructions/AArch64.kvm/A004.litmus.expected
@@ -1,6 +1,6 @@
 Test A004 Required
 States 1
- Fault(P0:L0,x,MMU:Translation);
+ Fault(P0:L0,x,D-MMU:Translation);
 Ok
 Witnesses
 Positive: 1 Negative: 0
--- a/herd/tests/instructions/AArch64.kvm/A013.litmus.expected
+++ b/herd/tests/instructions/AArch64.kvm/A013.litmus.expected
@@ -1,6 +1,6 @@
 Test A013 Required
 States 1
- Fault(P0:L0,UndefinedInstruction); Fault(P0:L1,x,MMU:Translation); Fault(P0:L2,y,MMU:AccessFlag);
+ Fault(P0:L0,UndefinedInstruction); Fault(P0:L1,x,D-MMU:Translation); Fault(P0:L2,y,D-MMU:AccessFlag);
 Ok
 Witnesses
 Positive: 1 Negative: 0
--- a/herd/tests/instructions/AArch64.kvm/F000.litmus.expected
+++ b/herd/tests/instructions/AArch64.kvm/F000.litmus.expected
@@ -1,6 +1,6 @@
 Test F000 Required
 States 1
- Fault(P0:L0,x,MMU:Translation);
+ Fault(P0:L0,x,D-MMU:Translation);
 Ok
 Witnesses
 Positive: 1 Negative: 0
--- a/herd/tests/instructions/AArch64.kvm/F002.litmus.expected
+++ b/herd/tests/instructions/AArch64.kvm/F002.litmus.expected
@@ -1,6 +1,6 @@
 Test F002 Required
 States 1
- Fault(P0:L0,x,MMU:Permission);
+ Fault(P0:L0,x,D-MMU:Permission);
 Ok
 Witnesses
 Positive: 1 Negative: 0
--- a/herd/tests/instructions/AArch64.kvm/MP-I2V+rel+dmb.ld.litmus.expected
+++ b/herd/tests/instructions/AArch64.kvm/MP-I2V+rel+dmb.ld.litmus.expected
@@ -1,6 +1,6 @@
 Test MP-I2V+rel+dmb.ld Allowed
 States 2
-1:X0=0; Fault(P1:L0,y,MMU:Translation);
+1:X0=0; Fault(P1:L0,y,D-MMU:Translation);
 1:X0=1;  ~Fault(P1:L0,y);
 No
 Witnesses
--- a/herd/tests/instructions/AArch64.kvm/MP-I2V+rel+noret-dmb.ld.litmus.expected
+++ b/herd/tests/instructions/AArch64.kvm/MP-I2V+rel+noret-dmb.ld.litmus.expected
@@ -1,6 +1,6 @@
 Test MP-I2V+rel+noret-dmb.ld Allowed
 States 2
-1:X0=0; Fault(P1:L0,y,MMU:Translation);
+1:X0=0; Fault(P1:L0,y,D-MMU:Translation);
 1:X0=1;  ~Fault(P1:L0,y);
 No
 Witnesses
--- a/herd/tests/instructions/AArch64.kvm/P019.litmus.expected
+++ b/herd/tests/instructions/AArch64.kvm/P019.litmus.expected
@@ -1,7 +1,7 @@
 Test P019 Allowed
 States 2
-[x]=0; [PTE(x)]=(oa:PA(x), af:0, db:0, dbm:1); Fault(P0,x,MMU:Permission);
-[x]=0; [PTE(x)]=(oa:PA(x), db:0, dbm:1); Fault(P0,x,MMU:Permission);
+[x]=0; [PTE(x)]=(oa:PA(x), af:0, db:0, dbm:1); Fault(P0,x,D-MMU:Permission);
+[x]=0; [PTE(x)]=(oa:PA(x), db:0, dbm:1); Fault(P0,x,D-MMU:Permission);
 Ok
 Witnesses
 Positive: 3 Negative: 6
--- a/herd/tests/instructions/AArch64.kvm/R019.litmus.expected
+++ b/herd/tests/instructions/AArch64.kvm/R019.litmus.expected
@@ -1,7 +1,7 @@
 Test R019 Allowed
 States 2
-[x]=42; [PTE(x)]=(oa:PA(x), af:0, db:0, dbm:1); Fault(P0,x,MMU:Permission);
-[x]=42; [PTE(x)]=(oa:PA(x), db:0, dbm:1); Fault(P0,x,MMU:Permission);
+[x]=42; [PTE(x)]=(oa:PA(x), af:0, db:0, dbm:1); Fault(P0,x,D-MMU:Permission);
+[x]=42; [PTE(x)]=(oa:PA(x), db:0, dbm:1); Fault(P0,x,D-MMU:Permission);
 Ok
 Witnesses
 Positive: 3 Negative: 6
--- a/herd/tests/instructions/AArch64.vmsa+ifetch/A006.litmus.expected
+++ b/herd/tests/instructions/AArch64.vmsa+ifetch/A006.litmus.expected
@@ -7,5 +7,5 @@ Positive: 1 Negative: 0
 Flag combining-vmsa-and-ifetch-is-not-supported
 Condition forall (fault(P0:Page0,I-MMU:Translation))
 Observation A006 Always 1 0
-Hash=e8bde7e3620aef33ad527b86e16a3165
+Hash=cc91e34143390ee14eeb05a9c0da6e90
 
--- a/herd/tests/instructions/AArch64.vmsa+ifetch/A008.litmus.expected
+++ b/herd/tests/instructions/AArch64.vmsa+ifetch/A008.litmus.expected
@@ -7,5 +7,5 @@ Positive: 1 Negative: 0
 Flag combining-vmsa-and-ifetch-is-not-supported
 Condition forall (fault(P0:Page0,I-MMU:AccessFlag))
 Observation A008 Always 1 0
-Hash=0c8f5aa3cdb65d2dccba1bdd75aa8203
+Hash=7c04c1b58c33689204712ff757a621ff
 
--- a/herd/tests/instructions/AArch64.vmsa+ifetch/A010.litmus.expected
+++ b/herd/tests/instructions/AArch64.vmsa+ifetch/A010.litmus.expected
@@ -6,5 +6,6 @@ Witnesses
 Positive: 1 Negative: 0
 Flag combining-vmsa-and-ifetch-is-not-supported
 Condition forall (fault(P0:Page0,I-MMU:Permission))
-Observation A008 Always 1 0
+Observation A010 Always 1 0
 Hash=cb0a8561acabcc5a329a5e61f8d9d457
+
--- a/herd/tests/instructions/AArch64.vmsa+ifetch/A011.litmus.expected
+++ b/herd/tests/instructions/AArch64.vmsa+ifetch/A011.litmus.expected
@@ -6,5 +6,6 @@ Witnesses
 Positive: 1 Negative: 0
 Flag combining-vmsa-and-ifetch-is-not-supported
 Condition forall (0:X0=1)
-Observation A007 Always 1 0
-Hash=664b4fb6731a307c4de1b8e71c757cb3
+Observation A011 Always 1 0
+Hash=9c258d3306eb6bfbf71ce598a396a20e
+
--- a/herd/tests/instructions/AArch64.vmsa+mte/A011.litmus.expected
+++ b/herd/tests/instructions/AArch64.vmsa+mte/A011.litmus.expected
@@ -1,7 +1,7 @@
 Test A011 Allowed
 States 3
 [x]=0; Fault(P0:L0,x:red,TagCheck);
-[x]=0; Fault(P0:L0,x:red,MMU:Permission);
+[x]=0; Fault(P0:L0,x:red,D-MMU:Permission);
 [x]=1;  ~Fault(P0:L0,x);
 Ok
 Witnesses
--- a/herd/tests/instructions/AArch64.vmsa+mte/A012.litmus.expected
+++ b/herd/tests/instructions/AArch64.vmsa+mte/A012.litmus.expected
@@ -1,7 +1,7 @@
 Test A012 Allowed
 States 3
 [x]=0; Fault(P0:L0,x:red,TagCheck);
-[x]=0; Fault(P0:L0,x:red,MMU:Permission);
+[x]=0; Fault(P0:L0,x:red,D-MMU:Permission);
 [x]=1;  ~Fault(P0:L0,x);
 Ok
 Witnesses


herd/tests/instructions/AArch64.vmsa+ifetch/A001.litmus
```diff
@@ -0,0 +1,5 @@
+AArch64 A001
+Variant=vmsa,ifetch
```
murzinv, line 2: Nitpick: it looks like `herd7` can handle both `ifetch` and `self`, yet `litmus7` recognizes only `self`. We can make `litmus7` recognize `ifetch` or we stick with`self` in litmus test, or we can have both - either way work for me, but happy to hear if anyone have preference

herd/tests/instructions/AArch64.vmsa+ifetch/A002.litmus
```diff
@@ -0,0 +1,7 @@
+AArch64 A002
+Variant=vmsa,ifetch
+{}
+P0          ;
+.pagealign  ;
+ADD X0,X0,#1;
```
murzinv: Nitpick: can we please use `W` registers (here and in other tests) so `litmus7` won't complain about different types for register
```
Register x0 has different types: <int> and <int64_t>
```

herd/tests/instructions/AArch64.vmsa+ifetch/A007.litmus
```diff
@@ -0,0 +1,9 @@
+AArch64 A005
```
murzinv: Isn't test name `A007`?

herd/tests/instructions/AArch64.vmsa+ifetch/A009.litmus
```diff
@@ -0,0 +1,13 @@
+AArch64 A011
```
murzinv: Isn't test name `A009`?

herd/tests/instructions/AArch64.vmsa+ifetch/A010.litmus
```diff
@@ -0,0 +1,10 @@
+AArch64 A008
```
murzinv: Isn't test name `A010`?

herd/tests/instructions/AArch64.vmsa+ifetch/A011.litmus
```diff
@@ -0,0 +1,9 @@
+AArch64 A007
```
murzinv: Isn't test name `A011`?

herd/tests/instructions/AArch64.vmsa+ifetch/A001.litmus
relokin: We should add `ifetch` as an alias for `self` in litmus7 please.

herd/AArch64Sem.ml
```diff
@@ -1531,23 +1576,10 @@ module Make
...
-- mop:        Function defining the memory operation.
-- perms:      Required permissions for Morello.
-- ma:         Virtual address to be accessed, represented as the value within the monad.
-- mv:         Value to be stored (for write operations), represented as the value in the monad.
-- an:         Annotation for the event structure.
-- ii:         Instruction metadata.
-*)
-      let lift_memop rA dir updatedb checked mop perms ma mv an ii =
+      let do_lift_memop rA (* Base address register *)
+            dir updatedb checked mop perms ma mv an ii (branch : 'a M.t -> branch M.t) =
```
diaolo01: The type is usually embedded in the function signature. Given the rest of the values don't have it, I propose to remove it for branch too.

herd/AArch64Sem.ml
```diff
@@ -1557,24 +1589,41 @@ Arguments:
...
+(*
+Arguments:
+- rA:         Base address register.
+- dir:        Access direction (Dir.R for reads, Dir.W for writes).
+- updatedb:   If true, update the Dirty Bit in table descriptors.
+- checked:    If true, perform memory tagging checks.
+- mop:        Function defining the memory operation.
+- perms:      Required permissions for Morello.
+- ma:         Virtual address to be accessed, represented as the value within the monad.
+- mv:         Value to be stored (for write operations), represented as the value in the monad.
```
diaolo01: This makes more sense above do_lift_memop and with an extra line to describe what branch does.

herd/AArch64Sem.ml
```diff
@@ -4655,57 +4659,165 @@ Arguments:
...
+           * fetching instructions from other relevant pages *)
+          if exposed_page then
+            let offset = (ii.A.addr mod Pseudo.proc_size) mod Pseudo.page_size in
+            relevant_pagelbls
+            |> List.map (fun (_,lbl) ->
+                let base = Label.Map.find lbl test.Test_herd.program in
+                let cand_a = base + offset in
+                let cand_i = match IntMap.find cand_a test.Test_herd.code_segment with
+                | (_,(_,i)::_) -> i
+                  | _ -> Warn.user_error "Instruction not found by the address %d" cand_a (* this case means that we have found a relevant page, but it does not have an instruction -- it may make sense to use NOP here rather than throw an error *)
```
diaolo01: Could the comment be on a separate line?

herd/AArch64Sem.ml
```diff
@@ -4655,57 +4659,165 @@ Arguments:
...
+                    (commit_pred ii >>*=
+                      fun () -> do_build_semantics test inst ii)
+                    k)
+              cands
+              begin
+              (* Anything else than a legit instruction is a failure *)
+                let (>>!) = M.(>>!) in
+                let m_fault = mk_fault None Dir.R Annot.N ii
+                    (Some FaultType.AArch64.UndefinedInstruction)
+                    (Some "Invalid") in
```
diaolo01: We have "invalid" in the context of PTE.v=0. I prefer to rename this to "Undefined instruction" or "Invalid instruction", although the semantics of I_UDF leave the msg value empty.

herd/AArch64Sem.ml
```diff
@@ -4655,57 +4659,165 @@ Arguments:
...
-                      let lbl_v = get_instr_label ii in
-                      commit_pred ii
-                        >>*= fun () -> m_fault >>| set_elr_el1 lbl_v ii
-                        >>! B.fault [AArch64Base.elr_el1, lbl_v]
-                    end
-        else do_build_semantics test inst ii
+                (get_exported_labels test))
+            labels in
+
+        let needs_vmsa_for_ifetch =
```
diaolo01: Checking understanding: `needs_vmsa_for_ifetch` might be needed when the PTE fields change, while `needs_ifetch` is needed for labels that would enable changing the instruction at an address?

herd/loader.ml
```diff
@@ -81,40 +86,104 @@ struct
...
+      assert false
+    | A.Skip n ->
+      let new_addr = addr + n in
+      load_code proc new_addr mem code
+
+  let make_padding old_addr new_addr =
+    let offset = new_addr - old_addr in
+    let immbranch_v = Option.get (A.mk_imm_branch offset) in
+    let immbranch_sz = A.size_of_ins immbranch_v in
+    assert (offset mod 4 = 0); (* FIXME: this check will fail when page alignment is implemented on architectures where instructions are not all 32-bit-sized *)
```
diaolo01: If this is meant as future work, could it be captured in the PR description similar to other items, please.

herd/mem.ml
```diff
@@ -1227,58 +1320,92 @@ let match_reg_events es =
...
+ let solve_mem_non_mixed test es rfm cns kont res =
+      (* The auxiliary functions *)
+      let is_to_codeloc e =
+        let open Constant in
+        match
+          Misc.seq_opt A.global (E.location_of e)
+        with
+        | Some (V.Var _) -> true
+        | Some (V.Val (Symbolic (Virtual {name=n;_}))) when Symbol.is_label n -> true
+        | Some (V.Val (Symbolic (Physical (s,_)))) when (s |> Symbol.of_string |> Symbol.is_label) -> true (* FIXME: this benefits from a proper exposition of a label as a physical address *)
```
diaolo01: If this is future work, can it be added in the PR description similar to the other items?

herd/semExtra.ml
```diff
@@ -330,6 +337,50 @@ module Make(C:Config) (A:Arch_herd.S) (Act:Action.S with module A = A)
...
+            match cst_to_pagelbl cst with
+            | Some v -> v::k
+            | None -> k
+            end
+          | A.Location_global _
+          | A.Location_reg _ -> k
+          )
+        st []
+
+    let get_exposed_codepage_mappings test =
```
diaolo01: What is this function for? I cannot find a use of it.

artkhyzha: Indeed, it's a duplicate of `get_exposed_codepages`.

herd/mem.ml
artkhyzha: Fair. With a bit of thinking, I wonder if the PAs could use Symbols instead of strings too. That would make the interface here a bit clearer. However, it's not necessary for the functionality of this PR, so I am tempted to leave it as a FIXME.

Specifically, right now we have:

In `mem.ml`:
```
let solve_mem_non_mixed test es rfm cns kont res =
  ...
  | Some (V.Val (Symbolic (Physical (s,_)))) when (s |> Symbol.of_string |> Symbol.is_label) -> true
  ...
```

In `semExtra.ml`:
```
let is_non_mixed_symbol test =
  ...
  | Physical (s,_) when (s |> Symbol.of_string |> Symbol.is_label) -> true
  ...
```

In `constant.ml`:
```
let is_non_mixed_symbol = function
  | Physical (s,_) when (s |> Symbol.of_string |> Symbol.is_label)
  ...
```

In all of these cases, I am slightly surprised by how I need to analyse the string to understand if it is a label. This is in constract to VAs, where I just look at the type of the name. The need to provide logic for PAs of labels will likely remain, but perhaps the interface for checking if a constant is a PA of a label could become simpler by letting the type of PA's symbol be Constant.Symbol rather than a string.

herd/loader.ml
artkhyzha: Not really a future work, I'll change the comment.

herd/AArch64Sem.ml
artkhyzha: Corrected to "empty" in line with UDF -- and, as far as I can tell, almost every function call to `mk_fault`.

herd/loader.ml
```diff
@@ -104,6 +111,13 @@ struct
         start,new_rets
     | A.Symbolic _
     | A.Macro (_,_) -> assert false
+    | A.Align p -> 
+      let new_addr = align_to_next addr p in
+      let diff = (new_addr - addr)/4 in
+      let padding = List.init diff (fun _ -> A.Instruction A.mynop) in
+      (* Printf.printf "    thought the addr would be %d, but need to proceed from %d, the diff being %d\n" addr new_addr diff; *)
```
relokin: Does it make sense to re-instantiate this protected by a debug flag? Otherwise should we remove the comment?

herd/loader.ml
```diff
@@ -104,6 +111,13 @@ struct
...
+      load_code proc addr mem rets (List.append padding code)
+      (* load_code proc (align_to_next addr p) mem rets code *)
```
relokin: Remove this?

herd/loader.ml
```diff
@@ -88,6 +94,7 @@ struct
        load_ins proc addr mem rets code ins
 
   and load_ins proc addr mem rets code = fun x ->
+    (* Printf.printf "Loading addr=%d\n" addr; *)
```
relokin: Should we remove this?

lib/AArch64Base.ml
```diff
@@ -1824,10 +1824,12 @@ type 'k kinstruction =
 type instruction = int kinstruction
 type parsedInstruction = MetaConst.k kinstruction
 
-let nop = Some I_NOP
 let mk_imm_branch off = Some (I_B (BranchTarget.Offset off)) 
+
+(* Nop is used by litmus7 as marker for start/stop of litmus assembly sequence *)
+let nop = Some (I_MOV(V32,ZR,K 0))
```
relokin, line 1828: I am not entirely sure about this change. We can go with this for now, but there are two alternatives:
* Just like we count NOPs we could have counted .pagealign as the adding the corresponding number of NOPs.
* A better alternative in my opinion would be to not rely on NOP for markers at all, what can't litmus7 emits internal labels instead of counting NOPs?

Perhaps we can leave a note with a TODO here. Or otherwise something for @diaolo01's backlog.

relokin (commented):
Thanks for this Artem. A couple of thoughts: 
* I think you can squash my commits into one, they don't stand on their own.
* You would have to squash the last commit into the relevant commits as well.
* This commit "Modify tests to match new fault syntax" should be squash with the commit that introduces the new syntax.

Overall, we should ensure that we can pass the regressions with every commit.

I am a little on the fence about D-MMU and I-MMU have we discussed and agreed on this syntax? In existing tests that we use MMU that implies D-MMU?

For litmus7 can we please check that we can run all catalogues in vmsa+ifetch mode?

Finally, variant=vmsa,ifetch now enables translation for data and for instructions. Do we see the need to have a mode where we have ifetch and translation only for data? Does this even make sense to support?

herd/loader.ml
artkhyzha: Adding a debug mode into Loader could be handy. However, it is also an invasive change (affects the hierarchy of modules). I'll remove the comments, and suggest that debugging info be added into Loader upon a refactor; otherwise, I am less certain what sort of debug info is helpful.

artkhyzha:
As a status update and an interim response to @relokin's feedback (thanks!):

> I am a little on the fence about D-MMU and I-MMU have we discussed and agreed on this syntax? In existing tests that we use MMU that implies D-MMU?

The current idea is to let `MMU = D-MMU | I-MMU`. With this approach, already written tests with Data Aborts do not need updating even though their faults would now become D-MMU faults. Outputs of `herd7` change, which is relevant for unit tests, but the tests in the catalogue and their verdicts remain the same. This seems like a good tradeoff, assuming we want to keep backwards compatibility and support the term `MMU` in dozens of already written tests and published materials, but also for it to mean `D-MMU`.

There is an open question whether `D-MMU` and `I-MMU` are our preferred terms. The names may get updated based on feedback.

> For litmus7 can we please check that we can run all catalogues in vmsa+ifetch mode?

Thank you for this suggestion. This is revealing some issues that will be investigated.

> Finally, variant=vmsa,ifetch now enables translation for data and for instructions. Do we see the need to have a mode where we have ifetch and translation only for data? Does this even make sense to support?

I agree that turning on/off translation for data accesses and/or instruction fetches could be helpful. This functionality is more about user interface, I think, and it strikes me as an "expert" feature that could be not enabled by default. We could support variants `NoTrForData` and `NoTrForInstr` (or something with less cryptic names) to disable address translation for certain accesses. By default, these variants would not be enabled.

That said, I wonder if this idea is okay to implement separately, in a follow-up PR.

herd/loader.ml
```diff
@@ -81,40 +86,106 @@ struct
...
+        | A.Nop -> code
+        | _ -> pseudo_ins::code
+        in
+        A.Label (lbl, A.Nop) :: (normalise_code addr next_code)
+    | A.Pagealign -> begin
+      try
+        let new_addr = next_addr_after_pagealign addr in
+        let padding = make_padding addr new_addr in
+        normalise_code addr (padding @ code)
+      with Invalid_argument _ -> Warn.fatal "Error in pre-processing litmus test"
```
fsestini: Instead of catching `Invalid_argument` here, could we instead pattern-match on `A.mk_imm_branch offset` at line 115 and `Warn.fatal` on `None`? The scope of this `try/with` seems broader than needed, so it could end up masking unrelated `Invalid_argument`s, and re‑raising can make stack traces less informative. Keeping the error local also keeps `normalise_code` in tail position here (minor win, but won't hurt).

herd/AArch64Sem.ml
```diff
@@ -4679,57 +4690,170 @@ Arguments:
...
+        let needs_vmsa_for_ifetch = (* checks if the logic for VA remapping needed *)
+          kvm && self && is_on_exported_page in
+        let needs_ifetch = (* checks if the logic for instruction overwrite needed *)
+          (lbl_exposed ii.A.addr) && self in
+        if needs_ifetch || needs_vmsa_for_ifetch then
+          try (
+            let mop_fetch = mk_mop_fetch is_on_exported_page lbl_exposed test ii in
+            let a_v =
+              if needs_vmsa_for_ifetch then begin
+                Option.get ii.A.rel_addr
```
fsestini: Similarly to `herd/loader.ml:153`, could we pattern‑match `ii.A.rel_addr` here instead of `Option.get` + catching `Invalid_argument` a few lines down below? This `Option.get` seems like the only likely `Invalid_argument` source, so handling None directly would make the error path clearer IMO. Also, I wonder if this should be reported as a user error or an internal invariant violation? In other words, is it possible for `ii.A.rel_addr` here to be `None` as a consequence of the user mis-using `herd7`?

lib/constant.ml
```diff
@@ -445,8 +445,19 @@ let do_mk_sym sym = match Misc.tr_pte sym with
     | Some s -> Physical (s,0)
     | None -> do_mk_virtual sym
 
+let mk_sym_physical_label_from_virt = function
```
fsestini: As a general rule, I try to avoid public APIs that expose a datatype in the `.mli` but then `assert false` on some of its constructors, since that makes the API implicitly partial and undermines the contract suggested by the interface. Would you consider making `mk_sym_physical_label_from_virt` and `unmk_sym_virtual_label_with_offset` total by restricting their domain? For example:

```
let mk_sym_physical_label_from_virt ~(label : Label.Full.full) ~offset = ...
```

Raising `Invalid_argument` would also be a viable alternative if it’s explicitly documented in the `.mli`/signature.

herd/mem.ml
```diff
@@ -355,6 +346,81 @@ module Make(C:Config) (S:Sem.Semantics) : S with module S = S	=
...
+                Warn.warn_always
+                  "On P%d, the instruction with address %d is on a page that does not start with a labelled instruction, which means that its address will not be translated" (proc_of_addr x) (x mod proc_size) ;
+              acc
+          ) IntMap.empty iaddrs in
+
+      let addr2va addr = IntMap.find_opt addr a2ra in
+
+      let is_exported_label lbl =
+        Label.Full.Set.exists
+          (fun (_,lbl0) -> Misc.string_eq lbl lbl0)
```
fsestini: Massive nitpick, but considering the stdlib has `String.equal`, and `Misc.string_eq` is just an alias to it, I'd generally stick to `String.equal` for consistency/clarity. (same for `Misc.int_eq` a few lines below)

lib/faultType.ml
```diff
@@ -80,29 +101,48 @@ module AArch64 = struct
...
+      | _ as s -> Warn.user_error "%s not a valid fault type" s
 
   let is s = try ignore (parse s); true  with  _ -> false
 
   let compare ft1 ft2 = Misc.polymorphic_compare ft1 ft2
+
+  let matches actual expected =
+    match actual, expected with
+    | MMU (_, m_actual), MMU (DISide.Any, m_expected) ->
+        m_actual = m_expected
```
fsestini: Not a big deal in this particular case since `mmu_t` is a small variant, but as a general rule I try to avoid polymorphic equality (`=`) on custom datatypes since it’s a bit fragile if the type ever grows to include non-comparable data. Perhaps using an explicit `let mmu_t_equal = (=)` could keep the intent clear.

herd/AArch64Sem.ml
artkhyzha: Good point. My understanding is that the "Invalid_argument" case should be unreachable code, but might be good to leave some warning just in case. Moreover, I can't see the reason for the Not_found case, so I am removing it too.

herd/loader.ml
artkhyzha: Right, hopefully this is addressed in the next commit.

lib/constant.ml
artkhyzha: I notice this function is no longer called anywhere. This makes it a bit trickier to understand what is best for its interface. Hence, removing it until the need arises.

lib/faultType.ml
artkhyzha: Interesting. Does the same concern apply to the FaultType.t itself? Would it be a good practice to have `equal : t -> t -> bool` in its interface?

herd/mem.ml
artkhyzha: Done, I think.

lib/faultType.ml
fsestini: It would apply to `FaultType.t` as well, yes, especially considering `FaultType` already exposes a `compare` function.

In general I would consider good practice for any module `M` exposing any type `t` that is intended to be compared for equality to also expose an explicit equality predicate.

PR #1520, opened 2025-10-24
[herd] Delay failure on non-symbolic memory accesses

maranget:
This PR completes PR #1513. 

The simulator **herd7** should fail when a memory access  to a non-symbolic address is attempted. This pre-existing (see PR #1513) check is now performed *after* the application of the memory model.
    
Consider the test:
```
AArch64 A260
(* Access to [16] avoided because of model, should succeed *)
{
int x=0;
int64_t y=16;
0:X1=x; 1:X1=x; 1:X3=y;
}

  P0         |  P1          ;
 MOV W0,#1   | LDR W0,[X1]  ;
 STR W0,[X1] | LDR W2,[X1]  ;
             | EOR W4,W0,#1 ;
             | ORR W4,W4,W2 ;
             | CBNZ W4,L1   ;
             | LDR X5,[X3]  ;
             | LDR W6,[X5]  ;
             |L1:           ;
forall 1:X4=1
```
The aarch64 memory model rejects the situation where `1:X0=1` and `1:X2=0`. As a consequence `1:X4` is always `1` and the load of address `16` is not attempted. Before this PR, running this test resulted in a fatal error `Non-symbolic memory access found on '[16]' (User error)`
Now the test completes:
```
% herd A260.litmus
Test A260 Required
States 1
1:X4=1;
Ok
Witnesses
Positive: 3 Negative: 0
Condition forall (1:X4=1)
Observation A260 Always 3 0
Time A260 0.02
Hash=7ed678244962720ad1fd67e9c88eed24


```

### DIFF
--- a/herd/tests/instructions/AArch64/A217.litmus.expected-failure
+++ b/herd/tests/instructions/AArch64/A217.litmus.expected-failure
@@ -1,1 +1,1 @@
-Warning: File "./herd/tests/instructions/AArch64/A217.litmus": Non-symbolic memory access found on '[2]' (User error)
+Warning: File "./herd/tests/instructions/AArch64/A217.litmus": Non symbolic address: [2] (User error)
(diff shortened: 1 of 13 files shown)


akiyks:
Hi @maranget,

Just confirmed this PR restores the previous behavior of herd7 with LKMM.
Thanks!

PR #655, opened 2023-09-10
[herd] fix tagloc initialisation for labels

artkhyzha:
This PR aims to resolve an issue arising with `-variant memtag`: whenever a register contains a label, herd7 creates an initial write for the instruction at that label, which the MTE variant then attempts to assign an initial tag to -- leading to an crash. For instance, consider the following minimal litmus test:
```
AArch64 AK001
{}
P0        ;
ADR X0,L0 ;
L0:       ;
 NOP      ;
```
Before this PR it used to crash with an error:
```
herd7 -cat aarch64.cat  -variant memtag AK001.litmus
Warning: File "AK001.litmus": Illegal tagloc on 0:L0 (User error)
```

To resolve the problem, I suggest to simply disable creation of initial tags for initial writes to labels.

PR #1524, opened 2025-10-27
[gen] Fix a problem on `vmsa/kvm`+`moreedges`, caused by `pp` conflict.

ShaleXIONG:
Fix a problem when `-variant kvm` and `-moredeges true`, e.g. `diyone7 -arch AArch64 -variant kvm -moreedges true -show edges`, the booting stage of `diy*` fails. This is due to a `pp` function, printing the same output/target string despite different input data.

We introduce separate data types to replace `Shootdown`. The new datatype is `ShootdownSync` and `ShootdownNoSync`. The former corresponds to `TLBI-sync*` syntax, and takes a `domain` and `operation`. The latter corresponds to `TLBI*` syntax and only takes `operation` now. 

Due to new data type we also fix a problem in the `compare_fence` all together, where previously `CMO` is not ordered correctly as it always return +1 even in the situation both side is `CMO`. Now it use the `stdlib.compare`, which is good enough to provide correct ordering based on the definition of fence datatype.

### DIFF
--- a/gen/AArch64Arch_gen.ml
+++ b/gen/AArch64Arch_gen.ml
@@ -621,39 +621,16 @@ let overwrite_value v ao w = match ao with
 type strength = Strong | Weak
 let fold_strength f r = f Strong (f Weak r)
 let fold_dirloc f r = f Next (f Prev r)
-type sync = Sync | NoSync
-type fence = | Barrier of barrier | CacheSync of strength * bool |
-               Shootdown of mBReqDomain * TLBI.op * sync | CMO of syncType * dirloc
+type fence = | Barrier of barrier | CacheSync of strength * bool
+             | ShootdownSync of mBReqDomain * TLBI.op
+             | ShootdownNoSync of TLBI.op
+             | CMO of syncType * dirloc
 
 let is_isync = function
   | Barrier ISB -> true
   | _ -> false
 
-let compare_fence b1 b2 = match b1,b2 with
-| (Barrier _,(CacheSync _|Shootdown _))
-| (CacheSync _,Shootdown _)
-  -> -1
-| Barrier b1,Barrier b2 -> barrier_compare b1 b2
-| CacheSync (s1,b1) ,CacheSync (s2,b2)->
-    begin match compare b1 b2 with
-   | 0 -> compare s1 s2
-   | r -> r
-    end
-| Shootdown (dom1,op1,sync1),Shootdown (dom2,op2,sync2) ->
-    begin match compare dom1 dom2 with
-    | 0 ->
-       begin
-         match compare op1 op2 with
-         | 0 -> compare sync1 sync2
-         | r -> r
-       end
-   | r -> r
-    end
-| (Shootdown _,(Barrier _|CacheSync _))
-| (CacheSync _,Barrier _)
-| (CMO _,_) | (_, CMO _)
- -> +1
-
+let compare_fence = compare
 
 let default = Barrier (DMB (SY,FULL))
 let strong = default
@@ -662,49 +639,40 @@ let add_dot f x = match f x with
 | "" -> ""
 | s -> "." ^ s
 
-let pp_sync = function
-  | NoSync -> ""
-  | Sync -> "-sync"
-
 let pp_fence f = match f with
 | Barrier f -> do_pp_barrier "." f
 | CacheSync (s,isb) ->
    sprintf "CacheSync%s%s"
      (match s with Strong -> "Strong" | Weak -> "")
      (if isb then "Isb" else "")
-| Shootdown (d,op,sync) ->
-   let tlbi = "TLBI" ^ pp_sync sync in
-   sprintf "%s%s%s" tlbi
+| ShootdownSync (d,op) ->
+   sprintf "TLBI-sync%s%s"
      (add_dot TLBI.short_pp_op op)
-     (match sync with
-      | NoSync -> ""
-      | Sync -> add_dot pp_domain d)
+     (add_dot pp_domain d)
+| ShootdownNoSync (op) ->
+   sprintf "TLBI%s"
+     (add_dot TLBI.short_pp_op op)
 | CMO (t,loc) ->
   sprintf "%s%s"
     (match t with DC_CVAU -> "DC.CVAU" | IC_IVAU -> "IC.IVAU")
     (match loc with Prev -> "p"| Next -> "n")
 
-
 let fold_cumul_fences f k =
    do_fold_dmb_dsb do_kvm C.moreedges (fun b k -> f (Barrier b) k) k
 
-let fold_shootdown =
-  if do_kvm then
+let fold_shootdown f acc =
+  if not do_kvm then acc
+  else
     let fold_domain =
       if C.moreedges then fold_domain
       else fun f k -> f ISH k
     and fold_op =
       if C.moreedges then TLBI.full_fold_op
       else TLBI.fold_op in
-    fun f k ->
-      fold_op
-        (fun op k ->
-          fold_domain
-            (fun d k ->
-              f (Shootdown(d,op,Sync))
-                (f (Shootdown(d,op,NoSync)) k)) k)
-        k
-  else fun _f k -> k
+    acc
+    |> fold_op ( fun op -> f (ShootdownNoSync(op)) )
+    |> fold_op ( fun op ->
+      fold_domain ( fun domain -> f (ShootdownSync(domain,op)) ) )
 
 let fold_cachesync =
   if do_self then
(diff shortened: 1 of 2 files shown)


gen/AArch64Arch_gen.ml
```diff
@@ -676,9 +653,7 @@ let pp_fence f = match f with
    let tlbi = "TLBI" ^ pp_sync sync in
    sprintf "%s%s%s" tlbi
```
murzinv: So, IIUC, before change we could have
```
TLBI
TLBI-sync.{NSH,ISH,OSH,SH}
```
after the change we have
```
TLBI.{NSH,ISH,OSH,SH}
TLBI-sync.{NSH,ISH,OSH,SH}
```
thus main question is `TLBI` alone is a valid annotation or it is always must come with `domain`?

@relokin may be you can give a clue?

gen/AArch64Arch_gen.ml
```diff
@@ -629,30 +629,7 @@ let is_isync = function
...
-         | 0 -> compare sync1 sync2
-         | r -> r
-       end
-   | r -> r
-    end
-| (Shootdown _,(Barrier _|CacheSync _))
-| (CacheSync _,Barrier _)
-| (CMO _,_) | (_, CMO _)
- -> +1
+let compare_fence = compare
```
relokin, line 633: Here you are changing this to use `Pervasives.compare`, aren't you? Not entirely sure why this was defined in the first place can you please elaborate?

gen/AArch64Arch_gen.ml
relokin: I agree this is not correct. For `TLBI` there shouldn't be a maintenance domain. Take this for example
```
$> diyone7 -arch AArch64 -variant kvm PteVA TLBI.ISHdWW P Rfe DMB.LDdRR FrePPteVA
LDdRR FrePPteVA
AArch64 A
"TLBI.ISHdWWPteVAP Rfe DMB.LDdRR FrePPteVA"
Com=Rf Fr
Orig=TLBI.ISHdWWPteVAP Rfe DMB.LDdRR FrePPteVA
{ int y=0; int x=4;
0:X0=PTE(x); 0:X1=(oa:PA(x), valid:0); 0:X3=y; 0:X4=x;
1:X3=y; 1:X4=x;
}
 STR X1,[X0]     | LDR W2,[X3] ;
 LSR X5,X4,#12   | DMB LD      ;
 TLBI VAAE1IS,X5 | L00:        ;
 MOV W2,#1       | LDR W5,[X4] ;
 STR W2,[X3]     |             ;
exists (1:X2=1 /\ ~Fault(P1:L00,x))
```
The `ISH` part of the TLBI is not used at all.

But more generally and this should be part of this but I don't think we want to ne enumerating all possible combinations of TLBI + DSB + ISB. I think TLBI should be insert edges like DCs and ICs.

gen/edge.ml
```diff
@@ -352,19 +352,9 @@ and type rmw = F.rmw = struct
...
-  let compare e1 e2 = match compare_atomo e1.a1 e2.a1 with
-  | 0 ->
-      begin match  compare_atomo e1.a2 e2.a2 with
-      | 0 -> compare e1.edge e2.edge
-      | r -> r
-      end
-  | r -> r
+  let compare_atomo = compare
+
+  let compare = compare
```
relokin: Again can you explain why you think this change is benign?

gen/AArch64Arch_gen.ml
ShaleXIONG: I see. Thanks for the explanation. I will find another way to fix it.

gen/AArch64Arch_gen.ml
ShaleXIONG: I think it will use the `stdlib.compare`, so it is the same thing as `Pervasives.compare`?. 

If I understand correctly, the behaviour of `sodlib.compare` based on how we define datatype. for example, `type = a of int | b of boolean`, it will order `a` before `b`, and within a constructor, for example `a` it will apply the `compare` recursively, here, to `int`. Given all the edges and atoms definition are not cyclic, the `stdlib.compare` should just work fine. All the `compare` function here is only used for building up the syntax lookup table, namely `String` to a `edge`, hence the detail ordering is not relevant as long as there is a order.

gen/edge.ml
ShaleXIONG: I have reverse this change since it is not immediately necessary for this pull.

gen/AArch64Arch_gen.ml
ShaleXIONG: I introduce two new data type `ShootdownSync` for `TLBI-sync.{domain}` and `ShootdownNoSync` for `TLBI.`. This fixes the problem namely `TLBI` by itself no longer needs a domain in the data type and hence solves the problem all together.

I have updated the pull message and commit.

gen/AArch64Arch_gen.ml
ShaleXIONG: I decide to keep this change by use the `sodlib.compare`, because we introduce new data type so I need to change this function anyway. Also I spotted the previous version is not correct, as you can see in the old line `#655`, `CMO` will always give +1 even in the situation where left-hand-side and right-hand-side are both `CMO`. This is not expected nor correct behaviour.

Do you think it makes sense here?

gen/AArch64Arch_gen.ml
```diff
@@ -621,39 +621,16 @@ let overwrite_value v ao w = match ao with
 type strength = Strong | Weak
 let fold_strength f r = f Strong (f Weak r)
 let fold_dirloc f r = f Next (f Prev r)
-type sync = Sync | NoSync
-type fence = | Barrier of barrier | CacheSync of strength * bool |
-               Shootdown of mBReqDomain * TLBI.op * sync | CMO of syncType * dirloc
+type fence = | Barrier of barrier | CacheSync of strength * bool
+             | ShootdownSync of mBReqDomain * TLBI.op
```
murzinv, line 625: This looks like right abstraction since `domain` is property of `sync`

gen/AArch64Arch_gen.ml
```diff
@@ -662,49 +639,40 @@ let add_dot f x = match f x with
...
-          fold_domain
-            (fun d k ->
-              f (Shootdown(d,op,Sync))
-                (f (Shootdown(d,op,NoSync)) k)) k)
-        k
-  else fun _f k -> k
+    acc
+    |> fold_op ( fun op -> f (ShootdownNoSync(op)) )
+    |> fold_op ( fun op ->
+      fold_domain ( fun demain -> f (ShootdownSync(demain,op)) ) )
```
murzinv: Nitpick: typo in `demain`

gen/AArch64Compile_gen.ml
```diff
@@ -2170,38 +2170,47 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
...
+      | Code _ -> Warn.user_error "TLBI/CacheSync" in
+      let open TLBI in
+      let r,init,csr,st =  match op.TLBI.typ with
+      | ALL|VMALL|VMALLS12 -> ZR,init,[],st
+      | ASID|VA|VAL|VAA|VAAL|IPAS2|IPAS2L ->
+        let r,init,st = U.next_init st p init loc in
+        let r1,st = tempo1 st in
+        let cs = [Instruction (lsri64 r1 r 12)] in
+        r1,init,cs,st in
+      let cs = emit_shootdown_nosync op r in
```
murzinv, line 2205: Nitpick: can we exact common code between `ShootdownSync` and `ShootdownNoSync` into helper?

gen/AArch64Arch_gen.ml
ShaleXIONG: ahh :) thanks !

gen/AArch64Compile_gen.ml
ShaleXIONG: That will work. Let me try it.

gen/AArch64Arch_gen.ml
relokin: Apologies, `Pervasives` was the name for `Stdlib`, but I shouldn't use the name `Pervasives` it makes me sound very outdated (it was deprecated in ocaml 5).
Ok in principle it sounds fine by me.

@fsestini can ask for your opinion on this?

fsestini: I agree that `Stdlib.compare` will give you the standard structural comparison for that type, which should make sense for this use case. The ordering produced by the previous `compare_fence` seems to be different from what `Stdlib.compare` produces (for example, the previous `compare_fence` orders `CacheSync` by comparing the second argument then the first). But indeed as @ShaleXIONG pointed out, any ordering will work here as long as the code using this function doesn't rely on any specific ordering but just needs _some_ ordering. That seems to be the case from a quick grep of the codebase, so I'd say it looks good to me.

ShaleXIONG: I believe this `compare` function are only used for building up the lookup table `lxm`. Since it builds a `Map` where a `compare` is needed.

relokin: Thanks both!

relokin (approved):

PR #1530, opened 2025-10-27
[ASL][Reference][ASLSpec] Error-related types and functions

Roman-Manevich:
Currently, error configurations for type errors and dynamic errors use strings, which is not very elegant. Instead, I defined  dedicated types for the various type error codes and dynamic error codes.

Currently, the `check` function is used both for type errors and dynamic errors, which will make specifying it in aslspec a bit tricky and later when type checking is introduced in aslspec it will be very hard to typecheck. This PR splits `check` into three functions: `be_check`, `te_check`, and `de_check`. Each function operates in a separate phase - build time, type-checking time, and dynamic execution time.

### DIFF
--- a/asllib/doc/TypeChecking.tex
+++ b/asllib/doc/TypeChecking.tex
@@ -185,7 +185,7 @@ The output configurations of type system assertions have two flavors:
 }
 \hypertarget{type-TypeErrorconfig}{}
 We define the shorthand $\TypeErrorConfig \triangleq \TypeError(\vc)$ for \typingerrorterm{} configurations
-with any error code $\vc \in \typeerrorcode$.
+with any type-error code $\vc \in \typeerrorcode$.
 
 When several \hyperlink{def-caserules}{case rules} for the same function use the same short-circuiting transition assertion,
 we do not repeat the\ProseOrTypeError, but rather include it only in the first rule.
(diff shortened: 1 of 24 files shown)


asllib/doc/asl.spec
```diff
@@ -1181,12 +1205,27 @@ typedef TDynError
...
 } =
-    DynamicError(error_code: Strings)
+    DynamicError(error_code: dynamic_error_code)
     { "dynamic error with error code {error_code}",
         math_macro = \DynamicError,
     }
 ;
 
+typedef dynamic_error_code { "dynamic error code" } =
+  | DE_UNR  { "Unreachable error Dynamic" }
```
HadrienRenaud: This reads badly. What about "Dynamic unreachable error"?

asllib/doc/TypeChecking.tex
```diff
@@ -184,7 +184,8 @@ \section{Typing Rule Configurations\label{sec:Typing Rule Configurations}}
 \]
 }
 \hypertarget{type-TypeErrorconfig}{}
-We define the shorthand $\TypeErrorConfig \triangleq \TypeError(\vs)$ for \typingerrorterm{} configurations with any code.
+We define the shorthand $\TypeErrorConfig \triangleq \TypeError(\vc)$ for \typingerrorterm{} configurations
+with any error code $\vc \in \typeerrorcode$.
```
HadrienRenaud: Maybe

```suggestion
with any type-error code $\vc \in \typeerrorcode$.
```

PR #1533, opened 2025-10-29
[litmus,test] Run tests during regression and check their outputs

maranget:
Add a validity check when tests compiled by **litmus7** are executed during regression testing. A run will be considered valid when all tests compiled  effectively produce an `Observation <name>....` line in output when executed.

The basic litmus test consists in compiling series of tests. Some series are also executed but not all of them. Tests that are not run by default would take too much time or raise errors.

  + `make litmus-x86_64-test`, for x86_64. All tests are executed by default.
  + `make litmus-aarch64-test`, for armv8. Some tests are executed some other are not.
      - `make litmus-aarch64-run` to compile and execute the test series that are executed by default.
      - `make litmus-aarch64-norun` to compile the test series that are not executed by default

Notice:
  1. One can use the command line to force execution (`make ... RUN_TESTS=true`) or to prevent execution (`make ... RUN_TESTS=false`).
  2. Litmus testing must be performed on machines whose processor matches the architecture tested.

### DIFF
--- a/README-tests.md
+++ b/README-tests.md
@@ -14,7 +14,13 @@
 
 ## Litmus testing
 
-By default. tests are compiled but do not run. To run the tests: `make
-`...` RUN_TESTS=true`.
-  + `make litmus-x86_64`, for x86_64.
-  + `make litmus-aarch64`, for armv8.
+The basic litmus test consists in compiling series of tests. Some series are also executed but not all of them. Tests that are not run by default would take too much time or raise errors.
+
+  + `make litmus-x86_64-test`, for x86_64. All tests are executed by default.
+  + `make litmus-aarch64-test`, for armv8. Some tests are executed some other are not.
+      - `make litmus-aarch64-run` to compile and execute the test series that are executed by default.
+      - `make litmus-aarch64-norun` to compile the test series that are not executed by default
+
+Notice:
+  1. One can use the command line to force execution (`make ... RUN_TESTS=true`) or to prevent execution (`make ... RUN_TESTS=false`).
+  2. Litmus testing must be performed on machines whose processor matches the architecture tested.


README-tests.md
```diff
@@ -14,7 +14,13 @@
...
-By default. tests are compiled but do not run. To run the tests: `make
-`...` RUN_TESTS=true`.
-  + `make litmus-x86_64`, for x86_64.
-  + `make litmus-aarch64`, for armv8.
+The basic litmus test consists in compiling series of tests. Some series are also executed but not all of them. Tests that are not run by default would take too much time or raise errors.
+
+  + `make litmus-x86_64-test`, for x86_64. All tests are executed by default.
+  + `make litmus-aarch64-test`, for armv8. Some tests are executed some other are not.
+      - `make litmus-aarch64-run` to compile and execute the test series that are executed by default.
+      - `make litmus-aarch64-nirun` to compile the test series that are not executed by default
```
HadrienRenaud:
```suggestion
      - `make litmus-aarch64-norun` to compile the test series that are not executed by default
```

maranget: Nice catch, thank you.

Makefile.aarch64
```diff
@@ -2,66 +2,90 @@ ifneq ($(OS),Darwin)
...
 litmus-cata-aarch64-test-cas: TEMP_DIR:=$(shell mktemp -d)
 litmus-cata-aarch64-test-cas: catalogue/aarch64-cas/tests/@all
 	$(LITMUS)					\
 		-set-libdir $(LITMUS_LIB_DIR)	        \
 		-gcc=$(AARCH64_PREFIX)gcc		\
 		-ccopts='-march=armv8-a+lse'		\
 		-o $(TEMP_DIR)				\
-		-mode std -a 4				\
+		-mode std -a 4	-s 1k -r 100		\
+                -driver C -ascall true                  \
```
relokin, line 40: That's a very good point, we should be testing `-driver C`. But what is the policy? Do we run some catalogues with `-driver shell` and some others with `-driver C`?

maranget: Generally `-driver C` is faster because startup cost is armortised on all tests in a serie. This is especially true for kvm tests and series consisting of many tests. Hence, I used `-driver C`,  as much as possible. However I kept a few `-driver shell` for the sake of coverage. Overall, I may not have been that systematic...

PR #1534, opened 2025-10-30
[gen] Update baseline configuration that generates forbidden litmus

ShaleXIONG:
Update baseline configuration that generates forbidden litmus.

- Add comment for auditing the connect between cat file and the forbidden file.
- Fix the problem on `Hat`.
- Add a missing `[A|Q Amo.* Po*W*]`.
- replace `[Pod** Amo.SwpAL]` by `[Po**R Amo.Swp L]` for example, where the latter is weaker then the former.
- Remove all `Amo.St*` and `Amo.Ld*` that causes problems due to commutative where the cycle is not correctly formed or observable.
- Add `DSB` edges too, whenever there is `DMB`.

### DIFF
--- a/gen/libdir/forbidden.conf
+++ b/gen/libdir/forbidden.conf
@@ -45,11 +45,171 @@
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 -arch AArch64
-
 -nprocs 2
-
--size 8
-
+-size 6
+# the generated file name is in format `Armv8-ext-forbidden********.litmus`,
+# where `********` are 8 digit numbers.
 -name Armv8-ext-forbidden
+-fmt 8
+# present value in hex
+#-hexa
+# If only want to generate cycles rather than litmus tests
+#-cycleonly true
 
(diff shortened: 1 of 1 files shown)


gen/libdir/forbidden.conf
```diff
@@ -45,11 +45,161 @@
...
 -arch AArch64
-
 -nprocs 2
+-size 6
+# the generate file name is in format `Armv8-ext-forbidden******.litmus`,
+# where `******` are 6 digit numbers.
+-name Armv8-ext-forbidden
+-fmt 6
+# present value in hex
+-hexa
```
relokin: I personally find `-hexa` useful only for the mixed size tests, is there a reason why you are adding it for these tests?

gen/libdir/forbidden.conf
```diff
@@ -45,11 +45,161 @@
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 -arch AArch64
-
 -nprocs 2
+-size 6
+# the generate file name is in format `Armv8-ext-forbidden******.litmus`,
```
relokin: *generated file

gen/libdir/forbidden.conf
```diff
@@ -45,11 +45,161 @@
...
+# where `******` are 6 digit numbers.
+-name Armv8-ext-forbidden
+-fmt 6
+# present value in hex
+-hexa
+# If only want to generate cycles rather than litmus tests
+#-cycleonly true
 
--size 8
+# If any relax input present, the cycles or litmus tests will contain at least one relaxation
```
relokin:
```suggestion
# If -relax is enabled tests will be required to have a relaxation from the list specified as its argument.
```

gen/libdir/forbidden.conf
```diff
@@ -45,11 +45,161 @@
...
+# let ca = fr | co
+# let Exp-obs =
+#   [Exp & M]; rf & ext; [Exp & M]
+#   | [Exp & M]; ca & ext; [Exp & M]
+-safe Rfe Fre Coe
+
+#let Exp-haz-ob = [Exp & R]; (po & same-loc); [Exp & R]; (ca & ext); [Exp & W]
+-safe [PosRR Coe] [PosRR Fre]
```
relokin:
```suggestion
-safe [PosRR Fre]
```
`[PosRR Coe]` doesn't make sense `Coe` starts (and ends) with a write.

gen/libdir/forbidden.conf
```diff
@@ -45,11 +45,161 @@
...
+
+### Data and address dependance
+# what is same instance?
```
relokin: This comment is unnecessary here.
```suggestion
```

gen/libdir/forbidden.conf
```diff
@@ -45,11 +45,161 @@
...
+# let dob = addr
+-safe DpAddrdR DpAddrdW DpAddrsR DpAddrsW
+
+# let ctrl = [Exp & R]; basic-dep; [BCC]; po
+# let dob = ctrl; [Exp & W | HU | TLBI | DC.CVAU | IC]
+# mcat2config7 aarch64.cat -let dob
+-safe DpCtrldW DpCtrlsW
+
+# let dob = addr; [Exp & M]; po; [Exp & W | HU]
+-safe [DpAddrdW PodWW] [DpAddrdW PosWW] [DpAddrsW PodWW] [DpAddrsW PosWW] [DpAddrdR PodRW] [DpAddrdR PosRW] [DpAddrsR PodRW] [DpAddrsR PosRW]
```
relokin, line 90: Not for this patch, but it seems it might be useful to add support for `DpAddrd*` `DpAddrs*` and similarly for `DpCtrld*` and `DpCtrls*` and this could be surprised by both diyone7 and diy7

gen/libdir/forbidden.conf
```diff
@@ -45,11 +45,161 @@
...
+# let bob = [L]; po; [A]
+-safe [L PodWR A] [L PosWR A]
+
+### Atomic operation
+# Amo.Swp
+# Amo.Cas
+# Amo.LdAdd Amo.LdClr Amo.LdEor Amo.LdSet Amo.LdSmax Amo.LdSmin Amo.LdUmax Amo.LdUmin
+# Amo.StAdd Amo.StClr Amo.StEor Amo.StSet
+# There is a known problem for the third and forth rows, due to commutative of those
+# edges.
```
relokin:
```suggestion
# There is a known problem for the Amo.Ld* and Amo.St* relaxations and for this reason are not included.
```

gen/libdir/forbidden.conf
```diff
@@ -45,11 +45,161 @@
...
+###
+# `LxSx` and its variants take more time for `herd7` to simulate due to loop.
+# It recommends pass `-unroll 0` to `herd7` for better performance.
+###
```
relokin:
```suggestion
```
I don't think this comment belongs here.

gen/libdir/forbidden.conf
```diff
@@ -45,11 +45,161 @@
...
+
+# let aob = [Exp & M]; rmw; [Exp & M]
+-safe LxSx Amo.Swp Amo.Cas
+
+# le aob = [Exp & M]; rmw; lrs; [A | Q]
+# let lrs = [W]; ((po & same-loc) & ~(intervening(W,(po & same-loc)))); [R]
+-safe [LxSx Rfi A] [Amo.Swp Rfi A] [Amo.Cas Rfi A]
+-safe [LxSx Rfi Q] [Amo.Swp Rfi Q] [Amo.Cas Rfi Q]
+
+#let bob = [range([A];amo;[L])]; po; [Exp & M | Imp & Tag & R | MMU & FAULT]
```
relokin, line 180: I think it would be add to add the following comment:
```suggestion
#let bob = [range([A];amo;[L])]; po; [Exp & M | Imp & Tag & R | MMU & FAULT]
# This relation cannot be directly mapped to a relaxation, however amo which in included in rmw is also in aob.
# For this reason, here, we map [A]; amo; [range([A];amo;[L])]; po; [Exp & M | Imp & Tag & R | MMU & FAULT]
```

gen/libdir/forbidden.conf
```diff
@@ -45,11 +45,161 @@
...
+-safe [A Amo.Swp L PodW*] [A Amo.Swp L PosW*]
+-safe [A Amo.Cas L PodW*] [A Amo.Cas L PosW*]
+#
+# let bob = [Exp & M | Imp & Tag & R]; po; [L] for `Pos*** A Amo.* L` Note that `A` is irrelevant
+# then let [range([A];amo;[L])]; po; [Exp & M | Imp & Tag & R | MMU & FAULT] for `A Amo.* L Po***`
+# This particular edges cannot be auto generated from other safe edges.
+-safe [Pod*R A Amo.Swp L PodW*] [Pod*R A Amo.Swp L PosW*] [Pos*R A Amo.Swp L PodW*] [Pos*R A Amo.Swp L PosW*]
+-safe [Pod*R A Amo.Cas L PodW*] [Pod*R A Amo.Cas L PosW*] [Pos*R A Amo.Cas L PodW*] [Pos*R A Amo.Cas L PosW*]
+
```
relokin, line 194: I think you are also missing `lwfs` which I think should map to `Pos*W`

relokin (commented):
Overall, looks good to me, I have a few mostly cosmetic comments.

gen/libdir/forbidden.conf
ShaleXIONG: I will comment it out to simplify the file. It was added since the value-picking strategy in atomic operation. While it is out of the score for now.

gen/libdir/forbidden.conf
ShaleXIONG: Done.

gen/libdir/forbidden.conf
ShaleXIONG: Done.

gen/libdir/forbidden.conf
ShaleXIONG: Removed.

gen/libdir/forbidden.conf
ShaleXIONG: Done.

gen/libdir/forbidden.conf
ShaleXIONG: Done.

gen/libdir/forbidden.conf
ShaleXIONG: I will move this to the end of this document as a note for consuming the generated tests.

gen/libdir/forbidden.conf
ShaleXIONG: Yes it will be supported in the wildcard change. (#1516)

There is some inconsistency currently caused by aliasing edges. Due to historical reason, I guessed, there are edges like `Dpd*` and `Dps*`, which becomes `DpDatad*` and `DpDatas*` effectively. However the latter two syntax are in fact illegal in the current version.

gen/libdir/forbidden.conf
ShaleXIONG: Done.

gen/libdir/forbidden.conf
ShaleXIONG: Will add it and re-check.

relokin (approved):
And just to double check with this configuration, diy7 doesn't generate any allowed tests?

relokin:
Can't you please check this version (with `-size 6`) as well?

murzinv:
> I am not sure if @murzinv have any insight here? I guess the file system gave up?

On linux machines it usually related to running out of filesystem's `inodes` (you can try to check with `df -i`). It is the case where `-cycleonly true` could help to make pipeline to process tests in batches. 

Anyway, I think we should be good with testing you already made.

PR #1537, opened 2025-10-30
[gen] New annotation for `vmsa`.

ShaleXIONG:
Let `{field}` denotes choices of `V` (for `VALID`), `AF`, `DB`, `DBM`.We introduce a list of new annotations to `diy` that enables better test generation for `vmsa`.
- `Pte{field}1`, and `Pte{field}0`. this enables precise manipulation of particular `pte` field. For example, the write memory event annotated by `PteAF1` will write `af` from `0` to `1`. Note that it may change the initial value of `pte` if necessary. It also means certainly cycle will be rejected because exists no initial value that satisfies the annotations. Given this new change, the old toggling-value annotation, `Pte{field}` has been removed.
- `PteHA`. It sets the initial `af` to `0`, and add `TTHM=HA` in the prelude. For the memory event annotated with this annotation, although it does not change the final assembly, but it hints this memory access will trigger automatic `af` set. Therefore a fault check will be added.
- `PteHD`. It sets the initial `db` to `0` and `dbm` to `1`, and add `TTHM=HD` in the prelude. The behaviour similar to `PteTTHMHA`, that the event label with such are hinted to trigger the automatic update to `db`.

Minor:
- Fix a misunderstanding on fault check condition: now load is check against `af` and `valid` and `store` is check against `af`, `valid` and `db`.
- Fix a mistake a load of `pte` does not appear in final exists check.
- Propagate fault check cross communication edges. consider a edge `(a,Coe,b)`, when there is a fault check against `a`, a extra fault check to `b` will be added, to correctly forms a cycle in the sense that `b` should observe `a`. Similar also apply to other communication edges `Fre` and `Rfe`.
- Fix an unexpected extra load `diyone7 -arch AArch64 -variant kvm PteAF0 PosWR Fre PteAF1 Coe -oneloc`. There was an extra load at the end of the thread of `PteAF1 Coe`.

A simple example: `diyone7 -arch AArch64 -variant kvm PteAF1 PodWR Fre PodWR Fre`

```
AArch64 A
"PodWRPteAF1P Fre PodWR FrePPteAF1"
Prefetch=0:x=F,0:y=T,1:y=F,1:x=T
Com=Fr Fr
Orig=PodWRPteAF1P Fre PodWR FrePPteAF1
{ int x=1; PTE(x)=(oa:PA(x), af:0); int y=5;
0:X0=PTE(x); 0:X1=(oa:PA(x)); 0:X2=y;
1:X2=y; 1:X4=x;
}
 P0          | P1          ;
 STR X1,[X0] | MOV W3,#6   ;
 LDR W3,[X2] | STR W3,[X2] ;
             | L00:        ;
             | LDR W5,[X4] ;
exists (0:X3=5 /\ Fault(P1:L00,x))
```

An example includes multiple `pte` change in one write: `diyone7 -arch AArch64 -variant kvm PteAF1 PteV0 PodWR Fre PodWR Fre`

    
```
AArch64 A
"PodWRPteAF1.V0P Fre PodWR FrePPteAF1.V0"
Prefetch=0:x=F,0:y=T,1:y=F,1:x=T
Com=Fr Fr
Orig=PodWRPteAF1.V0P Fre PodWR FrePPteAF1.V0
{ int x=1; PTE(x)=(oa:PA(x), af:0); int y=5;
0:X0=PTE(x); 0:X1=(oa:PA(x), valid:0); 0:X2=y;
1:X2=y; 1:X4=x;
}
 P0          | P1          ;
 STR X1,[X0] | MOV W3,#6   ;
 LDR W3,[X2] | STR W3,[X2] ;
             | L00:        ;
             | LDR W5,[X4] ;
exists (0:X3=5 /\ Fault(P1:L00,x))
```

A few examples where cycle will be rejected: 
- `diyone7 -arch AArch64 -variant kvm PteAF1 PteAF0 PodWR Fre PodWR Fre`, where `PteAF1 PteAF0` conflict.
- `diyone7 -arch AArch64 -variant kvm PteAF1 PodWR Fre PodWW PteAF1 Coe` where there are two `pte` writes to the same location, however both want to update `af` from `0` to `1`. In this case, there is no initial value will satisfy this scenario.

An example of `PteHD`, `diyone7 -arch AArch64 -variant kvm PteHD PodWR Fre PodWW Coe`

```
AArch64 A
"PodWRPteHDP Fre PodWW CoePPteHD"
Prefetch=0:x=F,0:y=T,1:y=F,1:x=W
Com=Fr Co
Orig=PodWRPteHDP Fre PodWW CoePPteHD
TTHM=HD
{ int x=1; PTE(x)=(oa:PA(x), db:0, dbm:1); int y=5;
0:X1=x; 0:X2=y;
1:X2=y; 1:X1=x;
}
 P0          | P1          ;
 MOV W0,#3   | MOV W0,#6   ;
 L01:        | STR W0,[X2] ;
 STR W0,[X1] | MOV W3,#2   ;
 LDR W3,[X2] | L00:        ;
             | STR W3,[X1] ;
exists ([x]=3 /\ 0:X3=5 /\ ~Fault(P0:L01,x) /\ ~Fault(P1:L00,x))
```

While `PteHD` can only apply to write event, `PteHA` can apply to both read and write events, for example `diyone7 -arch AArch64 -variant kvm PodWR PteHA Fre PodWW Coe`

```
AArch64 A
"PodWRPPteHA FrePteHAP PodWW Coe"
Prefetch=0:x=F,0:y=T,1:y=F,1:x=W
Com=Fr Co
Orig=PodWRPPteHA FrePteHAP PodWW Coe
TTHM=HA
{ int x=1; int y=5; PTE(y)=(oa:PA(y), af:0);
0:X1=x; 0:X2=y;
1:X2=y; 1:X1=x;
}
 P0          | P1          ;
 MOV W0,#3   | MOV W0,#6   ;
 STR W0,[X1] | L00:        ;
 L01:        | STR W0,[X2] ;
 LDR W3,[X2] | MOV W3,#2   ;
             | STR W3,[X1] ;
exists ([x]=3 /\ 0:X3=5 /\ ~Fault(P0:L01,y) /\ ~Fault(P1:L00,y))
```

An example where fault propagate cross communication edges: `diyone7 -arch AArch64 -variant kvm PteAF0 PosWW Coe PosWW Coe -oneloc -obs oo`

```
AArch64 A
"PosWWPteAF0P Coe PosWW CoePPteAF0"
Com=Co Co
Orig=PosWWPteAF0P Coe PosWW CoePPteAF0
{ int x=1;
0:X0=PTE(x); 0:X1=(oa:PA(x), af:0); 0:X3=x;
1:X3=x;
}
 P0          | P1          ;
 STR X1,[X0] | MOV W2,#3   ;
 MOV W2,#2   | L01:        ;
 L00:        | STR W2,[X3] ;
 STR W2,[X3] | MOV W4,#4   ;
             | STR W4,[X3] ;
exists ([x]=2 /\ ~Fault(P0:L00,x) /\ ~Fault(P1:L01,x)) (* <---- L01 is the new fault check *)
```

An example when read a `pte` via `Pte` annotation: `diyone7 -arch AArch64 -variant kvm PteAF0 PosWR Pte Fre PosWW Coe -oneloc`

```
AArch64 A
"PosWRPteAF0Pte FrePteP PosWW CoePPteAF0"
Com=Fr Co
Orig=PosWRPteAF0Pte FrePteP PosWW CoePPteAF0
{
pteval_t 0:X2;
 int x=1;
0:X0=PTE(x); 0:X1=(oa:PA(x), af:0);
1:X3=x;
}
 P0          | P1          ;
 STR X1,[X0] | MOV W2,#2   ;
 LDR X2,[X0] | L00:        ;
             | STR W2,[X3] ;
             | MOV W4,#3   ;
             | STR W4,[X3] ;
exists ([x]=2 /\ 0:X2=(oa:PA(x)) /\ ~Fault(P1:L00,x)) (* <- 0:X2=(oa:PA(x)) is the fix *)
```

### DIFF
--- a/gen/AArch64Compile_gen.ml
+++ b/gen/AArch64Compile_gen.ml
@@ -1754,6 +1754,7 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
       | None -> cs
 
     let emit_access st p init e =
+    let open WPTE in
     match e.C.dir,e.C.loc with
     | None,_ -> Warn.fatal "AArchCompile.emit_access"
     | Some d,Code lab ->
@@ -1919,9 +1920,31 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
             | _ -> assert false in
             let r,init,cs,st = emit A64.V64 st p init (Misc.add_pte loc) in
             Some r,init,cs,st
-        | R,Some (Pte (TTHM _),None) ->
+        (* A special case for TTHM HA on read. *)
+        | R,Some (Pte (Set pte),None) when WPTESet.mem HA pte ->
             let r,init,cs,st = LDR.emit_load st p init loc in
             Some r,init,cs,st
+        (* Special cases for TTHM.
+           - `HA` is on both read and write
+           - `HD` is only on write *)
+        | R,Some(Pte (Set pte),None) when pte = WPTESet.singleton HA ->
+            let r,init,cs,st = LDR.emit_load st p init loc in
+            Some r,init,cs,st
+        | R,Some(Pte ReadHAAcq,None) ->
+            let r,init,cs,st = LDAR.emit_load st p init loc in
+            Some r,init,cs,st
+        | R,Some(Pte ReadHAAcqPc,None) ->
+            let r,init,cs,st = LDAPR.emit_load st p init loc in
+            Some r,init,cs,st
+        | W,Some (Pte (Set pte),None) when is_tthm pte ->
+            let init,cs,st =
+              STR.emit_store st p init loc (Value.to_int e.C.v) None C.evt_null in
+            None,init,cs,st
+        | W,Some (Pte (SetRel pte),None) when is_tthm pte ->
+            let init,cs,st =
+              STLR.emit_store st p init loc (Value.to_int e.C.v) None C.evt_null in
+            None,init,cs,st
+        (* END special cases for TTHM. *)
         | W,Some (Pte (Set _),None) ->
             let init,cs,st =
               emit_set_pteval false st p init (Value.to_pte e.C.v) (Misc.add_pte loc) in
@@ -1929,10 +1952,6 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
         | W,Some (Pte (SetRel _),None) ->
             let init,cs,st =
               emit_set_pteval true st p init (Value.to_pte e.C.v) (Misc.add_pte loc) in
-            None,init,cs,st
-        | W,Some (Pte (TTHM _),None) ->
-            let init,cs,st =
-              STR.emit_store st p init loc (Value.to_int e.C.v) None C.evt_null in
             None,init,cs,st
         | d,Some (Pte _,_ as a) ->
             Warn.fatal
@@ -2573,7 +2592,7 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
                 let cs2 = pseudo cs0 in
                 let addi = [addi r2 r2 e.C.ord] in
                 r2,cs2,init,st,addi
-            | Some (Pte pte,None) when (not @@ is_tthm pte) ->
+            | Some (Pte (Set pte|SetRel pte),None) when (not @@ is_tthm pte) ->
                 let rA,init,st = U.emit_pteval st p init (Value.to_pte e.C.v) in
                 let cs,st =
                   match vdep with
@@ -2607,7 +2626,7 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
             | _ -> r2,cs2@pseudo addi,init,st in
           let loc = add_tag loc e.C.tag in
           begin match atom with
-          | None|Some (Pte (TTHM _),None) ->
+          | None ->
               let init,cs,st =
                 STR.emit_store_reg st p init loc r2 None C.evt_null in
               None,init,cs2@cs,st
@@ -2654,13 +2673,21 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
           | Some (Tag, None) ->
               let init,cs,st = STG.emit_store_reg st p init loc r2 in
               None,init,cs2@cs,st
+          | Some (Pte (Set pte),None) when is_tthm pte ->
+              let init,cs,st =
+                STR.emit_store_reg st p init loc r2 None C.evt_null in
+              None,init,cs2@cs,st
+          | Some (Pte (SetRel pte),None) when is_tthm pte ->
+              let init,cs,st =
+                STLR.emit_store_reg st p init loc r2 None C.evt_null in
+              None,init,cs2@cs,st
           | Some (Pte (Set _),None) ->
               let init,cs,st = emit_set_pteval_reg false st p init r2 (Misc.add_pte loc) in
               None,init,cs2@cs,st
           | Some (Pte (SetRel _),None) ->
               let init,cs,st = emit_set_pteval_reg true st p init r2 (Misc.add_pte loc) in
               None,init,cs2@cs,st
-          | Some ((Pte _,Some _)|(Pte (Read|ReadAcq|ReadAcqPc),_))
+          | Some ((Pte _,Some _)|(Pte (Read|ReadAcq|ReadAcqPc|ReadHAAcq|ReadHAAcqPc),_))
             -> assert false
           | Some (Plain _,None) -> assert false
           | Some (Tag,Some _) -> assert false
@@ -2887,13 +2914,15 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
     | Some _|None -> fun _ -> []
 
     let get_archinfo n =
+      let open WPTE in
       (* collect distinct tthm *)
       let tthm_value = C.fold ( fun node acc ->
         match node.C.edge.E.a1 with
-        | Some(Pte(TTHM e), _) -> WPTESet.union e acc
+        | Some(Pte (Set e|SetRel e), _) when is_tthm e -> WPTESet.union e acc
+        | Some(Pte (ReadHAAcq|ReadHAAcqPc), _) -> WPTESet.add HA acc
         | _ -> acc
         ) n WPTESet.empty
-      |> WPTESet.pp_str " " WPTE.pp_tthm in
+      |> WPTESet.pp_str " " WPTE.pp in
       if tthm_value = "" then []
       else [("TTHM",tthm_value)]
 
--- a/gen/common/AArch64Arch_gen.ml
+++ b/gen/common/AArch64Arch_gen.ml
@@ -156,60 +156,72 @@ type capa_opt = capa option
 
 module WPTE = struct
 
-  type pte_field = AF | DB | OA | DBM | VALID
-  let all_pte_field = [AF; DB; OA; DBM; VALID;]
+  type pte_field = AF | DB | DBM | VALID
+  let all_pte_field = [AF; DB; DBM; VALID;]
   let pp_pte_field = function
     | AF -> "AF"
     | DB -> "DB"
     | DBM -> "DBM"
-    | VALID -> "VALID"
-    | OA -> "OA"
+    | VALID -> "V"
 
           (* Toggle the value between 0 and 1 *)
-  type t = Base of pte_field
+  type t = OA
           (* Precise value of 0 to 1 *)
           | One of pte_field
           (* Precise value of 1 to 0 *)
           | Zero of pte_field
+          | HA
+          | HD
+
   let all =
-    List.map ( fun field -> [Base field; One field; Zero field;] ) all_pte_field
-    |> List.flatten
+    OA :: HA :: HD ::
+    ( List.map ( fun field -> [One field; Zero field;] ) all_pte_field
+      |> List.flatten )
   let compare = compare
   let pp = function
-    | Base p -> pp_pte_field p
-    | One p -> "One" ^ pp_pte_field p
-    | Zero p -> "Zero" ^ pp_pte_field p
-  let get_pte_field = function
-    | Base p | One p | Zero p -> p
-
-  let pp_tthm = function
-    | Base AF -> "HA"
-    | Base DB -> "HD"
-    | _ -> assert false
+    | OA -> "OA"
+    | One p -> ( pp_pte_field p ) ^ "1"
+    | Zero p -> ( pp_pte_field p ) ^ "0"
+    | HA -> "HA"
+    | HD -> "HD"
 end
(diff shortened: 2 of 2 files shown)


ShaleXIONG:
@relokin @murzinv I have not changed the syntax yet for example,`PteV1`. This pull request contains the necessary commits on the new annotations.

ShaleXIONG:
@relokin @murzinv I updated this pull via a force push (12th Nov. 2025), because other code structure change (already in main) we did together with @fsestini simplifies this change. I manually reviewed the commit history which match the committing messages.

gen/AArch64Compile_gen.ml
```diff
@@ -2194,12 +2194,12 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
...
-      pseudo
-        (I_FENCE (DSB(dom,FULL))::
-        I_TLBI(op,r)::I_FENCE (DSB(dom,FULL))::[])
+    let emit_shootdown_sync dom op r csr =
+      pseudo ([I_FENCE (DSB(dom,FULL))])
+        @ csr
+        @ pseudo([I_TLBI(op,r); I_FENCE (DSB(dom,FULL))])
 
-    let emit_shootdown_nosync op r = pseudo (I_TLBI(op,r)::[])
+    let emit_shootdown_nosync op r csr = csr @ pseudo ([I_TLBI(op,r)])
```
relokin, line 2222: We are we doing this? The order of the LSR with respect to the DSB doesn't matter. I don't understand the point of commit e5f2312adf7fcf2d877dd9f2406877e57463b55d

ShaleXIONG: This is a cosmatic change that @murzinv suggested. Also I have go through the `catalogues`, the sequence of `DSB ISH; LSR; TLBI; DSB ISH` is more common. This commit basic aligns to this more common pattern. Previously, it was `LSR; DSB ISH; TLBI; DSB ISH`. Notes that the old sequence has `LSR` before the fence.

relokin: I understand, but it really doesn't matter if the LSR is before or after the DSB.

ShaleXIONG: Sorry, if it is confused I can remove. I put in a separate commit so the history is clean here at least.

gen/CCompile_gen.ml
```diff
@@ -1139,9 +1139,10 @@ module Make(O:Config) : Builder.S
   else dump_c_test
  *)
 
-      let test_of_cycle name ?com ?(info=[]) ?(check=(fun _ -> true)) ?scope ?(init=[])
+      let test_of_cycle name ?com ?(info=[]) ?(check=(fun _ -> true)) ?scope ?(init=[]) ?(init_pte=[])
           es c =
         ignore (scope) ;
+        ignore (init_pte) ;
```
relokin: Can you help me understand this? Aren't you passing `init_pte` as an argument which is then unused?

gen/common/AArch64Arch_gen.ml
```diff
@@ -156,38 +156,53 @@ type capa_opt = capa option
...
+  let all =
+    List.map ( fun field -> [Base field; One field; Zero field;] ) all_pte_field
+    |> List.flatten
+  let compare = compare
+  let pp = function
+    | Base p -> pp_pte_field p
+    | One p -> "One" ^ pp_pte_field p
+    | Zero p -> "Zero" ^ pp_pte_field p
+  let get_pte_field = function
+    | Base p | One p | Zero p -> p
```
relokin: I don't think that `Base` is very useful and if you got rid of it, it would help with not listing `PteOneOA` and `PteZeroOA` as valid annotations (`-show annot`) and then throwing an error is a user actually used them.

As I've said before I think the following are unnecessarily verbose:
* `PteZero*` -> `Pte*0`
* `PteOne*` -> `Pte*1`
* `PteTTHMH{A,D}` -> `PteH{A,D}`
 
Lastly, the way this is structured I can combine L with PteOneValid but I cannot combine it with PteTTHMHA.

For these reasons, I think it would make more sense to write the `WPTE` module as follows:
```suggestion
  type pte_field = AF | DB |  DBM | VALID
  let all_pte_field = [AF; DB; DBM; VALID;]
  let pp_pte_field = function
    | AF -> "AF"
    | DB -> "DB"
    | DBM -> "DBM"
    | VALID -> "V"


  type t = OA
          (* Precise value of 0 to 1 *)
          | One of pte_field
          (* Precise value of 1 to 0 *)
          | Zero of pte_field
          | HA
          | HD
  let all = OA ::
    (List.map ( fun field -> One field ) all_pte_field) @
    (List.map ( fun field -> Zero field ) all_pte_field)
  let compare = compare
  let pp = function
    | OA -> "OA"
    | One p -> (pp_pte_field p) ^ "1"
    | Zero p -> (pp_pte_field p) ^ "0"
    | HA -> "HA"
    | HD -> "HD"
```

relokin (commented):
I have one general comment, please see in the code.

In addition, as we've discussed offline, can we please make sure that PTE(x) initialisations use pp_location_brk. Also, PTE(x) initialisations should appear at the top of the initialisations list, right after the initialisation of regular memory locations. See also, how `mprog7 -mode text <FILENAME.litmus>` would order the initialisations.

gen/common/AArch64Arch_gen.ml
ShaleXIONG: Let me change this. Thanks for the comment.

ShaleXIONG:
> I have one general comment, please see in the code.
> 
> In addition, as we've discussed offline, can we please make sure that PTE(x) initialisations use pp_location_brk. Also, PTE(x) initialisations should appear at the top of the initialisations list, right after the initialisation of regular memory locations. See also, how `mprog7 -mode text <FILENAME.litmus>` would order the initialisations.

I am looking for the where this code happen in `diy`. Once I identity the line, will change the render using `pp_location_brk`. Will also change the ordering of the initialisation.

gen/common/AArch64Arch_gen.ml
ShaleXIONG: I just made the change in a new commit, since it is a medium size rework. I have also updated the pull request message. Note that I have rebased on top of the current main; despite of the force push, only the last commit is new.

Following your comment and fix on `L` cannot apply `PteH{D,A}` in the case of `write` event. I also identify `A` and `Q` cannot apply to `PteHA`, while `PteHA` should be able to apply to `read` event. I have fixed it all together in the new commit.

gen/common/value_gen.ml
```diff
@@ -19,10 +19,18 @@ module type PteType = sig
   type pte
   val pp_pte : pte -> string
   val default_pte : string -> pte
+  (* Initial a valid pte_value that can be used to process atom list `pte_atom` *)
```
fsestini:
```suggestion
  (* Initialize a valid pte_value that can be used to process atom list `pte_atom` *)
```

gen/common/value_gen.ml
```diff
@@ -19,10 +19,18 @@ module type PteType = sig
...
+  val init_pte : string -> atom list -> pte
   val pte_compare : pte -> pte -> int
   val set_pteval : atom -> pte -> (unit -> string) -> pte
-  val can_fault : pte -> bool
+  (* Implicitly set pte value. Return indicates fault check and new pte vaule *)
```
fsestini: What does it mean to "implicitly" set a PTE value, in this context?

gen/top_gen.ml
```diff
@@ -1027,7 +1018,7 @@ let test_of_cycle name
   let (init,prog,final,env),(prf,coms) = compile_cycle check init c in
   let archinfo = Comp.get_archinfo c in
   let m_labs = num_labels prog in
-  let init = tr_labs m_labs init in
+  let init = (tr_labs m_labs init) in
```
fsestini: Why the parentheses?

ShaleXIONG: This is a mistake. I will remove.

gen/common/value_gen.ml
ShaleXIONG: I will clarify better in the code too:
it is special case when `PTE` is updated implicit by, for example, a `LDR` or `STR` to `x` rather than `PTE(x)`. For instance, `diyone7 -arch AArch64 -variant kvm PteHA PodWR Fre PodWR Fre`
```
AArch64 A
"PodWRPteHAP Fre PodWR FrePPteHA"
Prefetch=0:x=F,0:y=T,1:y=F,1:x=T
Com=Fr Fr
Orig=PodWRPteHAP Fre PodWR FrePPteHA
TTHM=HA
{ int x=1; PTE(x)=(oa:PA(x), af:0); int y=5;
0:X1=x; 0:X2=y;
1:X2=y; 1:X1=x;
}
 P0          | P1          ;
 MOV W0,#2   | MOV W0,#6   ;
 L00:        | STR W0,[X2] ;
 STR W0,[X1] | L01:        ;
 LDR W3,[X2] | LDR W3,[X1] ;
exists (0:X3=5 /\ 1:X3=1 /\ ~Fault(P0:L00,x) /\ ~Fault(P1:L01,x))
```
The value of `af:0` in `PTE(x)` will be update to `af:1` when excuse the instruction `L00:STR ...`, in fact, in this case `L00` will NOT fault because the `LDR` will observe `af:1`. Also notice there is a `TTHM=HA` prelude to enable such behaviour.

This function you comment, is basically used to deal with annotation `PteHA` and `PteHD`.

gen/common/value_gen.ml
ShaleXIONG: Will change. Thanks !

gen/common/value_gen.ml
fsestini: I see. I was confused because from the name of the function, it seemed like "implicit set" referred to the operation of the OCaml function itself, rather than referring to a domain-specific concept. Perhaps a name like `perform_implicit_pteval_set` or similar would help, but it's not a big deal.

ShaleXIONG:
@fsestini @relokin Thanks for reviews and comments. I have rebased on top of the current main and will merge once the CI passes.

gen/common/value_gen.ml
ShaleXIONG: ohhh I see. Let change to `implicitly_set_pteval` maybe, which is in contract to existing `set_pteval` function. Apology for my incorrect English grammar.

fsestini: I think the name is fine either way :) If needed, I can always propose an alternative naming scheme later on when I have a better grasp of the intent behind these functions

PR #1538, opened 2025-10-30
[asl] Add `--no-stdlib0` option to ignore ASL0 standard library

hrutvik:
Requested and implemented by @ldesnogu


HadrienRenaud (approved):
Looks good to me. One thought: should we activate this by default when there is no ASLv0 file in the arguments?

PR #1540, opened 2025-10-31
[aslref] Only put stdlib0 if there are ASL0 files

HadrienRenaud:
**NOTICE** this PR has for target #1538, not `master`.

This PR makes `--no-stdlib0` default when there are no ASLv0 files passed as argument.


ldesnogu:
> Should we add this commit to https://github.com/herd/herdtools7/pull/1538 @HadrienRenaud and @ldesnogu?

I think yes.

PR #1541, opened 2025-10-31
[ASL][ACL2] ACL2 ASL interpreter updates in order to pass ASLRef tests

solswords:
- added "prints" side effect descriptor for AST compatibility
 - switched order of checking for loop counter and limit conditions
 - find the main function with the correct signature instead of using the one called "main"
 - add the AsciiStr primitive
 - run tests with primitives allowed, to allow for primitive functions that don't have stdlib defs.


slobodova (approved):
Looks fine

PR #1546, opened 2025-11-03
[gen] Unify `pteval` and `int` into a `value` module.

ShaleXIONG:
Unify the `pteval` and `int` into a dedicated value module. This is a continuation of #1192.

- Fold interface `Pteval` into `value` type. Introduce a new constructor `PteValue of pte` where `pte` is the page table type based on the architecture back end.
- The old `Pteval_gen.ml` file renames and lifts to `value_gen.ml` which contains all information of value related type and functions. This means `Value` become an individual module which needs to be instantiate with an architecture-related pte information.
- Decouple `rmw` from the `Value`.


gen/cycle.ml
```diff
@@ -1012,7 +1000,10 @@ let set_same_loc st n0 =
                     E.set_pteval n.evt.atom pte_val next_loc
                   end else pte_val in
               let st = CoSt.set_pte_value st pte_val in
-              n.evt <- { n.evt with pte = pte_val; check_value; } ;
+              (* TODO UPDATE *)
```
fsestini: What is this TODO about?

gen/common/AArch64Arch_gen.ml
```diff
@@ -210,6 +202,60 @@ let  plain = Plain None
 
 type atom = atom_acc * MachMixed.t option
 
+module Value = struct
+
+  include Value.Make(struct
+    (* Do this on purpose since there is an outside `atom` type. *)
+    type _atom = atom
+    type atom = _atom
```
fsestini: You can use the `nonrec` keyword to have this refer to the `atom` already in scope.
```suggestion
    type nonrec atom = atom
```

gen/cycle.ml
ShaleXIONG: This should be removed. Just done.

gen/common/AArch64Arch_gen.ml
ShaleXIONG: Thanks ! I do not know the keyword before.

gen/common/archExtra_gen.ml
```diff
@@ -31,11 +32,18 @@ module type I = sig
...
 end
 
 module type S = sig
+
+  type arch_atom
   type arch_reg
 
+  (* It should be a `Value` module passed in, however, introduce
+     a different name to avoid conflict *)
+  module Value_extra : Value_gen.S with type atom = arch_atom
```
fsestini: I think this name clash should be dealt with at use site. In principle `S` is just an abstract interface. See other comments on how to resolve the clash.
```suggestion
  module Value : Value_gen.S with type atom = arch_atom
```
(note that if you go ahead with this suggestion, then the other occurrences of `Value_extra` should also be renamed to `Value` in the rest of this file.)

gen/arch_gen.mli
```diff
@@ -20,5 +20,6 @@ module type S =
     val tr_endian : int -> int
     module ScopeGen:ScopeGen.S
     include Fence.S
-    include ArchExtra_gen.S with type arch_reg = reg
+    (* Value is `included` from `Fence.S` which includes `Atom.S` *)
+    include ArchExtra_gen.S with type arch_reg = reg and type arch_atom = atom and module Value_extra = Value
```
fsestini: This is a quirk of OCaml module syntax. Writing `module Value = Value` creates a clash because the `=` is not actually stipulating an equality, but rather stipulating that all the types declared within the left-hand `Value` (from `ArchExtra_gen.S`) and the right-hand `Value` (from `Fence.S`) should be equal. The `module Value` definition from `ArchExtra_gen.S` is still there, hence the conflict with `module Value` from `Fence.S` when `ArchExtra_gen.S` is `include`d.

I think that what you really want here is to include `ArchExtra_gen.S` with its `Value` module _replaced_ by the `Value` module from `Fence.S`. This way, you get all the definitions from both `Fence.S` and `ArchExtra_gen.S`,  while both sharing a single common `Value` module. You can do this with the destructive update operator `:=`.

```suggestion
    include ArchExtra_gen.S with type arch_reg = reg and type arch_atom = atom and module Value := Value
```

gen/common/archExtra_gen.ml
```diff
@@ -112,8 +120,11 @@ with type arch_reg = I.arch_reg
 and type special = I.special
 and type special2 = I.special2
 and type special3 = I.special3
+and type arch_atom = I.arch_atom
+and module Value_extra = I.Value
```
fsestini: See other comment about `:=`.

It would also be slightly more principled to add the `Value := ...` constraint at every use site where `ArchExtra_gen.Make(...)` is `include`d somewhere, instead of constraining the functor signature here once and for all. However, currently this `Make` functor is always used to create modules that get `include`d, which means _every_ use site would need the `:=` constraint. Therefore we can just place the constraint here once for the sake of concision. 
```suggestion
and module Value := I.Value
```

ShaleXIONG: Amazing. `:=` is a syntax I do not know before.

gen/common/archExtra_gen.ml
```diff
@@ -148,6 +157,8 @@ and type special3 = I.special3
       end
   | Loc loc1,Loc loc2 -> compare loc1 loc2
 
+  module Value_extra = I.Value
```
fsestini: Can you please also change this (and all its references within this `struct`) to `Value`? The name itself doesn't actually matter since we the module is not being exposed, so `Value_extra` is also fine technically. But perhaps just re-using `Value` makes the intent clearer. I think just doing a search & replace `s/Value_extra/Value` on the whole file should suffice.
```suggestion
  module Value = I.Value
```

ShaleXIONG: Oh sorry, my bad, let me do a grep and replace.

fsestini: No problem!

PR #1552, opened 2025-11-06
[gen] Udpate the CI for testing baseline `diy`.

ShaleXIONG:
This pull introduces more CI for `diy`.

Add some extra edges and annotations in the `diy-test-aarch64` target in `Makefile` hence CI, to test baseline `diyone7` (via `diycross7`), which includes:
- `DpAddrCseld*`, `DpDataCseld*`, `DpCtrlCseld*`
- `DSB.SYd**`, `ISBd**`, `ISB` (pseudo-edge)
- `Amo.Cas`, `Amo.Swp`, `Amo.StAdd`, `LxSx`
- `Hat`
- `A`, `L`

Since we add `LxSx` in the CI, that generates warning on loop unrolling, we update the internal CI tool `herd_diycross_regression_test` so it will pattern match the warning too. 

Separately, we introduce `diy7` cycle generation test in `Makefile`, namely target `diy-baseline-cycleonly`. This includes a new internal/CI tool `diy_regression_test.ml` which run `diy7` in `cycleonly` mode and check the result. In particular, the new target checks whether `gen/libdir/forbidden.conf` generates the same cycle list but on `-size 4` configuration rather the `-size 6`. This smaller size is good enough to explore and check the `diy7` cycle exploration logic and still have a good performance to finish within a minute. A similar test is in for `forbidden_ifetch.conf`.

Given the updated tests on `diycross7` hence `diyone7` (because they share the same internal) and the new test for `diy7`+`cycleonly`. It should provide stronger guarantee on CI on `diy`.


.github/workflows/make-test-diy.yml
```diff
@@ -0,0 +1,77 @@
...
+      - name: Checkout tree
+        uses: actions/checkout@v4
+
+      - name: Set-up OCaml
+        uses: ocaml/setup-ocaml@v3
+        with:
+          ocaml-compiler: ocamlformat.0.28.1
+          dune-cache: true
+
+      - run: opam exec -- dune fmt
```
HadrienRenaud, line 76: I know I come after this is merged, but I'm wondering why this is here: as far as I know, `ocamlformat` is only activated for files in `asllib/`, so there would be no need to run this here?

fsestini: I'll let @ShaleXIONG chime in for the real answer, but we do plan to gradually enable ocamlformat for `gen/` in the near future. At least one part of `gen/`, namely [`cat2config`](https://github.com/herd/herdtools7/pull/1574) (which I'll hopefully merge soon), will have formatting enabled out of the box. You're right that this line is not strictly needed right now - just wanted to comment on why it could make sense to include it anyways in light of the near future.

ShaleXIONG: Yes, we might run ocaml format for the future for `gen/` part. The plan is to gradually introduce it behind the scenes, once we know certain file(s) in `gen/` become stable. We, (with @fsestini) also might change the files structure so it better serve `cat2config`.

HadrienRenaud: Ok this makes sense, thanks

maranget: Ah, I see this conversion after having introduced PR #1596. The current effect is to check the format of `asllib/` twice. One can always remove `check-fmt` now and re-introduce it later, when it becomes useful. The decision much depends on the timeframe of the planned extension of format checking.

ShaleXIONG: I see. Let me remove the `check-fmt` here and reinstall a separate CI for formatting check on `gen/`. Although I shall mention the CI is actually bound by the now `herd7`-only CI, which takes around 6min~ while `diy` side takes around 3min~.

maranget: I totally agree with your idea of installing a separated CI for diy. Thanks for having approved PR #1596, saving a (little) amount of CI and of electricity...

ShaleXIONG: @maranget Since you are here, I decide to not touch the part of existing CI which now is only for `herd7`. If you give me some guide so (1) I can make it only trigger on change on `herd` code, which I guess is `herd/**`, `lib/**` and possible `catalogue/**`, and (2) also possibly change the `makefile` target name so it reflect different tests.

PR #1553, opened 2025-11-06
Changes to Arm memory model that appear in that Arm ARM issue M.a

relokin:
This PR includes all the changes to the formal definition of the Arm memory model that correspond to the Arm Architecture Reference Manual Issue M.a.

These changes also appear in the Arm Architecture Reference Manual for A-profile architecture: Known issues in Issue L.b:
* 2.238 D24643: Strengthen the ordering requirements for Explicit-hazard-ordered-before.
* 2.231 D24571: Fixes relating to the ordering requirements of FEAT_HAFDBS
* 2.94 C23766 and 2.249 D24727: Clarify the definition of Same Location.
* 2.116 C23849: Replace the definition Reads-from-register with Local-register-read successor.
* 2.236 D24622: Revamp the definition of dependencies and pick dependencies.
* 2.111 C23842: Split the definition of Local write or MMU Fault successor to Local write successor and Local MMU Fault successor.


herd/libdir/aarch64.cat
```diff
@@ -108,13 +108,13 @@ let Tag-obs =
...
+(* TLBUncacheable-coherence-after *)
+let TLBuncacheable-ca =
+  [range([TLBUncacheable & FAULT]; tr-ib^-1; [Imp & TTD & R])]; ca; [Exp & W | HU]
 
-(* Hardware-update-predecessor *)
-let HU-pred =
-  (ca & ~intervening(W,ca)); [HU]
+(* Hardware-update-coherence-after *)
+let HU-ca =
+  [Exp & R]; ca; [HU]
```
ShaleXIONG, line 117: Quick question here, this `ca` is effectively a `fr`, because left is a read and right side is a write?

relokin: Indeed this is effectively an `fr`, it's a style choice that we use `ca` here.

herd/libdir/aarch64util.cat
```diff
@@ -44,6 +44,10 @@
  * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  *)
 
+let iico_data = iico_data+
+let ADDR = [ADDR]; iico_data
+let DATA = [DATA]; (iico_data|iico_ctrl)+
```
HadrienRenaud: Stylistically I find this a bit confusing, as previously all-caps names are effect sets, and other-capitalized names are relations.
Some suggestions:
- `ii_data`/`ii_addr`
- `from_data`/`from_addr`
- `from_data_port`/`from_addr_port`

relokin: that's a good point Hadrien, I will change this to one of your suggestions.

herd/libdir/aarch64.cat
```diff
@@ -123,8 +123,8 @@ let TLBI-ca =
 (* TTD-observed-by *)
 let TTD-obs =
     [Imp & TTD]; rf | rf; [Imp & TTD]
-    | TLBuncacheable-pred
-    | HU-pred
+    | TLBuncacheable-ca
+    | HU-ca
     | [HU]; ca; [W] | [W]; ca; [HU]
```
artkhyzha, line 128: This does not need changing, just an observation: these clauses could in future be merged into `HU-ca` on the basis of them all describing how explicit `ca` rules are adjusted to consider `HU`.

herd/libdir/aarch64deps.cat
```diff
@@ -45,29 +45,37 @@ catdep (* This option tells that the cat file computes dependencies *)
 
 include "aarch64util.cat"
 
-(* Local write or MMU Fault successor *)
-let lwfs =
+(* Notion of Same General Purpose Register *)
```
artkhyzha: Might be nice to put `same-gpr` belong in `aarch64loc.cat` just because that's where `loc`-based definitions get elaborated, but I am not sure it's an improvement necessarily.

herd/libdir/aarch64util.cat
```diff
@@ -44,6 +44,10 @@
  * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  *)
 
+let iico_data = iico_data+
```
HadrienRenaud, line 47: Might be interesting to also specify here that `iico_ctrl` and `iico_order` are transitive? This way the changes to `pick-data` would be more robust.

PR #1555, opened 2025-11-06
[ASL Reference][aslspec] macros and none

Roman-Manevich:
- Replaced the symbol for short-circuitting from `//` to `|`. This is because: 1. `//` looks like a line comment and I would like for aslspec to have both line comments and be able to specify short-circuit alternative for transitions, and 2. it is consistent with the notation for output configurations, which also uses `|`.
- Defined a constant for None - the empty optional.

[The ASL Reference build fine locally, but the `asldoc` workflow is temporarily broken.]

### DIFF
--- a/asllib/doc/ASLmacros.tex
+++ b/asllib/doc/ASLmacros.tex
@@ -1525,6 +1525,7 @@
 \newcommand\OrLexicalError[0]{\;\terminateas\;\LexicalErrorConfig}
 \newcommand\ParseError[0]{\hyperlink{def-parseerror}{\BuildErrorCode{PE}}}
 \newcommand\ParseErrorConfig[0]{\hyperlink{def-parseerror}{\texttt{\#}\BuildErrorCode{PE}}}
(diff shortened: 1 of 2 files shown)


asllib/doc/ASLFormal.tex
```diff
@@ -780,12 +779,12 @@ \section{Short-circuit Rule Macros\label{sec:ShortCircuitRuleMacros}}
 \hypertarget{def-terminateas}{}
 In the following, $\XP$ and $\XQ$ stand for, possibly empty, sequences of premises.
 %
-A rule macro includes the special premise form $C \rulearrow C' \terminateas E$,
-which introduces alternative output configurations $C'$ and short-circuit $E$:
+A rule macro includes the special premise form $C \rulearrow C'\; \terminateas E$,
+which introduces the alternative output configuration $C'$ and the short-circuiting alternative $E$:
```
hrutvik:
```suggestion
which introduces the output configuration $C'$ and the short-circuiting alternative $E$:
```

PR #1558, opened 2025-11-08
[aslspec] output variants

Roman-Manevich:
In order to automatically add short-circuit configurations to transition judgments, we need to automatically figure those for each relation. This PR makes a design desicison for aslspec: each relation has the form `name(arguments) -> T1 | T2 ... Tn` where `T1` is allowed to be any type term, but `T2..Tn` are allowed to be either type names or sets of constants (`constants_set`), and each such type name must define the `short_circuit_term` attribute. This way, each constant in a given set of constants is used as a short-circuit configuration, and for type names we have the attribute to use.

`spec.ml` has a new check for the condition described above.
`asl.spec` requires a few minor changes to adhere to the new design.

[The ASL Reference build fine locally, but the asldoc workflow is temporarily broken.]

### DIFF
--- a/asllib/aslspec/spec.ml
+++ b/asllib/aslspec/spec.ml
@@ -5,6 +5,27 @@ open AST
 module StringMap = Map.Make (String)
 module StringSet = Set.Make (String)
 
+let raise_not_type_name_error relation_name term =
+  let msg =
+    Format.asprintf
+      "In relation '%s', output type term '%a' is not a type name or a set of \
+       constants. All relation output types, except for the first one, must be \
+       type names or a set of constants."
+      relation_name PP.pp_type_term term
+  in
+  raise (SpecError msg)
+
+let raise_missing_short_circuit_attribute relation_name term type_name =
+  let msg =
+    Format.asprintf
+      "In relation '%s', output type term '%a' references type '%s' which does \
+       not have a short-circuit macro defined. All relation output types, \
+       except for the first one, must reference types with short-circuit \
+       macros."
+      relation_name PP.pp_type_term term type_name
+  in
+  raise (SpecError msg)
+
 (** [vars_of_type_term term] returns the list of term-naming variables that
     occur at any depth inside [term]. *)
 let rec vars_of_type_term term =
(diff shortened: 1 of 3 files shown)


asllib/aslspec/spec.ml
```diff
@@ -353,6 +353,54 @@ module Check = struct
     in
     List.iter (check_no_undefined_ids_in_elem id_to_defining_node) elements
 
+  let check_relations_outputs elems id_to_defining_node =
+    let raise_not_type_name_error relation_name term =
+      let msg =
+        Format.asprintf
+          "In relation '%s', output type term '%a' is not a type name or a set \
+           of constants. All relation outputs types, except for the first one, \
```
hrutvik: nit:
```suggestion
           of constants. All relation output types, except for the first one, \
```

asllib/aslspec/spec.ml
```diff
@@ -353,6 +353,54 @@ module Check = struct
...
+                    match Type.short_circuit_macro typedef with
+                    | None ->
+                        let msg =
+                          Format.asprintf
+                            "In relation '%s', output type term '%a' \
+                             references type '%s' which does not have a \
+                             short-circuit macro defined. All relation output \
+                             types, except for the first one, must reference \
+                             types with short-circuit macros."
+                            name PP.pp_type_term term id
```
hrutvik: nit: could this be factored out into a function too, as you did for `raise_not_type_name_error`?

asllib/aslspec/spec.ml
```diff
@@ -353,6 +353,54 @@ module Check = struct
...
+    in
+    List.iter
+      (fun { Relation.name; output } ->
+        let alternative_outputs =
+          match output with
+          | [] -> assert false (* Guaranteed by the parser. *)
+          | _ :: alternative_outputs -> alternative_outputs
+        in
+        List.iter
+          (fun term ->
```
hrutvik: nit: I wonder if this inner-loop function would be clearer as a factored-out helper, e.g. `check_alternative_output`

hrutvik (approved):

asllib/aslspec/spec.ml
HadrienRenaud: Maybe those should be separated from the rest of the code so that they don't interrupt too much the code?

asllib/doc/asl.spec
```diff
@@ -3136,16 +3136,20 @@ typing relation get_for_constraints(
...
+    | TThrowing
+    | TDynError
+    | TDiverging
 {
    prose_description = "evaluates a statement {s} in an environment {env},
                         resulting in one of four types of configurations (see
                         more details in
                         \secref{KindsOfSemanticConfigurations}):
                         \begin{itemize}
-                        \item returning configurations with values {vs}, execution graph {new_g}, and a modified environment {new_env};
```
HadrienRenaud, line 3147: Just to double check: the variable names that were introduced here (`vs`, `new_g`, `new_env`) are defined somewhere else right?

asllib/aslspec/spec.ml
```diff
@@ -353,6 +353,54 @@ module Check = struct
...
+      List.filter_map
+        (function Elem_Relation def -> Some def | _ -> None)
+        elems
+    in
+    List.iter
+      (fun { Relation.name; output } ->
+        let alternative_outputs =
+          match output with
+          | [] -> assert false (* Guaranteed by the parser. *)
+          | _ :: alternative_outputs -> alternative_outputs
```
HadrienRenaud: Maybe simply `List.tail`?

asllib/aslspec/spec.ml
Roman-Manevich: I hoisted them to the top level.

asllib/aslspec/spec.ml
Roman-Manevich: It's now in `check_alternative_outputs`

asllib/doc/asl.spec
Roman-Manevich: They are names used in the first output of the relation `Continuing(new_g: XGraphs, new_env: envs)`

PR #1521, opened 2025-10-24
[herd,asl] Generate more asl code from published xml

HadrienRenaud:
This PR uses 2 other resources published by Arm Ltd to automatically generate
some asl code.

Main positive advantage is not to have to change the ASL code every release.
It also decluters our hand-written ASL files.

- **[herd,asl] Implement bunlder.py for systerm registers**
- **[asl] Improve Pretty-printer**
- **[asl,herd] Extract features from released json file**
- **[git] Update gitignore with new file names**
- **[asl] Cleanup unused executable**

PR #1559, opened 2025-11-09
[build] Remove dependency of clean in asl-pseudocode directory

maranget:
The clean target depended on build target. As a consequence `make clean-asl-pseudocode` triggers the download and extraction of the ASL files before erasing them.


maranget:
Hi @HadrienRenaud. Would you check that this PR does not contradict your intent in PR #1521?

maranget:
What about:
```
herd/libdir/asl-pseudocode/shared_pseudocode.asl:
	@ $(MAKE) -j 1 -C $(@D) all clean-tmp
```

PR #1565, opened 2025-11-12
[ci][asl] Use non-latest `texlive` image for ASL reference CI

hrutvik:
Attempt to fix recent breakages due to a change in the `texlive` image by fixing on (hopefully) a more stable tag. In particular, we saw errors when OCaml installation was running `./configure`:
```
  # checking whether the C compiler works... no
  # configure: error: in `/__w/herdtools7/herdtools7/_opam/.opam-switch/build/ocaml-compiler.5.4.0':
  # configure: error: C compiler cannot create executables
```

### DIFF
--- a/.github/workflows/build-asl-reference.yml
+++ b/.github/workflows/build-asl-reference.yml
@@ -19,7 +19,8 @@ concurrency:
 jobs:
   make-asldoc:
     runs-on: ubuntu-latest
-    container: texlive/texlive:latest-full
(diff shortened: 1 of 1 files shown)


HadrienRenaud:
Could we keep an open PR that would switch back to the latest texlive so that we try every now and then to see if the latest texlive works?

HadrienRenaud (approved):
It looks good to me otherwise

Roman-Manevich (approved):
Looks good to me!

PR #1570, opened 2025-11-14
[aslspec] refactoring aslspec

Roman-Manevich:
This PR performs refactors the code for two main reasons: improving code reusability for rendering and preparing for the [introduction of inference rules ](https://github.com/herd/herdtools7/pull/1482):
- In `AST.ml`: renamed `Operator` to `TypeOperator` to distinguish it from operators on expressions, which will be used by inference rules.
- Moved helper functions related to layouts to the new `LayoutUtils.ml`.
- Moved text color business to a module in `main.ml`, since that is where it is used.
- `SpecParser.mly`: moved `SEMI` to `element` instead of having it throughout the various definitions, factored the attributes into their own productions to make it easier to reuse them and see which attributes are used for which elements.
- `latex.ml` introduces various helper functions that represent the different latex patterns needed for rendering elements, which are now used by `render.ml`.
- `render.ml` separates the rendering of an element from the rendering of a macro for that elements. This enables rendering elements for debug purposes, which helps spotting problems faster.

### DIFF
--- a/asllib/aslspec/LayoutUtils.ml
+++ b/asllib/aslspec/LayoutUtils.ml
@@ -2,36 +2,38 @@
 
 open AST
 
-let horizontal_if_unspecified layout list =
-  match layout with
-  | Horizontal _ | Vertical _ -> layout
-  | _ -> Horizontal (List.map (fun _ -> Unspecified) list)
+let unspecified_for_elements elements = List.map (fun _ -> Unspecified) elements
 
-let vertical_if_unspecified layout list =
+let horizontal_if_unspecified layout elements =
   match layout with
+  | Unspecified -> Horizontal (unspecified_for_elements elements)
   | Horizontal _ | Vertical _ -> layout
-  | _ -> Vertical (List.map (fun _ -> Unspecified) list)
+
+let vertical_if_unspecified layout elements =
+  match layout with
+  | Unspecified -> Vertical (unspecified_for_elements elements)
+  | Horizontal _ | Vertical _ -> layout
 
 let rec contains_vertical = function
   | Unspecified -> false
   | Horizontal layout_list -> List.exists contains_vertical layout_list
   | Vertical _ -> true
 
-(** [apply_layout_to_list layout list] distributes the layouts listed in
-    [layout] to the list of elements in [list]. If [layout] is [Unspecified],
-    then all elements in [list] are assigned the [Unspecified] layout. For
-    example:
+(** [apply_layout_to_list layout elements] distributes the layouts listed in
+    [layout] to the list of elements in [elements]. If [layout] is
+    [Unspecified], then all elements in [elements] are assigned the
+    [Unspecified] layout. For example:
     {[
       apply_layout_to_list
         (Horizontal [ Unspecified; Vertical [ Unspecified; Unspecified ] ])
         [ a; b ]
       = [ (a, Unspecified); (b, Vertical [ Unspecified; Unspecified ]) ]
     ]} *)
-let apply_layout_to_list layout list =
+let apply_layout_to_list layout elements =
   let args_layout =
     match layout with
     | Horizontal l | Vertical l -> l
-    | Unspecified -> List.map (fun _ -> Unspecified) list
+    | Unspecified -> unspecified_for_elements elements
   in
-  assert (List.compare_lengths list args_layout = 0);
-  List.combine list args_layout
+  assert (List.compare_lengths elements args_layout = 0);
+  List.combine elements args_layout
--- a/asllib/aslspec/PP.ml
+++ b/asllib/aslspec/PP.ml
@@ -28,7 +28,7 @@ let pp_attribute fmt = function
   | MathMacroAttribute macro -> pp_print_string fmt macro
 
 let pp_attribute_key_value fmt (key, value) =
-  fprintf fmt {|%a = %a,|} pp_attribute_key key pp_attribute value
+  fprintf fmt "%a = %a," pp_attribute_key key pp_attribute value
 
 let pp_attribute_key_values fmt attributes =
   if Utils.list_is_empty attributes then ()
--- a/asllib/aslspec/SpecParser.mly
+++ b/asllib/aslspec/SpecParser.mly
@@ -118,13 +118,13 @@ let list1(x) :=
 (* Actual AST rules                         *)
 (* ---------------------------------------- *)
 
-let spec := elems=terminated(list(elem), EOF); { elems }
+let spec := elems=terminated(list(terminated(elem, SEMI)), EOF); { elems }
 
 let elem :=
-    | ~=type_definition; SEMI; { type_definition }
-    | ~=relation_definition; SEMI; { relation_definition }
-    | ~=constant_definition; SEMI; { constant_definition }
-    | ~=render_types; SEMI; { render_types }
+    | type_definition
+    | relation_definition
+    | constant_definition
+    | render_types
 
 let type_kind := TYPEDEF; { TypeKind_Generic }
     | AST; { TypeKind_AST }
--- a/asllib/aslspec/latex.ml
+++ b/asllib/aslspec/latex.ml
@@ -6,37 +6,46 @@ open LayoutUtils
 
 let regex_underscore = Str.regexp_string "_"
 let regexp_spaces_segment = Str.regexp "[ \t]+"
+let regexp_only_letters = Str.regexp "^[a-zA-Z]+$"
 let union_macro_name = "cup"
 
 (** A LaTeX underscore escape sequence. *)
 let escaped_underscore = {|\_|}
 
-let shrink_whitespace str =
-  let regexp_newline_segment = Str.regexp "[ \n\r]+" in
-  Str.global_replace regexp_newline_segment " " str
-
 let escape_underscores str =
   Str.global_replace regex_underscore escaped_underscore str
 
 let underscore_to_space str = Str.global_replace regex_underscore " " str
+let spec_var_to_prose var_str = underscore_to_space var_str
+let to_math_mode str = Printf.sprintf "$%s$" str
+let spec_var_to_template_var var_str = "{" ^ var_str ^ "}"
+let remove_underscores str = Str.global_replace (Str.regexp_string "_") "" str
+let ignore_macro = "\\Ignore"
+
+let shrink_whitespace str =
+  let regexp_newline_segment = Str.regexp "[ \n\r]+" in
+  Str.global_replace regexp_newline_segment " " str
 
 type font_type = Text | TextTT | TextSF | TextSC | TextIT
 
-(** [pp_one_arg_macro fmt macro_name pp_arg arg] renders a LaTeX macro named
+(** [pp_one_arg_macro macro_name pp_arg fmt arg] renders a LaTeX macro named
     [macro_name] with one argument, applied to the argument is [arg], which is
     formatted using [pp_arg]. *)
-let pp_one_arg_macro fmt macro_name pp_arg arg =
+let pp_one_arg_macro macro_name pp_arg fmt arg =
+  assert (Str.string_match regexp_only_letters macro_name 0);
   fprintf fmt "\\%s{%a}" macro_name pp_arg arg
 
 (** [pp_macro fmt macro_name] renders a LaTeX macro by prefixing it with a
     backslash. *)
-let pp_macro fmt macro_name = fprintf fmt "\\%s" macro_name
+let pp_macro fmt macro_name =
+  assert (Str.string_match regexp_only_letters macro_name 0);
+  fprintf fmt "\\%s" macro_name
 
 (** [spec_var_to_latex_var ~font_type var_str] returns a version of the
     specification variable [var_str] suitable for LaTeX math mode using the
     specified [font_type]. *)
 let spec_var_to_latex_var ~font_type var_str =
-  if String.equal var_str "_" then "\\Ignore"
+  if String.equal var_str "_" then ignore_macro
   else
     let font_macro_name =
       match font_type with
@@ -46,22 +55,23 @@ let spec_var_to_latex_var ~font_type var_str =
       | TextSC -> "textsc"
       | TextIT -> "textit"
     in
-    Format.asprintf {|%a{%s}|} pp_macro font_macro_name
+    Format.asprintf "%a"
+      (pp_one_arg_macro font_macro_name pp_print_string)
       (escape_underscores var_str)
 
-let spec_var_to_prose var_str = underscore_to_space var_str
-let to_math_mode str = "$" ^ str ^ "$"
-let spec_var_to_template_var var_str = "{" ^ var_str ^ "}"
-let remove_underscores str = Str.global_replace (Str.regexp_string "_") "" str
-let elem_name_to_math_macro elem_name = "\\" ^ remove_underscores elem_name
+let elem_name_to_math_macro elem_name =
+  Format.asprintf "%a" pp_macro (remove_underscores elem_name)
+
 let pp_comma = fun fmt () -> fprintf fmt ", "
 
 (** [substitute_spec_vars_by_latex_vars math_mode s vars] returns a string [s]
-    with every variable like [my_var] is substituted into [\texttt{my\_var}],
-    which makes the returned string suitable to typsetting with LaTeX. If
-    [math_mode] is true then the result is surrounded by [$$]. *)
+    with every instance of [{my_var}] substituted by [\texttt{my\_var}]. If
+    [math_mode] is true [{my_var}] is substituted by [$\texttt{my\_var}$]. This
+    makes the returned string suitable to typesetting with LaTeX. *)
 let substitute_spec_vars_by_latex_vars ~math_mode str vars =
   let module StringMap = Map.Make (String) in
+  (* Create a map from template variables to LaTeX variables
+    to make replacement efficient. *)
   let substitutions_map =
     List.fold_left
       (fun acc_map var_str ->
@@ -74,26 +84,22 @@ let substitute_spec_vars_by_latex_vars ~math_mode str vars =
       StringMap.empty vars
   in
   let template_var_regexp = Str.regexp "{[a-zA-Z0-9_']+}" in
-  let blocks = Str.full_split template_var_regexp str in
-  List.fold_left
-    (fun acc block ->
-      match block with
-      | Str.Text txt -> acc ^ txt
-      | Str.Delim var -> (
-          match StringMap.find_opt var substitutions_map with
-          | Some latex_var -> acc ^ latex_var
-          | None -> acc ^ var))
-    "" blocks
+  Utils.string_replace_all template_var_regexp
+    (fun var ->
+      match StringMap.find_opt var substitutions_map with
+      | Some latex_var -> latex_var
+      | None -> var)
+    str
 
 (** [pp_texthypertarget fmt target_str] renders a hypertarget for [target_str]
     suitable for inclusion in LaTeX text mode, using the formatter [fmt]. *)
 let pp_texthypertarget fmt target_str =
-  fprintf fmt {|\texthypertarget{%s}|} target_str
+  (pp_one_arg_macro "texthypertarget" pp_print_string) fmt target_str
 
 (** [pp_mathhypertarget fmt target_str] renders a hypertarget for [target_str]
     suitable for inclusion in LaTeX math mode, using the formatter [fmt]. *)
 let pp_mathhypertarget fmt target_str =
-  fprintf fmt {|\mathhypertarget{%s}|} target_str
+  (pp_one_arg_macro "mathhypertarget" pp_print_string) fmt target_str
 
 (** A type enumerating different kinds of parentheses. *)
 type parenthesis = Parens | Braces | Brackets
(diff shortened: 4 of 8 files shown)


asllib/aslspec/LayoutUtils.ml
```diff
@@ -0,0 +1,37 @@
+(** A module defining utility functions for layouts. *)
+
+open AST
+
+let horizontal_if_unspecified layout list =
```
hrutvik: nit: throughout this file, use of `list` as a variable name does funny things to syntax highlighting

asllib/aslspec/SpecParser.mly
```diff
@@ -117,32 +121,32 @@ let list1(x) :=
 let spec := elems=terminated(list(elem), EOF); { elems }
 
 let elem :=
-    | type_definition
-    | relation_definition
-    | constant_definition
-    | render_definition
+    | ~=type_definition; SEMI; { type_definition }
```
hrutvik: nit: refactor into a `terminated`?

asllib/aslspec/latex.ml
```diff
@@ -0,0 +1,267 @@
...
+  let right = if large then {|\right|} else "" in
+  match parenthesis with
+  | Parens -> fprintf fmt "%s(%a%s)" left pp_elem elem right
+  | Braces -> fprintf fmt "%s\\{%a%s\\}" left pp_elem elem right
+  | Brackets -> fprintf fmt "%s[%a%s]" left pp_elem elem right
+
+(** [over_text fmt fmt_top top fmt_bottom bottom] renders
+    [\overtext{top}{bottom}] where [top] and [bottom] are formatted using
+    [fmt_top] and [fmt_bottom] respectively. *)
+let over_text fmt fmt_top top fmt_bottom bottom =
```
hrutvik: nit: would `overtext` be more in line with the LaTeX?

asllib/aslspec/latex.ml
```diff
@@ -0,0 +1,267 @@
...
+  fprintf fmt {|\overtext{%a}{%a}|} fmt_top top fmt_bottom bottom
+
+(** [pp_field_name_to_latex field_name] formats [field_name] - a field name
```
hrutvik, line 125: nit: comment doesn't match function name. Should the function be in the `pp_*` style for consistency?

asllib/aslspec/latex.ml
```diff
@@ -0,0 +1,267 @@
...
+  fprintf fmt "@[<v>\\begin{array}{%s}@ %a@ \\end{array}@]" alignment
+    (pp_print_list ~pp_sep:(fun fmt () -> fprintf fmt {|\\@ |}) pp_one_row)
+    pp_fun_rows
+
+(** [pp_sep_elements ~pp_sep ~alignment pp_elem layout fmt elements] formats a
+    list of [elements] separated by [pp_sep] according to [layout]. Each element
+    is formatted using [pp_elem]. The [alignment] string specifies the alignment
+    of each column and is copied directly to the array environment when the
+    [layout] is vertical. *)
+let pp_sep_elements ~pp_sep ~alignment pp_elem layout fmt elements =
```
hrutvik: `pp_sep_elements` and `pp_connect_elements` seem very similar, both in description and code. What is their difference, and could they be combined?

asllib/aslspec/latex.ml
```diff
@@ -0,0 +1,267 @@
...
+      let elem_pp_funs =
+        List.mapi
+          (fun elem_counter elem_with_layout ->
+            [
+              (fun fmt ->
+                pp_elem fmt elem_with_layout;
+                if elem_counter < List.length elements_with_layouts - 1 then
+                  pp_sep fmt ()
+                else ());
+            ])
```
hrutvik: Related to discussion on a previous PR - could the `pp_sep` part be handled by `pp_latex_array` with a suitable (optional) argument?

asllib/aslspec/tests.t/run.t
```diff
@@ -7,7 +7,7 @@
   Generated LaTeX macros into generated_macros.tex
 
   $ aslspec type_name.bad
-  Syntax Error: illegal element-defining identifier: t2 around type_name.bad line 2 column 1
+  Syntax Error: illegal element-defining identifier: t2 around type_name.bad line 1 column 41
```
hrutvik, line 10: Is this change significant?

asllib/aslspec/LayoutUtils.ml
Roman-Manevich: changed to `elements`.

asllib/aslspec/tests.t/run.t
Roman-Manevich: As the grammar changed and became more modular, the syntax error is discovered for a different (deeper nested) non-terminal, which is actually slightly closer to where the error is. So, it's a slight improvement.

asllib/aslspec/latex.ml
Roman-Manevich: The difference is whether the separator goes on a column of its own (`pp_connect_elements`) or not (`pp_sep_elemnets`). I'll refactor.

asllib/aslspec/latex.ml
Roman-Manevich: The thing is `pp_latex_array` accepts lists of lists of no-op functions. That makes the calls very cumbersome. I feel like capturing the actual patterns and offering simple functions which take formatting functions (while under the hood they use `pp_latex_array`) is more ergonomic.

asllib/aslspec/PP.ml
```diff
@@ -28,19 +28,19 @@ let pp_attribute fmt = function
   | MathMacroAttribute macro -> pp_print_string fmt macro
 
 let pp_attribute_key_value fmt (key, value) =
-  fprintf fmt {|%a = %a|} pp_attribute_key key pp_attribute value
+  fprintf fmt {|%a = %a,|} pp_attribute_key key pp_attribute value
```
HadrienRenaud: Why keep the `{|`?

```suggestion
  fprintf fmt "%a = %a," pp_attribute_key key pp_attribute value
```

asllib/aslspec/main.ml
```diff
@@ -91,6 +91,12 @@ let pp_std spec =
   PP.pp_spec Format.std_formatter spec;
   Format.print_newline ()
 
+module TextColor = struct
+  let red = "\027[31m"
+  let green = "\027[32m"
+  let reset_color = "\027[0m"
+end
```
HadrienRenaud, line 98: Probably not in the scope of this PR, but I just wanted to flag that there is a nice way of doing this in OCaml with the "semantic tags" thing in `Format`.

asllib/aslspec/spec.ml
```diff
@@ -5,26 +5,117 @@ open AST
...
-       not have a short-circuit macro defined. All relation output types, \
-       except for the first one, must reference types with short-circuit \
-       macros."
-      relation_name PP.pp_type_term term type_name
-  in
-  raise (SpecError msg)
+module Error = struct
+  let undefined_reference id context =
+    let msg = Format.asprintf "Undefined reference to '%s' in %s" id context in
+    raise (SpecError msg)
```
HadrienRenaud: Nit: why not extract `raise (SpecError foo)` in a separate function?

```suggestion
  let spec_error msg = raise (SpecError msg)

  let undefined_reference id context =
    spec_error @@ Format.asprintf "Undefined reference to '%s' in %s" id context
```

asllib/aslspec/spec.ml
```diff
@@ -5,26 +5,117 @@ open AST
...
+
+  let instantiation_failure_not_a_type term label =
+    let msg =
+      Format.asprintf
+        "The type term `%a` cannot be instantiated since '%s' is not a type"
+        PP.pp_type_term term label
+    in
+    raise (SpecError msg)
+
+  let bad_layout term layout consistent_layout =
```
HadrienRenaud: suggestion: put a label to differentiate 2 arguments of the same type

asllib/aslspec/spec.ml
```diff
@@ -235,9 +330,7 @@ let make_id_to_definition_node definition_nodes =
   List.fold_left
     (fun acc_map node ->
       let name = definition_node_name node in
-      if StringMap.mem name acc_map then
-        let msg = Format.sprintf "Duplicate definitions found for '%s'" name in
-        raise (SpecError msg)
+      if StringMap.mem name acc_map then Error.duplicate_definition name
       else StringMap.add name node acc_map)
```
HadrienRenaud: Consider using `update`? Maybe to be refactored in a separate function

```suggestion
      StringMap.update name (function
      | Some _ -> Error.duplicate_definition name
      | None -> Some node) acc_map)
```

asllib/aslspec/spec.ml
```diff
@@ -251,61 +344,45 @@ type t = {
...
+    let consistent_layout = Layout.default_for_type_term term in
     let open Layout in
     match (term, layout) with
     | Label _, Unspecified -> ()
-    | Label _, _ -> raise (SpecError msg)
+    | Label _, _ -> Error.bad_layout term layout consistent_layout
     | TypeOperator { term = _, t }, _ -> check_layout t layout
     | LabelledTuple { components }, Horizontal cells
     | LabelledTuple { components }, Vertical cells ->
         if List.length components <> List.length cells then
```
HadrienRenaud: nit: consider using `List.compare_lengths`

```suggestion
        if List.compare_lengths components cells <> 0 then
```

There are at least 4 occurrences of this pattern

asllib/aslspec/spec.ml
```diff
@@ -790,23 +851,12 @@ module Check = struct
                       (List.length def_components)
```
HadrienRenaud: Why not using `List.compare_lengths`?

asllib/aslspec/spec.ml
```diff
@@ -897,16 +930,43 @@ module Check = struct
   end
 end
 
+(** [from_ast ast] converts an AST into a specification after ensuring its
+    correctness. *)
 let from_ast ast =
   let definition_nodes = list_definition_nodes ast in
-  let defined_ids = List.map definition_node_name definition_nodes in
+  let defined_ids =
+    List.map definition_node_name definition_nodes |> List.rev
```
HadrienRenaud: Can you add a comment explaining why the `rev` is needed?

Why not using `List.rev_map`?

```suggestion
    List.rev_map definition_node_name definition_nodes
```

asllib/aslspec/LayoutUtils.ml
```diff
@@ -0,0 +1,37 @@
...
+  match layout with
+  | Horizontal _ | Vertical _ -> layout
+  | _ -> Horizontal (List.map (fun _ -> Unspecified) list)
```
HadrienRenaud: Consider matching on the wanted item, this makes it confusing otherwise

Same for `vertical_if_unspecified`

```suggestion
  match layout with
  | Unspecified -> Horizontal (List.map (fun _ -> Unspecified) list)
  | Horizontal _ | Vertical _ -> layout
```

asllib/aslspec/LayoutUtils.ml
```diff
@@ -0,0 +1,37 @@
...
+      = [ (a, Unspecified); (b, Vertical [ Unspecified; Unspecified ]) ]
+    ]} *)
+let apply_layout_to_list layout list =
+  let args_layout =
+    match layout with
+    | Horizontal l | Vertical l -> l
+    | Unspecified -> List.map (fun _ -> Unspecified) list
+  in
+  assert (List.compare_lengths list args_layout = 0);
+  List.combine list args_layout
```
HadrienRenaud:
```suggestion
  match layout with
  | Horizontal li' | Vertical li' -> List.combine li li'
  | Unspecified -> List.map (fun elt -> (elt, Unspecified)) li
```

asllib/aslspec/latex.ml
```diff
@@ -0,0 +1,267 @@
...
+    let font_macro_name =
+      match font_type with
+      | Text -> "text"
+      | TextTT -> "texttt"
+      | TextSF -> "textsf"
+      | TextSC -> "textsc"
+      | TextIT -> "textit"
+    in
+    Format.asprintf {|%a{%s}|} pp_macro font_macro_name
+      (escape_underscores var_str)
```
HadrienRenaud: Why not use `pp_one_arg_macro`:

```suggestion
    let pp_escape_underscore fmt s = pp_print_string fmt (escape_underscore s) in
    pp_one_arg_macro fmt font_macro_name pp_escape_underscore var_str
```

asllib/aslspec/latex.ml
```diff
@@ -0,0 +1,267 @@
...
+
+let spec_var_to_prose var_str = underscore_to_space var_str
+let to_math_mode str = "$" ^ str ^ "$"
+let spec_var_to_template_var var_str = "{" ^ var_str ^ "}"
```
HadrienRenaud: Consider using `sprintf` to do concatenation:

```suggestion
let spec_var_to_template_var var_str = Printf.sprintf "{%s}" var_str
```

asllib/aslspec/latex.ml
```diff
@@ -0,0 +1,267 @@
...
+let remove_underscores str = Str.global_replace (Str.regexp_string "_") "" str
```
HadrienRenaud: Should this be moved with the rest of the underscore handling?

asllib/aslspec/latex.ml
```diff
@@ -0,0 +1,267 @@
...
+let elem_name_to_math_macro elem_name = "\\" ^ remove_underscores elem_name
```
HadrienRenaud: Consider using `pp_macro`:

```suggestion
let pp_math_macro fmt elem_name = pp_macro fmt (remove_underscores elem_name)
```

asllib/aslspec/latex.ml
```diff
@@ -0,0 +1,267 @@
...
+let pp_comma = fun fmt () -> fprintf fmt ", "
+
+(** [substitute_spec_vars_by_latex_vars math_mode s vars] returns a string [s]
+    with every variable like [my_var] is substituted into [\texttt{my\_var}],
+    which makes the returned string suitable to typsetting with LaTeX. If
+    [math_mode] is true then the result is surrounded by [$$]. *)
+let substitute_spec_vars_by_latex_vars ~math_mode str vars =
+  let module StringMap = Map.Make (String) in
```
HadrienRenaud, line 72: Maybe this is something that could be put in `utils`?

asllib/aslspec/latex.ml
```diff
@@ -0,0 +1,267 @@
...
+        let template_var = spec_var_to_template_var var_str in
+        StringMap.add template_var latex_var acc_map)
+      StringMap.empty vars
+  in
+  let template_var_regexp = Str.regexp "{[a-zA-Z0-9_']+}" in
+  let blocks = Str.full_split template_var_regexp str in
+  List.fold_left
+    (fun acc block ->
+      match block with
+      | Str.Text txt -> acc ^ txt
```
HadrienRenaud: consider putting it into a list and using `String.concat` at the end, this way you allocate the string only once.

asllib/aslspec/latex.ml
```diff
@@ -0,0 +1,267 @@
...
+  let substitutions_map =
+    List.fold_left
+      (fun acc_map var_str ->
+        let latex_var = spec_var_to_latex_var ~font_type:TextTT var_str in
+        let latex_var =
+          if math_mode then to_math_mode latex_var else latex_var
+        in
+        let template_var = spec_var_to_template_var var_str in
+        StringMap.add template_var latex_var acc_map)
+      StringMap.empty vars
```
HadrienRenaud: nit: often `String.of_list bindings` is faster than `List.fold_left String.add bindings`

```suggestion
    List.map (fun var_str ->
        let latex_var = spec_var_to_latex_var ~font_type:TextTT var_str in
        let latex_var = if math_mode then to_math_mode latex_var else latex_var in
        let template_var = spec_var_to_template_var var_str in
        (template_var, var_str)) vars
      |> StringMap.of_list
```

asllib/aslspec/latex.ml
```diff
@@ -0,0 +1,267 @@
...
+  let blocks = Str.full_split template_var_regexp str in
+  List.fold_left
+    (fun acc block ->
+      match block with
+      | Str.Text txt -> acc ^ txt
+      | Str.Delim var -> (
+          match StringMap.find_opt var substitutions_map with
+          | Some latex_var -> acc ^ latex_var
+          | None -> acc ^ var))
+    "" blocks
```
HadrienRenaud: Consider extracting the `Str.replace` with a function as argument logic in a separate function

```ocaml
let replace_fun regex f s =
  Str.full_split regex s
  |> List.map (function Str.Text txt -> txt | Str.Delim d -> f d)
  |> String.concat ""
```

Then:
```suggestion
  replace_fun template_var_regex (fun var ->
    match StringMap.find_opt var subsitutions_map with
    | Some latex_var -> latex_var
    | None -> var) str
```

asllib/aslspec/latex.ml
```diff
@@ -0,0 +1,267 @@
...
+
+(** [pp_one_arg_macro fmt macro_name pp_arg arg] renders a LaTeX macro named
+    [macro_name] with one argument, applied to the argument is [arg], which is
+    formatted using [pp_arg]. *)
+let pp_one_arg_macro fmt macro_name pp_arg arg =
+  fprintf fmt "\\%s{%a}" macro_name pp_arg arg
+
+(** [pp_macro fmt macro_name] renders a LaTeX macro by prefixing it with a
+    backslash. *)
+let pp_macro fmt macro_name = fprintf fmt "\\%s" macro_name
```
HadrienRenaud: Could we add a check taht there are no non-escaped underscores in `macro_name`?

asllib/aslspec/latex.ml
```diff
@@ -0,0 +1,267 @@
...
+      | Str.Delim var -> (
+          match StringMap.find_opt var substitutions_map with
+          | Some latex_var -> acc ^ latex_var
+          | None -> acc ^ var))
+    "" blocks
+
+(** [pp_texthypertarget fmt target_str] renders a hypertarget for [target_str]
+    suitable for inclusion in LaTeX text mode, using the formatter [fmt]. *)
+let pp_texthypertarget fmt target_str =
+  fprintf fmt {|\texthypertarget{%s}|} target_str
```
HadrienRenaud: Use `pp_one_arg_macro`?

asllib/aslspec/latex.ml
```diff
@@ -0,0 +1,267 @@
...
+    a LaTeX array environment. The [alignment] string specifies the alignment of
+    each column and is copied directly to the array environment. The
+    [pp_fun_rows] is a list of lists, where the outer list represents the array
+    rows and the inner lists represent the columns. *)
+let pp_latex_array alignment fmt pp_fun_rows =
+  let () = assert (Str.string_match (Str.regexp "[lcr]+$") alignment 0) in
+  let num_columns = String.length alignment in
+  let pp_elt = fun fmt pp_fun -> pp_fun fmt in
+  let pp_one_row fmt pp_funs =
+    let () = assert (List.compare_length_with pp_funs num_columns = 0) in
```
HadrienRenaud: Use `List.compare_lengths`?

asllib/aslspec/spec.ml
Roman-Manevich: I rewrote the code to maintain the order without having to reverse.

asllib/aslspec/LayoutUtils.ml
Roman-Manevich: I think it's more confusing this way. RIght now there are two stages: 1) get the list, 2) combine, and I'd like to keep this separation..

asllib/aslspec/latex.ml
Roman-Manevich: Sure, I'll use it. But since I'm producing a string and there's no formatter at hand, I think this is as elegant as I can make it: `Format.asprintf "%a" (pp_one_arg_macro font_macro_name pp_print_string) (escape_underscores var_str)`

asllib/aslspec/latex.ml
Roman-Manevich: I need to get a string so I went with `Format.asprintf "%a" pp_macro (remove_underscores elem_name)`

asllib/aslspec/latex.ml
Roman-Manevich: I think this is very rendering-specific. I'm trying to keep utils for generic data type stuff. I don't want to look for rendering-specific stuff elsewhere.

asllib/aslspec/latex.ml
Roman-Manevich: I added a check that the macro name contains only letters.

asllib/aslspec/latex.ml
Roman-Manevich: But `alignment` is not a list - it's a string.

HadrienRenaud: Ah ooops that makes sense sorry

asllib/aslspec/LayoutUtils.ml
HadrienRenaud: ok fair enough

asllib/aslspec/latex.ml
HadrienRenaud: ok, I think ideally we'd used as little points as possible to produce macros

asllib/aslspec/latex.ml
HadrienRenaud: I was talking about the `StringMap` module, right?

Ok otherwise.

asllib/aslspec/latex.ml
HadrienRenaud: Ah looks like it's not in Stdlib before 5.1. Maybe we can forget about this one.

asllib/aslspec/latex.ml
Roman-Manevich: I added `string_replace_all` to `utils.ml`

PR #1572, opened 2025-11-15
[asl reference] removed backup macros for types and relations

Roman-Manevich:
I used a python script generated by Github Co-pilot to remove all call to the macros `\BackupOriginalAST{...}`, `\BackupOriginalType{...}`, and
`\BackupOriginalRelation{...}`.
All files were scanned manually post-transformation.
Some macro calls mistakenly included references to examples, which I've pulled out (the script checked for this by making sure macro calls don't contain "example" in a case-insensitive way).
The script and a PDF diff will be shared offline.


hrutvik (approved):
On the basis of a PDF diff sent offline

PR #1574, opened 2025-11-17
[tools, gen] Introduce new tool `cat2config`

fsestini:
This PR introduces a new tool `cat2config`, which can be used to generate `diy7` relaxations corresponding to relations defined in a `.cat` model. While supporting various architectural features like MTE and VMSA is planned, the tool will initially focus on supporting a subset of so-called "baseline" relations as defined in `aarch64.cat`.

The tool is in a good spot now, but many more features need to be added. In particular,
the subset of the cat language that this tool supports is limited, although large enough to handle the memory model's "baseline" features.
(The `gen/libdir/forbidden.conf` file is used as reference to define what "baseline" means.)

## Usage

The tool takes as input a cat file and a list of options, most notably `--let STR`, which instructs the tool to generate diy7 relaxations corresponding to `STR`'s definition from the input cat file.
For example:

```sh
mcat2config.exe --let IFB-ob herd/libdir/aarch64.cat
[DpCtrl, ISB]
[DpCtrlCsel, ISB]
[DpAddr, ISB***]
[DpAddrCsel, ISB***]
[DSB.SY***, ISB]
[DSB.LD*R*, ISB]
[DSB.ST*W*, ISB]
```

Results can also be printed in diy7 configuration file-compatible format, and with additional metadata.
For example:

```sh
mcat2config.exe --let IFB-ob --conf --dump origin herd/libdir/aarch64.cat
### IFB-ob
## [Exp & R]; ctrl; [IFB]; po
-safe [DpCtrl, ISB]
## [Exp & R]; pick-ctrl-dep; [IFB]; po
-safe [DpCtrlCsel, ISB]
## [Exp & R]; addr; [Exp & M]; po; [IFB]; po
-safe [DpAddr, ISB***]
## [Exp & R]; pick-addr-dep; [Exp & M]; po; [IFB]; po
-safe [DpAddrCsel, ISB***]
## DSB-ob; [IFB]; po
-safe [DSB.SY***, ISB] [DSB.LD*R*, ISB] [DSB.ST*W*, ISB]
```

It is possible to enable per-module logging. This is especially useful for debugging.
For example, here is how to run the tool with standard logging enabled for the normalisation step:

```sh
mcat2config.exe --log 'translation' ...
```

Here is how to get full debugging logs for the translation step, but only warnings for the normalization step:

```sh
mcat2config.exe --log 'translation:debug' --log 'normalization:warning' ...
```

## Tests

The tool comes with a test suite in `gen/cat2config/test`, which can be run via `dune runtest` or `dune build @gen/cat2config/test/runtest`.
These are end-to-end tests that check the full output of the tool against an expected text.
`gen/cat2config/test/baseline_conf.t` in particular aims to verify that the tool can recreate `gen/libdir/forbidden.conf`.

The tests are implemented as [dune cram tests](https://dune.readthedocs.io/en/latest/reference/cram.html).
They rely on the cat definitions of the AArch64 memory model from `herd/libdir`. ~~The cat files are _copied_ over to `gen/cat2config/test`, not _symlinked_, to allow the tool and the model to be developed independently.~~

## Design decisions

The code is structured in two main parts: evaluation and translation. `cat`
expressions are not translated directly into `diy7` relaxation; instead, we use
a shared data type of expressions that acts as an intermediate representation
(IR) between the _evaluation_ step and the _translation_ step.

- evaluation: `cat --> IR`
- translation: `IR --> diy7 relaxations`

`cat2config` consumes `cat` files as input, but the `cat` language is, currently, significantly more expressive than the `diy7` relaxation syntax. The two-step approach outlined above is an attempt to bridge the gap while favouring maintainability and separation of concerns:

- _evaluation_ is concerned with taking the raw `cat` input and reducing it/simplifying it down to a core subset of `cat` that is more amenable to being translated into `diy7` relaxation. This step is realised in `normalization.ml` and `ir.ml`. The actual normalization in done in `normalization.ml` against an abstract signature of "set and relation expressions", which `ir.ml` implements.
- _translation_ is concerned with translating the reduced IR expressions into `diy7` relaxations. It's implemented in `translation.ml`.

### Intermediate representation (IR)

The IR data structure thus fulfils two main roles:

- To precisely define the subset of the cat language that cat2config is able to handle.

  cat2config works by evaluating each relation and set expression from the input cat model into a simplified, equivalent expression that is built out of just three operations:

  - intersection of primitive values
  - sequence of intersections
  - union of intersections

  For example, this `cat` definition adheres to the format described above:

  ```
  let Exp-obs =
      [Exp & M]; rf & ext; [Exp & M]
    | [Exp & M]; ca & ext; [Exp & M]
  ```

  In most cases, we can use the properties of set-theoretic operations to rewrite expressions
  that don't follow this structure into equivalent expressions that do. For example:

  ```
  let aob = [Exp & M]; (lxsx | amo); [Exp & M]
  ```

  is equivalent to:

  ```
  let aob =
      [Exp & M]; lxsx; [Exp & M]
    | [Exp & M]; amo; [Exp & M]
  ```

  Some `cat` expressions cannot be translated into this format. For those, currently, `cat2config` will return an error.

  The reason for using this structure as the basis for `cat2config`'s translation is that:

  - expressions in this form have a more or less clear mapping into `diy7` relaxations
  - mapping expressions _not_ in this form appears to be very challenging

  As the tool and our understanding develops, this "supported subset" might grow to include more of what can be expressed in `cat`, but for now this is what the tool can handle.

- To provide a "normal form" that equivalent cat expressions can be reduced to.

  The IR data structure used internally by `cat2config` is designed to be a "normal form" that tries to eliminate as much duplication as possible, by construction.

  For example, these two equivalent expressions have the same IR representation:

  ```
  let aob = [Exp & M]; (lxsx | amo); [Exp & M]
  let aob = [Exp & M]; lxsx; [Exp & M] | [Exp & M]; amo; [Exp & M]
  ```

Implementing the translation part of the tool as map from IR to diy7 (as opposed to a map from raw cat AST to diy7) allows for some simplification in the translation logic. This is especially the case for those parts of the translation logic that work by performing purely syntactic pattern matching on the input expressions. For example, let's say our implementation of `cat2config` includes a translation rule where every occurrence of the cat expression `[M]; po & loc; [M]` is translated into the diy7 relaxation `Pos**` by performing a syntactic matching on the cat expression. Note that this expression is equivalent to this other expression: `([M]; po; [M]) & loc`. However, this second expression would fail to trigger the translation rule just described if we were to pass it to the tool as-is, since our translation rule performs a _syntactic_ matching. We want `cat2config` to be, as much as possible, stable under "semantic equivalence" (that is, the tool should not produce different outputs from equivalent inputs). One way towards this is to use normal forms that provide a single canonical representation for equivalent inputs.

It should be noted that neither of these points _require_ the use of a separate IR datastructure. Since the expression language that cat2config handles is a strict subset of the full cat language, we could just have used the existing cat AST datatype to also represent cat2config's IR. The biggest advantage with using a separate IR data structure is in the static guarantees it brings. By encoding the AST structure that `cat2config` expects as an OCaml type, the compiler can help verify that all the assumptions we are making about these structures are not inadvertently violated.

For now, I think the IR datatype is tiny (only a handful of constructors) and private (it is not exposed to other `herdtools7` modules outside of `cat2config`) enough that its advantages outweigh the maintenance costs.

### Logging

As this is a completely new tool, I've taken the liberty to rely on the [logs](https://opam.ocaml.org/packages/logs/logs.0.8.0/) library to implement logging. This is a tiny well-established logging library that makes it particularly easy to define different logging sources (for example, to implement `--log 'normalization'` vs `--log 'translation'` as shown above), each with their own independent logging levels (debugging, info, warnings, etc.).

While I personally think the project would benefit from gradually introducing this library, this decision is of course up for discussion. As introducing this extra library is a project-wide change, it doesn't _need_ to happen on this PR, or any time soon. Therefore, please feel free to push back on this. If needed, I can very quickly refactor the tool to remove dependency on `logs`.

## Further work (for future PRs):

- Use `Amo.Cas`/`Amo.Swp` relaxations to "implement" read/write effects `R`/`W`, in addition to plain loads/stores
- Do not output relaxations between access types `X` and `Y` if a "weaker" relaxation between the same access types has already been generated. For example, we should not include `DMB.STsWR` in the generated list of `PosWR` is already present.
- Add regression test suite that integrates `cat2config` with `diy7` and `herd7`, to verify that the litmus test generated for a cat relation `r` do actually "hit" that relation when simulated via `herd7`.
- Add support for MTE, Ifetch, VMSA


gen/cat2config/.ocamlformat
ShaleXIONG, line 1: Feel free to switch on formatting, maybe after we settle other files.

fsestini: Within `gen/cat2config`, formatting is indeed already enabled, as per `disable = false` (a roundabout way to say it's enabled) which overrides the `disable = true` that is globally set for the repo.

ShaleXIONG: Ohhh I see.

gen/cat2config/ast_utils.ml
```diff
@@ -0,0 +1,47 @@
+module Pprinters : sig
+  val show_exp : AST.exp -> string
+end = struct
+  type op1 = [%import: AST.op1] [@@deriving show { with_path = false }]
```
fsestini: I will be removing these uses of `ppx` (and removing `ppx` as a dependency overall, unless needed for something else), as I realised I will be needing more control over how AST expressions get pretty printed, therefore I'll need to write them by hand anyways.

gen/cat2config/test/baseline.t
```diff
@@ -0,0 +1,135 @@
...
+  [DpCtrlCsels*, ISB]
+  [DpAddrd*, ISBd**]
+  [DpAddrs*, ISBd**]
+  [DpAddrd*, ISBs**]
+  [DpAddrs*, ISBs**]
+  [DpAddrCseld*, ISBd**]
+  [DpAddrCsels*, ISBd**]
+  [DpAddrCseld*, ISBs**]
+  [DpAddrCsels*, ISBs**]
+  [DSB.SYd**, ISB]
```
ShaleXIONG: Those test cases on `DSB.SYd**, ISB` probably need to be `DSB.SYd**, ISB***`.
I think the case is from `| DSB-ob;[IFB];po`, while in transition we are effectively translate `| DSB-ob;[IFB];po; [M]`

gen/cat2config/test/baseline.t
```diff
@@ -0,0 +1,135 @@
...
+  [Amo.Swp, PosWRPQ]
+  [Amo.Cas, PosWRPQ]
+  $ mcat2config7 -set-libdir ./libdir -let bob libdir/aarch64.cat
+  DMB.SYd**
+  DMB.SYs**
+  DMB.LDdR*
+  DMB.LDsR*
+  DMB.STdWW
+  DMB.STsWW
+  [Amo.SwpAL, Pod**LP]
```
ShaleXIONG: I do not know which syntax will be easier for you to translate:
`Amo.SwpAL, Pod**LP` vs `A, Amo.Swp, L, Pod**`. I personally think the latter match better to the `cat` file. However I can guess the existing code chosen the former, so we might just push the question to future?

gen/cat2config/test/baseline.t
fsestini: Yes, this translation comes from `DSB-ob; [IFB]; po`, which I expand into `[M]; po; [dsb.full]; po; [ISB]; po; [M]` by inlining the definition of `DSB-ob` and adding a `[M]` at the end.

I don't follow where `DSB.SYd**, ISB***` would come from. 
Wouldn't `DSB.SYd**, ISB***` imply that there is a `[M]` between the DSB and the ISB? Since there isn't one, then the two barriers should be contiguous right?

ShaleXIONG: [update] actually `DSB-ob = [M | DC.CVAU | IC]; po; [dsb.full]; po; [~(Imp & TTD & M | Imp & Instr & R)]` so it becomes `[M | DC.CVAU | IC]; po; [dsb.full]; po; [~(Imp & TTD & M | Imp & Instr & R)]; po; [ISB]`. If we add a `[M]` to the end. it will becomes `DSB.SY*** ISB***`. There are 3 memory access here.

gen/cat2config/test/baseline.t
fsestini: I think the `Amo.SwpAL` syntax might be slightly more convenient for me when doing the translation, but it should be possible to convert the relaxation to the other format post-facto. I'll keep that in mind for the future.

gen/cat2config/test/baseline.t
fsestini: If you inline:
```
DSB-ob = [M | DC.CVAU | IC]; po; [dsb.full]; po; [~(Imp & TTD & M | Imp & Instr & R)]
```
into `DSB-ob; [IFB]; po`, shouldn't that become:
```
[M | DC.CVAU | IC]; po; [dsb.full]; po; [~(Imp & TTD & M | Imp & Instr & R)]; [IFB]; po
```
It seems to me that you have inserted an extra `po` relation before `[IFB]`.

ShaleXIONG: Ahh, I see.

gen/cat2config/diy_utils.ml
```diff
@@ -0,0 +1,31 @@
+[@@@warning "-40-42"]
```
ShaleXIONG: Can we remove this warning disable?

fsestini: Ah yes, I put that pragma there temporarily for ease of development. Will remove once the PR stabilises.

I do think however that those two warnings should be disabled by default, at least for `cat2config`. I think they are mainly relevant to old versions of OCaml (for example, [-42](https://discuss.ocaml.org/t/can-warning-42-be-ignored/2870) is only relevant for OCaml <= 4.0.0, which is older than the oldest version we officially support).

gen/cat2config/translation.ml
```diff
@@ -0,0 +1,328 @@
...
+      l
+  in
+  if st.core = [] then st.relaxs else []
+
+let pp_relax : relax -> string = function
+  | Relax [ rlx ] -> E.pp_edge rlx
+  | Relax rlxs ->
+      Format.sprintf "[%s]" (String.concat ", " (List.map E.pp_edge rlxs))
+
+let pp_relax_list : relax list -> string =
```
ShaleXIONG: I assume there is no different between (1) put the parameter after the function name but before the `=`  vs (2) the lambda-style you use here?

gen/common/util.ml
```diff
@@ -61,14 +61,78 @@ let parse_cmdline options get_cmd_arg =
...
     (sprintf "Usage %s [options] [arg]*\noptions are:" Sys.argv.(0))
 
+module type Monad = sig
+  type 'a t
+  val pure : 'a -> 'a t
+  val bind : 'a t -> ('a -> 'b t) -> 'b t
+
+  module Infix : sig
+    val (>>=) : 'a t -> ('a -> 'b t) -> 'b t
+    val (let*) : 'a t -> ('a -> 'b t) -> 'b t
```
ShaleXIONG: These two functions `(>>=)` and `(let*)` are always the same?

gen/common/util.mli
```diff
@@ -32,7 +32,38 @@ val arch_opt : Archs.t ref -> spec
 
 val parse_cmdline : spec list -> (string -> unit) -> unit
 
+module type Monad = sig
+  type 'a t
+  val pure : 'a -> 'a t
```
ShaleXIONG: Not sure if it worths to use the function name from Ocaml official tutorial (https://ocaml.org/docs/monads), namely, `return` rather than `pure`.

ShaleXIONG: I assume `pure` and`return` are the same concept, the latter name is the function name in the ocaml tutorial ? (https://ocaml.org/docs/monads)

gen/cat2config/translation.ml
fsestini: In this instance, the difference is just more conciseness.

gen/common/util.ml
fsestini: They are expected to be implemented in the same exact way, yes. With the difference that `let*` is intended to be used to implement monadic binding:
```
let* x = foo in
bar x
```
vs
```
foo >>= fun x -> bar x
```
Does that answer your question?

ShaleXIONG: Ahh I see. Thanks !

gen/common/util.mli
fsestini: Good point, will change. `pure` is often how `return` is called for "applicative functors" and you can often find it as an [alias of `return`](https://ocaml.org/p/containers/latest/doc/containers/CCOption/index.html#val-pure). But I agree with you here.

gen/cat2config/mcat2config.ml
```diff
@@ -0,0 +1,246 @@
+[@@@warning "-40-42"]
+
+(* Create a set of relaxations for diy using a cat file *)
+
+module StringMap = Map.Make (String)
```
ShaleXIONG: There is a StringMap in `lib/stringmap.ml` although it is exactly as what you did with both one line.

gen/cat2config/mcat2config.ml
```diff
@@ -0,0 +1,246 @@
...
+
+  (* Parse a cat model by file path, recursively following included iles. *)
+  let rec find_parse_deep (file_path : string) : AST.ins list =
+    let _, (_, _, ast) = Parser.find_parse file_path in
+    let includes = get_includes ast in
+    let included_asts = List.concat_map find_parse_deep includes in
+    included_asts @ ast
+end
+
+module Log = struct
```
ShaleXIONG: This is the only place where Logger.ml is used/implemented?

fsestini: It is also passed as a functor argument to `Normalization.Make`:

```
gen/cat2config/mcat2config.ml
128:module Log = struct
145:module Nf = Normalization.Make (Ir) (Log)
```
It will soon also be used in `translation.ml`, as it's quite likely we'd want logging in there too.

gen/cat2config/mcat2config.ml
```diff
@@ -0,0 +1,246 @@
...
+                let expanded = Ir.expand_domain_range compressed in
+                (compressed, expanded, ast_expr))
+          in
+          (var, nfs))
+  in
+  if opts.conf then (
+    let open Format in
+    printf "-arch AArch64@.";
+    printf "-nprocs 2@.";
+    printf "-size 6@.";
```
ShaleXIONG: A discussion for the future. If we can parametrise some of those arguments for better usability. But I do not think it is anything urgent at all.

gen/cat2config/mcat2config.ml
fsestini: I believe the string map in `lib/stringmap.ml` just extends the standard `Map`.

gen/cat2config/mcat2config.ml
fsestini: Yes definitely, this is just for early experimentation. In fact I don't think it's necessary for `cat2config` to actually print out those parameters at all, since they can just be pre-pended with a script:
```
echo "-arch AArch64 etc..." > forbidden.conf
cat2config [opts] aarch64.cat >> forbidden.conf
```

gen/cat2config/mcat2config.ml
fsestini: Actually I'll just use the map from `lib/stringmap.ml`, thanks for reminding me. Otherwise I would need to place the `module StringMap = ...` declaration in some common location if I want multiple files to share the same map type.

fsestini:
Hi @relokin  @ShaleXIONG, I've marked this ready to review now. Please could you take a look at your earliest convenience.
Also let me know your thoughts on my comment from the PR description about the `logs` library.

Note that the commit history reflects my development workflow. I can squash if needed.

gen/cat2config/translation.ml
```diff
@@ -0,0 +1,334 @@
...
+type prim_rel = Ir.prim_rel
+type seq_item = Ir.seq_item
+
+let initial_effect : partial_effect =
+  { extr = Code.Irr; atom = None; explicit_mem = false }
+
+let initial_edge : partial_edge =
+  { tedges = None; ie = UnspecCom; sd = UnspecLoc }
+
+let apply_prim_set : partial_effect -> prim_set -> partial_effect option =
```
ShaleXIONG, line 117: There are several functions here that mapping cat syntax to internal data structures.
I was wondering do we have way to check we cover every possible syntax in cat-language.

fsestini: At present, cat2config most definitely does _not_ cover every possible syntactic construct or builtin symbol of cat.
Even in the long term however, I don't think exhaustive coverage of the whole cat language is a property that is in scope for this tool, or even achievable at all considering the expressive gap between cat and diy.

fsestini: Related to this, however, is the question "what parts of the cat language _are_ supported by the tool?"
Currently we only document this in the test suite, which gives a list of some of the cat relations from `aarch64.cat` that are supported.
We could improve discoverability of this aspect though, for sure. For example, we could have some `-show` option that prints all supported cat constructs, similarly to what `diy7` does with `-show edges` etc.

fsestini:
Update: squashed commits and replaced the libdir copy with a symlink to the model snapshot from `herd/libdir/aarch64`.

relokin:
I am looking at this right now and before merging we need to add copyright and licence headers to all source files.

gen/cat2config/astUtils.ml
```diff
@@ -0,0 +1,79 @@
+(****************************************************************************)
+(*                           the diy toolsuite                              *)
+(*                                                                          *)
+(* Jade Alglave, University College London, UK.                             *)
+(* Luc Maranget, INRIA Paris-Rocquencourt, France.                          *)
+(*                                                                          *)
+(* Copyright 2021-present Institut National de Recherche en Informatique et *)
+(* en Automatique and the authors. All rights reserved.                     *)
```
relokin:
```suggestion
(* Copyright 2025 Arm Limited and/or its affiliates <open-source-office@arm.com> *)
```

fsestini: I've wrapped the `<open-source-office@arm.com>` part in a following newline to keep it within the header's width. I can put everything in one line if needed, though.

fsestini:
Thanks for the review @relokin !

> I also like the new style in command line argument syntax but I also acknowledge that it deviates from other tools. One notable difference will be that most other tools will recognise `-variant` whereas cat2config will only accept `--variant`.

That is a very fair point. I will consider adding support, in a following PR, for both styles for established options like `-variant`, or at the very least print a helpful message like "did you mean --variant?" or similar.

relokin (commented):
Thanks for the changes Filippo! Please rebase on top of master and feel free to merge!

relokin (approved):

PR #1577, opened 2025-11-18
[all] Turn labels into symbols

artkhyzha:
This PR is a preparation for several strands of work, #1507 and #1518, which depend on treatment of labels as symbols rather than strings. With this PR, they would be treated more like virtual addresses, referred to as symbols like "x".

The use of labels in the syntax of litmus tests remains the same. No user-facing changes to the semantics of any feature of the tools are intended with this PR.

### DIFF
--- a/herd/AArch64Sem.ml
+++ b/herd/AArch64Sem.ml
@@ -3438,7 +3438,7 @@ Arguments:
 (*********************)
 
       let make_label_value proc lbl_str =
-        A.V.cstToV (Constant.Label (proc, lbl_str))
+        A.V.cstToV (Constant.mk_sym_virtual_label proc lbl_str)
 
       let read_loc_instr a ii =
         M.read_loc Port.No (mk_fetch Annot.N) a ii
@@ -3450,7 +3450,7 @@ Arguments:
       let v2tgt =
         let open Constant in
         function
-        | M.A.V.Val(Label (_, lbl)) -> Some (B.Lbl lbl)
+        | M.A.V.Val (Symbolic (Virtual {name=Symbol.Label (_, lbl); _})) -> Some (B.Lbl lbl)
         | M.A.V.Val (Concrete i) -> Some (B.Addr (M.A.V.Cst.Scalar.to_int i))
         | _ -> None
 
@@ -3628,7 +3628,7 @@ Arguments:
             >>= do_indirect_jump test [] i ii
 
         | I_ERET ->
-            let eret_to_addr v =
+           let eret_to_addr v =
               match v2tgt v with
               | Some tgt -> B.faultRetT tgt
               | _ ->
--- a/herd/mem.ml
+++ b/herd/mem.ml
@@ -259,8 +259,11 @@ module Make(C:Config) (S:Sem.Semantics) : S with module S = S	=
             try
               match A.V.as_virtual v with
               | Some s ->
-                 let sym = Constant.mk_sym_virtual s in
-                 A.Location_global (A.V.Val sym)::locs
+                  if A.V.is_instrloc v then
+                    locs
+                  else
+                    let sym = Constant.mk_sym_virtual s in
+                    A.Location_global (A.V.Val sym)::locs
               | None -> locs
             with V.Undetermined -> locs)
           init [] in
@@ -275,7 +278,8 @@ module Make(C:Config) (S:Sem.Semantics) : S with module S = S	=
            | A.Location_global _ -> true
            | A.Location_reg _ -> false)
           (S.observed_locations test)
-      and locs_init = get_all_locs_init test.Test_herd.init_state in
+      and locs_init = get_all_locs_init test.Test_herd.init_state
+      in
       let () =
         if dbg then begin
           let pp_locs locs =
@@ -308,7 +312,8 @@ module Make(C:Config) (S:Sem.Semantics) : S with module S = S	=
         A.LocSet.fold
           (fun loc env ->
             try
-              let v = A.look_address_in_state test.Test_herd.init_state loc in
+              let v = A.look_address_in_state test.Test_herd.init_state loc
+              in
               (loc,v)::env
             with A.LocUndetermined -> assert false)
           locs [] in
@@ -428,12 +433,12 @@ module Make(C:Config) (S:Sem.Semantics) : S with module S = S	=
                   match Label.norm lbls with
                   | None -> assert false (* as lbls is non-empty *)
                   | Some lbl ->
-                      let loc =
-                        A.Location_global
-                          (A.V.cstToV (Constant.Label (proc, lbl)))
+                      let symb = Constant.mk_sym_virtual_label proc lbl in
+                      let loc = A.Location_global (A.V.cstToV symb)
                       and v = A.V.instructionToV i in
                       A.state_add env loc v)
-                init_state lbls2i in
+                init_state lbls2i
+            in
             { test with init_state; } in
 
 (*****************************************************)
@@ -493,7 +498,7 @@ module Make(C:Config) (S:Sem.Semantics) : S with module S = S	=
 
         let addr2v proc s =
           try (* Look for label to overwritable instruction *)
-            V.Val (Constant.Label (proc,norm_lbl s))
+            V.Val (Constant.mk_sym_virtual_label proc (norm_lbl s))
           with
             e -> (* No, look for data location *)
               let v =  A.V.nameToV s in
@@ -1152,7 +1157,7 @@ let match_reg_events es =
             let ws =
               let v = S.E.global_loc_of er in
               match v with
-              | None|Some (V.Val (Constant.(Symbolic _|Label _))) -> ws
+              | None|Some (V.Val (Constant.Symbolic _)) -> ws
               | Some v -> NoSymbol v::ws in
             (* Add reading from nowhere for speculated reads *)
             let ws = if is_spec es er then NoWrite::ws else ws in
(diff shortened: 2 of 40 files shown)


artkhyzha:
In principle, the intent is to merge #1518 soon, but since there is a bit of work to do on it, and since this specific commit has been creating a massive amount of conflicts, and since it is useful for another PR, I quite like @murzinv's idea to merge it quicker. 

@murzinv , @relokin -- what do you think, what do we need to merge this PR? My thinking is, this PR would likely affect:
1) ifetch in herd7, but if the regressions pass, this seems okay
2) ifetch in litmus7 -- let's try and run tests for extra assurance
3) recent ASL-related changes (in case if conflicts were resolved not carefully enough)

herd/AArch64Sem.ml
```diff
@@ -1374,6 +1374,9 @@ module Make
             fun (r,_) -> M.unitT r
         else m
 
+      let set_elr_el1 v ii =
```
murzinv: This override `set_elr_el1` defined several lines above

herd/AArch64Sem.ml
```diff
@@ -3612,7 +3615,8 @@ Arguments:
             >>= do_indirect_jump test [] i ii
 
         | I_ERET ->
-            let eret_to_addr v =
+           let open Constant in
```
murzinv: Unnecessary change

herd/mem.ml
```diff
@@ -259,8 +259,11 @@ module Make(C:Config) (S:Sem.Semantics) : S with module S = S	=
             try
               match A.V.as_virtual v with
               | Some s ->
-                 let sym = Constant.mk_sym_virtual s in
-                 A.Location_global (A.V.Val sym)::locs
+                  if A.V.is_instrloc v then
+                    locs
+                  else
+                    let sym = Constant.mk_sym_virtual s in
+                      A.Location_global (A.V.Val sym)::locs
```
murzinv: Nitpick: indention

litmus/archExtra_litmus.ml
```diff
@@ -116,9 +116,9 @@ module Make(O:Config)(I:I) : S with module I = I
   let tr_global (c:ParsedConstant.v) =
     let open Constant in
     match c with
-    | Symbolic sym -> Global_litmus.tr_symbol sym
-    | Tag _|Concrete _|ConcreteVector _|ConcreteRecord _
-    | Label _|PteVal _|AddrReg _|Instruction _
+    | Symbolic sym when not (is_label c) -> Global_litmus.get_base_symbol sym
```
murzinv: Change from `tr_symbol` to `get_base_symbol` has side effect that later tolerate non-zero symbol offset... it might be ok, but I'd better check that it is intention

artkhyzha: Good point. I don't think there was an intent here; I will treat this as an accidental change and unroll it. (The "get_base_symbol" part.)

lib/constant.ml
```diff
@@ -24,6 +24,43 @@ open Printf
 (*vector metadata - prim size (arg1) and total array size (arg2) *)
 (*needed for is-non-mixed-symbol and global vectors *)
 
+module Symbol = struct
```
diaolo01, line 27: I wonder if it would be worth moving this into a dedicated Symbol module, similar to how we treat PAC in the symbolic_data record.

lib/constant.ml
```diff
@@ -215,8 +255,11 @@ let oa2symbol oa =
      end
 
 let virt_match_phy s1 s2 = match s1,s2 with
-| Virtual {name=s1; offset=i1;_},Physical (s2,i2) ->
-    Misc.string_eq s1 s2 && Misc.int_eq i1 i2
+(* | Virtual {name=s1; offset=i1;_},Physical (s2,i2) ->
```
diaolo01: Does this need to be kept?

lib/constant.ml
```diff
@@ -338,8 +373,9 @@ let collision s1 s2 = match s1,s2 with
         vs;
       Buffer.add_char b '}';
       Buffer.contents b
+  | Symbolic (Virtual {name=Symbol.Label(p,lbl);_}) -> pp_label p lbl
   | Symbolic sym -> pp_symbol sym
-  | Label (p, lbl) -> pp_label p lbl
+  (* | Label (p, lbl) -> pp_label p lbl *)
```
diaolo01: Probably this needs to be deleted?

lib/constant.ml
```diff
@@ -456,18 +498,32 @@ let mk_replicate sz v = ConcreteVector (Misc.replicate sz v)
 
 let is_symbol = function
   | Symbolic _ -> true
-  | Concrete _ | ConcreteVector _ | ConcreteRecord _ | Label _ | Tag _
+  | Concrete _ | ConcreteVector _ | ConcreteRecord _ | Tag _
   | PteVal _ | AddrReg _ | Instruction _ | Frozen _ ->
       false
 
+let is_data = function
```
diaolo01, line 502: What's the use of this?

lib/constant.ml
```diff
@@ -456,18 +498,32 @@ let mk_replicate sz v = ConcreteVector (Misc.replicate sz v)
...
+    false
+
 let is_label = function
-  | Label _ -> true
+  | Symbolic (Virtual ({name=n; _})) -> Symbol.is_label n
+  | Concrete _ | ConcreteVector _ | ConcreteRecord _ | Symbolic _ | Tag _
+  | PteVal _ | AddrReg _ | Instruction _ | Frozen _ ->
+      false
+
+let is_label_pa = function
```
diaolo01: What's the use of this?

lib/stateParser.mly
```diff
@@ -21,9 +21,6 @@ open LocationsItem
 open MiscParser
 open ConstrGen
 
-let mk_sym_tag s t =
```
diaolo01, line 24: What's the reasoning behind storing this in constant.ml and mk_lab here?

lib/stateParser.mly
```diff
@@ -39,29 +36,7 @@ let mk_sym_tagloc s o =
 
 let mk_sym_tagloc_zero s = do_mk_sym_tagloc s 0
 
-let mk_sym_morello p s t =
```
diaolo01, line 42: Same question as for `mk_sym_tag`.

diaolo01 (commented):
Hello @artkhyzha,
I had a look and the PR makes sense to me. I mostly focused on `constant.ml` and the parser changes, which seemed to be most significant.

murzinv (commented):
> **UPD.** It occurred to me that the impact on `litmus7` could be tested too -- and there are regressions that show up when tests from the catalogue are run:
> 
> ```
> File "catalogue/aarch64-ifetch/tests/DIC0-IDC0/MP.RF+cachesync+switch-ctrl.litmus" litmus cannot handle symbol '1:Lend'
> File "catalogue/aarch64-ifetch/tests/DIC0-IDC0/MP.RF+cachesync+switch-ctrlisb.litmus" litmus cannot handle symbol '1:Lend'
> File "catalogue/aarch64-ifetch/tests/DIC0-IDC0/MP.RF+dc-dsb-ic-dsb+addr.litmus" litmus cannot handle symbol '1:Lend'
> File "catalogue/aarch64-ifetch/tests/DIC0-IDC0/MP.RF+dc-dsb-ic-dsb+ctrlisb.br.litmus" litmus cannot handle symbol '1:Lend'
> File "catalogue/aarch64-ifetch/tests/DIC0-IDC0/miniJit01.litmus" litmus cannot handle symbol '1:L0'
> File "catalogue/aarch64-ifetch/tests/DIC0-IDC0/miniJit02.litmus" litmus cannot handle symbol '1:L0'
> ```
> 
> Something to look into.

Here is (slightly tested) diff which makes problem go away
```
iff --git a/litmus/compile.ml b/litmus/compile.ml
index 60c986d9d..60d227cd3 100644
--- a/litmus/compile.ml
+++ b/litmus/compile.ml
@@ -210,7 +210,7 @@ module Generic
         List.fold_left
           (fun env (loc,(t,v)) -> match loc,v with
           | _,Constant.Concrete _ -> env
-          | A.Location_global _,Symbolic s ->
+          | A.Location_global _,Symbolic s when not (Constant.is_label v) ->
               let a = A.Location_global (G.tr_symbol s) in
               begin try
                 ignore (A.LocMap.find a env) ;
diff --git a/litmus/skel.ml b/litmus/skel.ml
index 91d83289e..088c30055 100644
--- a/litmus/skel.ml
+++ b/litmus/skel.ml
@@ -368,11 +368,8 @@ module Make
 (* Right value, all cases *)
       let rec dump_a_v = function
         | Concrete i ->  A.V.Scalar.pp  Cfg.hexa i
-        | Symbolic _ as v when is_label v ->
-           Warn.user_error
-             "Labels cannot be used as initial values of memory locations"
-        | Symbolic (Virtual {name=s;tag=None; offset=0;_}) ->
-            dump_a_addr (Symbol.pp s)
+        | Symbolic (Virtual {name=Symbol.Data s;tag=None; offset=0;_}) ->
+            dump_a_addr s
         | ConcreteVector vs ->
            let pps =
              List.map

```

lib/constant.ml
artkhyzha: Fair! Functions `is_data` and `is_label_pa` are useful, but their uses do not exist in this PR. I see they are not useful for the GCS PR. With that in mind, they could be dropped from here and introduced later when they are used. (With the benefit that this will let the author choose better names. :-) )

lib/constant.ml
artkhyzha: Ditto.

lib/stateParser.mly
artkhyzha: Actually, having `mk_lab` here was an oversight (as a similar function got added into `constant.ml`). However, the tiny problem I have with removing it is that I'd need to "uncurry" the function in `constant.ml`: it does not straightforwardly apply to tuples. I haven't thought of a shorter solution to this.

lib/constant.ml
artkhyzha: I ended up finding places to use `is_data` in (including a small edit to @murzinv's patch).

lib/stateParser.mly
artkhyzha: Same answer! Strictly speaking, this patch does not need this change, but when I picked up, it already moved some of these `mk_*` functions. This seems to be in line with `constant.mli` interface, so I am keeping the change as is unless there are strong opinions.

lib/constant.ml
artkhyzha: I do not have a strong opinion. Right now the code uses `Constant.Symbol` in various places. I don't think it's fundamental for it to be exposed through the interface of constants though. It could be hidden in side `Constant`, with wrapper functions made available.

Presently I do not clearly see what would be an improvement, so I'd opt for no action.

artkhyzha:
This PR currently passes `herd7` regressions and appears to have the same behaviour in `litmus7`. I suggest considering this ready for merging.

@maranget, would you have time to have a quick look at this PR? Essentially, the last concern I have is that this PR had conflicts with the changes that have been happening due to ASL+VMSA (I think). It would be helpful to know if any special care was needed in resolving the conflicts.

artkhyzha:
Merged after a bit of grace period.

Thank you, @diaolo01 and @murzinv, for the reviews, and for preparing these changes, @relokin and @murzinv.

PR #1569, opened 2025-11-14
[gen] Update the baseline forbidden configuration file on `Pos*W`.

ShaleXIONG:
We remove relaxation that is stronger than `Pos*W` in the baseline configuration file, which generates smaller test suites.
This removed edges includes:
- Fence on same location and end with write, e.g. `DMB.SY*W`.
- `Data`, `Addr`, and `Ctrl` dependency, with or without `Csel`, on the same location and end with write, e.g. `DpDatasW`.
- We remove some relaxations that `diy` can automatically generate.

PR #1578, opened 2025-11-18
[WIP] [gen] Update baseline configuration for `rmw`

ShaleXIONG:
Add several missing `rmw` relaxations since such relaxation can match memory access such as `M`, `R`, and `W`.

this pull requisition contains #1569 too.


gen/libdir/forbidden.conf
```diff
@@ -71,80 +71,157 @@
...
 # let IFB-ob = [Exp & R]; pick-ctrl-dep; [IFB]; po
 # let IFB = ISB | EXC-ENTRY-IFB | EXC-RET-IFB
--safe [DpCtrlCseldR ISB] [DpCtrlCseldW ISB] [DpCtrlCselsR ISB] [DpCtrlCselsW ISB]
+-safe [DpCtrlCseldR ISB] [DpCtrlCseldW ISB] [DpCtrlCselsR ISB]
+# [DpCtrlCselsW ISB] is stronger than `Pos*W`
+#-safe [DpCtrlCselsW ISB]
 
 # let IFB-ob = [Exp & R]; addr; [Exp & M]; po; [IFB]; po
--safe [DpAddrdR ISBd**] [DpAddrdR ISBs**] [DpAddrdW ISBd**] [DpAddrdW ISBs**] [DpAddrsR ISBd**] [DpAddrsR ISBs**] [DpAddrsW ISBd**] [DpAddrsW ISBs**]
+-safe [DpAddrdR ISBdRR] [DpAddrdR ISBsRR] [DpAddrdW ISBdWR] [DpAddrsR ISBdRR] [DpAddrsR ISBsRR] [DpAddrsW ISBsWR]
```
chacon01: I think you can remove `[DpAddrsW ISBsWR]` because it is stronger than `[DpAddrsW PosWR]` which is included in line 113.

ShaleXIONG: Fixed. Thanks for reviewing again. The file becomes difficult to editing and track now. I will try to speed up another change on introducing more wildcard `*` syntax , #1516 , so hope to simplify the auditing in the future.

gen/libdir/forbidden.conf
```diff
@@ -71,80 +71,157 @@
...
+-safe [DpAddrdR ISBdRR] [DpAddrdR ISBsRR] [DpAddrdW ISBdWR] [DpAddrsR ISBdRR] [DpAddrsR ISBsRR] [DpAddrsW ISBdWR]
+# `[DpAddrsW ISBsWW] [DpAddrsR ISBsRW]` are stronger than `Pos*W` as the former are a read followed by a write to the same location.
+#-safe [DpAddrsW ISBsWW] [DpAddrsR ISBsRW]
+# `[DpAddr** ISB**W]` is stronger than `[DpAddr** Po**W]`; the latter can be combined by `diy`.
+#-safe [DpAddrdR ISBdRW] [DpAddrdR ISBsRW] [DpAddrdW ISBdWW] [DpAddrdW ISBsWW] [DpAddrsR ISBdRW] [DpAddrsW ISBdWW]
+# `[DpAddrdW ISBsWR]` is stronger than `[DpAddrdW PosWR]`.
+#-safe [DpAddrdW ISBsWR]
+-safe [DpAddrdR Amo.Swp ISBdWR] [DpAddrdR Amo.Swp ISBsWR] [DpAddrsR Amo.Swp ISBdWR] [DpAddrsR Amo.Swp ISBsWR]
+-safe [DpAddrdR Amo.Cas ISBdWR] [DpAddrdR Amo.Cas ISBsWR] [DpAddrsR Amo.Cas ISBdWR] [DpAddrsR Amo.Cas ISBsWR]
+-safe [DpAddrdR LxSx ISBdWR] [DpAddrdR LxSx ISBsWR] [DpAddrsR LxSx ISBdWR] [DpAddrsR LxSx ISBsWR]
```
chacon01: The relaxations here with `ISBsWR` are stronger than the `[DpAddr*R rmw PosWR]` relaxations from lines 114-116.

gen/libdir/forbidden.conf
```diff
@@ -71,80 +71,157 @@
...
+# `[DpAddrCselsW PosWW] [DpAddrCselsR PosRW]` are stronger than `Pos*W` as the former are a read followed by a write to the same location.
+#`[DpAddrCseldW PosWW]` can be combined by `diy`.
+#-safe [DpAddrCselsW PosWW] [DpAddrCselsR PosRW]
+-safe [DpAddrCseldR Amo.Swp PodWW] [DpAddrCselsR Amo.Swp PodWW]
+-safe [DpAddrCseldR Amo.Swp PodWW] [DpAddrCselsR Amo.Swp PodWW]
+-safe [DpAddrCseldR Amo.Swp PodWW] [DpAddrCselsR Amo.Swp PodWW]
+# `[DpAddrCseldR rmw PosWW]` can be combined by `diy` as `[DpAddrCseldR rmw]` and `PosWW`
+-safe [DpAddrCseldR PodRR Amo.Swp] [DpAddrCseldR PosRR Amo.Swp] [DpAddrCseldW PodWR Amo.Swp] [DpAddrCseldW PosWR Amo.Swp] [DpAddrCselsR PodRR Amo.Swp] [DpAddrCselsW PodWR Amo.Swp]
+-safe [DpAddrCseldR PodRR Amo.Cas] [DpAddrCseldR PosRR Amo.Cas] [DpAddrCseldW PodWR Amo.Cas] [DpAddrCseldW PosWR Amo.Cas] [DpAddrCselsR PodRR Amo.Cas] [DpAddrCselsW PodWR Amo.Cas]
+-safe [DpAddrCseldR PodRR LxSx] [DpAddrCseldR PosRR LxSx] [DpAddrCseldW PodWR LxSx] [DpAddrCseldW PosWR LxSx] [DpAddrCselsR PodRR LxSx] [DpAddrCselsW PodWR LxSx]
```
chacon01: This is missing the cases where the `Amo`/`LxSx` is both the consumer of the pick-address-dependency and the subsequent write in program order. For example `[DpAddrCseldR Amo.Swp PodWR Amo.Swp]`. Unfortunately, this ought to cross on the `Amo`/`LxSx` types, for example `[DpAddrCseldR Amo.Swp PodWR LxSx]`

gen/libdir/forbidden.conf
```diff
@@ -160,36 +237,50 @@
...
--safe [A Amo.Swp PodW*] [A Amo.Swp PosW*]
--safe [A Amo.Cas PodW*] [A Amo.Cas PosW*]
--safe [Q Amo.Swp PodW*] [Q Amo.Swp PosW*]
--safe [Q Amo.Cas PodW*] [Q Amo.Cas PosW*]
+-safe [A Amo.Swp PodW*] [A Amo.Swp PosWR]
+-safe [A Amo.Cas PodW*] [A Amo.Cas PosWR]
+-safe [A LxSx PodW*] [A LxSx PosWR]
+-safe [Q Amo.Swp PodW*] [Q Amo.Swp PosWR]
+-safe [Q Amo.Cas PodW*] [Q Amo.Cas PosWR]
+-safe [Q LxSx PodW*] [Q LxSx PosWR]
```
chacon01: The `Q` annotation cannot apply to `Amo` or `LxSx`

chacon01 (changes requested):
Thanks @ShaleXIONG! A couple of comments.

gen/libdir/forbidden.conf
ShaleXIONG: I will change it.

gen/libdir/forbidden.conf
ShaleXIONG: I will manually add all the components.

gen/libdir/forbidden.conf
ShaleXIONG: Removed those clauses

gen/libdir/forbidden.conf
```diff
@@ -205,72 +272,77 @@
 # For this reason, these relaxations are not included.
 
 # let bob = [A | Q]; po; [Exp & M | Imp & Tag & R | MMU & FAULT]
--safe [A PodR*] [A PosRR] [Q PodR*] [Q PosRR]
+-safe [A PodR* ExpObs] [A PosRR ExpObs] [Q PodR* ExpObs] [Q PosRR ExpObs]
```
chacon01: This prevents over-ordering, which is good, but it also means that the generator won't be able to compose this with `Amo` or `LxSx`. Consequently, in addition to these, we now need
* `[Q PodRR Amo|LxSx ExpObs]`
* `[A PodRR Amo|LxSx ExpObs]`
We don't need to consider `PosRR` here because the `Amo` or `LxSx` contains a write and would result in something stronger than `PosRW`.

gen/libdir/forbidden.conf
```diff
@@ -205,72 +272,77 @@
...
+-safe [A Amo.LdAdd PodW* ExpObs] [A Amo.LdAdd PosWR ExpObs]
+-safe [A LxSx PodW* ExpObs] [A LxSx PosWR ExpObs]
+-safe [Q LxSx PodW* ExpObs] [Q LxSx PosWR ExpObs]
 #[A|Q rmw PosWW] is stronger than `Pos*W`
-#-safe [A Amo.Swp PosWW] [A Amo.Cas PosWW] [Q Amo.Swp PosWW] [Q Amo.Cas PosWW] [A LxSx PosWW] [Q LxSx PosWW]
+#-safe [A Amo.Swp PosWW] [A Amo.Cas PosWW] [A Amo.StAdd PosWW] [A Amo.LdAdd PosWW] [A LxSx PosWW] [Q LxSx PosWW]
 
 # let bob = [Exp & M | Imp & Tag & R]; po; [L]
--safe [Pod*W L]
+-safe [ExpObs Pod*W L]
```
chacon01, line 293: Similar to the acquire case above, this won't compose with `Amo` or `LxSx`. In addition to these, we now need
* `[ExpObs Amo|LxSx PodWW L]`

gen/libdir/forbidden.conf
```diff
@@ -81,64 +82,97 @@
...
 # let dob = addr
--safe DpAddrdR DpAddrdW DpAddrsR
+-safe DpAddrd* DpAddrsR
 # `DpAddrsW` is stronger than `Pos*W`.
 #-safe DpAddrsW
 
 # let ctrl = [Exp & R]; basic-dep; [BCC]; po
 # let dob = ctrl; [Exp & W | HU | TLBI | DC.CVAU | IC]
--safe DpCtrldW
+-safe [DpCtrldW ExpObs]
```
chacon01: Be aware that tying to `ExpObs` here will prevent it from combining with other relaxations that begin with `ExpObs` or `Hat`. We will need some way to allow `ExpObs` to combine with itself and with `Hat` e.g. `[DpCtrldW ExpObs]` + `[ExpObs PodRW L]` -> `DpCtrldW Fre PodRWPL` or `[DpCtrlIsbdR ExpObs]` + `[Hat LxSx]` -> `DpCtrlIsbdR Hat LxSx`. I think expressing the possible combinations in the configuration file would be tedious and error-prone, so it'll likely need to be some option for the generator.

gen/libdir/forbidden.conf
```diff
@@ -81,64 +82,97 @@
...
+-safe [DpAddr*R Amo.Cas PodWW]
+-safe [DpAddr*R Amo.StAdd PodWW]
+-safe [DpAddr*R Amo.LdAdd PodWW]
+-safe [DpAddr*R LxSx PodWW]
+# `[DpAddrdR rmw PosWW]` can be combined by `diy` as `[DpAddrdR rmw]` and `PosWW`
+-safe [DpAddrd* Po**R Amo.Swp] [DpAddrs* Pod*R Amo.Swp]
+-safe [DpAddrd* Po**R Amo.Cas] [DpAddrs* Pod*R Amo.Cas]
+-safe [DpAddrd* Po**R Amo.StAdd] [DpAddrs* Pod*R Amo.StAdd]
+-safe [DpAddrd* Po**R Amo.LdAdd] [DpAddrs* Pod*R Amo.LdAdd]
+-safe [DpAddrd* Po**R LxSx] [DpAddrs* Pod*R LxSx]
```
chacon01: For `addr; [Exp & M]; po; [Exp & W | HU]`, the above covers the case where the `[Exp & M]` is from an `Amo` or `LxSx` and it covers the case where the `[Exp & W | HU]` from an `Amo` or `LxSx`, but there isn't anything that covers both of them being `Amo` or `LxSx`. We also need
* `[DpAddrdR Amo|LxSx PodWR Amo|LxSx]` (note the `PosWR` version of this could be achieved from the composition of `[DpAddrdR Amo|LxSx PosWR]` above and `LxSx|Amo` below.

gen/libdir/forbidden.conf
```diff
@@ -205,72 +272,77 @@
...
+-safe [ExpObs Pod*R Amo.Cas L]
+-safe [ExpObs Pod*R Amo.StAdd L]
+-safe [ExpObs Pod*R Amo.LdAdd L]
+-safe [ExpObs Pod*R LxSx L]
 # `[Pos*R rmw L]` is stronger than `Pos*W`
-#-safe  [Pos*R Amo.Swp L] [Pos*R Amo.Cas L] [Pos*R LxSx L]
+#-safe [Pos*R Amo.Swp L] [Pos*R Amo.Cas L] [Pos*R LxSx L]
 
 # let aob = [Exp & M]; rmw; [Exp & M]
 -safe LxSx Amo.Swp Amo.Cas
```
chacon01: Missing `Amo.LdAdd` and `Amo.StAdd`

gen/libdir/forbidden.conf
```diff
@@ -149,51 +183,84 @@
...
 # let bob = [Exp & W]; po; [dmb.st]; po; [Exp & W | MMU & FAULT]
 -safe DMB.STdWW
 # `DMB.STsWW` is stronger than `Pos*W`.
 #-safe DMB.STsWW
+-safe [DMB.STdWR Amo.Swp] [DMB.STdWR Amo.Cas] [DMB.STdWR Amo.StAdd] [DMB.STdWR Amo.LdAdd] [DMB.STdWR LxSx]
 
 # let DSB-ob = [M | DC.CVAU | IC]; po; [dsb.full]; po; [~(Imp & TTD & M | Imp & Instr & R)]
 # DSB.ISH*** DSB.OSH*** in `-moreedges` in `diy7`; these two edges also satisfy `dsb.full`
--safe DSB.SYd** DSB.SYs*R
+-safe [ExpObs DSB.SYd** ExpObs] [ExpObs DSB.SYs*R ExpObs]
```
chacon01: Because the `ExpObs` prevents composition with `Amo` or `LxSx`, we now additionally need
* `[ExpObs Amo|LxSx DMB.SYdW* ExpObs]`
* `[ExpObs DMB.SYd*R Amo|LxSx ExpObs]`
* `[ExpObs Amo|LxSx DMB.SYdWR Amo|LxSx ExpObs]`

gen/libdir/forbidden.conf
```diff
@@ -149,51 +183,84 @@
...
--safe DSB.LDdR* DSB.LDsRR
+-safe [DSB.LDdR* ExpObs] [DSB.LDsRR ExpObs]
 # `DSB.LDsRW` is stronger than `Pos*W`.
 #-safe DSB.LDsRW
+-safe [Amo.Swp DSB.LDdW*] [Amo.Cas DSB.LDdW*] [Amo.LdAdd DSB.LDdW*] [LxSx DSB.LDdW*]
+-safe [Amo.Swp DSB.LDsWR] [Amo.Cas DSB.LDsWR] [Amo.LdAdd DSB.LDsWR] [LxSx DSB.LDsWR]
 
 # let DSB-ob = [Exp & W]; po; [dsb.st]; po; [~(Imp & TTD & M | Imp & Instr & R)]
--safe DSB.STdW* DSB.STsWR
+-safe [DSB.STdW* ExpObs] [DSB.STsWR ExpObs]
```
chacon01: Because the ExpObs prevents composition with `Amo` or `LxSx`, we now additionally need
* `[DSB.STdWR Amo|LxSx ExpObs]`

gen/libdir/forbidden.conf
```diff
@@ -149,51 +183,84 @@
...
 # `DMB.SYs*W` is stronger than `Pos*W`.
 #-safe DMB.SYs*W
 
 # let bob = [Exp & (R \ NoRet) | Imp & Tag & R]; po; [dmb.ld]; po; [Exp & M | Imp & Tag & R | MMU & FAULT]
--safe DMB.LDdR* DMB.LDsRR
+-safe [DMB.LDdR* ExpObs] [DMB.LDsRR ExpObs]
 # `DMB.LDsRW` is stronger than `Pos*W`.
 #-safe DMB.LDsRW
+-safe [Amo.Swp DMB.LDdW*] [Amo.Cas DMB.LDdW*] [Amo.LdAdd DMB.LDdW*] [LxSx DMB.LDdW*]
+-safe [Amo.Swp DMB.LDsWR] [Amo.Cas DMB.LDsWR] [Amo.LdAdd DMB.LDsWR] [LxSx DMB.LDsWR]
```
chacon01: These can be strengthened by adding `ExpObs` to the end (like was done on line 229)

gen/libdir/forbidden.conf
```diff
@@ -205,72 +272,77 @@
...
--safe [A LxSx PodW*] [A LxSx PosWR]
--safe [Q Amo.Swp PodW*] [Q Amo.Swp PosWR]
--safe [Q Amo.Cas PodW*] [Q Amo.Cas PosWR]
--safe [Q LxSx PodW*] [Q LxSx PosWR]
+-safe [A Amo.Swp PodW* ExpObs] [A Amo.Swp PosWR ExpObs]
+-safe [A Amo.Cas PodW* ExpObs] [A Amo.Cas PosWR ExpObs]
+-safe [A Amo.StAdd PodW* ExpObs] [A Amo.StAdd PosWR ExpObs]
+-safe [A Amo.LdAdd PodW* ExpObs] [A Amo.LdAdd PosWR ExpObs]
+-safe [A LxSx PodW* ExpObs] [A LxSx PosWR ExpObs]
+-safe [Q LxSx PodW* ExpObs] [Q LxSx PosWR ExpObs]
```
chacon01: `Q` can't be used with `LxSx`

gen/libdir/forbidden.conf
```diff
@@ -205,72 +272,77 @@
...
--safe [A Amo.Cas PodW*] [A Amo.Cas PosWR]
--safe [A LxSx PodW*] [A LxSx PosWR]
--safe [Q Amo.Swp PodW*] [Q Amo.Swp PosWR]
--safe [Q Amo.Cas PodW*] [Q Amo.Cas PosWR]
--safe [Q LxSx PodW*] [Q LxSx PosWR]
+-safe [A Amo.Swp PodW* ExpObs] [A Amo.Swp PosWR ExpObs]
+-safe [A Amo.Cas PodW* ExpObs] [A Amo.Cas PosWR ExpObs]
+-safe [A Amo.StAdd PodW* ExpObs] [A Amo.StAdd PosWR ExpObs]
+-safe [A Amo.LdAdd PodW* ExpObs] [A Amo.LdAdd PosWR ExpObs]
+-safe [A LxSx PodW* ExpObs] [A LxSx PosWR ExpObs]
```
chacon01: This will no longer compose with `Amo` or `LxSx`. In addition to these, we now need
* `[A Amo|LxSx PodWR Amo|LxSx ExpObs]`

gen/libdir/forbidden.conf
```diff
@@ -205,72 +272,77 @@
...
 # -safe [Pos*W L]
 # A memory access followed by a release atomic operation also matches above.
--safe [Pod*R Amo.Swp L]
--safe [Pod*R Amo.Cas L]
--safe [Pod*R LxSx L]
+-safe [ExpObs Pod*R Amo.Swp L]
+-safe [ExpObs Pod*R Amo.Cas L]
+-safe [ExpObs Pod*R Amo.StAdd L]
+-safe [ExpObs Pod*R Amo.LdAdd L]
+-safe [ExpObs Pod*R LxSx L]
```
chacon01: Similarly, we now need
* `[ExpObs Amo|LxSx PodWR Amo|LxSx L]`

gen/libdir/forbidden.conf
```diff
@@ -149,51 +183,84 @@
...
 #-safe DMB.LDsRW
+-safe [Amo.Swp DMB.LDdW*] [Amo.Cas DMB.LDdW*] [Amo.LdAdd DMB.LDdW*] [LxSx DMB.LDdW*]
+-safe [Amo.Swp DMB.LDsWR] [Amo.Cas DMB.LDsWR] [Amo.LdAdd DMB.LDsWR] [LxSx DMB.LDsWR]
+# `[DMB.LD*R* rmw]` can be combined by `diy`
 
 # let bob = [Exp & W]; po; [dmb.st]; po; [Exp & W | MMU & FAULT]
 -safe DMB.STdWW
 # `DMB.STsWW` is stronger than `Pos*W`.
 #-safe DMB.STsWW
+-safe [DMB.STdWR Amo.Swp] [DMB.STdWR Amo.Cas] [DMB.STdWR Amo.StAdd] [DMB.STdWR Amo.LdAdd] [DMB.STdWR LxSx]
```
chacon01: We additionally need
* `[DMB.STdWR A Amo|LxSx PodW* ExpObs]`
* `[DMB.STdWR A Amo|LxSx PosWR ExpObs]`

which cannot be composed from other relaxations.

gen/libdir/forbidden.conf
```diff
@@ -81,64 +82,97 @@
...
+
+# let dob = data; [Exp & M]; lmrs; [Exp & R | Imp & Tag & R]
+# let lmrs = [W]; ((po & same-loc) & ~(intervening(W,(po & same-loc)))); [R]
+-safe [DpData*W PosWR]
 
 # let ctrl = [Exp & R]; basic-dep; [BCC]; po
 # let IFB-ob = [Exp & R]; ctrl; [IFB]; po
 # let IFB = ISB | EXC-ENTRY-IFB | EXC-RET-IFB
--safe [DpCtrldR ISB] [DpCtrldW ISB] [DpCtrlsR ISB]
+-safe [DpCtrld* ISB ExpObs] [DpCtrlsR ISB ExpObs]
```
chacon01: Can replace `[DpCtrld* ISB ExpObs]` with `[DpCtrldR ISB ExpObs]` because `[DpCtrldW ISB ExpObs]` is stronger than `[DpCtrldW ExpObs]` which is already present.

gen/libdir/forbidden.conf
```diff
@@ -81,64 +82,97 @@
...
 # [DpCtrlsW ISB] is stronger than `Pos*W`
 #-safe [DpCtrlsW ISB]
 
 # let IFB-ob = [Exp & R]; pick-ctrl-dep; [IFB]; po
 # let IFB = ISB | EXC-ENTRY-IFB | EXC-RET-IFB
--safe [DpCtrlCseldR ISB] [DpCtrlCseldW ISB] [DpCtrlCselsR ISB]
+-safe [DpCtrlCseld* ISB ExpObs] [DpCtrlCselsR ISB ExpObs]
```
chacon01: Can replace `[DpCtrlCseld* ISB ExpObs]` with `[DpCtrlCseldR ISB ExpObs]` because `[DpCtrlCseldW ISB ExpObs]` is stronger than `DpCtrlCseldW` which is already present in `pob`

gen/libdir/forbidden.conf
```diff
@@ -205,72 +272,77 @@
...
 # For this reason, here, we map
 # `[A]; amo; [range([A];amo;[L])]; po; [Exp & M | Imp & Tag & R | MMU & FAULT]`
--safe [A Amo.Swp L PodW*] [A Amo.Swp L PosWR]
--safe [A Amo.Cas L PodW*] [A Amo.Cas L PosWR]
--safe [A LxSx L PodW*]    [A LxSx L PosWR]
+-safe [A Amo.Swp L PodW* ExpObs] [A Amo.Swp L PosWR ExpObs]
+-safe [A Amo.Cas L PodW* ExpObs] [A Amo.Cas L PosWR ExpObs]
+-safe [A Amo.StAdd L PodW* ExpObs] [A Amo.StAdd L PosWR ExpObs]
+-safe [A Amo.LdAdd L PodW* ExpObs] [A Amo.LdAdd L PosWR ExpObs]
+-safe [A LxSx L PodW* ExpObs]    [A LxSx L PosWR ExpObs]
```
chacon01: `[range([A];amo;[L])]; po; [Exp & M | Imp & Tag & R | MMU & FAULT]` does not apply for `LxSx`, so this relaxation can be removed. While this will not produce allowed tests because we would need it to compose with a free `Pod*R` (that doesn't exist), it is inconsistent with the justifying comment.

For the `Amo` variants (and `LxSx` too), the generator will only compose these with the other relaxations which are properly ordered into the acquire read effect, so all the forbidden combinations will also be forbidden without the `L`.

gen/libdir/forbidden.conf
```diff
@@ -205,72 +272,77 @@
...
 
 # let bob = [Exp & M | Imp & Tag & R]; po; [L] for `Pos*** A Amo.* L` Note that `A` is irrelevant
 # then let [range([A];amo;[L])]; po; [Exp & M | Imp & Tag & R | MMU & FAULT] for `A Amo.* L Po***`
 # This particular edges cannot be auto generated from other safe edges.
--safe [Pod*R A Amo.Swp L PodW*] [Pod*R A Amo.Swp L PosW*] [Pos*R A Amo.Swp L PodW*] [Pos*R A Amo.Swp L PosW*]
--safe [Pod*R A Amo.Cas L PodW*] [Pod*R A Amo.Cas L PosW*] [Pos*R A Amo.Cas L PodW*] [Pos*R A Amo.Cas L PosW*]
+-safe [ExpObs Po**R A Amo.Swp L Po*W* ExpObs]
+-safe [ExpObs Po**R A Amo.Cas L Po*W* ExpObs]
+-safe [ExpObs Po**R A Amo.StAdd L Po*W* ExpObs]
+-safe [ExpObs Po**R A Amo.LdAdd L Po*W* ExpObs]
```
chacon01: The `ExpObs` here prevents these from composing with `Amo` or `LxSx` so they need to be specified individually.
* `[ExpObs Amo|LxSx Po*WR Amo L Po*W* ExpObs]`
* `[ExpObs Po**R Amo L Po*WR Amo|LxSx ExpObs]`
* `[ExpObs Amo|LxSx Po*WR Amo L Po*WR Amo|LxSx ExpObs]`

gen/libdir/forbidden.conf
```diff
@@ -205,72 +272,77 @@
...
--safe [Hat DMB.LDdR*]
--safe [Hat DSB.LDdR*]
--safe [Hat A PodR*] [Hat Q PodR*]
--safe [Hat Amo.Swp] [Hat Amo.Cas] [Hat LxSx]
+-safe [Hat DpCtrldW ExpObs]
+-safe [Hat DMB.SYdR* ExpObs]
+-safe [Hat DSB.SYdR* ExpObs]
+-safe [Hat DMB.LDdR* ExpObs]
+-safe [Hat DSB.LDdR* ExpObs]
+-safe [Hat A PodR* ExpObs] [Hat Q PodR* ExpObs]
```
chacon01: While these produce forbidden tests, they don't produce useful ones. The extra instructions produced by these don't contribute to the cycle in ordered-before. All uses of `Hat` that are not followed by `LxSx` or `Amo` can be removed.

chacon01 (changes requested):
`Amo` and `LxSx` become tedious to enumerate the more the configuration file is restricted to prevent additional ordering. I see two issues:
* Several cases where an existing effect could be an `Amo` or `LxSx` without changing the meaning
* We can't create a general `Rmw` or `Amo` relaxation because it will include e.g. `Amo.LdClr` which has the value collision problem.

Perhaps the generator could support "user-defined" relaxations to help here e.g.
```
-define LdRmw LxSx|Amo.Swp|Amo.Cas|Amo.LdAdd
-define Rmw LdRmw|Amo.StAdd
...
-safe [A LdRmw PodW* ExpObs]
-safe [ExpObs PodWR Rmw L]
```
which would reduce the duplication everywhere else.

Additionally, perhaps an operation to make a relaxation "optional" e.g.
```
-safe [A Rmw? Pod** Rmw? ExpObs]
```
which would be equivalent to
```
-safe [A Pod** ExpObs]
-safe [A Rmw Pod** ExpObs]
-safe [A Pod** Rmw ExpObs]
-safe [A Rmw Pod** Rmw ExpObs]
```

gen/libdir/forbidden.conf
ShaleXIONG: Done.

gen/libdir/forbidden.conf
ShaleXIONG: Yes, I will update and check for other similar situation.

gen/libdir/forbidden.conf
ShaleXIONG: Done.

gen/libdir/forbidden.conf
ShaleXIONG: Done.

gen/libdir/forbidden.conf
ShaleXIONG: Done.

gen/libdir/forbidden.conf
ShaleXIONG: I will add `[DpAddr*R Amo.Safe|LxSx Po*WR Amo.Safe|LxSx]` note that `[DpAddrdR Amo|LxSx PosWR]` is NOT a safe relaxation as the `cat` file say it must ends with a write event.

chacon01: `[DpAddrdR Amo|LxSx PosWR]` satisifes `dob` as `addr; [Exp & M]; lmrs; [Exp & R | Imp & Tag & R]` so it ought to be a safe relaxation.

ShaleXIONG: Ahh I see. Yes I have that in the next line.

gen/libdir/forbidden.conf
ShaleXIONG: Done.

gen/libdir/forbidden.conf
ShaleXIONG: Done.

gen/libdir/forbidden.conf
ShaleXIONG: Done.

gen/libdir/forbidden.conf
ShaleXIONG: Will do.

gen/libdir/forbidden.conf
ShaleXIONG: Will update to `[ExpObs [Amo.Safe|LxSx]? Po**R A Amo.Safe L Po*W* [Amo.Safe|LxSx]? ExpObs]`

PR #1579, opened 2025-11-18
[ASL-891] Fix bugs in bitfield subtyping

hrutvik:
- one `assert false` should have just been a return of `false`
- a subtype-satisfies check compared annotated types against unannotated types, which would therefore likely fail even when it shouldn't have

### DIFF
--- a/asllib/Typing.ml
+++ b/asllib/Typing.ml
@@ -3903,7 +3903,7 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
     let here x = add_pos_from ~loc:ty x in
     let+ () = check_var_not_in_genv ~loc genv name in
     let env = with_empty_local genv in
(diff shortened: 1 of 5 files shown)


asllib/Typing.ml
```diff
@@ -3903,33 +3903,33 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
     let here x = add_pos_from ~loc:ty x in
     let+ () = check_var_not_in_genv ~loc genv name in
     let env = with_empty_local genv in
-    let env1, t1, s' =
+    let t1, ses_t = annotate_type ~decl:true ~loc env ty in
```
HadrienRenaud: Maybe `loc` could be just `t1`? Thinking about the error message for `TypingRule.AnnotateExtraFields.bad.asl`

PR #1580, opened 2025-11-18
[ASL-888] Fix a implementation/documentation mismatch for ATCs

hrutvik:
For the semantics of ATCs, if a type is not an integer/bits/tuple then the dynamic check should pass (as it has already been checked during type-checking). Also fix a typo in the reference, and add a test.

### DIFF
--- a/asllib/Interpreter.ml
+++ b/asllib/Interpreter.ml
@@ -843,7 +843,11 @@ module Make (B : Backend.S) (C : Config) = struct
             (i + 1, m)
           in
           List.fold_left fold (0, m_true) tys |> snd
-      | _ -> fatal_from loc env TypeInferenceNeeded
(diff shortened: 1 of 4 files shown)


asllib/Interpreter.ml
```diff
@@ -843,7 +843,7 @@ module Make (B : Backend.S) (C : Config) = struct
             (i + 1, m)
           in
           List.fold_left fold (0, m_true) tys |> snd
-      | _ -> fatal_from loc env TypeInferenceNeeded
+      | _ -> m_true
```
HadrienRenaud: Can we leave a comment justifying this? This is quite counter-intuitive.

Roman-Manevich (approved):
Looks good to me!

PR #1581, opened 2025-11-18
[asl] Some minor lexer/parser tidying

hrutvik:
Partly from suggestions by @acjf3 and @ldesnogu

### DIFF
--- a/asllib/Parser.mly
+++ b/asllib/Parser.mly
@@ -522,8 +522,6 @@ let global_keyword ==
   | global_keyword_non_config
   | CONFIG; { GDK_Config }
 
-let pass == { S_Pass }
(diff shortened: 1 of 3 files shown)


asllib/Parser.mly
```diff
@@ -522,8 +522,6 @@ let global_keyword ==
   | global_keyword_non_config
   | CONFIG; { GDK_Config }
 
-let pass == { S_Pass }
```
HadrienRenaud, line 525: Just by curiosity, is there any difference for `pass`?

Roman-Manevich (approved):
Looks good to me!

asllib/Parser.mly
hrutvik: Discussed offline - no difference, just a refactor in an effort to improve readability

PR #1582, opened 2025-11-18
[gen] Rework on printing generated litmus tests

ShaleXIONG:
This pull request replaces the `diy` manual test printing by `lib/simpledumper.ml`. It fixes two main problems for the initialisation list.
- Incorrect array initialisation, for example, changing `int y[2]; y=5` to `int y[2] = {5,5}`.
- Using square bracket syntax for `pte` initialisation `[PTE(x)]`.

A good example is `diyone7 -arch AArch64 -variant kvm  PodRW Pa Rfe PodRW PteV1 Rfe`. The new result shows at the end of this message.

Separately, we made _major_ rework to `top_gen.ml` and `final.ml` in the internal code of printing litmus test. It now use `lib/SimpleDumper` directly rather than manually print out all the information. This simplifies the code and ensure the formatting is consistent with other tools in this repo. The new formatting can be seen below. Notice that 
- the `"PodRWPPa RfePaP PodRWPPteV1 RfePteV1P"` is now at the end of prelude,
- all the white space are the result of `SimpleDumper`,
- the code label, e.g. `L00`, now is on the same line as the instruction, and
- the `fault` in post-condition now is in lower case and logical negation is `not` rather than `~`.

Last, we add automatic `Variant= ...` based on the `diy*` argument `-variant ...`. This enables `herd7` directly configure the expected `variant`.
```
AArch64 LB+popptev1+poppa
Com=Rf Rf
Orig=PodRWPPa RfePaP PodRWPPteV1 RfePteV1P
Variant=vmsa
"PodRWPPa RfePaP PodRWPPteV1 RfePteV1P"
{
 [x]=1;
 [PTE(x)]=(oa:PA(x), valid:0);
 int y[2]={5,5};
 0:X0=x; 0:X2=y;
 1:X2=y; 1:X3=PTE(x); 1:X4=(oa:PA(x));
}
 P0               | P1          ;
 L00: LDR W1,[X0] | LDR W1,[X2] ;
 MOV W3,#7        | STR X4,[X3] ;
 SUB W4,W3,#1     |             ;
 STP W4,W3,[X2]   |             ;

exists (1:X1=6 /\ not (fault(P0:L00,x)))
```
It now has the same syntax as the `mprog7 -mode text` tool.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -694,10 +694,16 @@ test.vmsa+ifetch:
 	@ echo "herd7 AArch64 VMSA+ifetch instructions tests: OK"
 
 ### Diy tests, includes
+### - A `diyone7` generated syntax check
 ### - A `diy7` with `cycleonly` instance checks the cycle generations
 ### - Several `diycross7` + `herd7` instances, check if the generated litmus tests
 ###   are equivalent based on `herd7` result.
 diy-test:: | build
+diy-test:: diyone-basic-test
+diyone-basic-test:
+	@ echo
+	dune test gen/tests
+	@ echo "diyone7 basic test: OK"
 diy-test:: diy-baseline-cycleonly
 diy-baseline-cycleonly::
 	@ echo
--- a/gen/tests/diyone7-check.t
+++ b/gen/tests/diyone7-check.t
@@ -1,0 +1,251 @@
+metadata-false
+  $ diyone7 -arch AArch64 -variant vmsa PteOA PosWW PteOA PteV1 PteAF0 PosWR PteHA Fri -oneloc -metadata false
+  AArch64 CoWR0+posWpteoapteoa.v1.af0-pospteoa.v1.af0pteha-friptehapteoa
+  Variant=vmsa
+  TTHM=HA
+  {
+   [x]=1;
+   [PTE(x)]=(oa:PA(x), af:0, valid:0);
+   [y]=5;
+   0:X0=PTE(x); 0:X1=(oa:PA(y), af:0, valid:0); 0:X2=(oa:PA(y)); 0:X3=x;
+  }
+   P0               ;
+   STR X1,[X0]      ;
+   STR X2,[X0]      ;
+   L00: LDR W4,[X3] ;
+  
+  exists (fault(P0:L00,x))
+vmsa-neg-exists
+  $ diyone7 -arch AArch64 -variant kvm Amo.Cas TLBI-sync.ISHdWW PteV1 PteAF0 PteOA Rfe Pte PodRW PteHD Rfe -neg true -info "User-define=User-define"
+  AArch64 LB+popteptehd+amo.cas-tlbi-sync.ishppteoa.v1.af0
+  Variant=vmsa
+  TTHM=HD
+  Generator=diyone7 (version 7.58+1)
+  Prefetch=0:x=F,0:y=W,1:y=F,1:x=W
+  Com=Rf Rf
+  Orig=Amo.Cas TLBI-sync.ISHdWWPPteOA.V1.AF0 RfePteOA.V1.AF0Pte PodRWPtePteHD RfePteHDP
+  User-define=User-define
+  "Amo.Cas TLBI-sync.ISHdWWPPteOA.V1.AF0 RfePteOA.V1.AF0Pte PodRWPtePteHD RfePteHDP"
+  {
+   [x]=1;
+   [PTE(x)]=(oa:PA(x), db:0, dbm:1);
+   [y]=5;
+   [PTE(y)]=(oa:PA(y), valid:0);
+   0:X0=x; 0:X3=PTE(y); 0:X4=(oa:PA(x), af:0);
+   1:X0=x; pteval_t 1:X1=0; 1:X3=PTE(y);
+  }
+   P0                  | P1               ;
+   MOV W1,#2           | LDR X1,[X3]      ;
+   MOV W2,#3           | MOV W2,#2        ;
+   L01: CAS W1,W2,[X0] | L00: STR W2,[X0] ;
+   DSB ISH             |                  ;
+   LSR X5,X0,#12       |                  ;
+   TLBI VAAE1IS,X5     |                  ;
+   DSB ISH             |                  ;
+   STR X4,[X3]         |                  ;
+  
+  ~exists ([x]=3 /\ 0:X1=2 /\ 1:X1=(oa:PA(x), af:0) /\ not (fault(P0:L01,x)) /\ not (fault(P1:L00,x)))
+vmsa-forall
+  $ diyone7 -arch AArch64 -variant kvm Amo.Cas TLBI-sync.ISHdWW PteV1 PteAF0 PteOA Rfe Pte PodRW PteHD Rfe -info "User-define=User-define" -cond observe
+  AArch64 LB+popteptehd+amo.cas-tlbi-sync.ishppteoa.v1.af0
+  Variant=vmsa
+  TTHM=HD
+  Generator=diyone7 (version 7.58+1)
+  Prefetch=0:x=F,0:y=W,1:y=F,1:x=W
+  Com=Rf Rf
+  Orig=Amo.Cas TLBI-sync.ISHdWWPPteOA.V1.AF0 RfePteOA.V1.AF0Pte PodRWPtePteHD RfePteHDP
+  User-define=User-define
+  "Amo.Cas TLBI-sync.ISHdWWPPteOA.V1.AF0 RfePteOA.V1.AF0Pte PodRWPtePteHD RfePteHDP"
+  {
+   [x]=1;
+   [PTE(x)]=(oa:PA(x), db:0, dbm:1);
+   [y]=5;
+   [PTE(y)]=(oa:PA(y), valid:0);
+   0:X0=x; 0:X3=PTE(y); 0:X4=(oa:PA(x), af:0);
+   1:X0=x; pteval_t 1:X1=0; 1:X3=PTE(y);
+  }
+   P0                  | P1               ;
+   MOV W1,#2           | LDR X1,[X3]      ;
+   MOV W2,#3           | MOV W2,#2        ;
+   L01: CAS W1,W2,[X0] | L00: STR W2,[X0] ;
+   DSB ISH             |                  ;
+   LSR X5,X0,#12       |                  ;
+   TLBI VAAE1IS,X5     |                  ;
+   DSB ISH             |                  ;
+   STR X4,[X3]         |                  ;
+  
+  locations [x; 0:X1; 1:X1; fault(P0:L01,x); fault(P1:L00,x);]
+  forall (true)
+vmsa-location
+  $ diyone7 -arch AArch64 -variant kvm Amo.Cas TLBI-sync.ISHdWW PteV1 PteAF0 PteOA Rfe Pte PodRW PteHD Rfe -info "User-define=User-define" -cond unicond
+  AArch64 LB+popteptehd+amo.cas-tlbi-sync.ishppteoa.v1.af0
+  Variant=vmsa
+  TTHM=HD
+  Generator=diyone7 (version 7.58+1)
+  Prefetch=0:x=F,0:y=W,1:y=F,1:x=W
+  Com=Rf Rf
+  Orig=Amo.Cas TLBI-sync.ISHdWWPPteOA.V1.AF0 RfePteOA.V1.AF0Pte PodRWPtePteHD RfePteHDP
+  User-define=User-define
+  "Amo.Cas TLBI-sync.ISHdWWPPteOA.V1.AF0 RfePteOA.V1.AF0Pte PodRWPtePteHD RfePteHDP"
+  {
+   [x]=1;
+   [PTE(x)]=(oa:PA(x), db:0, dbm:1);
+   [y]=5;
+   [PTE(y)]=(oa:PA(y), valid:0);
+   0:X0=x; 0:X3=PTE(y); 0:X4=(oa:PA(x), af:0);
+   1:X0=x; pteval_t 1:X1=0; 1:X3=PTE(y);
+  }
+   P0                  | P1               ;
+   MOV W1,#2           | LDR X1,[X3]      ;
+   MOV W2,#3           | MOV W2,#2        ;
+   L01: CAS W1,W2,[X0] | L00: STR W2,[X0] ;
+   DSB ISH             |                  ;
+   LSR X5,X0,#12       |                  ;
+   TLBI VAAE1IS,X5     |                  ;
+   DSB ISH             |                  ;
+   STR X4,[X3]         |                  ;
+  
+  forall (not (fault(P0:L01,x)) /\ not (fault(P1:L00,x)) /\ ([y]=(oa:PA(x), af:0) /\ (0:X1=2 /\ ([x]=3 /\ (1:X1=(oa:PA(x), af:0) \/ 1:X1=0)) \/ 0:X1=0 /\ (1:X1=(oa:PA(x), af:0) /\ ([x]=3 \/ [x]=2) \/ 1:X1=0 /\ ([x]=3 \/ [x]=2)))))
+memtag
+  $ diyone7 -arch AArch64 -variant memtag DpDatadW T PosWW T Rfe PodRW Rfe T -info "Variant=memtag"
+  AArch64 LB+po+dataWtt-postt
+  Variant=memtag memtag
+  Generator=diyone7 (version 7.58+1)
+  Prefetch=0:x=F,0:y=W,1:y=F,1:x=W
+  Com=Rf Rf
+  Orig=DpDatadWTT PosWWTT RfeTP PodRW RfePT
+  "DpDatadWTT PosWWTT RfeTP PodRW RfePT"
+  {
+   0:X1=x:green; 0:X3=y:red; 0:X5=y:green; 0:X6=y:blue;
+   1:X1=x:green; 1:X6=y:blue;
+  }
+   P0                | P1               ;
+   MOV X0,X1         | L00: LDR W0,[X6] ;
+   LDG X0,[X1]       | MOV W2,#1        ;
+   EOR X2,X0,X0      | STR W2,[X1]      ;
+   ADD X4,X3,W2,SXTW |                  ;
+   STG X4,[X5]       |                  ;
+   STG X6,[X3]       |                  ;
+  
+  exists ([tag(y)]=:blue /\ 0:X0=x:green /\ 1:X0=0 /\ not (fault(P1:L00,y)))
+ifetch
+  $ diyone7 -arch AArch64 -variant ifetch CacheSyncStrongIsbdWRPI FreIP PodWR Fre
+  AArch64 SB+cachesyncstrongisbpi+po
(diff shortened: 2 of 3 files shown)


gen/common/variant_gen.ml
```diff
@@ -91,5 +96,14 @@ let pp = function
   | SME -> "sme"
   | ConstrainedUnpredictable -> "ConstrainedUnpredictable"
 
+let pp_herd_variant = function
+  | AsAmo | ConstsInInit | NoVolatile | NoFault | Neon | SVE | SME -> None
```
murzinv: `Neon`, `SVE` and `SME` are supported and valid variants for `herd7`

murzinv (commented):
How automatic insert of variant deal with
```
-info add metadata to generated test(s)
```
?
The reason I'm asking is that another way to inject variant into litmus test is using `-info`, for instance `-info "variant=memtag"`. Now if combined with automatic variant insertion is would add two variants into the test and upset `herd7`
```
Warning: File "T.litmus": Duplicated meta-data on key Variant
 (User error)
```

gen/common/variant_gen.ml
ShaleXIONG: Done. Thanks for the comment !

gen/common/archExtra_gen.ml
relokin: Ideally, this function would be producing the same output as `pp_location` in `lib/location.ml`. Could you do that?

Also, it would be good to understand whether we can reuse the code in `lib/location.ml` instead of redefining a new `location` type.

gen/top_gen.ml
```diff
@@ -1018,25 +1016,39 @@ let tr_labs m init =
...
+  :: [] )
+
+let merge_info lhs rhs =
+  let mk_map info =
+    List.map ( fun (k,v) -> String.capitalize_ascii k,v ) info
+      |> List.to_seq |> StringMap.of_seq in
+  StringMap.union
+    ( fun v1 v2 -> v1 ^ " " ^ v2 ) (mk_map lhs) (mk_map rhs)
+  |> StringMap.to_seq
+  |> List.of_seq
```
relokin: I will leave this to you but I find that the new code is not an improvement over the old. I understand that you want to add the variant line but I don't understand why you are refactoring the old code.

gen/common/archExtra_gen.ml
ShaleXIONG: I add a new commit to line it up with the `lib/location.ml`

gen/top_gen.ml
ShaleXIONG: Let me revisit this. My initial intention was to move all the update to the `info` into one function, so people can clearly see the ordered and what information has been added.

relokin:
Two more comments:
* The same variant might appear twice, if I do:
```
$> diyone7 -oneloc -variant mixed,MixedDisjoint -arch AArch64 h2 PosWW h0 Rfe Fre  -hexa  -metadata false
AArch64 WW+R+posh2h0+w0
Com=Rf Fr
Orig=PosWWh2h0 Rfeh0w0 Frew0h2
Variant=mixed mixed
"PosWWh2h0 Rfeh0w0 Frew0h2"
{
 uint32_t 0:X0=0; 0:X1=x; uint32_t 0:X2=0;
 uint32_t 1:X0=0; 1:X1=x;
}
 MOV W0,#514     | LDR W0,[X1] ;
 STRH W0,[X1,#2] |             ;
 MOV W2,#257     |             ;
 STRH W2,[X1]    |             ;

exists (([x]=0x2020101) /\ (1:X0=0x101))
```
The variant mixed appears twice.
* The variant should appear before the metadata, if not after all the metadata.
* As you see above, when I do `-metadata false`, I still get the metadata.

gen/CCompile_gen.ml
```diff
@@ -818,17 +818,18 @@ module Make(O:Config) : Builder.S
...
                   let evts =
                     List.map
                       (List.map (fun n -> n.C.evt))
                       splitted in
-                  F.run evts m
-              | Cycle -> F.check f
-              | Observe -> F.observe f in
-            (add_args env c,f (F.FaultSet.empty,F.FaultSet.empty)),
+                  F.run evts m flts
+              | Cycle -> F.check ~is_pos:true f flts
```
fsestini: Why are we hardcoding `~is_pos` to `true` here?

With this change, the following commands generate the same final condition:

```
dune exec diyone7 -- -arch C PodWW Rfe PodRR Fre
dune exec diyone7 -- -arch C PodWW Rfe PodRR Fre -neg true
```

whereas with `diyone7` from latest `master`, the final condition is negated when `-neg true`.

gen/CCompile_gen.ml
```diff
@@ -818,17 +818,18 @@ module Make(O:Config) : Builder.S
...
+              | Observe -> F.exist_true in
```
fsestini: Similarly, why is this set to `F.exist_true` in all cases?

Prior to this change, the command `dune exec diyone7 -- -arch C PodWW Rfe PodRR Fre -cond observe` would generate locations in the final condition, but now only generates `exists (true)`.

Actually, other architectures as well now seem to always include this `exists (true)` line:

```
$ dune exec diyone7 -- -arch AArch64 PodWW Rfe PodRR Fre -cond observe
AArch64 MP
Com=Rf Fr
...
STR W2,[X3] |             ;

locations [x; y; 1:X0; 1:X2;]
exists (true)
```

gen/final.ml
```diff
@@ -239,10 +229,15 @@ module Make : functor (O:Config) -> functor (C:ArchRun.S) ->
         else finals
     | None -> finals
 
+    open Printf
```
fsestini: Could I suggest to not open `Printf` (and similar large modules) globally, if possible. It pollutes the global scope unnecessarily and doesn't play well with LSPs. I would just use it fully qualified, or open it only within the scope of pretty printing functions.

gen/top_gen.ml
```diff
@@ -1008,30 +1005,49 @@ let tr_labs m env =
...
+  The key is case inssensitive. Yet the final output result will
+  capitalise the first letter.*)
+let merge_info lhs rhs =
+  let mk_map info =
+    List.map ( fun (k,v) -> String.capitalize_ascii k,v ) info
+      |> List.to_seq |> StringMap.of_seq in
+  StringMap.union
+    ( fun v1 v2 -> v1 ^ " " ^ v2 ) (mk_map lhs) (mk_map rhs)
+  |> StringMap.to_seq
+  |> List.of_seq
```
fsestini: Here the variants are sorted by key, which means the following comment by @relokin is still not implemented:

> The variant should appear before the metadata, if not after all the metadata.

For example, if I run:

```
$ dune exec diyone7 -- -oneloc -variant mixed,MixedDisjoint -arch AArch64 h2 PosWW h0 Rfe Fre -hexa -info 'Z=foo'
AArch64 WW+R+posh2h0+w0
Com=Rf Fr
Orig=PosWWh2h0 Rfeh0w0 Frew0h2
Variant=mixed
Z=foo
"PosWWh2h0 Rfeh0w0 Frew0h2"
{
 uint32_t 0:X0=0; 0:X1=x; uint32_t 0:X2=0;
...
```

The `Variant` key/value pair appears in between other metadata.

Code-wise, I don't see why we couldn't sort the keys via `List.sort` or similar, rather than `StringMap`.

gen/top_gen.ml
```diff
@@ -858,121 +861,115 @@ let max_set = IntSet.max_elt
...
+                    |> Option.value ~default:TestType.TyDef in
+          (loc,(typing,Option.value ~default:(A.S "0") init_opt))
+          ) inits in
+      (* Add those location exists in `env` but not in `inits`
+         to the head of `type_inits` *)
+      let locs = List.split inits |> fst |> A.LocSet.of_list in
+      let extra_type_declaration =
+        A.LocMap.filter ( fun k _ -> not @@ A.LocSet.mem k locs ) env
+        (* to_list only exists after ocaml 5 *)
+          |> A.LocMap.to_seq |> List.of_seq
```
fsestini: Might be worth adding this operation to `MyMap.S`. I think it's common enough.

gen/top_gen.ml
```diff
@@ -858,121 +861,115 @@ let max_set = IntSet.max_elt
...
+          (* give default 0 value *)
+          |> List.map ( fun (loc, t) -> (loc, (typ_to_testtype t, A.S "0")) ) in
+      List.sort (fun (l, _) (r, _) -> A.location_compare l r) (type_inits @ extra_type_declaration)
+          |> List.filter_map ( fun (loc, (typ, value)) ->
+              match typ,value with
+              (* fix the array value from `v` -> `{v, v, ..}` *)
+              | TestType.TyArray (_,size), A.S v ->
+                Some (loc, (typ, A.S ( sprintf "{%s}" ( List.init size ( fun _ -> v ) |> String.concat "," ) ) ) )
+              (* fix the array only support basic value *)
```
fsestini: What does this comment mean?

gen/final.ml
```diff
@@ -101,17 +102,7 @@ module Make : functor (O:Config) -> functor (C:ArchRun.S) ->
...
-          | I i1,I i2 -> compare i1 i2
-          | S s1,S s2 -> String.compare s1 s2
-          | P p1,P p2 -> C.A.Value.pte_compare p1 p2
-          | ((P _|S _),I _)
-          | (P _,S _)
-            -> -1
-          | (I _,(S _|P _))
-          | (S _,P _)
-            -> +1
+          let compare = compare
```
fsestini: Nitpick for extra clarity.
```suggestion
          let compare = Stdlib.compare
```

gen/final.ml
```diff
@@ -255,105 +250,96 @@ module Make : functor (O:Config) -> functor (C:ArchRun.S) ->
...
-          end
+    let fault_atom_to_prop is_pos flt =
+      let atom = Atom (FF flt) in
+      if is_pos then atom else Not (atom)
+
+    let fault_atoms_to_prop pos_flts neg_flts =
+      let pos = FaultAtomSet.elements pos_flts
+      |> List.map (fault_atom_to_prop true) in
+      let neg = FaultAtomSet.elements neg_flts
+      |> List.map (fault_atom_to_prop false) in
```
fsestini: I would indent these, otherwise it look like they are at the same scope level as `let`.
```suggestion
      let pos = FaultAtomSet.elements pos_flts
          |> List.map (fault_atom_to_prop true) in
      let neg = FaultAtomSet.elements neg_flts
          |> List.map (fault_atom_to_prop false) in
```

gen/final.ml
```diff
@@ -255,105 +250,96 @@ module Make : functor (O:Config) -> functor (C:ArchRun.S) ->
...
+
+    let check ~is_pos f (pos_flts,neg_flts) =
+      let value_prop = fenv_to_prop f in
+      let fault_prop = fault_atoms_to_prop pos_flts neg_flts in
+      let prop = match value_prop, fault_prop with
+      | (Atom _ as l), (Atom _ as r) -> And [l; r;]
+      | (Atom _ as l), And r -> And (l :: r)
+      | And l, (Atom _ as r) -> And (l @ [r])
+      | And l, And r -> And (l @ r)
+      | _,_ -> Warn.fatal "error in connect fault and value propositions" in
```
fsestini: As above.
```suggestion
      let prop = match value_prop, fault_prop with
        | (Atom _ as l), (Atom _ as r) -> And [l; r;]
        | (Atom _ as l), And r -> And (l :: r)
        | And l, (Atom _ as r) -> And (l @ [r])
        | And l, And r -> And (l @ r)
        | _,_ -> Warn.fatal "error in connect fault and value propositions" in
```

gen/final.ml
```diff
@@ -255,105 +250,96 @@ module Make : functor (O:Config) -> functor (C:ArchRun.S) ->
...
+      value_location @ fault_location
+
+    let rec run_cond_to_constr_gen_cond = function
+      | Run.Atom (loc,v) -> Atom ( LV (Loc loc, I v))
+      | Run.Or (cond) -> Or (List.map run_cond_to_constr_gen_cond cond)
+      | Run.And (cond) -> And (List.map run_cond_to_constr_gen_cond cond)
+
+    let run evts m (pos_flts,neg_flts) =
+      let value_prop = Run.run evts m
+      |> run_cond_to_constr_gen_cond in
```
fsestini:
```suggestion
      let value_prop = Run.run evts m
        |> run_cond_to_constr_gen_cond in
```

gen/final.ml
```diff
@@ -67,30 +70,28 @@ module Make : functor (O:Config) -> functor (C:ArchRun.S) ->
...
-    val check : fenv -> faults * faults -> final
-    val observe : fenv -> faults * faults -> final
-    val run : C.C.event list list -> C.A.location C.C.EventMap.t -> faults * faults -> final
+    type locations = (C.A.location, v, FaultType.No.t) LocationsItem.t
 
-    val dump_final : out_channel -> final -> unit
+    val check : is_pos:bool -> fenv -> faults * faults -> final
+    val exist_true : final
+    val observe : fenv -> faults * faults -> locations list
+    val run : C.C.event list list -> C.A.location C.C.EventMap.t -> faults * faults -> final
```
fsestini: I find the names `check`, `observe`, `run` etc. quite confusing for these functions. To me, these look like "smart constructors" that build `final` terms. In other words, they are not "checking" or "running" anything, really.

gen/final.ml
```diff
@@ -67,30 +70,28 @@ module Make : functor (O:Config) -> functor (C:ArchRun.S) ->
...
 
-    type final
+    type prop = (C.A.location, v, FaultType.No.t) ConstrGen.prop
+    type final = prop ConstrGen.constr
+    val pp_prop_atom : (C.A.location, v, FaultType.No.t) ConstrGen.atom -> string
 
-    val check : fenv -> faults * faults -> final
-    val observe : fenv -> faults * faults -> final
-    val run : C.C.event list list -> C.A.location C.C.EventMap.t -> faults * faults -> final
+    type locations = (C.A.location, v, FaultType.No.t) LocationsItem.t
```
fsestini, line 83: Do we actually need to expose the definition of `locations` and `final`, instead of making them abstract?
Do we need to expose `prop` at all?
`pp_prop_atom` seems to be added only for the dumper module's convenience, but otherwise has no real connection with `Final`. I'd just move it to the dumper module.
I'm also not sure we even need to include the whole `Fault.S` signature inside `Final`, since users of `Final` only access a fraction of what exposed by `Fault.S`. I understand `Fault.S` was already getting included prior to this PR, but since this PR is already doing some significant refactoring, it would have been nice if we could've also simplified things a bit in the process.

gen/top_gen.ml
```diff
@@ -1008,30 +1005,49 @@ let tr_labs m env =
...
           end
       | _ -> bd)
-    env
-
-let do_self =  O.variant Variant_gen.Self
+    init
+
+let basic_info scope prefetch com_edges cycle_description =
+  let append_key_string key value list =
+    match value with "" -> list | _ -> list @ [(key,value)] in
```
fsestini: This function traverses and rebuilds `list` in memory at every invocation. While the performance impact is probably negligible in this particular context, it still seems unnecessary.

Moreover, here the empty string `""` is used to signal absence of metadata value. That is brittle and obscures intent. I would use a `string option` value instead.

gen/final.ml
ShaleXIONG: Yes. Let me change it.

gen/final.ml
ShaleXIONG: Will do.

gen/final.ml
ShaleXIONG: Will do.

gen/final.ml
ShaleXIONG: Will do

gen/final.ml
ShaleXIONG: Will do.

gen/top_gen.ml
ShaleXIONG: I will update:
The first `v` -> `{v,v,v}` basically change initial value of array from for example `0`, to `{0,0}` based on size. Only the latter is an accepted syntax for `herd7`.
The second basically preserved the previous behaviour where array can only apple to plain value. Note that there is `A.P` for the `value` which corresponds to `Pte` value.

gen/top_gen.ml
ShaleXIONG: Done.

gen/final.ml
ShaleXIONG: Will change them to `exists_condition` `observe_condition` and `forall_condition`. Hope it will reduce some confusion.

gen/final.ml
ShaleXIONG: It will not be straight forwards to remove them.
The Fault signature comes with a `type fault` and a `module FaultAtomSet` (where the element is `type fatom`) which are used in `top_gen.ml`. I try to only declaim `type fault` and `type faults`, then ad
The details of `location` is needed to around the line 970, `MiscParser.locations = t.obs` in `top_gen.ml`. The compiler needs to know the type detail for type check.

I think I will push this problem for the future as the potentially further rework is to unify many type between `lib` and `gen`. For example `location` type you see here and `C.A.location`. Also the `fault` in `lib` are different from `fault` used internal in `diy`.

fsestini: Ok. Maybe for now we can just remove what's easy to remove (like `type prop` and `pp_prop_atom`, probably?) and consider further simplification of this module for a follow-up PR.

ShaleXIONG: Sorry, I forgot to mention `type prop` and the `pp_prop_atom`. Yes I managed to abstract away `type prop`.

gen/CCompile_gen.ml
ShaleXIONG: Done.

gen/top_gen.ml
ShaleXIONG: I update it in a way that the `rhs` will be merged into the `lhs`. Also I realised if the key collides, for example internal and user-defined, or between two user-defined. The value will be concatenate. 

In the case of `Variant` it is in fact ok to have `Variant = memtag memtag` for `herd7`.

gen/top_gen.ml
ShaleXIONG: Will change to `string option` then `filter_map`.

gen/CCompile_gen.ml
ShaleXIONG: One of the goal of this change is to line up with `mprog7 -mode text`. In the case where only exists `locations [...]`, the tool `mprog7` actually append a `forall (true)`. I will match the behaviour there. 

Regarding questions on when `location[...]` yields a different result in `herd7`, the example is `diyone7 -variant vmsa -arch AArch64 PteV1 PodWW Rfe PodRR Fre -cond observe`. The `location` contains `fault(P1:L00,x)`. This requests `herd7` to show the `fault` information in all stages.

gen/top_gen.ml
relokin: One more request please. Can we make sure that the variant is always printed. That's even when I do `-metadata false`.

relokin: And similarly for TTHM, `-metadata false` shouldn't hide. Also TTHM should appear right after the Variant=....

ShaleXIONG: All good point ! I will update now.

PR #1586, opened 2025-11-20
[aslspec] Automatically assign names to expressions in the output configurations of rules

Roman-Manevich:
The signature of a function/relation names various subexpressions to enable referencing them later. Specifically, the names assigned to subexpressions of the output configurations are used to render the output configurations with those names on top of them.
This PR uses the first output term to assign names to the output configuration of a rule.
For example, for
```
typing relation annotate_stmt(tenv: static_envs, s: stmt) -> (new_s: stmt, new_tenv: static_envs, ses: powerset(TSideEffect)) | type_error
=
  case SPass {
    s = S_Pass;
    --
    (S_Pass, tenv, empty_set);
  }
  ...
 ```
the output configuration is effectively as follows: `(new_s : S_Pass, new_tenv, ses: empty_set)`.

### DIFF
--- a/asllib/aslspec/spec.ml
+++ b/asllib/aslspec/spec.ml
@@ -98,6 +98,13 @@ module Error = struct
     spec_error
     @@ Format.asprintf "The rule for relation '%s' is empty." relation_name
 
+  let missing_relation_argument_name relation_name =
+    spec_error
+    @@ Format.asprintf
+         "All arguments in the relation '%s' must have names, since it \
+          specifies a rule."
+         relation_name
+
   let multiple_output_judgments relation_name rule_name_opt =
     let pp_name_opt fmt = function
       | Some name -> Format.fprintf fmt ": %s" name
@@ -323,8 +330,38 @@ type t = {
 
 let ast spec = spec.ast
 
-(** A module for performing resolution transformations on rules to make them
-    amenable for further checking and for rendering. *)
+(** [update_symbol_table ast] creates a symbol table from [ast]. *)
+let make_symbol_table ast =
+  let definition_nodes = list_definition_nodes ast in
+  make_id_to_definition_node definition_nodes
+
+(** A module for normalizing rules within relations to make them amenable for
+    further checking and for rendering. There are three transformations:
+    {ul
+     {- Aggregates consecutive [Case] blocks into a single [Case] block. }
+     {- Resolves [Unresolved] in applicators to either [Node_Relation] or
+        [Node_TypeVariant]
+     }
+     {- Transforms the conclusion judgment from an expression to a transition.
+        That is, given
+        {[
+          relation r(a: A) -> (b:B) =
+            p1; ... pk;
+            --
+            c_expr;
+        ]}
+        we transform it to
+        {[
+          relation r(a: A) -> (b:B) =
+            p1; ... pk;
+            --
+            r(a) -> c_expr;
+        ]}
+        The result is the re-written AST along with an updated map from
+        identifiers to definition nodes, updated for the newly created relation
+        nodes.
+     }
+    } *)
 module ResolveRules = struct
   open Rule
 
@@ -344,7 +381,9 @@ module ResolveRules = struct
         in
         Record { label; fields = resolved_fields }
     | Application ({ applicator; args } as app) -> (
-        let resolved_args = List.map (resolve_application_expr id_to_defining_node) args in
+        let resolved_args =
+          List.map (resolve_application_expr id_to_defining_node) args
+        in
         match applicator with
         | EmptyApplicator | Relation _ | ExprOperator _ | Fields _ ->
             Application { app with args = resolved_args }
@@ -389,53 +428,16 @@ module ResolveRules = struct
         in
         NamedExpr (resolved_sub_expr, name)
 
-  (** [extend_with_names expr type_term] recursively transforms [expr] by adding
-      names from [type_term] to sub-expressions of [expr]. Currently, only
-      tuples (labelled or unlabelled) are supported, which is sufficient for
-      most output configurations. *)
-  let rec extend_with_names expr type_term =
-    let opt_extend expr opt_name =
-      match opt_name with None -> expr | Some name -> NamedExpr (expr, name)
-    in
-    let applicator_matches_label applicator expr_label =
-      match (applicator, expr_label) with
-      | EmptyApplicator, None -> true
-      | TupleLabel label1, Some label2 -> label1 = label2
-      | _ -> false
-    in
-    match (expr, type_term) with
-    | _, LabelledTuple { label_opt = None; components = [ (opt_name, _) ] } ->
-        (* An unlabelled tuple with a single component serves as a named reference
-           to any type.*)
-        opt_extend expr opt_name
-    | Application { applicator; args }, LabelledTuple { label_opt; components }
-      when applicator_matches_label applicator label_opt ->
-        let extended_args =
-          List.map2
-            (fun arg (opt_name, arg_type) ->
-              opt_extend (extend_with_names arg arg_type) opt_name)
-            args components
-        in
-        Application { applicator; args = extended_args }
-    | _ -> expr
-
   (** [resolve_rule_element id_to_defining_node rule_element] performs
       resolution transformations on [rule_element] using [id_to_defining_node]
       to lookup type variants and relations. *)
-  let rec resolve_rule_element id_to_defining_node conclusion_lhs
-      first_output_type rule_element =
+  let rec resolve_rule_element id_to_defining_node conclusion_lhs rule_element =
     let open Rule in
     match rule_element with
     | Judgment ({ expr; is_output } as judgment) ->
         let resolved_expr = resolve_application_expr id_to_defining_node expr in
         let resolved_expr =
           if is_output then
-            (* TODO: we currently use the first output type term to assign names,
-            but a more flexible approach is to find the output type term
-            that resolved_expr corresponds to. *)
-            let resolved_expr =
-              extend_with_names resolved_expr first_output_type
-            in
             Transition
               {
                 lhs = conclusion_lhs;
@@ -463,14 +465,14 @@ module ResolveRules = struct
 
   (** [lhs_of_conclusion] returns an expression representing the LHS of the
       conclusion judgment for relation definition with the given [name] and
-      [input] arguments. *)
+      [input] arguments. This function assumes
+      [relation_names_arguments_if_exists_rule] has been called on the AST to
+      ensure that all input arguments have names. *)
   let lhs_of_conclusion { Relation.name; input } =
     let input_vars =
       List.map (fun (name_opt, _) -> Var (Option.get name_opt)) input
     in
     Application { applicator = Relation name; args = input_vars }
-
-  let first_output_type { Relation.output } = List.hd output
 
   (** In text, a list of [case] elements without non-[case] elements in between
       them are considered to be a single case element with multiple cases.
@@ -524,32 +526,6 @@ module ResolveRules = struct
     in
     aggregate rule_elements
 
-  (** Transforms relations in three ways:
-      {ul
-       {- Aggregates consecutive [Case] blocks into a single [Case] block. }
-       {- Resolves [Unresolved] in applicators to either [Node_Relation] or
-          [Node_TypeVariant]
-       }
-       {- Transforms the conclusion judgment from an expression to a transition.
-          That is, given
-          {[
-            relation r(a: A) -> (b:B) =
-              p1; ... pk;
-              --
-              c_expr;
-          ]}
-          we transform it to
-          {[
-            relation r(a: A) -> (b:B) =
-              p1; ... pk;
-              --
-              r(a) -> c_expr;
-          ]}
-          The result is the re-written AST along with an updated map from
-          identifiers to definition nodes, updated for the newly created
-          relation nodes.
-       }
-      } *)
   let resolve ast id_to_defining_node =
     let open Rule in
     let ast =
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: asllib/aslspec/render.ml [hunk 1]
```diff
@@ -387,12 +387,258 @@ module Make (S : SPEC_VALUE) = struct
...
+          | [] -> ()
+          | _ ->
+              fprintf fmt {|\;%a\;%a|} pp_macro
+                short_circuit_separator_macro_name
+                (PP.pp_sep_list ~sep:", " pp_print_string)
+                alternative_macros)
+      | Some alternatives ->
+          if Utils.list_is_empty alternatives then ()
+          else
+            let layout =
```
```suggestion
      | Some [] -> ()
      | Some alternatives ->
          let layout =
```

asllib/aslspec/spec.ml
```diff
@@ -291,6 +334,372 @@ type t = {
...
+        | Unresolved (FieldAccess path) ->
+            Application { applicator = Fields path; args = resolved_args }
+        | Unresolved _ ->
+            Error.invalid_application_of_expression_to_arguments expr)
+    | Transition { lhs; rhs; short_circuit } ->
+        let resolved_lhs = resolve_application_expr id_to_defining_node lhs in
+        let resolved_rhs = resolve_application_expr id_to_defining_node rhs in
+        let resolved_short_circuit =
+          Option.map
+            (List.map (fun e -> resolve_application_expr id_to_defining_node e))
```
HadrienRenaud: nit:
```suggestion
            (List.map (resolve_application_expr id_to_defining_node))
```

asllib/aslspec/spec.ml
```diff
@@ -291,6 +334,372 @@ type t = {
...
+      tuples (labelled or unlabelled) are supported, which is sufficient for
+      most output configurations. *)
+  let rec extend_with_names expr type_term =
+    let opt_extend expr opt_name =
+      match opt_name with None -> expr | Some name -> NamedExpr (expr, name)
+    in
+    let applicator_matches_label applicator expr_label =
+      match (applicator, expr_label) with
+      | EmptyApplicator, None -> true
+      | TupleLabel label1, Some label2 -> label1 = label2
```
HadrienRenaud: Can we avoid polymorphic comparaison when types are not immediately obvious please?

```suggestion
      | TupleLabel label1, Some label2 -> String.equal label1 label2
```

asllib/aslspec/spec.ml
Roman-Manevich: I don't like doing eta reductions when the function has more than one argument. I think it makes the code less readable. How strongly do you feel about this?

HadrienRenaud: I don't feel strongly about this, it takes less space I guess

asllib/aslspec/spec.ml
```diff
@@ -291,6 +334,372 @@ type t = {
...
+    in
+    match (expr, type_term) with
+    | _, LabelledTuple { label_opt = None; components = [ (opt_name, _) ] } ->
+        (* An unlabelled tuple with a single component serves as a named reference
+           to any type.*)
+        opt_extend expr opt_name
+    | Application { applicator; args }, LabelledTuple { label_opt; components }
+      when applicator_matches_label applicator label_opt ->
+        let extended_args =
+          List.map2
```
HadrienRenaud: Here we know that the number of elements in `args` and `components` are the same?

asllib/aslspec/spec.ml
```diff
@@ -291,6 +334,372 @@ type t = {
 
 let ast spec = spec.ast
 
+(** A module for performing resolution transformations on rules to make them
```
HadrienRenaud: Sorry could you expand a little bit on what a `resolution transformation` is?

asllib/aslspec/spec.ml
```diff
@@ -291,6 +334,372 @@ type t = {
...
+        let resolved_body = resolve_application_expr id_to_defining_node body in
+        Indexed { indexed_expr with body = resolved_body }
+    | NamedExpr (sub_expr, name) ->
+        let resolved_sub_expr =
+          resolve_application_expr id_to_defining_node sub_expr
+        in
+        NamedExpr (resolved_sub_expr, name)
+
+  (** [extend_with_names expr type_term] recursively transforms [expr] by adding
+      names from [type_term] to sub-expressions of [expr]. Currently, only
```
HadrienRenaud: Is `type_term` typically coming from the type definition of a rule?

asllib/aslspec/spec.ml
```diff
@@ -291,6 +334,372 @@ type t = {
...
+            cases
+        in
+        Cases resolved_cases
+
+  (** [lhs_of_conclusion] returns an expression representing the LHS of the
+      conclusion judgment for relation definition with the given [name] and
+      [input] arguments. *)
+  let lhs_of_conclusion { Relation.name; input } =
+    let input_vars =
+      List.map (fun (name_opt, _) -> Var (Option.get name_opt)) input
```
HadrienRenaud, line 473: Here `Option.get` cannot fail because the name has been resolved?

asllib/aslspec/spec.ml
```diff
@@ -291,6 +334,372 @@ type t = {
...
+                    let resolved_elements =
+                      List.map
+                        (fun rule_elem ->
+                          resolve_rule_element id_to_defining_node
+                            conclusion_lhs first_output_type rule_elem)
+                        elements
+                    in
+                    { def with rule_opt = Some resolved_elements }
+              in
+              Elem_Relation resolved_def)
```
HadrienRenaud:
```suggestion
          | Elem_Relation { rule_opt = None } -> term
          | Elem_Relation ({ rule_opt = Some elements } as def) ->
              let elements = aggregate_case_blocks elements in
              let conclusion_lhs = lhs_of_conclusion def in
              let first_output_type = first_output_type def in
              let resolved_elements =
                List.map
                  (fun rule_elem ->
                    resolve_rule_element id_to_defining_node
                      conclusion_lhs first_output_type rule_elem)
                  elements
                in
              let resolved_def =
                { def with rule_opt = Some resolved_elements }
              in
              Elem_Relation resolved_def)
```

You could also isolate this case into its own function.

asllib/aslspec/spec.ml
```diff
@@ -291,6 +334,372 @@ type t = {
...
+      | Some prefix_name, Some suffix_name ->
+          Some (Rule.join_case_names [ prefix_name; suffix_name ])
+    in
+    let category_opt =
+      match (prefix_category_opt, suffix_category_opt) with
+      | Some category, None | None, Some category -> Some category
+      | Some prefix_category, Some suffix_category ->
+          if prefix_category = suffix_category then Some prefix_category
+          else assert false
+            (* This should not happen: the same category is set by [expand]. *)
```
HadrienRenaud: Changing a polymorphic equal.

```suggestion
          (* the same category is set by [expand]. *)
          assert
            (Relation.relation_category_equal prefix_category suffix_category);
          Some prefix_category
```

asllib/aslspec/spec.ml
```diff
@@ -895,11 +1343,25 @@ let from_ast ast =
...
+  let () = Check.CheckRules.check ast in
   { ast; id_to_defining_node; defined_ids }
 
 let defined_ids self = self.defined_ids
 
-let defining_node_for_id self id =
+let defining_node_opt_for_id self id =
   match StringMap.find_opt id self.id_to_defining_node with
+  | Some def -> Some def
+  | None -> None
```
HadrienRenaud: Looks like it's directly the call to `StringMap.find_opt`?


```suggestion
StringMap.find_opt id self.id_to_defining_node
```

asllib/aslspec/spec.ml
```diff
@@ -895,11 +1343,25 @@ let from_ast ast =
...
+
+let defining_node_for_id self id =
+  match defining_node_opt_for_id self id with
   | Some def -> def
   | None -> Error.undefined_element id
+
+let relation_for_id self id =
+  match defining_node_for_id self id with
+  | Node_Relation def -> def
+  | _ -> assert false
```
HadrienRenaud, line 1414: This does not correspond to the comment on the `mli`:
```ocaml
val relation_for_id : t -> string -> Relation.t
(** [relation_for_id spec id] returns the relation definition for [id] in
    [spec]. Raises [SpecError] if [id] is not defined as a relation. *)
```

asllib/aslspec/spec.ml
```diff
@@ -496,6 +906,44 @@ module Check = struct
...
+            (fun { Rule.is_output } ->
+              if is_output then
+                Error.multiple_output_judgments relation_name
+                  expanded_rule.name_opt
+              else ())
+            prefix_fules
+      | _ -> Error.missing_output_judgment (Option.get expanded_rule.name_opt)
+
+    let check_rule_for_relation { Relation.name; category } elements =
+      let expanded_rules = ExpandRules.expand category elements in
```
HadrienRenaud: Couldn't we do the expansion just once instead of doing it: once to do the checks and once to render?

asllib/aslspec/spec.ml
```diff
@@ -496,6 +906,44 @@ module Check = struct
...
+      List.iter
+        (fun elem ->
+          match elem with
+          | Elem_Type _ | Elem_Constant _ | Elem_RenderTypes _
+          | Elem_RenderRule _ ->
+              ()
+          | Elem_Relation ({ rule_opt } as def) -> (
+              match rule_opt with
+              | None -> ()
+              | Some elements -> check_rule_for_relation def elements))
```
HadrienRenaud:
```suggestion
          | Elem_Type _ | Elem_Constant _ | Elem_RenderTypes _
          | Elem_RenderRule _ | Elem_Relation { rule_opt = None } ->
              ()
          | Elem_Relation ({ rule_opt = Some elements } as def) ->
              check_rule_for_relation def elements
```

asllib/aslspec/spec.ml
Roman-Manevich: This whole business has now been removed.

asllib/aslspec/spec.ml
Roman-Manevich: ah yes, a copy-paste error. thanks for catching this.

asllib/aslspec/spec.ml
Roman-Manevich: We don't. This will be checked in a future PR. However, I've factored name extension to a separate transformation phase so we can indeed insert a checker in between the other transformations (that do not require checking) and this one, which does require checking.

asllib/aslspec/spec.ml
Roman-Manevich: There is a detailed documentation for the `resolve` function itself. I pulled it up to describe the entire module.

asllib/aslspec/spec.ml
Roman-Manevich: It's a type appearing in the output of the relation signature.

asllib/aslspec/spec.ml
Roman-Manevich: No. This needs to be checked. I intended to add checks in a future PR, but so I don't forget I'm adding it now.

asllib/aslspec/spec.ml
Roman-Manevich: I thought about this and decided against. The reason is that it would mean caching the expanded rules for the entire specification before going to render it, which would consume non-trivial memory. Given that rules are not that large, I think it's most efficient to expand them only when needed and then release the memory.

asllib/aslspec/spec.ml
```diff
@@ -1240,6 +1365,26 @@ module Check = struct
...
+          List.iter
+            (fun (opt_name, _) ->
+              if Option.is_none opt_name then
+                Error.missing_relation_argument_name name
+              else ())
+            input
+    in
+    List.iter
+      (fun elem ->
+        match elem with Elem_Relation def -> check_relation def | _ -> ())
```
HadrienRenaud:
```suggestion
      (function Elem_Relation def -> check_relation def | _ -> ())
```

asllib/aslspec/spec.ml
```diff
@@ -845,10 +971,9 @@ module Check = struct
           | Elem_Type _ | Elem_Constant _ | Elem_RenderTypes _
           | Elem_RenderRule _ ->
               ()
-          | Elem_Relation ({ rule_opt } as def) -> (
-              match rule_opt with
-              | None -> ()
-              | Some elements -> check_rule_for_relation def elements))
+          | Elem_Relation ({ rule_opt = Some elements } as def) ->
+              check_rule_for_relation def elements
+          | Elem_Relation { rule_opt = None } -> ())
```
HadrienRenaud: Maybe this case could be merged with the previous ones?

asllib/aslspec/spec.ml
```diff
@@ -1240,6 +1365,26 @@ module Check = struct
...
+    let open Rule in
+    let check_relation { Relation.name; input; rule_opt } =
+      match rule_opt with
+      | None -> ()
+      | Some _ ->
+          List.iter
+            (fun (opt_name, _) ->
+              if Option.is_none opt_name then
+                Error.missing_relation_argument_name name
+              else ())
```
HadrienRenaud: technically the `else ()` is optional. Do you want to keep it?

asllib/aslspec/spec.ml
```diff
@@ -512,27 +535,130 @@ module ResolveRules = struct
...
+              let first_output_type = List.hd output in
+              (* TODO: we currently use the first type variant in the list of
+               output types for the relation to assign names.
+               A more flexible approach is to find the output type term
+               that rhs corresponds to. *)
+              let extended_elements =
+                List.map (extend_rule_element first_output_type) elements
+              in
+              Elem_Relation { def with rule_opt = Some extended_elements }
+          | Elem_Relation { rule_opt = None } as elem -> elem)
```
HadrienRenaud: Maybe this should be merged with the previous cases? (i.e. the ones `Elem_Type _` and other ones, not the other `Elem_Relation _` one)

Also:

```suggestion
          | Elem_Relation { rule_opt = None } -> elem)
```

PR #1587, opened 2025-11-23
[aslspec] allow defining operators

Roman-Manevich:
This PR allows defining operators - relation-like definitions that can be used to construct expressions.
Another change is allowing operators, but not relations just yet, to define type parameters.
An operator like
```
operator member[T](x: T, s: powerset(T)) -> Bool
{
  math_macro = \member,
};
```
allows rendering `x \in S`.
The rendering of operators follows some rules:
- If an operator is unary it expects a unary macro and rendered as `\macro{arg}`
- If an operator is binary is expects a nullary macro and rendered as `a \macro b`
- If an operator takes a list argument and has `associative = true` then it is considered a binary operator. It expects a nullary macro and rendered as `a1 \macro .... \macro an`
- If an operator has more than two arguments or has `custom = true` it is rendered as `\macro[H/V]{a1}...{an}` where `H`/`V` stands for horizontal/vertical layout. The layout argument should be optional for the macro.

Operators can always be specified with the `operator(a1,...,an)` syntax, but some binary operators and the `if_then_else` operators are directly supported by the parser as binary/ternary operators.
 
This PR defines a list of operators in `asl.spec`.

In terms of checking operators with type parameters - the checker modifies the symbol table while checking an operator by adding the type parameters for the duration of the check.

### DIFF
--- a/asllib/aslspec/AST.ml
+++ b/asllib/aslspec/AST.ml
@@ -34,6 +34,12 @@ module AttributeKey = struct
             format [{var}]. *)
     | Math_Macro  (** A LaTeX macro name for the element. *)
     | Math_Layout  (** The visual layout of the element. *)
+    | Associative
+        (** Whether an operator over a list of arguments should be rendered as a
+            list of expressions separated by the operator macro. *)
+    | Custom
+        (** Whether an operator over a list of arguments should be rendered as a
+            by a custom multi-argument macro. *)
     | Short_Circuit_Macro
         (** A LaTeX macro name to succinctly denote any value of a type [T].
             This is used to denote the short-circuit result of a relation
@@ -52,6 +58,8 @@ module AttributeKey = struct
       | Math_Layout -> 3
       | Short_Circuit_Macro -> 4
       | LHS_Hypertargets -> 5
+      | Associative -> 6
+      | Custom -> 7
     in
     let a_int = key_to_int a in
     let b_int = key_to_int b in
@@ -66,6 +74,8 @@ module AttributeKey = struct
     | Math_Layout -> "math_layout"
     | Short_Circuit_Macro -> "short_circuit_macro"
     | LHS_Hypertargets -> "lhs_hypertargets"
+    | Associative -> "associative"
+    | Custom -> "custom"
 end
 
 (** A value associated with an attribute key. *)
@@ -107,7 +117,8 @@ module Attributes = struct
   (** Helper functions for typed attribute access *)
 
   (** [find_string key attrs] returns [Some s] if [key] maps to a
-      [StringAttribute s] in [attrs], [None] otherwise. *)
+      [StringAttribute s] in [attrs], [None] otherwise. This assumes that [key]
+      is indeed a string attribute. *)
   let find_string key attrs =
     match find_opt key attrs with
     | Some (StringAttribute s) -> Some s
@@ -123,14 +134,15 @@ module Attributes = struct
     | _ -> ""
 
   (** [get_string_exn key attrs] returns the string value for [key] in [attrs],
-      or raises [Assert_failure] if not found or wrong type. *)
+      or asserts false if not found or wrong type. *)
   let get_string_exn key attrs =
     match find_opt key attrs with
     | Some (StringAttribute s) -> s
     | _ -> assert false
 
   (** [find_math_macro key attrs] returns [Some s] if [key] is bound to the math
-      macro [s], and [None] otherwise. *)
+      macro [s], and [None] otherwise. This assumes that [key] is indeed a math
+      macro attribute. *)
   let find_math_macro key attrs =
     match find_opt key attrs with
     | Some (MathMacroAttribute s) -> Some s
@@ -138,7 +150,8 @@ module Attributes = struct
     | _ -> None
 
   (** [find_layout key attrs] returns [Some layout] if [key] is bound to
-      [layout] in [attrs], and [None] otherwise. *)
+      [layout] in [attrs], and [None] otherwise. This assumes that [key] is
+      indeed a math layout attribute. *)
   let find_layout key attrs =
     match find_opt key attrs with
     | Some (MathLayoutAttribute layout) -> Some layout
@@ -366,41 +379,13 @@ end
 
 (** A datatype for a set of inference rules for a given relation. *)
 module Rule = struct
-  (** An operator over expressions. *)
-  type expr_operator =
-    | Operator_Assign
-    | Operator_Equal
-    | Operator_Iff
-    | Operator_List
-    | Operator_Size
-    | Operator_Set
-    | Operator_Union
-    | Operator_UnionList
-    | Operator_Some
-
-  (** [is_infix_operator] returns [true] if the operator is a binary operator
-      that is typically written as [e1 op e2]. *)
-  let is_infix_operator = function
-    | Operator_Assign | Operator_Equal | Operator_Iff -> true
-    | Operator_List | Operator_Set | Operator_Size | Operator_Union
-    | Operator_UnionList | Operator_Some ->
-        false
-
-  (** [is_prefix_operator] returns [true] if the operator is a unary operator
-      that is typically written as [op(e1,...,ek)]. *)
-  let is_prefix_operator op = not (is_infix_operator op)
-
-  (** [is_associative_operator op] returns [true] if [op(e1,...,ek)] is
-      typically written as [e1 op e2 ... op ek]. *)
-  let is_associative_operator = function Operator_Union -> true | _ -> false
-
   (** The left-hand side of an application expression. *)
   type applicator =
     | EmptyApplicator
         (** This is used for tuples, for which there is no applicator. *)
     | Relation of string
     | TupleLabel of string
-    | ExprOperator of expr_operator
+    | ExprOperator of string
     | Fields of string list
     | Unresolved of expr
 
@@ -444,11 +429,11 @@ module Rule = struct
   let make_application lhs exprs =
     Application { applicator = Unresolved lhs; args = exprs }
 
-  let make_prefix_operator_application op args =
-    Application { applicator = ExprOperator op; args }
-
-  let make_infix_operator_application op lhs rhs =
-    Application { applicator = ExprOperator op; args = [ lhs; rhs ] }
+  (** [make_operator_application op_name args] constructs an application
+      expression representing the application of operator [op_name] to [args].
+  *)
+  let make_operator_application op_name args =
+    Application { applicator = ExprOperator op_name; args }
 
   let make_record label fields = Record { label; fields }
   let make_list_index var index = ListIndex { var; index }
@@ -498,6 +483,9 @@ module Relation : sig
 
   type t = {
     name : string;
+    parameters : string list;
+        (** Type parameters. Currently, only available to operators. *)
+    is_operator : bool;
     property : relation_property;
     category : relation_category option;
     input : opt_named_type_term list;
@@ -516,10 +504,26 @@ module Relation : sig
     Rule.t option ->
     t
 
+  val make_operator :
+    string ->
+    string list ->
+    opt_named_type_term list ->
+    type_term ->
+    (AttributeKey.t * attribute) list ->
+    t
+
   val attributes_to_list : t -> (AttributeKey.t * attribute) list
   val prose_description : t -> string
   val math_macro : t -> string option
   val prose_application : t -> string
+
+  val is_associative_operator : t -> bool
+  (** [is_associative_operator t] tests whether the operator represented by [t]
+      has the [associative] attributes set to [true]. *)
+
+  val is_custom_operator : t -> bool
+  (** [is_custom_operator t] tests whether the operator represented by [t] has
+      the [custom] attributes set to [true]. *)
 
   val math_layout : t -> layout option
   (** The layout used when rendered as a stand-alone relation definition. *)
@@ -534,6 +538,8 @@ end = struct
 
   type t = {
     name : string;
+    parameters : string list;
+    is_operator : bool;
     property : relation_property;
     category : relation_category option;
     input : opt_named_type_term list;
@@ -545,12 +551,27 @@ end = struct
   let make name property category input output attributes rule_opt =
     {
       name;
+      parameters = [];
+      is_operator = false;
       property;
       category;
       input;
       output;
       att = Attributes.of_list attributes;
       rule_opt;
+    }
+
+  let make_operator name parameters input output_type attributes =
+    {
+      name;
+      parameters;
+      is_operator = true;
+      property = RelationProperty_Function;
+      category = None;
+      input;
+      output = [ output_type ];
+      att = Attributes.of_list attributes;
+      rule_opt = None;
     }
 
   let attributes_to_list self = Attributes.bindings self.att
@@ -568,6 +589,12 @@ end = struct
 
   let math_layout self =
     Attributes.find_layout AttributeKey.Math_Layout self.att
+
+  let is_associative_operator self =
+    Attributes.get_bool AttributeKey.Associative ~default:false self.att
+
+  let is_custom_operator self =
+    Attributes.get_bool AttributeKey.Custom ~default:false self.att
 end
 
 (** A datatype for grouping (subsets of) type definitions. *)
--- a/asllib/aslspec/LayoutUtils.ml
+++ b/asllib/aslspec/LayoutUtils.ml
@@ -32,8 +32,15 @@ let rec contains_vertical = function
 let apply_layout_to_list layout elements =
   let args_layout =
     match layout with
+    | (Horizontal [ _ ] | Vertical [ _ ]) when List.length elements > 1 ->
+        unspecified_for_elements elements
     | Horizontal l | Vertical l -> l
     | Unspecified -> unspecified_for_elements elements
   in
-  assert (List.compare_lengths elements args_layout = 0);
+  if List.compare_lengths elements args_layout <> 0 then
+    failwith
+      (Format.asprintf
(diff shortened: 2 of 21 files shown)


asllib/aslspec/LayoutUtils.ml
```diff
@@ -32,8 +32,16 @@ let rec contains_vertical = function
...
     match layout with
-    | Horizontal l | Vertical l -> l
+    | Horizontal l | Vertical l ->
+        if List.length l == 1 && List.length elements > 1 then
+          unspecified_for_elements elements
+        else l
     | Unspecified -> unspecified_for_elements elements
   in
-  assert (List.compare_lengths elements args_layout = 0);
+  if List.compare_lengths elements args_layout <> 0 then
```
Roman-Manevich, line 40: This is temporary until a check is in place.

asllib/doc/asl.spec
```diff
@@ -193,10 +423,10 @@ ast expr { "expression" } =
     { "variable expression for {name}" }
     | E_ATC(source: expr, type: ty)
     { "asserting type conversion for the source expression {source} and type {type}" }
-    | E_Binop(operator: binop, left: expr, right: expr)
```
Roman-Manevich, line 196: I had to rename `operator`, since it is now a keyword.

asllib/doc/asl.spec
```diff
@@ -1539,11 +1769,12 @@ semantics relation eval_multi_assignment(env: envs, lelist: list0(expr), vmlist:
         | ResultLexpr(new_g: XGraphs, new_env: envs)
         | TThrowing
         | TDynError
+        | TDiverging
```
Roman-Manevich, line 1797: This is a fix. `TDiverging` should have been specified in an earlier PR.

asllib/doc/asl.spec
```diff
@@ -3828,7 +4059,8 @@ typing relation annotate_exprs(tenv: static_envs, exprs: list0(expr)) ->
 };
 
 semantics relation eval_call(env: envs, name: Identifier, params: list0(expr), args: list0(expr)) ->
-    ResultCall(vms2: (list0(value_read_from), XGraphs), new_env: envs) | TDynError | TDiverging
+    ResultCall(vms2: (list0(value_read_from), XGraphs), new_env: envs)
+    | TThrowing | TDynError | TDiverging
```
Roman-Manevich, line 4088: This is a fix. `TThrowing` should have been specified in an earlier PR.

asllib/aslspec/spec.ml
```diff
@@ -328,7 +327,42 @@ type t = {
       (** The list of identifiers defined in the spec in order of appearance. *)
 }
 
-let ast spec = spec.ast
+let defined_ids self = self.defined_ids
```
Roman-Manevich, line 337: The next four functions were moved here from a different position in the file.

asllib/aslspec/tests.t/operators.expected
```diff
@@ -0,0 +1,50 @@
...
+
+
+
+
+
+
+
+
+
+
```
hrutvik, line 34: Is all this whitespace intentional?

asllib/aslspec/tests.t/operators.spec
```diff
@@ -0,0 +1,38 @@
...
+  associative = true,
+  math_macro = \intplus,
+};
+
+typing relation f(a: N, b: N, c: N, S: powerset(N)) -> N {} =
+    a + b = c;
+    d := if a in S then b else c;
+    --
+    d;
+;
```
hrutvik, line 36: I'm not sure I follow - how does this use the operators defined above?

asllib/aslspec/AST.ml
```diff
@@ -498,6 +483,9 @@ module Relation : sig
 
   type t = {
     name : string;
+    parameters : string list;
+        (** Type parameters. Currently, only available to operators. *)
+    is_operator : bool;
```
hrutvik, line 488: nit: I wonder, if these two are linked, perhaps a distinct type for them might be useful - e.g. something isomorphic to `is_operator : string list option`? (I don't feel strongly about this)

asllib/aslspec/PP.ml
```diff
@@ -194,12 +175,32 @@ let relation_keyword_to_string =
   | RelationProperty_Relation -> tok_str RELATION
   | RelationProperty_Function -> tok_str FUNCTION
 
+(** [pp_parameters fmt parameters] pretty-prints the list of parameters
+    [parameters] with the formatter [fmt]. If the list is empty, nothing is
+    printed. *)
+let pp_parameters fmt parameters =
+  if Utils.list_is_empty parameters then ()
+  else fprintf fmt "[%a]" (pp_sep_list ~sep:", " pp_print_string) parameters
```
hrutvik: nit:
```suggestion
  if not (Utils.list_is_empty parameters) then
    fprintf fmt "[%a]" (pp_sep_list ~sep:", " pp_print_string) parameters
```

asllib/aslspec/spec.ml
```diff
@@ -391,16 +425,20 @@ module ResolveRules = struct
...
             | Some (Node_TypeVariant _) ->
                 Application { applicator = TupleLabel id; args = resolved_args }
             | Some (Node_Constant { Constant.name })
             | Some (Node_Type { Type.name }) ->
                 Error.invalid_application_of_symbol_in_expr name expr
-            | _ ->
-                Application { applicator = Fields [ id ]; args = resolved_args }
-            )
+            | None | Some (Node_RecordField _) ->
+                Format.eprintf "unspected application LHS: %a@." PP.pp_expr expr;
```
hrutvik: typo:
```suggestion
                Format.eprintf "unexpected application LHS: %a@." PP.pp_expr expr;
```
Perhaps separately, maybe this should become some form of dedicated "internal error"?

asllib/aslspec/spec.mli
```diff
@@ -74,6 +74,13 @@ val is_defined_id : t -> string -> bool
 (** [is_defined_id spec id] returns [true] if [id] is defined in [spec] and
     [false] otherwise. *)
 
+val is_operator : elem -> bool
+(** [is_operator elem] returns [true] if [elem] is a relation of category
+    [RelationCategory_Operator] in [spec], and [false] otherwise. *)
+
+(** A module for expanding rules with cases into multiple rules without cases.
```
hrutvik, line 81: Is this doc comment in the right place? It doesn't seem attached to the `module` below

asllib/aslspec/SpecParser.mly
```diff
@@ -345,24 +399,34 @@ let expr :=
...
 
-let prefix_expr_operator ==
-    | UNION; { Rule.Operator_Union }
-    | UNION_LIST; { Rule.Operator_UnionList }
-    | LIST; { Rule.Operator_List }
-    | SET; { Rule.Operator_Set }
-    | SIZE; { Rule.Operator_Size }
-    | SOME; { Rule.Operator_Some }
-
 let infix_expr_operator ==
```
hrutvik, line 412: Ah I see now that various operators have hard-coded notations. I wonder if this seems a bit unusual? In particular, the fact that the operators are "defined" in the `.spec` file, but actually they are used via a different, hard-coded notation?

asllib/doc/asl.spec
```diff
@@ -85,6 +85,261 @@ typedef def_use_name { "subprogram identifier kind" } =
...
+{
+  math_macro = \member,
+};
+
+operator not_member[T](x: T, s: powerset(T)) -> Bool
+{
+  math_macro = \notmember,
+};
+
+operator suset[T](A: powerset(T), B: powerset(T)) -> Bool
```
hrutvik: typo
```suggestion
operator subset[T](A: powerset(T), B: powerset(T)) -> Bool
```

asllib/doc/ASLFormal.tex
```diff
@@ -57,12 +57,16 @@ \section{Mathematical Definitions and Notations\label{sec:Mathematical Definitio
...
 \end{array}
 \]
-\hypertarget{def-neg}{}
+\hypertarget{operator-neg}{}
+\hypertarget{operator-IFF}{}
+\hypertarget{operator-implies}{}
 For a pair of Boolean values $a,b\in\Bool$, we define $a \land b \triangleq \land\{a, b\}$
 and $a \lor b \triangleq \lor\{a, b\}$.
+We define implication as $a \implies b \triangleq (\neg a) \lor b$, and
+bi-implication as $a \IFF b \triangleq (a \land b) \lor (\neg a \land \neg b)$.
```
hrutvik: nit: I think it's more common to go for `(a \implies b) \land (b \implies a)`

asllib/doc/ASLmacros.tex
```diff
@@ -276,9 +270,96 @@
...
+\newcommand\notmember[0]{\mathrel{\not\in}} % DO NOT LINT
+\newcommand\unionlist[0]{\hyperlink{def-unionlist}{\textfunc{union\_list}}}
+\newcommand\UNIONLIST[1]{\unionlist\left({#1}\right)} % A union of a list of sets
+
+\newcommand\intplus[0]{\mathbin{+}} % DO NOT LINT
+\newcommand\intminus[0]{\mathbin{-}} % DO NOT LINT
+\newcommand\intnegate[1]{\mathbin{\smash{-}}{#1}} % DO NOT LINT
+\newcommand\inttimes[0]{\mathbin{\times}} % DO NOT LINT
+\newcommand\intdivide[0]{\mathbin{\div}} % DO NOT LINT
+\newcommand\intexponent[0]{\mathbin{{^}}} % DO NOT LINT
```
hrutvik, line 334: This seems odd - what happens with superscripts in LaTeX? Is this currently used?

asllib/aslspec/tests.t/operators.expected
Roman-Manevich: unfortunately, my pretty-printer does that. I will fix this and other pretty-printing issues in a future PR.

asllib/aslspec/tests.t/operators.spec
Roman-Manevich: Some operators, like `=`, `+`, and `if then else` are supported by the parser and then mapped to `equal`, `int_plus` and `if_then_else`.

asllib/aslspec/AST.ml
Roman-Manevich: They're not. I intend to let relations have parameters, which is actually needed to specify some relations that I am currently unable to specify. However, I want to introduce parameters gradually.

asllib/aslspec/tests.t/operators.spec
```diff
@@ -0,0 +1,38 @@
+typedef N;
+typedef Bool;
+
+operator assign[T](T, T) -> Bool
```
HadrienRenaud, line 4: Wondering how the parser make the association between `=` and `assign`

asllib/aslspec/latex.ml
```diff
@@ -38,11 +38,16 @@ let pp_one_arg_macro macro_name pp_arg fmt arg =
...
 
-(** [pp_macro fmt macro_name] renders a LaTeX macro by prefixing it with a
-    backslash. *)
+(** [pp_macro fmt macro_name] renders a LaTeX macro by making sure it starts
+    with a backslash. *)
 let pp_macro fmt macro_name =
+  let macro, macro_name =
+    if Utils.string_starts_with ~prefix:"\\" macro_name then
+      (macro_name, String.sub macro_name 1 (String.length macro_name - 1))
+    else ("\\" ^ macro_name, macro_name)
```
HadrienRenaud, line 47: Wondering about having a function
```ocaml
(** [remove_prefix_opt ~prefix s] is the part after the prefix if [s] starts with [prefix], [None] otherwise. *)
val remove_prefix_opt -> prefix:string -> string -> string
```

Then

```suggestion
    match remove_prefix_opt ~prefix:"\\" macro_name with
    | Some suffix -> (macro_name, suffix)
    | None -> ("\\" ^ macro_name, macro_name)
```

asllib/aslspec/spec.mli
Roman-Manevich: fixed. also fixed the first comment.

asllib/aslspec/latex.ml
```diff
@@ -66,6 +71,7 @@ let elem_name_to_math_macro elem_name =
   Format.asprintf "%a" pp_macro (StringOps.remove_underscores elem_name)
 
 let pp_comma = fun fmt () -> fprintf fmt ", "
+let pp_noop = fun fmt () -> fprintf fmt ""
```
HadrienRenaud:
```suggestion
let pp_noop = fun _fmt () -> ()
```

asllib/aslspec/LayoutUtils.ml
```diff
@@ -32,8 +32,16 @@ let rec contains_vertical = function
 let apply_layout_to_list layout elements =
   let args_layout =
     match layout with
-    | Horizontal l | Vertical l -> l
+    | Horizontal l | Vertical l ->
+        if List.length l == 1 && List.length elements > 1 then
+          unspecified_for_elements elements
+        else l
```
HadrienRenaud: I find this more explicit

```suggestion
    | Horizontal [ _ ] | Vertical [ _ ] when List.length elements > 1 ->
        unspecified_for_elements elements
    | Horizontal l | Vertical l ->
        l
```

asllib/aslspec/SpecParser.mly
Roman-Manevich: Right. This is not ideal, but I don't know a good way to specify operator syntax in a very flexible way. At least not one that wouldn't take a lot of time at this state. One can always opt for the `op(a1,...,an)` syntax, which always works.

asllib/doc/ASLmacros.tex
Roman-Manevich: I haven't used it yet. It might need fixing when I get to use it.

asllib/aslspec/render.ml
```diff
@@ -524,14 +500,57 @@ module Make (S : SPEC_VALUE) = struct
...
+      let is_prefix_list_operator ({ Relation.input } as def) =
+        match input with
+        | [ (_, TypeOperator { op = List0 | List1 }) ] ->
+            not (Relation.is_associative_operator def)
+        | _ -> false
+      in
+      let op_macro = get_or_gen_math_macro op_name in
+      let operator = Spec.relation_for_id S.spec op_name in
+      let { Relation.input } = operator in
+      match input with
```
HadrienRenaud: my personal style preference would be

```suggestion
      match operator.Relation.input with
```

asllib/aslspec/render.ml
```diff
@@ -524,14 +500,57 @@ module Make (S : SPEC_VALUE) = struct
...
-        operator [op] applied to [args] with [fmt] and laid out according to
-        [layout]. For example, [UNION(a, b, c)] is rendered as
-        [a \cup b \cup c]. *)
-    and pp_associative_operator op layout fmt args =
+    (** [pp_operator op_name layout fmt args] renders an operator named
+        [op_name] applied to [args] with [fmt] and laid out according to
+        [layout]. *)
+    and pp_operator op_name layout fmt args =
+      let is_prefix_list_operator ({ Relation.input } as def) =
+        match input with
```
HadrienRenaud: my personal style preference would be


```suggestion
      let is_prefix_list_operator def =
        match def.Relation.input with
```

asllib/aslspec/render.ml
```diff
@@ -524,14 +500,57 @@ module Make (S : SPEC_VALUE) = struct
...
+          let layout_arg = if vertical then "{V}" else "{H}" in
+          fprintf fmt "%a%s%a" pp_macro op_macro layout_arg
+            (PP.pp_sep_list ~sep:"" pp_arg)
+            args_with_layouts
+
+    (* and pp_associative_operator op layout fmt args =
       pp_aligned_elements_and_operators
         ~pp_sep:(fun fmt () -> pp_macro fmt (operator_to_macro_name op))
-        ~alignment:"l" pp_expr layout fmt args
+        ~alignment:"l" pp_expr layout fmt args *)
```
HadrienRenaud: Is there a good reason to keep this?

asllib/aslspec/render.ml
```diff
@@ -524,14 +500,57 @@ module Make (S : SPEC_VALUE) = struct
...
+            ~pp_sep:(fun fmt () -> pp_macro fmt op_macro)
+            ~alignment:"c" pp_expr layout fmt args
+      | [ _ ] ->
+          (* A simple unary operator. *)
+          let arg = List.hd args in
+          fprintf fmt "%a{%a}" pp_macro op_macro pp_expr (arg, layout)
+      | [ _; _ ] when not (Relation.is_custom_operator operator) ->
+          (* A simple binary operator. *)
+          let lhs_arg = List.hd args in
+          let rhs_arg = List.hd (List.tl args) in
```
```suggestion
          let lhs_arg, rhs_arg =
            match args with
            | [lhs; rhs] -> (lhs, rhs)
            | _ -> assert false
          in
```

asllib/aslspec/tests.t/operators.spec
Roman-Manevich: The `infix_expr_operator` non-terminal connects the tokens to the operator names:
```
let infix_expr_operator ==
    | COLON_EQ; { "assign" }
    | EQ; { "equal" }
    | MINUS; { "int_minus" }
    ...
```

asllib/aslspec/render.ml
Roman-Manevich: removed

asllib/aslspec/latex.ml
Roman-Manevich: I don't want to do it in this PR, but I listed this in my task list for this project.

PR #1590, opened 2025-11-25
[aslspec] default rule renders

Roman-Manevich:
For some relations, like expressions and statements, we want to render different cases of their rules separately. For example, for statements, we want to render the subset of cases for declaration statements in the section for declarations statements. However, for most relations, we want to render all of their cases at once. Currently, this requires specifying `render rule <relation-name>;`, which is tedious and noisy.
This PR automatically adds those for relations that have rules but don't already have any `render rule <relation-name>...` associated with them.

After this addition, we do not longer need `render rule annotate_literal;` and `render rule lookup_constant;`, since they are now generated automatically.

### DIFF
--- a/asllib/aslspec/AST.ml
+++ b/asllib/aslspec/AST.ml
@@ -275,7 +275,7 @@ end = struct
   let make name attributes = { name; att = Attributes.of_list attributes }
 
   let prose_description self =
-    Attributes.get_string_exn AttributeKey.Prose_Description self.att
+    Attributes.get_string_or_empty AttributeKey.Prose_Description self.att
 
   let math_macro self =
     Attributes.find_math_macro AttributeKey.Math_Macro self.att


asllib/aslspec/SpecParser.mly
```diff
@@ -374,9 +438,9 @@ let judgment_attribute := math_layout_attribute
 
 let render_rule :=
   | RENDER; RULE; relation_name=IDENTIFIER; EQ; relation=IDENTIFIER; rule_name=pared(rule_name);
-    { check_definition_name relation_name; Elem_RenderRule (RuleRender.make relation_name relation rule_name) }
+    { check_definition_name relation_name; Elem_RenderRule (RuleRender.make ~name:relation_name ~relation_name:relation rule_name) }
   | RENDER; RULE; relation_name=IDENTIFIER;
-    { check_definition_name relation_name; Elem_RenderRule (RuleRender.make relation_name relation_name []) }
+    { check_definition_name relation_name; Elem_RenderRule (RuleRender.make ~name:relation_name ~relation_name []) }
```
hrutvik: nit:
```suggestion
| RENDER; RULE; name=IDENTIFIER; EQ; relation_name=IDENTIFIER; rule_name=pared(rule_name);
    { check_definition_name name; Elem_RenderRule (RuleRender.make ~name ~relation_name rule_name) }
| RENDER; RULE; name=IDENTIFIER;
    { check_definition_name name; Elem_RenderRule (RuleRender.make ~name ~relation_name []) }
```
(i.e. rename identifiers from the production to better match the `~name` and `~relation_name` arguments)

asllib/aslspec/spec.ml
```diff
@@ -1383,6 +1431,41 @@ module Check = struct
...
+let add_default_rule_renders ast =
+  (* First, gather the relations that have rule renders. *)
+  let relations_with_rules =
+    List.fold_left
+      (fun acc elem ->
+        match elem with
+        | Elem_RenderRule { RuleRender.relation_name } ->
+            StringSet.add relation_name acc
+        | _ -> acc)
+      StringSet.empty ast
```
hrutvik, line 1459: nit: based on previous discussion, I wonder if it's better to do a `List.filter_map` followed by a `StringSet.of_list`?

asllib/aslspec/spec.ml
```diff
@@ -1383,6 +1431,41 @@ module Check = struct
...
+  in
+  (* For each relation without a rule render, add a render for all its cases,
+      with the same name as the relation. *)
+  let generated_elems =
+    List.filter_map
+      (fun elem ->
+        match elem with
+        | Elem_Relation { Relation.name; rule_opt = Some _ } ->
+            if StringSet.mem name relations_with_rules then None
+            else
```
```suggestion
        | Elem_Relation { Relation.name; rule_opt = Some _ } when not (StringSet.mem relations_with_rules) ->
```

HadrienRenaud (approved):
LGTM, modulo #1587

PR #1592, opened 2025-11-26
[aslspec] phase3 - specified some typing rules

Roman-Manevich:
- Currently, aslspec uses the first output variant of a relation to name the output configuration in the conclusion. This is fine most of the time, but when the actual configuration matches an output variant that is not the first one (like an error configuration), we need a way to disable the naming. Therefore, I've added an `auto_name` attribute that can be set to `false`. This will be resolved once aslspec performs type inference.
- Specified rules for `annotate_block` and `unop_literals`. Specifically, `not_bits` has been simplified into just one generic case (for both empty and non-empty bitvectors).
- aslspec doesn't currently support numbers so they need to be manually introduced as constants. Added `one` for 1.
- Generalized `int_negate` to `negate` so that it can operate on any number type.
- Introduced a constant for the relation `unop_signatures`. Although the expression language that aslspec supports is rich enough for such relations, aslspec does not currently associate constants with expressions (future PR).

### DIFF
--- a/asllib/doc/doclint.py
+++ b/asllib/doc/doclint.py
@@ -101,6 +101,9 @@ def check_hyperlinks_and_hypertargets(latex_files: list[str]):
     labels defined in `\hypertarget` definitions, print the mismatches
     to the console.
     """
+    # Labels to exclude from the check
+    excluded_labels: set[str] = {"constant-one"}
+
     hyperlink_labels: set[str] = set()
     hypertarget_labels: set[str] = set()
     for latex_source in latex_files:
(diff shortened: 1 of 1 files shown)


HadrienRenaud (approved):
LGTM

asllib/doc/asl.spec
```diff
@@ -2331,7 +2340,41 @@ typing function unop_literals(op: unop, l: literal) ->
...
+
+  case not_bool {
+    op = BNOT;
+    l = L_Bool(b);
+    --
+    L_Bool(not(b));
+  }
+
+  case not_bits {
+    op = NOT;
```
hrutvik, line 2372: Is this missing a premise like:
```
l = L_Bits(bits)
```
?

PR #1593, opened 2025-11-26
[gen] Restructure on module `RMW`, `Fence` and `Atom`.

ShaleXIONG:
This is a code structure rework so it reduce confusion. This is a continuation of reducing the number of files and mixed `include` between signatures.

- Move the interface definition of `RMW` into `Atom`, because the former is a part of the latter. Previous template implementations,`NoRmw` (change to `No`), `Exch` and `Lxsx` has been moved into one location, `rmw.ml` hence they will be self explained by the module name. We also make `RMW` as module when being used, rather than a simple `include` all interface as part of the signature.
- Separate the interface of `Atom` and `Fence`, where, previously, `Fence` included `Atom`. This reduces the confusion as `fence` is for `edges` while `atom` is for `annotation`.

Minor:
- We update the `debug_node` function in `cycle.ml` which prints more information.

### DIFF
--- a/gen/cycle.ml
+++ b/gen/cycle.ml
@@ -251,13 +251,16 @@ module Make (O:Config) (E:Edge.S) :
     String.concat ", " @@ List.map debug_val @@ Array.to_list v
 
   let debug_evt e =
-    sprintf "%s%s %s %s%s%s%s%s"
+    sprintf "#[%d] %s%s %s %s %s%s%s%s%s fault_check:%s value_check:%s"
+      e.idx
       (debug_dir e.dir)
       (debug_atom e.atom)
       (Code.pp_loc e.loc)
-      (match debug_vec e.cell with
-       | "" -> "" | s -> "cell=[" ^ s ^"] ")
+      ( if e.rmw then "rmw" else "" )
+      ( match debug_vec e.cell with | "" -> "" | s -> "cell=[" ^ s ^"] ")
       (debug_val e.v) (debug_tag e) (debug_morello e) (debug_vector e)
+      ( match e.check_fault with | Some (_,b) -> sprintf "%b" b | None -> "none" )
+      ( match e.check_value with | Some b -> sprintf "%b" b | None -> "none" )
 
   let debug_edge = E.pp_edge
 
@@ -1537,7 +1540,7 @@ let merge_changes n nss =
     if
       not O.allow_back &&
       List.exists proc_back nss
-    then Warn.fatal "Forbidden po vs. com" ;
+    then Warn.fatal "Forbidden po vs. com";
     if O.verbose > 1 then begin
       eprintf "SPLITTED:\n" ; debug_procs nss
     end ;


gen/common/atom.mli
```diff
@@ -25,14 +26,39 @@ module type SIMD = sig
...
-  module SIMD : SIMD
+  val pp_rmw : bool (* backward compatibility *) -> rmw -> string
+  val is_one_instruction : rmw -> bool
+  (* The first boolean indicates whether wildcard syntax is included in the fold *)
+  val fold_rmw : bool -> (rmw -> 'a -> 'a) -> 'a -> 'a
+  (* Second round of fold, for rmw with back compatible name *)
+  val fold_rmw_compat : (rmw -> 'a -> 'a) -> 'a -> 'a
+  val applies_atom_rmw : rmw -> atom option -> atom option -> bool
+  val show_rmw_reg : rmw -> bool
+  val compute_rmw : rmw  -> int (* old *) -> int (* operand *) -> int
```
fsestini: In situations like this, where a function takes multiple arguments of the same type and their meaning is not clear from context, it is better to disambiguate them using names arguments (which are enforced by the compiler) rather than comments (which aren't enforced and may drift away from the actual implementation with time).
 
```suggestion
  val compute_rmw : rmw  -> old:int -> operand:int -> int
```

gen/common/rmw.ml
```diff
@@ -46,18 +64,18 @@ module
     let expand_rmw rmw = [rmw]
   end
 
-module LxSx(A:sig type arch_atom end) = struct
+module LxSx(A:RmwType) = struct
   include Make
     (struct
-      type atom = A.arch_atom
+      type arch_atom = A.arch_atom
```
fsestini: Why `arch_atom` instead of `atom`?

gen/common/rmw.ml
```diff
@@ -14,19 +14,37 @@
 (* "http://www.cecill.info". We also give a copy in LICENSE.txt.            *)
 (****************************************************************************)
 
-(** The only RMW is exchange *)
+module type RmwType = sig
```
fsestini: I wonder if we actually need this `RmwType` sig, considering it's just one type.

gen/arch_gen.mli
```diff
@@ -14,12 +14,18 @@
 (* "http://www.cecill.info". We also give a copy in LICENSE.txt.            *)
 (****************************************************************************)
 
+module type FenceAtom = sig
```
fsestini: I think part of the objective of this PR is to make things less confusing by separating the interface of `Fence` and `Atom`, so re-introducing a new module signature which only purpose is to combine `Fence` and `Atom` seems at odds with that goal. Can we just use `Atom.S` and `Fence.S` separately?

gen/common/rmw.ml
fsestini: I wonder if we actually need this `RmwType` signature here, considering it's just a single type.

Also: why `arch_atom` and not `atom`?

gen/arch_gen.mli
fsestini: I think part of the objective of this PR is to make things less confusing by separating the interface of `Fence` and `Atom`, so re-introducing a new module signature which only purpose is to combine `Fence` and `Atom` seems at odds with that goal. Can we just use `Atom.S` and `Fence.S` separately?

gen/common/atom.mli
fsestini: In situations like this, where a function takes multiple arguments of the same type and their meaning is not clear from context, it is better to disambiguate them using names arguments (which are enforced by the compiler) rather than comments (which aren't enforced and may drift away from the actual implementation with time).
```suggestion
  val compute_rmw : rmw  -> old:int -> operand:int -> int
```

ShaleXIONG:
> @ShaleXIONG Left some comments.
> 
> From now on, if you make changes to this PR, can you please push them as their own separate commits on top of the current ones (in other words, _without_ squashing or force pushing)? Otherwise, it's very difficult to understand what has been changed in-between reviews. We can think about cleaning up the history later, once the PR is approved.

Will address the comment in separate commit for sure.

gen/common/atom.mli
ShaleXIONG: Done.

gen/common/rmw.ml
ShaleXIONG: To be honest I cannot remember too much now. There are some part of the `diy` code use `arch_atom` which means to refer those annotations/atoms that are specifically to an architecture. But I think we do not need to use different name. I will systematically review at least for the part of code I changed and use `atom` consistently.

fsestini: Thank you!

gen/common/rmw.ml
ShaleXIONG: Will change to 'atom' and remove 'RmwType'

gen/common/atom.mli
ShaleXIONG: Done

gen/common/rmw.ml
ShaleXIONG: I removed the type.

gen/arch_gen.mli
ShaleXIONG: I removed this but directly used `Fence.S` and `Atom.S` where they are used.

ShaleXIONG: I removed this but directly used Fence.S and Atom.S where they are used.

gen/atoms.ml
```diff
@@ -78,35 +78,33 @@ let () =
...
+    | `RISCV ->
+        let module M = RISCVArch_gen.Make(RISCVArch_gen.Config) in
+        (module M),(module M)
+    | `LISA ->
+        let module BellConfig = Config.ToLisa(Config) in
+        let module M = BellArch_gen.Make(BellConfig) in
+        (module M),(module M)
+    | `C -> (module CArch_gen),(module CArch_gen)
+    | `ASL -> Warn.fatal "ASL arch in atoms"
+    | `BPF -> Warn.fatal "BPF arch in atomize"
```
fsestini: Could this be a typo? (See also similar pattern match branches below)
```suggestion
    | `BPF -> Warn.fatal "BPF arch in atoms"
```

gen/common/edge.ml
```diff
@@ -31,13 +31,11 @@ module type S = sig
...
-  module SIMD : Atom.SIMD
+  include Atom.AtomType
 
-  type atom
-  module Value : Value_gen.S with type atom = atom
-  type rmw
-  type value = Value.v
   val wildcard : bool
+  type value = Value.v
+  type rmw
```
fsestini: Since `Edge.S` now includes a type `RMW.rmw` via  `include Atom.AtomType`, this additional `type rmw` declaration feels a bit redundant, especially since `rmw` is always aliased to `RMW.rmw` at every use of `Edge.S`. Maybe we can just get rid of this line and replace `rmw` with an explicit `RMW.rmw` within `Edge.S`. Will also need very small updates to `archRun.mli`, `compileCommon.ml`, and `namer.ml` accordingly (basically just following the type errors).

gen/atoms.ml
ShaleXIONG: Yes, it is a copy-paste type. Let do double check on other cases too.

gen/common/edge.ml
ShaleXIONG: Will do. Thanks for the further simplification.

PR #1594, opened 2025-11-27
[aslspec] phase4

Roman-Manevich:
This PR introduces rules for relations defined in the Section 13.8 "Basic Type Attributes".
It also includes some fixes and extensions to aslspec:
- Custom macros need a Horizontal/Vertical layout parameter, which is not correctly passed to them.
- Expression variables named the same as types were rmistakenly endered as the types rather than as expressions.
- Operators can now have a `prose_description` attribute, which is useful for documentation and perhaps later for rendering them.
- Expressions like `L(fields)` were considered illegal, since `L` was a variable rather than a type constructor/operator/relation. However, we use this pattern extensively so I'm allowing it now. The downside is that, until more checking is added, more mistakes can slip through.

### DIFF
--- a/asllib/doc/asl.spec
+++ b/asllib/doc/asl.spec
@@ -92,7 +92,18 @@ typedef def_use_name { "subprogram identifier kind" } =
 // Some of the operators below will be removed once type parameters
 // are made available to ordinary relations.
 
-// Assignment is special as it defines the left-hand side.
+operator list_from_indices[A,B](index: Identifier, list0(A), partial N -> B) -> list0(B)
+{
+  math_macro = \listfromindices
+};
+
+operator list_map[A,B](elem: A, elements: list0(A), elem_operation: partial N -> B) -> (new_elements: list0(B))
+{
+  "forms a new list where for each {elem} of {elements}, {new_elements} has an element obtained
+   by applying {elem_operation} to {elem}",
+  math_macro = \listmap
+};
+
 operator assign[T](T, T) -> Bool
 {
   math_macro = \eqdef,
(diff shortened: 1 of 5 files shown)


asllib/doc/asl.spec
```diff
@@ -92,7 +92,18 @@ typedef def_use_name { "subprogram identifier kind" } =
 // Some of the operators below will be removed once type parameters
 // are made available to ordinary relations.
 
-// Assignment is special as it defines the left-hand side.
+operator list_from_indices[A,B](index: Identifier, list0(A), partial N -> B) -> list0(B)
```
hrutvik, line 95: nit: the type here is a bit odd - I guess `Identifier` as there are no numbers yet, but why `N`? Similarly for `list_map` below

Roman-Manevich: `Identifier` is the name of the index variable.
Numbers exist - the order of definitions does not affect meaning.
The `partial N -> B` stands for an expression that takes an index value and returns a new element.
Think of it as an OCaml closure `fun index -> ...`.
This will slightly change in a future PR for directly supporting expressions as closures.

hrutvik: Ah perhaps I am getting confused here between types and non-types - I would have assumed something like `index : natural` or similar, and `partial A -> B` (rather than `N -> B`) to match the `list0(A)`?

Roman-Manevich: `index`  is a variable local to the operator, it's like `fun index -> ...`. Operators are a bit meta in that they allow constructing quite arbitrary compound expressions so the semantics is different that that of functions/relations.
I intend to document this and implement a proper checker in a future PR.

PR #1596, opened 2025-11-27
[action] Do not check code formatting as part of diy test action

maranget:
PR #1552 introduced  separated CI actions for  testing **diy**, which is a good thing.

The new action file  checks the formatting of code, as does the original one. As a result, code formatting is checked twice by the CI in some situations, typically when `lib/` is modified.

This PR is a naive attenpt to avoid this duplication of work,


ShaleXIONG:
@maranget Thanks for the quick fix. I am happy on this.

Issue #1549, opened 2025-11-05
[diyone7] Make `-norm` print to stdout by default, and use `-o` to save to file


fsestini:
`diyone7` takes an optional flag `-norm`, which has the effect of producing _no_ output on stdout, but instead saving a copy of the generated litmus test to disk with a normalised name.

I think it would be nicer to have `-norm` just print the test with normalised name to stdout, as:

- it would align it to what `diyone7` does by default if no `-norm` flag is set
- it would align it to what passing the option `-name XYZ` does
- there is no way, AFAICT, to control where the litmus test is saved on disk. It is always saved on the current working directory

With `-norm` printing the test to stdout by default, we could then use an option like `-o` to specify a path to which to save the file. Under that setup, the current behaviour of `-norm` could then be recovered by passing `-norm -o .` to the CLI.

Name normalisation and the choice of stdout vs file seem to me to be orthogonal axes, and all combinations of them are useful. So IMO it makes sense to have them under two different options that can be mixed and matched.

PS: Note that currently `-o` is actually accepted by `diyone7` but seems to have no effect.

CC @ShaleXIONG

PR #1597, opened 2025-11-27
[gen] Wire in `-o` and `-stdout` argument in `diyone7`.

ShaleXIONG:
This PR wires in`-o` and `-stdout` arguments in `diyone7`, which maintains the same behaviour as in `diy7`. (ref #1549)

- `-stdout true` redirects output to `stdout` rather than file, when `-norm` is on or `-name {test_name}` is specified. 
- `-o` will have the expected behaviour as documented in https://diy.inria.fr/doc/gen.html#sec72. That is,
```
Redirect output to <dest>. This option applies when tools generate 
a set of tests and an index file @all, .i.e. in all situations except 
for diyone7 simplest operating mode.

If argument <dest> is an archive (extension .tar) or a compressed 
archive (extension .tgz), the tool builds an archive. Otherwise, 
<dest> is interpreted as the name of an existing directory. 
Default is “.”, that is tool output goes into the current directory.
```

Example:
- `diyone7 -arch AArch64  PodWR Fre PodWR Fre -norm` generates a normalised test and save to a normalised file name under the current directory. This preserves the existing behaviour. Similar for `-name specific_file_name`.
- `diyone7 -arch AArch64  PodWR Fre PodWR Fre -norm -o dir` assuming `dir` is a directory, the generated file will be put under the `dir`.
- `diyone7 -arch AArch64  PodWR Fre PodWR Fre -norm -o pack.tar` the generated file will be wrap into `pack.tar`
- `diyone7 -arch AArch64  PodWR Fre PodWR Fre -norm -stdout`  generates a normalised test yet it will print to `stdout` rather than the file system.

Minor:
- A minor code simplification on parsing `es` in `do_zyva` in `diyone7`.

### DIFF
--- a/gen/common/config.ml
+++ b/gen/common/config.ml
@@ -271,18 +271,18 @@ let common_specs () =
 let name_spec is_diyone =
   if is_diyone then
    ("-name",Arg.String (fun s -> name := Some s),
-     "<s> specify the name of test and output to file (`-stdout false`). Note `-norm` overrides this flag.")
+     "<s> specify the name of the test. Outputs to file by default, unless `-stdout` is set. Note: `-norm` overrides this flag.")
   else
    ("-name",Arg.String (fun s -> name := Some s),
      "<s> specify base name of tests.")
 
 let stdout_spec is_diyone =
   if is_diyone then
-    ("-stdout", Arg.Bool (fun b ->  stdout := b),
-     "output to stdout, when `-name` and `-norm` is specified. It overrides `-o`. (default true)")
+    ("-stdout", Arg.Set stdout,
+     "when `-name` or `-norm` is set, control whether output is written to stdout, error otherwise.")
   else
-    ("-stdout", Arg.Bool (fun b ->  stdout := b),
-     "output to stdout. It overrides `-o`. If `-cycleonly` is true, then this is implied to be true. (default false)")
+    ("-stdout", Arg.Set stdout,
+     "output to stdout. It overrides `-o`. If `-cycleonly` is true, then this is implied to be true.")
 
 let varatomspec =
   ("-varatom", Arg.String (fun s -> varatom := !varatom @ [s]),
@@ -294,7 +294,7 @@ let diyone_spec () =
   common_specs () @
   name_spec true
   :: stdout_spec true
-  :: ("-norm",Arg.Set norm, "a normalised name for test name and output to file (`-stdout false`)")
+  :: ("-norm",Arg.Set norm, "find a normalised name for the test. Outputs to file by default, unless `-stdout` is set.")
   :: []
 
 let diycross_spec () =
@@ -356,6 +356,13 @@ let diy_spec () =
    varatomspec ::
    []
 
+let valid_stdout_flag is_diyone =
+  if !stdout &&
+  (* it should only be used with `-norm` or `-name`, and no `-o` in `diyone7` *)
+    ( ( is_diyone && ( (!name = None && not !norm) || !tarfile <> None ) )
+  (* it should not be used with `-o` and `-cycleonly` in `diy7` and `diycross7`*)
+      || ( !cycleonly || !tarfile <> None ) )
+  then Warn.user_error "-stdout will be ignored."
 
 let prog = if Array.length Sys.argv > 0 then Sys.argv.(0) else "XXX"
 let baseprog = sprintf "%s (version %s)" (Filename.basename prog) (Version.version)
--- a/gen/diy.ml
+++ b/gen/diy.ml
@@ -203,6 +203,7 @@ let () =
   | None -> ()
   | Some s -> exec_conf s
   end;
+  Config.valid_stdout_flag false ;
   let relax_list = split_cands !Config.relaxs
   and safe_list = split_cands !Config.safes
   and reject_list = split !Config.rejects in
--- a/gen/diycross.ml
+++ b/gen/diycross.ml
@@ -84,7 +84,8 @@ let pp_es = ref []
 let () =
   Util.parse_cmdline
     (Config.diycross_spec ())
-    (fun x -> pp_es := x :: !pp_es)
+    (fun x -> pp_es := x :: !pp_es);
+  Config.valid_stdout_flag false
 
 let pp_es = List.rev !pp_es
 
--- a/gen/diyone.ml
+++ b/gen/diyone.ml
@@ -180,7 +180,8 @@ let pp_es = ref []
 let () =
   Util.parse_cmdline
     (Config.diyone_spec ())
-    (fun x -> pp_es := x :: !pp_es)
+    (fun x -> pp_es := x :: !pp_es) ;
+    Config.valid_stdout_flag true
 
 let pp_es = List.rev !pp_es
 let cpp = match !Config.arch with


maranget:
> > Hi @ShaleXIONG. I agree with separating the functionality of `-norm` and `-o` (before `-norm` triggered systematic creation of a test file). Interface is clearer now: `-norm` for computing the name `-o` for saving.
> > I am less convinced by this change:
> > > * `-norm` and `-name` now always print to `stdout` alongside to the file system.
> > 
> > 
> > I do not see the reason behind this change. In other words, why is it useful to have both output systematically ?
> 
> This is more for @fsestini request, as he wants `-norm` print to both file and `stdout` in a offline chat. So there are several approaches we can do:
> 
> * Add extra behaviour: always print to `stdout` alongside whether to print to files, which we pick. Separately, `-o` behaviour here is to specify an optional `directory` of which contains the generated file.
> * Use `-o` to trigger print to file but here `-o` will be the file name. However this will remove a behaviour in `-norm`, namely it will print to `stdout` by default now.
> * Potentially introduce a flag `-stdout` so to indicate print to `stdout` in extra.

I have a slight preference for the last suggestion. I have no strong opinion other than avoiding the double output as the default behaviour.

gen/common/config.ml
```diff
@@ -251,7 +251,7 @@ let common_specs () =
    sprintf "<bool> complete test name with number when identical (default %b)"
      !addnum)::
    ("-name",Arg.String (fun s -> name := Some s),
-     "<s> specify base name of tests")::
+     "<s> specify base name of tests. Note `-norm` overwrites this flag.")::
```
fsestini:
```suggestion
     "<s> specify base name of tests. Note `-norm` overrides this flag.")::
```

fsestini (commented):
A couple of nitpicks and one slightly more pressing concern.

- There's a typo in the description of the `-o` flag ("curent"). (I know this wasn't part of this PR, but might as well fix it here since it's relevant)
- Can we please also add a note to the description of `-norm` saying something along the lines of "this flag overrides `-name`"?
- When both `-o` and `-stdout true` are set, the test is not saved to disk. Perhaps that should happen (effectively treating `-o` and `-stdout` as orthogonal flags)? If setting them both is not intended behaviour, I think we should then throw an error in this case, rather than silently ignoring one of the two. Thoughts?

My slightly major concern is that the description of the `-stdout` flag, displayed when the user does `diyone7 -help`, reads:
> -stdout output to stdout. If Cycleonly is true, then this is implied to be true. (default false)

For `diyone7`, this is not accurate, since the default is actually to print to stdout unless `-norm` or `-o` is set. I think the description of the flag should reflect this.

I understand that technically this a consequence of `diy7` and `diyone7` sharing some of the config fields internally, hence the description mentioning "default false" (which is indeed the case for `diy7`) and "cycleonly" (which is only valid for `diy7` but displayed by `diyone7 -help` nonetheless). Not on this PR, but in the future I think we should move towards separating the configs for these two tools a bit better. Perhaps what we _can_ do on this PR, however, is to revisit how the `-stdout` flag is wired up so that its description is different for `diy7` and `diyone7`, and thus accurate for both.

maranget (commented):
Code is fine. 

The question of what to do when the options `-o` and `-stdout` are both specified is still pending.

gen/common/config.ml
ShaleXIONG: done

fsestini:
Thanks @ShaleXIONG , this is very useful.

One issue I see with the `stdout` option as currently implemented is that, for example, `diyone7 -arch AArch64 PodWR Fre PodWR Fre -stdout false` will surprisingly print its results to stdout even though `-stdout false` is set. I think that for `diyone`, setting `-stdout false` without also setting `-o` should be an error, or at least a warning.

Moreover, `diyone7`'s help message says that `stdout` is `false` by default, but it's actually `true` by default.

Perhaps this means we should modify `gen/common/config.ml` to define a dedicated `stdout` option for `diyone`, instead of defining a single `stdout` option shared among all `diy*` tools.

gen/common/config.ml
```diff
@@ -260,18 +262,56 @@ let common_specs () =
...
     "output only cycle, i.e. no litmus body (default false)")::
+  ("-num", Arg.Bool (fun b -> numeric := b),
+   sprintf "<bool> use numeric names (default %b)" !numeric)::
   []
 
-let numeric = ref true
+let name_spec is_diyone =
+  if is_diyone then
+   ("-name",Arg.String (fun s -> name := Some s),
+     "<s> specify the name of test and output to file (`-stdout false`). Note `-norm` overrides this flag.")
```
fsestini: From the message, it sounds like `-name` will always output to file.
If this description is only intended to be true when `-stdout false`, then there should also be a description for the `-stdout true` case.

As for the text `(-stdout false)`: by itself, I personally find it a big ambiguous, as it could suggest various scenarios:
* `-stdout false` is a precondition for `-name` to take effect
* setting `-name` _strongly_ implies `-stdout false`, making `-stdout true` incompatible
* setting `-name` implies `-stdout false` _unless_ explicitly set to `true` by the user

Perhaps we could adjust the text slightly so that it's more clear we mean the 3rd interpretation and _only_ that one:
```suggestion
     "<s> specify the name of the test. Outputs to file by default, unless `-stdout true` is set. Note: `-norm` overrides this flag.")
```

gen/common/config.ml
```diff
@@ -260,18 +262,56 @@ let common_specs () =
...
-let speclist () =
+let diyone_spec () =
+  nprocs := 1000 ;
+  numeric := true ;
   common_specs () @
-  ("-num", Arg.Bool (fun b -> numeric := b),
-   sprintf "<bool> use numeric names (default %b)" !numeric)::
+  name_spec true
+  :: stdout_spec true
+  :: ("-norm",Arg.Set norm, "a normalised name for test name and output to file (`-stdout false`)")
```
fsestini: See my comment about `-stdout`.

```suggestion
  :: ("-norm",Arg.Set norm, "find a normalised name for the test. Outputs to file by default, unless `-stdout true` is set.")
```

gen/common/config.ml
```diff
@@ -260,18 +262,56 @@ let common_specs () =
...
+   ("-name",Arg.String (fun s -> name := Some s),
+     "<s> specify the name of test and output to file (`-stdout false`). Note `-norm` overrides this flag.")
+  else
+   ("-name",Arg.String (fun s -> name := Some s),
+     "<s> specify base name of tests.")
+
+let stdout_spec is_diyone =
+  if is_diyone then
+    ("-stdout", Arg.Bool (fun b ->  stdout := b),
+     "output to stdout, when `-name` and `-norm` is specified. It overrides `-o`. (default true)")
```
fsestini: I get what you're trying to say here, however personally I find it unclear to simultaneously state:
* `-stdout` is only effective when `-name` or `-norm` are set
* `-stdout` is true by default

That is because _if_ we limit ourselves to the case where `-norm` or `-name` is set, then the default value of `-stdout` is actually `false`. Moreover, the user can set `-stdout` when running `diyone7` without neither `-norm` nor `-name` and no errors/warnings will be raised, which seems to suggest that `-stdout` is indeed admitted outside of `-norm`/`-name` (albeit not always with the effect that one might expect).

I was assuming we would advertise `-stdout` as a global flag (my suggestion that `-stdout` is `true` by default rests on that assumption). I'm fine with properly restricting it to `-name`/`-norm`, but in that case I would perhaps phrase the help message slightly differently:
```suggestion
     "when `-name` or `-norm` is set, control whether output is written to stdout (ignored otherwise). It overrides `-o`.")
```
As per my earlier comment, just to make things very clear, I would also print out a _warning_ to the CLI whenever `-stdout` is used inappropriately, telling the user that the flag has been ignored.

ShaleXIONG:
> Sorry for the back and forth @ShaleXIONG , but I still find some of the `-help` messages a bit confusing. I've left more specific comments in the code itself.
> 
> I understand that documentation is tricky, and we don't _have_ to get it perfect right now. But since this PR is not super urgent for now, maybe we can think about this a little bit longer.
> 
> As a future PR, it might also be worth taking up on @maranget 's earlier suggestion of using `-o -` instead of `-stdout`, at least for `diyone7`. i think that would remove at the roots some of the issues we are facing currently with incompatible combinations of options.

I think many flags and arguments need some tide up probably similar to what have been done to `herd7`. I will spend some extra time to investigate this after this quick patch.

gen/common/config.ml
ShaleXIONG: I guess the best way here is to make `-stdout` a flag rather than a boolean argument. In this case you only have two possible inputs here:
- `-name a` prints to file of `a.litmus` 
- `-name a -stdout` prints to `stdout`.

gen/common/config.ml
ShaleXIONG: Similarly the less confusing way is to make `-stdout` a flag rather than a boolean argument.

gen/common/config.ml
ShaleXIONG: I just updated to refuse generating anything but output a warning in `diyone7` if `-stdout` does not pair with `-name` or `-norm`.

PR #1599, opened 2025-11-27
[catalogue,aarch64] Specify PE for TFSR_ELx register

murzinv:
Otherwise, it is treated as locatation...


diaolo01:
@relokin @murzinv I will investigate whether a warning can be introduced and whether the tests make sense.

catalogue/aarch64-MTE/tests/asorted-async/S+dmb.sttp+data_asynctagfault.litmus
relokin, line 1: Can we please remove this test. As Artem suggested this is a duplicate of `catalogue/aarch64-MTE/tests/to-tagcheck-async/S+dmb.sttp+data--async.litmus"

PR #1601, opened 2025-11-28
[asl,herd] Introduce primitive_to_ml

HadrienRenaud:
This PR introduces a way of checking the signature of primitive functions.

For a given file `primitives.asl`:
```
func Foo(first_argument: integer, second_argument: integer) => integer
begin return 0; end; // unused implementation
```
the newly introduced tool `primitive_to_ml` will produce a `primitives.ml` file with a functor that accept a input module of the form:
```ocaml
module type I = sig
  type ii (* state *)
  type v
  type 'a m

  val foo : ii (* state *) -> first_argument: v m -> second_argument: v m -> v m
end
```
and produces a output module of the form:
```ocaml
module type O = sig
  val foo : _ (* state *) -> v m list (* parameters *) -> v m list (* arguments *) -> v m list (* output *)
end
```

The output function does the arity check and the argument matching for free.

For the moment, it only supports single output functions, i.e. functions that don't return tuples.

For the moment, it is only used in `herd`, but it should be possible to do this in `asllib` too.

---

We take the opportunity to document the primitives used by herd in `herd/ASLSemPrimitives.asl`.

### DIFF
--- a/herd/AArch64ASLSem.ml
+++ b/herd/AArch64ASLSem.ml
@@ -1055,36 +1055,27 @@ module Make (TopConf : AArch64Sig.Config) (V : Value.AArch64ASL) :
     let pstate_default_fields =
       let open Asllib.AST in
       lazy
-        (let proc_state_decl =
-           C.libfind "asl-pseudocode/patches.asl"
-           |> ASLBase.build_ast_from_file ~ast_type:`Ast `ASLv1
-           |> List.find (fun d ->
-                  match d.desc with
-                  | D_GlobalStorage { keyword = GDK_Var; name = "PSTATE"; ty = Some _ ; _ } -> true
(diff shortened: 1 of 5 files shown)


maranget:
> the newly introduced tool `primitive_to_ml` will produce a `primitives.ml` file with a functor that accept a input module of the form:
> 
> ```ocaml
> module type I = sig
>   type v
>   type 'a m
> 
>   val foo : _ (* state *) -> first_argument: v m -> second_argument: v m -> v m
> end
> ```

Nitpicking the type of state `ii` is part of the input signature:
```ocaml
module type I = sig
  type ii
  type v
  type 'a m
 
  val foo : ii  (* state *) -> first_argument: v m -> second_argument: v m -> v m
end
 ```

PR #1603, opened 2025-11-29
[aslspec] enhancements + eval_stmt

Roman-Manevich:
- Refactored the grammar to allow transitions to be nested only inside `INDEX` constructs, and `INDEX` constructs to appear only as judgments. This avoids things like `f(a) -> g(b) -> x`. Functions/relations call can be nested in expressions but without the `->` notation, which means only ones without short-circuiting alternatives should be used that way (will be checked in a future PR).
- The grammar used to allow only a single short-circuit expressions, and it now supports a list:
`eval_stmt(env1, s2) -> Continuing(g2, new_env) | ThrowingConfig(), DynErrorConfig(), DivergingConfig()`.
- Generalized list indexing to support expressions as indices, e.g., `envs[i + one]`.
- Fixed a bug that confused a rule name like `aa.bb` as a prefix of `aa.bbb`. The notion of a prefix now splits case names along `.` and compares prefixes with respect to the list of names.
- Added support for zero-argument operators, which are needed, for example, to express short-circuit configurations like `ThrowingConfig()`.
- Specified the rules for `eval_stmt`. This required introduced various operators for: short-circuit configurations, forall/exist over  lists, `fresh_identifier`,  operators to generate read/write effects, operators to generate native Booleans/integers/strings.
- A bug fix to the definition of `Throwing`
- Added the error code for dynamic assertion failure (`DE_DAF`), which was missing from the spec.

### DIFF
--- a/asllib/doc/asl.spec
+++ b/asllib/doc/asl.spec
@@ -1564,6 +1564,16 @@ render xgraphs_and_components = XGraphs(-), Nodes(-), Labels(-), effect_type(-);
 
 constant empty_graph { "empty execution graph", math_macro = \emptygraph };
 
+constant return_var_prefix
+{
+  math_macro = \returnvarprefix,
+};
+
+operator string_of_nat(n: Z) -> Strings
+{
+  math_macro = \stringofnat,
+};
+
 ast symdom { "\symbolicdomain{}" } =
     | Finite(powerset_finite(Z))
     { "symbolic finite set integer domain" }
@@ -4676,16 +4686,22 @@ semantics relation eval_stmt(env: envs, s: stmt) ->
     s =: S_Seq(s1, s2);
     eval_stmt(env, s1) -> Continuing(g1, env1);
     case continuing {
-      eval_stmt(env1, s2) -> Continuing(g2, new_env) | ThrowingConfig(), DynErrorConfig(), DivergingConfig();
+      eval_stmt(env1, s2) -> Continuing(g2, new_env) | DynErrorConfig(), DivergingConfig();
       new_g := ordered_po(g1, g2);
       --
       Continuing(new_g, new_env);
     }
     case returning {
-      eval_stmt(env1, s2) -> Returning((vs, g2), new_env) | ThrowingConfig(), DynErrorConfig(), DivergingConfig();
+      eval_stmt(env1, s2) -> Returning((vs, g2), new_env) | DynErrorConfig(), DivergingConfig();
       new_g := ordered_po(g1, g2);
       --
       Returning((vs, new_g), new_env);
+    }
+    case throwing {
+      eval_stmt(env1, s2) -> Throwing(v, t, g2, new_env) | DynErrorConfig(), DivergingConfig();
+      new_g := ordered_po(g1, g2);
+      --
+      Throwing(v, t, new_g, new_env);
     }
   }
 
@@ -4703,16 +4719,22 @@ semantics relation eval_stmt(env: envs, s: stmt) ->
     v =: nvbool(b);
     s' := if b then s1 else s2;
     case continuing {
-      eval_block(env1, s') -> Continuing(g2, new_env) | ThrowingConfig(), DynErrorConfig(), DivergingConfig();
+      eval_block(env1, s') -> Continuing(g2, new_env) | DynErrorConfig(), DivergingConfig();
       new_g := ordered_ctrl(g1, g2);
       --
       Continuing(new_g, new_env);
     }
     case returning {
-      eval_block(env1, s') -> Returning((vs, g2), new_env) | ThrowingConfig(), DynErrorConfig(), DivergingConfig();
+      eval_block(env1, s') -> Returning((vs, g2), new_env) | DynErrorConfig(), DivergingConfig();
       new_g := ordered_ctrl(g1, g2);
       --
       Returning((vs, new_g), new_env);
+    }
+    case throwing {
+      eval_block(env1, s') -> Throwing(v, t, g2, new_env) | DynErrorConfig(), DivergingConfig();
+      new_g := ordered_ctrl(g1, g2);
+      --
+      Throwing(v, t, new_g, new_env);
     }
   }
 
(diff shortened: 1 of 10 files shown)


asllib/aslspec/SpecParser.mly
```diff
@@ -398,11 +421,7 @@ let expr :=
     | list_var=IDENTIFIER; LBRACKET; index=IDENTIFIER; RBRACKET;
       { Rule.make_list_index list_var index }
     | label_opt=ioption(IDENTIFIER); LBRACKET; fields=tclist1(field_and_value); RBRACKET;
-      { Rule.make_record label_opt fields }
-    | lhs=expr; ARROW; rhs=expr; ~=short_circuit;
-      { Rule.Transition { lhs; rhs; short_circuit } }
-    | INDEX; LPAR; index=IDENTIFIER; COMMA; list=IDENTIFIER; COLON; body=expr; RPAR;
-      { Rule.Indexed { index; list; body } }
+      { Rule.make_record label_opt fields }    | lhs=expr; ARROW; rhs=expr; ~=short_circuit;
```
HadrienRenaud: Misplaced line ending?

asllib/doc/asl.spec
```diff
@@ -4569,7 +4632,275 @@ semantics relation eval_stmt(env: envs, s: stmt) ->
...
+    Continuing(new_g, new_env);
+  }
+
+  case SCond {
+    s =: S_Cond(e, s1, s2);
+    eval_expr(env, e) -> ResultExpr((v, g1), env1);
+    v =: nvbool(b);
+    s' := if b then s1 else s2;
+    case continuing {
+      eval_block(env1, s') -> Continuing(g2, new_env) | ThrowingConfig(), DynErrorConfig(), DivergingConfig();
```
HadrienRenaud: I'm surprised that `ThrowingConfig` doesn't also get a `ordered_ctrl`

asllib/doc/asl.spec
```diff
@@ -4569,7 +4632,275 @@ semantics relation eval_stmt(env: envs, s: stmt) ->
...
+    }
+    case returning {
+      eval_block(env, s1) -> Returning((vs, g), new_env) | DynErrorConfig(), DivergingConfig();
+      eval_catchers(env, catchers, otherwise_opt, Returning((vs, g), new_env)) -> C | ;
+    }
+    case throwing {
+      eval_block(env, s1) -> Throwing(v, t, g, new_env) | DynErrorConfig(), DivergingConfig();
+      eval_catchers(env, catchers, otherwise_opt, Throwing(v, t, g, new_env)) -> C | ;
+    }
+    eval_catchers(env, catchers, otherwise_opt, sm) -> C | ;
```
HadrienRenaud: This is confusing. why is `eval_catchers` called multiple times?

asllib/doc/asl.spec
```diff
@@ -4569,7 +4632,275 @@ semantics relation eval_stmt(env: envs, s: stmt) ->
...
+      Returning((empty_list, empty_graph), env);
+    }
+    case one {
+      s =: S_Return(some(e));
+      eval_expr(env, e) -> ResultExpr((v, g1), new_env);
+      wid := fresh_identifier();
+      write_identifier(wid, v) -> g2;
+      new_g := ordered_data(g1, g2);
+      --
+      Returning((make_list(v), new_g), new_env);
```
HadrienRenaud, line 4885: Aren't you suppose to return `wid`? otherwise what's the point?

asllib/aslspec/SpecParser.mly
Roman-Manevich: Yes, but it was fixed in a later commit.

asllib/doc/asl.spec
Roman-Manevich: Fixed by a later commit.

asllib/doc/asl.spec
Roman-Manevich: The last one is a mistake. Removed. Thanks for catching.

asllib/doc/asl.spec
Roman-Manevich: It's inside `g2` now, so it is getting returned.

HadrienRenaud: In `asllib/Interpreter.ml`, it is a specific identifier that is built for this, see function `return_identifier`. The goal is to be able to reconstruct it back in `eval_subprogram`

asllib/doc/asl.spec
```diff
@@ -1452,11 +1513,12 @@ typedef envs
 ////////////////////////////////////////////////////////////////////////////////
 // Concurrent Execution Graphs
 
-constant Read
-{ "read effect", math_macro = \Read };
-
-constant Write
-{ "write effect", math_macro = \Write };
+typedef effect_type =
```
hrutvik, line 1516: What's the relation between this and the `operator`s `ReadEffect`/`WriteEffect`?

asllib/doc/asl.spec
```diff
@@ -4569,7 +4632,300 @@ semantics relation eval_stmt(env: envs, s: stmt) ->
...
+    Continuing(new_g, new_env);
+  }
+
+  case SThrow {
+    s =: S_Throw(e, t);
+    eval_expr(env, e) -> ResultExpr((v, g1), new_env);
+    name := fresh_identifier();
+    g2 := WriteEffect(name);
+    new_g := ordered_data(g1, g2);
+    ex := (v, name_fresh);
```
hrutvik: I notice this used to be `value_read_from(v, name)` - so should this be `name` instead of `name_fresh`? And is the deletion of `value_read_from` intentional?

asllib/doc/asl.spec
```diff
@@ -4569,7 +4632,300 @@ semantics relation eval_stmt(env: envs, s: stmt) ->
...
+    --
+    Throwing(ex, t, new_g, new_env);
+  }
+
+  case STry {
```
hrutvik, line 4854: Did these need splitting up, or could they still have a variable for the configuration produced by `eval_block` and be streamlined into a single case?

asllib/doc/asl.spec
```diff
@@ -4569,7 +4632,300 @@ semantics relation eval_stmt(env: envs, s: stmt) ->
...
+      Returning((vs, new_g), new_env);
+    }
+  }
+
+ case SPrint {
+   case print {
+     s =: S_Print(elist, False);
+     eval_expr_list(env, e_list) -> ResultExprList((v_list, g), env_1);
+     envs[one] = env;
+     INDEX(i, list: output_to_console(envs[i], v_list[i]) -> envs[i + one]);
```
hrutvik: Should this be:
```suggestion
     envs[one] = env_1;
     INDEX(i, v_list: output_to_console(envs[i], v_list[i]) -> envs[i + one]);
```
?
I also wonder if you could just have `envs[one]` returned from `eval_expr_list`, avoiding the extra indirection?

asllib/aslspec/render.ml
```diff
@@ -536,7 +540,15 @@ module Make (S : SPEC_VALUE) = struct
...
-            match args with [ lhs; rhs ] -> (lhs, rhs) | _ -> assert false
+            match args with
+            | [ lhs; rhs ] -> (lhs, rhs)
+            | _ ->
+                let msg =
+                  Format.asprintf
+                    "Expected exactly two arguments for binary operator %s"
+                    op_name
+                in
+                failwith msg
```
HadrienRenaud, line 551: Is there not an arity check somewhere that would catch this earlier? It is strange that this is the first time that this is encountered.

asllib/doc/asl.spec
HadrienRenaud: Once again this is strange that this is not caught statically by the checks.

HadrienRenaud (approved):
In general, after fixes to @hrutvik's comment, I think this is good

asllib/aslspec/render.ml
Roman-Manevich: There are no static checks yet (future PR).

asllib/doc/asl.spec
Roman-Manevich: The operators use these labels:
image

asllib/doc/asl.spec
Roman-Manevich: Right, this should be `name`. I mistakenly used `value_read_from`, which is a type name for a pair, as a constructor. So this is a fix.
This isn't caught statically, since there is no static check yet :-)

asllib/doc/asl.spec
Roman-Manevich: Unfortunately, they do need splitting up. Putting a single variable `C` for any output configuration means that we attempt to manipulate graphs for dynamic error configurations and diverging configurations, which don't have graphs.

asllib/doc/asl.spec
Roman-Manevich: discussed offline and fixed accordingly.

PR #1639, opened 2025-12-16
[all] Add LDAP/STLP instructions

RaghavRoy145:
Adds support for LDAP (Load-Acquire Pair) and STLP (Store-Release Pair) instructions to herd.

#### Summary

LDAP Xt1, Xt2, [Xn]:  Load-Acquire Pair. Both loads take Acquire ordering, subsequent memory accesses on the same thread cannot be reordered before the pair.
STLP Xt1, Xt2, [Xn]: Store-Release Pair. The pair takes Release ordering, earlier memory accesses cannot be reordered after it.

The existing `pair_opt = Pa | PaN | PaI` is refactored into polymorphic variants split by direction:  `ld_pair_opt` gains \`PaA (LDAP) and `st_pair_opt` gains \`PaL (STLP), while the old \`PaI is split into \`PaIQ (load side) and \`PaIL (store side). Polymorphic variants let load-only and store-only tags be enforced at the instruction level on I_LDP / I_STP without duplicating constructors like `Pa` and `PaN`, which are used on both the store and load directions.

#### Test Added:
`Y006.litmus` is a message-passing test with STLP as writer and LDAP as reader:

If the Release on the STLP and the Acquire on the LDAP compose correctly, then observing the second element of the pair (1:X3=1) must imply observing the preceding store, so 1:X5=0 should be impossible.

### DIFF
--- a/gen/AArch64Compile_gen.ml
+++ b/gen/AArch64Compile_gen.ml
@@ -1461,8 +1461,8 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
       let load ar (r1,r2) rA =
         let a =
           match ar with
-          | Pair (Pa,_),None -> XP
-          | Pair (PaI,_),None -> AXP
+          | Pair ((`Pa),_),None -> XP
+          | Pair ((`PaIQ),_),None -> AXP
           | _ ->
              Warn.fatal
                "Illegal %s annotaton on load exclusive pair" (pp_atom ar)  in
@@ -1476,8 +1476,8 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
       let store aw r (r1,r2) rA =
         let a =
           match aw with
-          | Pair (Pa,_),_ -> YY
-          | Pair (PaI,_),_ -> LY
+          | Pair ((`Pa),_),_ -> YY
+          | Pair ((`PaIL),_),_ -> LY
           | _ ->
              Warn.fatal
                "Illegal %s annotaton on store exclusive pair" (pp_atom aw)  in
@@ -1695,7 +1695,7 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
          | _ -> LDN.emit_load n
        in
        emit_load
-    | Code.Pair -> emit_ldp Pa UnspecLoc
+    | Code.Pair -> emit_ldp (`Pa) UnspecLoc
 
 
     let emit_obs_not_value = OBS.emit_load_not_value
@@ -1858,8 +1858,12 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
             Some r,init,cs,st
         | R,Some (Neon _,Some _) -> assert false
         | R,Some (Pair (opt,idx),None) ->
-          let r,init,cs,st = emit_ldp opt idx st p init loc in
-          Some r,init,cs,st
+          let ld_opt = match opt with
+            | `Pa -> `Pa | `PaN -> `PaN | `PaIQ -> `PaIQ
+            | `PaIL -> assert false
+          in
+    let r,init,cs,st = emit_ldp ld_opt idx st p init loc in
+    Some r,init,cs,st
         | R,Some (Pair _,Some _) -> assert false
         | W,None ->
             let init,cs,st =
@@ -1899,8 +1903,12 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
             let init,cs,st = STG.emit_store st p init e in
             None,init,cs,st
         | W,Some (Pair (opt,idx),None) ->
-            let init,cs,st = emit_stp opt idx st p init loc e in
-            None,init,cs,st
+            let st_opt = match opt with
+              | `Pa -> `Pa | `PaN -> `PaN | `PaIL -> `PaIL
+              | `PaIQ -> assert false
+            in
+    let init,cs,st = emit_stp st_opt idx st p init loc e in
+    None,init,cs,st
         | W,Some (Pair _,Some _) -> assert false
         | (R|W), Some (Instr, _) -> Warn.fatal "Instr annotation did not create code location %s" (C.debug_evt e)
         | R,Some (Pte (Read|ReadAcq|ReadAcqPc as rk),None) ->
@@ -2356,7 +2364,7 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
               Some rB,init,pseudo cs0@cs,st
           | R,Some (Pair (opt,idx),None) ->
               let r,init,cs,st =
-                emit_ldp_idx_var opt idx vdep st p init loc r2 in
+                emit_ldp_idx_var (pair_opt_to_ld opt) idx vdep st p init loc r2 in
               Some r,init, pseudo cs0@cs,st
           | R,Some ((Neon _|Pair _),Some _) -> assert false
           | W,None ->
@@ -2439,9 +2447,9 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
               None,init,pseudo cs0@cs,st
           | W,Some (Pair (opt,idx),None) ->
               let init,cs,st =
-                emit_stp_idx_var opt idx vdep st p init loc e r2 in
+                emit_stp_idx_var (pair_opt_to_st opt) idx vdep st p init loc e r2 in
               None,init, pseudo cs0@cs,st
-          | W,Some (Pair _,_) -> assert false
+          | W,Some (Pair _,Some _) -> assert false
           | (W,(Some (Pte (Set _),None))) ->
               let init,cs,st =
                 emit_set_pteval_idx false vdep r2 st p init (Value.to_pte e.C.v) (Misc.add_pte loc) in
@@ -2539,7 +2547,7 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
       let ar,_ = tr_none er.C.atom
       and aw,_ = tr_none ew.C.atom in
       match ar,aw with
-      | (Pair _,Pair _)->
+      | (Pair _,Pair _) ->
          emit_exch_dep_addr22 csel vdep st p init er ew rd
       | (Pair _,_) ->
          check_cu (not A64.do_cu);
@@ -2713,7 +2721,7 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
              None,init,cs2@cs,st
           | Some (Neon _,Some _) -> assert false
           | Some (Pair (opt,idx),None) ->
-             let init,cs,st = stp_emit_store_reg opt idx st p init loc r2 in
+             let init,cs,st = stp_emit_store_reg (pair_opt_to_st opt) idx st p init loc r2 in
              None,init,cs2@cs,st
           | Some (Pair _,Some _) -> assert false
           end
--- a/gen/AArch64Compile_gen.ml
+++ b/gen/AArch64Compile_gen.ml
@@ -1463,6 +1463,7 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
           match ar with
           | Pair ((`Pa),_),None -> XP
           | Pair ((`PaIQ),_),None -> AXP
+          | Pair ((`PaA),_),None -> AXP
           | _ ->
              Warn.fatal
                "Illegal %s annotaton on load exclusive pair" (pp_atom ar)  in
@@ -1478,6 +1479,7 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
           match aw with
           | Pair ((`Pa),_),_ -> YY
           | Pair ((`PaIL),_),_ -> LY
+          | Pair ((`PaL),_),_ -> LY
           | _ ->
              Warn.fatal
                "Illegal %s annotaton on store exclusive pair" (pp_atom aw)  in
@@ -1858,12 +1860,8 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
             Some r,init,cs,st
         | R,Some (Neon _,Some _) -> assert false
         | R,Some (Pair (opt,idx),None) ->
-          let ld_opt = match opt with
-            | `Pa -> `Pa | `PaN -> `PaN | `PaIQ -> `PaIQ
-            | `PaIL -> assert false
-          in
-    let r,init,cs,st = emit_ldp ld_opt idx st p init loc in
-    Some r,init,cs,st
+          let r,init,cs,st = emit_ldp (pair_opt_to_ld opt) idx st p init loc in
+          Some r,init,cs,st
         | R,Some (Pair _,Some _) -> assert false
         | W,None ->
             let init,cs,st =
@@ -1903,12 +1901,8 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
             let init,cs,st = STG.emit_store st p init e in
             None,init,cs,st
         | W,Some (Pair (opt,idx),None) ->
-            let st_opt = match opt with
-              | `Pa -> `Pa | `PaN -> `PaN | `PaIL -> `PaIL
-              | `PaIQ -> assert false
-            in
-    let init,cs,st = emit_stp st_opt idx st p init loc e in
-    None,init,cs,st
+          let init,cs,st = emit_stp (pair_opt_to_st opt) idx st p init loc e in
+          None,init,cs,st
         | W,Some (Pair _,Some _) -> assert false
         | (R|W), Some (Instr, _) -> Warn.fatal "Instr annotation did not create code location %s" (C.debug_evt e)
         | R,Some (Pte (Read|ReadAcq|ReadAcqPc as rk),None) ->
@@ -2547,7 +2541,7 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
       let ar,_ = tr_none er.C.atom
       and aw,_ = tr_none ew.C.atom in
       match ar,aw with
-      | (Pair _,Pair _) ->
+      | (Pair _,Pair _)->
          emit_exch_dep_addr22 csel vdep st p init er ew rd
       | (Pair _,_) ->
          check_cu (not A64.do_cu);
--- a/gen/common/AArch64Arch_gen.ml
+++ b/gen/common/AArch64Arch_gen.ml
@@ -477,12 +477,21 @@ let is_tthm fields =
      | Some Capability -> "c"
 
    let pp_pair_opt = function
-     | Pa -> ""
-     | PaN -> "N"
-     | PaI -> "I"
+     | `Pa -> ""
+     | `PaN -> "N"
+     | `PaIQ -> "IQ"
+     | `PaIL -> "IL"
 
    and pp_pair_idx = function
      | UnspecLoc -> ""
+
+   let pair_opt_to_ld : pair_opt -> ld_pair_opt = function
+     | `Pa -> `Pa | `PaN -> `PaN | `PaIQ -> `PaIQ
+     | `PaIL -> assert false
+
+   let pair_opt_to_st : pair_opt -> st_pair_opt = function
+     | `Pa -> `Pa | `PaN -> `PaN | `PaIL -> `PaIL
+     | `PaIQ -> assert false
 
    let pp_atom_acc = function
      | Atomic rw -> sprintf "X%s" (pp_atom_rw rw)
@@ -575,15 +584,16 @@ let is_tthm fields =
      else
        fun _ r -> r
 
-      let fold_pair f r =
-        if do_mixed then r
-        else
-          let f opt idx r =
-            f (Pair (opt,idx)) r in
-          r |>
-          f Pa UnspecLoc |>
-          f PaN UnspecLoc |>
-          f PaI UnspecLoc
+    let fold_pair f r =
+      if do_mixed then r
+      else
+        let f opt idx r =
+          f (Pair (opt, idx)) r in
+        r |>
+        f `Pa UnspecLoc |>
+        f `PaN UnspecLoc |>
+        f `PaIQ UnspecLoc |>
+        f `PaIL UnspecLoc
 
       let fold_acc_opt o f r =
         let r = f (Acq o) r in
--- a/gen/common/AArch64Arch_gen.ml
+++ b/gen/common/AArch64Arch_gen.ml
@@ -240,7 +240,7 @@ type pair_idx = UnspecLoc
 type atom_acc =
   | Plain of capa_opt | Acq of capa_opt | AcqPc of capa_opt | Rel of capa_opt
   | Atomic of atom_rw | Tag | CapaTag | CapaSeal | Pte of atom_pte | Neon of neon_opt
-  | Pair of pair_opt * pair_idx | Instr
+  | Pair of [ld_pair_opt | st_pair_opt] * pair_idx | Instr
 
 let  plain = Plain None
 
@@ -446,8 +446,10 @@ let applies_atom (a,_) d =
   | Rel _,W
   | Pte (Read|ReadAcq|ReadAcqPc),R
   | Instr, R
-  | (Plain _|Atomic _|Tag|CapaTag|CapaSeal|Neon _|Pair _),(R|W)
+  | (Plain _|Atomic _|Tag|CapaTag|CapaSeal|Neon _),(R|W)
     -> true
+  | Pair ((`Pa|`PaN|`PaIQ|`PaA),_),R -> true
+  | Pair ((`Pa|`PaN|`PaIL|`PaL),_),W -> true
   (* special case for TTHM HA for read *)
   | Pte (Set p),R when WPTESet.mem HA p -> true
   | Pte (ReadHAAcq|ReadHAAcqPc),R -> true
@@ -481,17 +483,19 @@ let is_tthm fields =
      | `PaN -> "N"
      | `PaIQ -> "IQ"
      | `PaIL -> "IL"
+     | `PaA -> "A"
+     | `PaL -> "L"
 
    and pp_pair_idx = function
      | UnspecLoc -> ""
 
-   let pair_opt_to_ld : pair_opt -> ld_pair_opt = function
-     | `Pa -> `Pa | `PaN -> `PaN | `PaIQ -> `PaIQ
-     | `PaIL -> assert false
+   let pair_opt_to_ld : [ld_pair_opt | st_pair_opt] -> ld_pair_opt = function
+     | `Pa -> `Pa | `PaN -> `PaN | `PaIQ -> `PaIQ | `PaA -> `PaA
+     | `PaIL | `PaL -> assert false
 
-   let pair_opt_to_st : pair_opt -> st_pair_opt = function
-     | `Pa -> `Pa | `PaN -> `PaN | `PaIL -> `PaIL
-     | `PaIQ -> assert false
+   let pair_opt_to_st : [ld_pair_opt | st_pair_opt] -> st_pair_opt = function
+     | `Pa -> `Pa | `PaN -> `PaN | `PaIL -> `PaIL | `PaL -> `PaL
+     | `PaIQ | `PaA -> assert false
 
    let pp_atom_acc = function
      | Atomic rw -> sprintf "X%s" (pp_atom_rw rw)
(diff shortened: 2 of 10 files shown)


herd/AArch64Sem.ml
```diff
@@ -1879,6 +1879,34 @@ Arguments:
...
+        let>= base = read_reg_ord ra ii in
+        let>= v1 = do_read_mem_ret sz Annot.A aexp Access.VIR base ii in
+        
+        (* Calculate second address *)
+        let>= addr2 = M.op1 (Op.AddK width) base in
+        let>= v2 = do_read_mem_ret sz Annot.A aexp Access.VIR addr2 ii in
+        
+        (* Perform the writes *)
+        (write_reg_dest r1 v1 ii >>| write_reg_dest r2 v2 ii)
+        >>= M.ignore >>= B.next1T
```
relokin: Can you please reuse the module LoadPair instead?

herd/AArch64Sem.ml
```diff
@@ -1879,6 +1879,34 @@ Arguments:
...
+
+      let stlp sz r1 r2 ra ii =
```
relokin: Could you check if you could use/extend the existing stp function?

relokin (commented):
Thanks for this! I've got two comments.

herd/AArch64Sem.ml
RaghavRoy145: Thanks for the reviews! Yes, I have made this change.

herd/AArch64Sem.ml
RaghavRoy145: Good call, I didn't notice the other load variants do the same thing by changing the read_mem.  I'm not sure if my new implementation is idiomatic. Happy to make it so based on comments

herd/AArch64Sem.ml
```diff
@@ -2064,6 +2072,9 @@ Arguments:
             end >>!  ())
         sz t rr rd ii
 
+      let stlp sz r1 r2 ra ii =
+        stp AArch64.PaI sz r1 r2 ra (0, AArch64.Idx) ii
```
relokin: This is not quite right. `AArch64.PaI` is doing the right thing for STILP but won't work for STLP. Much like STILP you want the Release semantics, but you don't want to order the two Explicit Memory Write Effects. This means that you need to use the `>>|` monad.

herd/AArch64Sem.ml
```diff
@@ -1881,6 +1881,14 @@ Arguments:
             end) in
         LDPSW.ldp AArch64.Pa MachSize.Word
 
+      let ldap =
+        let module LDAP_Module = LoadPair(struct
+            let read_mem sz _ expl ac reg addr ii =
+              do_read_mem sz Annot.A expl ac reg addr ii
+          end) in
+        fun sz r1 r2 ra ii ->
+          LDAP_Module.ldp AArch64.Pa sz r1 r2 ra (0, AArch64.Idx) ii
```
relokin: `AArch64.Pa` is not correct because it will mean that the load pair is not going to have Acquire semantics. I think you need to introduce a new annotation `AArch64.PaA`

relokin (changes requested):

herd/AArch64Sem.ml
RaghavRoy145: Is it clean to add the PaA mapping directly to the Ldp semantics like this in that case?

```
match tnt with
              | Pa -> N
              | PaN -> NTA
              | PaI -> Q 
              | PaA -> A in
              ...
```

herd/AArch64Sem.ml
RaghavRoy145: I see, does it then make sense to add a PaL analogous to the PaA, and then changing the stp semantics to something like this?:

```
match tnt with
              | AArch64.(Pa | PaN | **PaL**) -> (>>|)
              | AArch64.PaI -> M.seq_mem in
```

herd/AArch64Sem.ml
relokin: yes please

herd/AArch64Sem.ml
relokin: yes please

herd/AArch64Sem.ml
RaghavRoy145: Done, thanks for the reviews!

herd/AArch64ASLSem.ml
relokin, line 1: @HadrienRenaud can I please ask your opinion on the changes to this file?

lib/AArch64Base.ml
```diff
@@ -1226,7 +1226,25 @@ let tr_simd_variant = function
 let simd_variant_nbytes v = tr_simd_variant v |> MachSize.nbytes
 
 type temporal = TT | NT
-type pair_opt = Pa | PaN | PaI
+type pair_opt = [`Pa | `PaN | `PaIQ | `PaIL | `PaA | `PaL]
```
relokin: I don't think we should keep this type

lib/AArch64Base.ml
```diff
@@ -1226,7 +1226,25 @@ let tr_simd_variant = function
...
+type ld_pair_opt = [`Pa | `PaN | `PaIQ | `PaA]
+type st_pair_opt = [`Pa | `PaN | `PaIL | `PaL]
```
relokin, line 1236: why are we turning this into a Polymorphic type?

lib/AArch64Base.ml
```diff
@@ -1414,16 +1434,18 @@ type 'k kinstruction =
   | I_ADD_SIMD of reg * reg * reg
   | I_ADD_SIMD_S of reg * reg * reg
   (* More loads *)
-  | I_LDP of pair_opt * variant * reg * reg * reg * 'k idx
+  | I_LDP of ld_pair_opt * variant * reg * reg * reg * 'k idx
   | I_LDPSW of reg * reg * reg * 'k idx
   | I_LDAR of variant * ld_type * reg * reg
   | I_LDXP of variant * ldxp_type * reg * reg * reg
+  | I_LDAP of variant * reg * reg * reg
```
relokin: do you really need this?

lib/AArch64Base.ml
```diff
@@ -1414,16 +1434,18 @@ type 'k kinstruction =
...
   (* Stores *)
   | I_STR of variant * reg * reg * 'k MemExt.ext
-  | I_STP of pair_opt * variant * reg * reg * reg * 'k idx
+  | I_STP of st_pair_opt * variant * reg * reg * reg * 'k idx
   | I_STLR of variant * reg * reg
   | I_STXR of variant * st_type * reg * reg * reg
   | I_STXP of variant * st_type * reg * reg * reg * reg
+  | I_STLP of variant * reg * reg * reg
```
relokin: do you really need this?

gen/common/AArch64Arch_gen.ml
```diff
@@ -477,13 +477,24 @@ let is_tthm fields =
      | Some Capability -> "c"
 
    let pp_pair_opt = function
```
relokin, line 481: shouldn't we also split this to a ld and st variant?

gen/AArch64Compile_gen.ml
```diff
@@ -2069,15 +2079,15 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
...
+          check_cu (not A64.do_cu) ;
+          emit_exch21 st p init er ew
       | (_,Pair _) ->
-         check_cu (not A64.do_cu) ;
-         emit_exch12 st p init er ew
+          check_cu (not A64.do_cu) ;
+          emit_exch12 st p init er ew
       | _,_ ->
-         emit_exch1 st p init er ew
+          emit_exch1 st p init er ew
```
relokin: can you please revert these changes?

gen/AArch64Compile_gen.ml
```diff
@@ -2545,19 +2555,19 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
...
+          | (Pair _,Pair _) ->
+              emit_exch_dep_addr22 csel vdep st p init er ew rd
+          | (Pair _,_) ->
+              check_cu (not A64.do_cu);
+              emit_exch_dep_addr21 csel vdep st p init er ew rd
+          | (_,Pair _) ->
+              check_cu (not A64.do_cu);
+              emit_exch_dep_addr12 csel vdep st p init er ew rd
+          | _,_ ->
+              emit_exch_dep_addr1 csel vdep st p init er ew rd
```
relokin: can you please revert these changes?

relokin (commented):
I am not sure we will retain backwards compatibility, but I think diyone7 will need to support the new instructions. This means we need the annotations PaL -> STLP and PaA -> LDAP. Now I think we can either keep interpreting PaI as either LDIAPP or STILP or we can split that to PaIQ and PaIL. Not that right now I can't generate a test with STILP. Neither this works:
```
$> diyone7 -metadata false -arch AArch64 PodWW PaIL Rfe PodRR Fre
```
nor this:
```
$> diyone7 -metadata false -arch AArch64 PodWW PaI Rfe PodRR Fre
```
and indeed neither annotation is valid
```
$> diyone7 -arch AArch64 -show annotations
 P X XL XA XAL L Q A PaL PaA PaIQ PaN Pa
```

herd/AArch64ASLSem.ml
```diff
@@ -866,42 +866,62 @@ module Make (TopConf : AArch64Sig.Config) (V : Value.AArch64ASL) :
...
+      let wback, postindex = decode_wback_postindex idx in
+      let fname =
+        match pa with
+        | `Pa -> "ldstpair_post/STP_32_ldstpair_post.opn"
+        | `PaIL -> "ldiappstilp/STILP_32SE_ldiappstilp.opn"
+        | `PaN -> "ldstnapair_offs/STNP_32_ldstnapair_offs.opn"
+        | `PaL -> "ldiappstilp/STLP_64_ldiappstilp.opn"
+      and acqrel = false
+      and offset = if pa = `PaIL then liti k else litbv 64 k in
+  Some
```
HadrienRenaud: nit: wrong intendation.

lib/AArch64Base.ml
RaghavRoy145: - I've kept `pair_opt` because having separate `Ld_Pair` and `St_Pair` caused assertion failures in both `add_lxm_atom` and `add_lxm_edge` as the pretty-printers for shared variants like Pa and PaN produce identical strings, but `compare_atomo` sees them as different constructors. 
- Using prefixes (LdPa/StPa) breaks backward compatibility with diyone7 and user-facing tests.
- The unified `Pair` at the was the only approach that preserved backward compatibility while avoiding these issues, and this needs the `pair_opt` type. The split into `ld_pair_opt/st_pair_opt` is preserved at the instruction level (I_LDP/I_STP) where direction is known, with `pair_opt_to_ld `and `pair_opt_to_st` conversion functions used in `emit_access` where the direction is relevant again.

lib/AArch64Base.ml
RaghavRoy145: I used polymorphic variants so that `ld_pair_opt` and `st_pair_opt` could share common tags (`Pa, `PaN) while having separate type constraints on `I_LDP` vs `I_STP`.

With regular variants, the shared constructors (Pa, PaN) would need to be duplicated across both types (e.g. LdPa/StPa, LdPaN/StPaN), and every conversion and pattern match becomes more verbose. If regular variants are preferred, it's a straightforward change. Happy to switch if there's a preference.

gen/common/AArch64Arch_gen.ml
RaghavRoy145: Currently `atom_acc `uses the `Pair of pair_opt * pair_idx`. 

At the instruction level, the pretty-printing of `I_LDP/I_STP` doesn't go through `pp_pair_opt` and it uses the instruction printer directly. So there's no need for split `pp_ld_pair_opt/pp_st_pair_opt ` functions. 

If we were to remove the unified `pair_opt` and go back to `Ld_Pair/St_Pair` in `atom_acc`, we cpould split printers, but then we'd hit the `add_lxm_atom` and `add_lxm_edge` assertion failures described above since both would produce the same strings for shared variants like `Pa` and `PaN`.

RaghavRoy145:
> I am not sure we will retain backwards compatibility, but I think diyone7 will need to support the new instructions. This means we need the annotations PaL -> STLP and PaA -> LDAP. Now I think we can either keep interpreting PaI as either LDIAPP or STILP or we can split that to PaIQ and PaIL. Not that right now I can't generate a test with STILP. Neither this works:
> 
> ```
> $> diyone7 -metadata false -arch AArch64 PodWW PaIL Rfe PodRR Fre
> ```
> 
> nor this:
> 
> ```
> $> diyone7 -metadata false -arch AArch64 PodWW PaI Rfe PodRR Fre
> ```
> 
> and indeed neither annotation is valid
> 
> ```
> $> diyone7 -arch AArch64 -show annotations
>  P X XL XA XAL L Q A PaL PaA PaIQ PaN Pa
> ```


This I think is because of a bug in my PR, I failed to include "PaIL" in `fold_pair` (I overwrote it in [the second commit](https://github.com/herd/herdtools7/pull/1639/changes/cb76a56ec27d8d7ec4dc0ee6c257a3fd31d3d338#diff-9dda1bfaa29914247a91cd78d17def95cfca44cdbc1b45e52ee2e264175be1c3L596))

Here is a working example after I fixed it locally:

```
herdtools7 % _build/install/default/bin/diyone7 -metadata false -arch AArch64 PodWW PaIL Rfe PodRR Fre
AArch64 MP+poppail+po
{
 int y[2]={0,0};
 0:X1=x; 0:X2=y;
 1:X1=x; 1:X2=y;
}
 P0               | P1          ;
 MOV W0,#1        | LDR W0,[X2] ;
 STR W0,[X1]      | LDR W3,[X1] ;
 MOV W3,#2        |             ;
 SUB W4,W3,#1     |             ;
 STILP W4,W3,[X2] |             ;

exists (1:X0=1 /\ 1:X3=0)

```

And here is one with STLP:

```
herdtools7 % _build/install/default/bin/diyone7 -metadata false -arch AArch64 PodWW PaL Rfe PodRR Fre 
AArch64 MP+poppal+po
{
 int y[2]={0,0};
 0:X1=x; 0:X2=y;
 1:X1=x; 1:X2=y;
}
 MOV W0,#1       | LDR W0,[X2] ;
 STR W0,[X1]     | LDR W3,[X1] ;
 MOV W3,#2       |             ;
 SUB W4,W3,#1    |             ;
 STLP W4,W3,[X2] |             ;

exists (1:X0=1 /\ 1:X3=0)

```

lib/AArch64Base.ml
RaghavRoy145: I can work on removing these instances and only use LDP/STP variants for these instructions.

herd/AArch64ASLSem.ml
```diff
@@ -866,42 +866,62 @@ module Make (TopConf : AArch64Sig.Config) (V : Value.AArch64ASL) :
...
-                ])
+      | I_LDP (pa, v, rt, rt2, rn, (k, idx)) ->
+        let wback, postindex = decode_wback_postindex idx in
+        let fname =
+          match pa with
+          | `Pa -> "ldstpair_post/LDP_32_ldstpair_post.opn"
+          | `PaIQ -> "ldiappstilp/LDIAPP_32LE_ldiappstilp.opn"
+          | `PaN -> "ldstnapair_offs/LDNP_32_ldstnapair_offs.opn"
+          | `PaA -> "ldiappstilp/LDAP_64_ldiappstilp.opn"
+        and acqrel = pa = `PaIQ && rt != ZR && rt2 != ZR
```
HadrienRenaud: Here you need:

```suggestion
        and acquire = pa = `PaA && rt != ZR && rt2 != ZR
        and acqrel = pa = `PaIQ && rt != ZR && rt2 != ZR
```

RaghavRoy145: Thanks! I'll try this out

herd/AArch64ASLSem.ml
```diff
@@ -866,42 +866,66 @@ module Make (TopConf : AArch64Sig.Config) (V : Value.AArch64ASL) :
...
+                "ispair" ^= litb true;
+                "t" ^= reg rt;
+                "t2" ^= reg rt2;
+                "n" ^= reg rn;
+                "datasize" ^= variant v;
+                "nontemporal" ^= litb (pa = `PaN);
+                "offset" ^= offset;
+                "tagchecked" ^= litb (wback || rn <> SP);
+                (match pa with
+                  | `Pa | `PaN -> with_pos S_Pass
```
HadrienRenaud: Maybe we can inline `with_pos S_Pass` to make it easier to read. This would override the `pass` from `Asllib`, but that's not really a problem:

```suggestion
                  | `Pa | `PaN -> pass
```

herd/AArch64ASLSem.ml
HadrienRenaud: I am happy with the file as it is now.

herd/AArch64ASLSem.ml
RaghavRoy145: I could add a declaration `let pass = with_pos S_Pass` that's scoped to the I_LDP and I_STP branches of the match, does that sound good? Unsure what the scope should be

herd/tests/instructions/AArch64/asl.cfg
```diff
@@ -61,4 +61,4 @@ nonames L086
 # Missing SVC
 nonames A259
 # Mixed mode non implemented
-nonames M001
+nonames M001
```
diaolo01: What is the change here?

herd/tests/instructions/AArch64/Y006.litmus
```diff
@@ -0,0 +1,15 @@
...
+ 1:X0=y; 1:X2=x;
+ uint64_t 1:X3; uint64_t 1:X4;
+}
+ P0              | P1              ;
+ MOV X2,#1       | LDAP X3,X4,[X0] ;
+ STR X2,[X0]     | LDR X5,[X2]     ;
+ MOV X3,#1       |                 ;
+ MOV X4,#99      |                 ;
+ STLP X3,X4,[X1] |                 ;
+exists (1:X3=1 /\ 1:X5=0)
```
diaolo01: Would it make sense for this test to require observing both X3 and X4 on P1, so it it clearly testing observation of the STLP pair rather than a standard MP pattern?

diaolo01 (commented):
Thanks for this @RaghavRoy145. Generally, this looks good to me.

herd/tests/instructions/AArch64/Y006.litmus
RaghavRoy145: Ah good call, though the behaviour would not change (no witnesses) I should still strengthen the post-condition to include 1:X4=99 so don't permit partial writes

herd/tests/instructions/AArch64/asl.cfg
RaghavRoy145: Thanks for the reviews! I think this is an artefact from my rebase, I will revert this file.

lib/AArch64Base.ml
RaghavRoy145: I have instead replaced the usage of this type with `[ld_pair_opt | st_pair_opt]`

gen/common/AArch64Arch_gen.ml
```diff
@@ -240,7 +240,7 @@ type pair_idx = UnspecLoc
 type atom_acc =
   | Plain of capa_opt | Acq of capa_opt | AcqPc of capa_opt | Rel of capa_opt
   | Atomic of atom_rw | Tag | CapaTag | CapaSeal | Pte of atom_pte | Neon of neon_opt
-  | Pair of pair_opt * pair_idx | Instr
+  | Pair of [ld_pair_opt | st_pair_opt] * pair_idx | Instr
```
relokin, line 243: @ShaleXIONG can you please have a look at the changes (at least to `gen`) of this PR?

lib/AArch64Parser.mly
```diff
@@ -566,9 +567,11 @@ zeroopt:
 
 ldp_instr:
 | LDP
-  { (fun v r1 r2 (r3,idx) -> I_LDP (Pa,v,r1,r2,r3,idx)) }
+  { (fun v r1 r2 (r3,idx) -> I_LDP ((`Pa),v,r1,r2,r3,idx)) }
 | LDNP
-  { (fun v r1 r2 (r3,idx) -> I_LDP (PaN,v,r1,r2,r3,idx)) }
+  { (fun v r1 r2 (r3,idx) -> I_LDP ((`PaN),v,r1,r2,r3,idx)) }
+| LDAP
+  { (fun v r1 r2 (r3,idx) -> I_LDP ((`PaA),v,r1,r2,r3,idx)) }
```
relokin, line 574: LDAP doesn't have a post-index variant, so this
`LDAP W0,W1,[X2],#4`
should return an error
and it doesn't have a pre-index variant, so this
`LDAP W0,W1,[X2,#4]!`
should return an error
or any offset other than 0 should also return an error.
`LDAP W0,W1,[X2,#4]`
see https://developer.arm.com/documentation/ddi0602/2026-03/Base-Instructions/LDAP--Load-acquire-pair-of-registers-?lang=en
as opposed to https://developer.arm.com/documentation/ddi0602/2026-03/Base-Instructions/LDP--Load-pair-of-registers-

lib/AArch64Parser.mly
```diff
@@ -589,9 +592,11 @@ ldp_simd_instr:
 
 stp_instr:
 | STP
-  { (fun v r1 r2 (r3,idx) -> I_STP (Pa,v,r1,r2,r3,idx)) }
+  { (fun v r1 r2 (r3,idx) -> I_STP ((`Pa),v,r1,r2,r3,idx)) }
 | STNP
-  { (fun v r1 r2 (r3,idx) -> I_STP (PaN,v,r1,r2,r3,idx)) }
+  { (fun v r1 r2 (r3,idx) -> I_STP ((`PaN),v,r1,r2,r3,idx)) }
+| STLP
+  { (fun v r1 r2 (r3,idx) -> I_STP ((`PaL),v,r1,r2,r3,idx)) }
```
relokin, line 599: The same variant restrictions apply for STLP as for LDAP.

relokin (commented):
Two comments otherwise the herd side looks good to me. The diyone side looks good to me as well but there is also an opportunity to refactor the code there as well, but I will leave that decision to @ShaleXIONG.

lib/AArch64Parser.mly
relokin: one last point, I just realised that LDAP is only defined for reading a pair of 64bit values, so
```
LDP W0,W1,[X2]
```
is not a valid instruction, but
```
LDP X0,X1,[X2]
```
is a valid instruction.

relokin: if you can add these checks in the parser and throw a ParserError when they fail that would be fantastic.

lib/AArch64Parser.mly
```diff
@@ -600,7 +605,7 @@ stp_instr:
       | (V32,(MetaConst.Int (-8),PreIdx))
       | (V64,(MetaConst.Int (-16),PreIdx))
           ->
-            I_STP (PaI,v,r1,r2,r3,idx)
+            I_STP ((`PaIL),v,r1,r2,r3,idx)
       | _, _ -> raise Parsing.Parse_error)
```
relokin, line 609: This is a good example, of the type of checks you need to do.

gen/common/AArch64Arch_gen.ml
ShaleXIONG: On it.

gen/common/AArch64Arch_gen.ml
ShaleXIONG: This implementation is fine for me. 

However can you extend the function `applies_atom`, which checks the validity between annotation, `a` and memory event, `d` (of `R` and `W`). It should be around line `439` in your branch, in this file. This enable an earlier filter in building cycle, rather an error raise in the final compilation time.

herd/tests/instructions/AArch64/Y006.litmus
ShaleXIONG, line 1: Question in general: should we add more test on the new instruction, `STNP`,`LDNP`, `STILP` and `LDIAPP`?

lib/AArch64Base.ml
ShaleXIONG: From the `diy` side, I have no strong opinion if you want to _rework_ constructors `Pa` and `PaN`, which originally can be apply to both `read` and `write` event. For example if we introduce different new name and hence remove the polymorphic type implementation here.

ShaleXIONG:
> @ShaleXIONG can you please review and help with the changes to `gen/`?

Ref: @relokin. I left a few comments. Thanks for @RaghavRoy145 on also extending the `diy` side. if you find it difficult to understand `gen/`, please feel free to ping me and I can help.

lib/AArch64Parser.mly
murzinv: > if you can add these checks in the parser and throw a ParserError when they fail that would be fantastic.

I'd suggest using existing infrastructure to reject invalid instructions in user-friendly way - just place checks for invalid combo  under `is_valid` function (`lib/AArch64Base.ml`)

RaghavRoy145: Thanks for the reviews! with your suggested change @murzinv I am able to reproduce these helpful warnings:

```
Warning: File "./herd/tests/instructions/AArch64/Y006.litmus": Illegal instruction 'LDAP W3,W4,[X0]' (User error)

Warning: File "./herd/tests/instructions/AArch64/Y006.litmus": Illegal instruction 'STLP W3,W4,[X1]' (User error)

Warning: File "./herd/tests/instructions/AArch64/Y006.litmus": Illegal instruction 'STLP X3,X4,[X1,#4]' (User error)
Warning: File "./herd/tests/instructions/AArch64/Y006.litmus": Illegal instruction 'LDAP X3,X4,[X0,#4]' (User error)
```

Though the syntax `LDAP X3,X4,[X0,#0]` or `STLP X3,X4,[X1,#0]` remains allowed

gen/common/AArch64Arch_gen.ml
RaghavRoy145: I have made the extension to add the direction based matching in `applies_atom`

Though I think the above change should be made as it is a stronger check, this does not mean I can get rid of the above type-conversion functions (`pair_opt_to_ld `and `pair_opt_to_st`) unfortunately, as:

`emit_ldp` expects a ld_pair_opt (and equivalently the emit store analogue), but `opt` as it comes from Pair() has the following type:
`| Pair of [ld_pair_opt | st_pair_opt] * pair_idx`

So it is the union of the two pair types which `emit_ldp` does not like

We could keep the type-conversion functions, as well as make the change @ShaleXIONG  wants me to in `applies_atom` 

With the changes to `applies_atom` I get this error for a wrong cycle (example courtesy of @ShaleXIONG ):

```
diyone7 -arch AArch64 PaL PosRW Pa Rfi -oneloc      
diyone7: Fatal error: Test CoRW1+pospalpa-rfipapal [PosRWPaLPa RfiPaPaL] failed:
annotation mismatch on edge PosRWPaLPa, annotation 'PaL' on R
```

lib/AArch64Parser.mly
murzinv: Perhaps both `STILP` and `LDIAPP` deserve moving checks to `is_valid` function, but that for another patch/PR 😄

relokin:
@RaghavRoy145 can you please rebase your PR?

herd/AArch64ASLSem.ml
```diff
@@ -866,42 +867,66 @@ module Make (TopConf : AArch64Sig.Config) (V : Value.AArch64ASL) :
...
+                "t2" ^= reg rt2;
+                "n" ^= reg rn;
+                "datasize" ^= variant v;
+                "nontemporal" ^= litb (pa = `PaN);
+                "offset" ^= offset;
+                "tagchecked" ^= litb (wback || rn <> SP);
+                (match pa with
+                  | `Pa | `PaN -> pass
+                  | (`PaA) -> "acquire" ^= litb (rt != ZR && rt2 != ZR)
+                  | (`PaIQ) -> "acqrel" ^= litb (rt != ZR && rt2 != ZR));
```
relokin:
```suggestion
                  | (`PaA) -> "acquire" ^= litb true
                  | (`PaIQ) -> "acqrel" ^= litb true;
```

herd/AArch64ASLSem.ml
```diff
@@ -866,42 +867,66 @@ module Make (TopConf : AArch64Sig.Config) (V : Value.AArch64ASL) :
...
+                "t2" ^= reg rt2;
+                "n" ^= reg rn;
+                "datasize" ^= variant v;
+                "nontemporal" ^= litb (pa = `PaN);
+                "offset" ^= offset;
+                "tagchecked" ^= litb (wback || rn <> SP);
+                (match pa with
+                  | `Pa | `PaN -> pass
+                  | (`PaL) -> "acquire" ^= litb (rt != ZR && rt2 != ZR)
+                  | (`PaIL) -> "acqrel" ^= litb (rt != ZR && rt2 != ZR));
```
relokin:
```suggestion
                  | (`PaL) -> "acquire" ^= litb true
                  | (`PaIL) -> "acqrel" ^= litb true;
```

relokin (commented):
Two minor comments, if you can fix these, we can merge this PR.

herd/AArch64ASLSem.ml
RaghavRoy145: Thanks for this review, I've pushed this change!

relokin:
Can you also please rebase?

relokin (approved):
LGTM, thanks Raghav! Will merge this now.

PR #1685, opened 2026-01-27
[gen] Introduce a new relaxation/edge/atom parser in `diy` tool.

ShaleXIONG:
This is a first phrase to rework on the `diy*` lexer and parser on input relaxations or cycles. In this pull request we introduce a more conventional way, that is, 

- update `lexutil.mll` tokenises input to an AST, specifically type `t` in new `ast.ml` and 
- new `parser.mly` parses the AST to internal data structure.

In the `Ast.t`, individual primitive relaxation is represented by 
- `One of string` for individual relaxation
- `Seq of t list` for a sequence of relaxations. This corresponds to `,` syntax or in some situation, white space. For example `diyone7 PosWR PosRR Fri` is equivalent to `diyone7 PosWR,PosRR,Fri`. 
- A new choice constructor `Choice of t list`, together with a new syntax `|`. For example `PosWR|DpAddrdW` means either `PosWR` or `DpAddrdW`. 
- A new option constructor `Opt of t`, together with a new syntax `?`. For example `PosWR?` means either `PosWR` or `(empty)`.

The new constructors can be used in `diycross7` and `diy7`. In `diyone7`, although it will be parsed however an error occurs due to input is not a precise one cycle. An example is `diy7 -arch AArch64 -safe "[Po|DMB.SY*** DpAddr?] Coe" -cycleonly true -size 4 -nprocs 2`, which gives:
``` 
Generator produced 10 tests
2+2W000: PosWR DpAddrdW Coe PosWR DpAddrdW Coe
2+2W001: PodWW Coe PosWR DpAddrdW Coe
2+2W002: PosWR DpAddrdW Coe PodWR DpAddrsW Coe
2+2W003: PosWR DpAddrdW Coe PodWR DpAddrdW Coe
2+2W004: DMB.SYsWR DpAddrdW Coe PosWR DpAddrdW Coe
2+2W005: DMB.SYdWW Coe PosWR DpAddrdW Coe
2+2W006: DMB.SYdWR DpAddrsW Coe PosWR DpAddrdW Coe
2+2W007: DMB.SYdWR DpAddrdW Coe PosWR DpAddrdW Coe
2+2W008: PodWW Coe PodWW Coe
2+2W009: PodWW Coe PodWR DpAddrsW Coe
2+2W010: PodWW Coe PodWR DpAddrdW Coe
2+2W011: PodWW Coe DMB.SYsWR DpAddrdW Coe
2+2W012: DMB.SYdWW Coe PodWW Coe
2+2W013: PodWW Coe DMB.SYdWR DpAddrsW Coe
2+2W014: PodWW Coe DMB.SYdWR DpAddrdW Coe
2+2W015: PodWR DpAddrsW Coe PodWR DpAddrsW Coe
2+2W016: PodWR DpAddrsW Coe PodWR DpAddrdW Coe
2+2W017: DMB.SYsWR DpAddrdW Coe PodWR DpAddrsW Coe
2+2W018: DMB.SYdWW Coe PodWR DpAddrsW Coe
2+2W019: DMB.SYdWR DpAddrsW Coe PodWR DpAddrsW Coe
2+2W020: DMB.SYdWR DpAddrdW Coe PodWR DpAddrsW Coe
2+2W021: PodWR DpAddrdW Coe PodWR DpAddrdW Coe
2+2W022: DMB.SYsWR DpAddrdW Coe PodWR DpAddrdW Coe
2+2W023: DMB.SYdWW Coe PodWR DpAddrdW Coe
2+2W024: DMB.SYdWR DpAddrsW Coe PodWR DpAddrdW Coe
2+2W025: DMB.SYdWR DpAddrdW Coe PodWR DpAddrdW Coe
2+2W026: DMB.SYsWR DpAddrdW Coe DMB.SYsWR DpAddrdW Coe
2+2W027: DMB.SYdWW Coe DMB.SYsWR DpAddrdW Coe
2+2W028: DMB.SYsWR DpAddrdW Coe DMB.SYdWR DpAddrsW Coe
2+2W029: DMB.SYsWR DpAddrdW Coe DMB.SYdWR DpAddrdW Coe
2+2W030: DMB.SYdWW Coe DMB.SYdWW Coe
2+2W031: DMB.SYdWW Coe DMB.SYdWR DpAddrsW Coe
2+2W032: DMB.SYdWW Coe DMB.SYdWR DpAddrdW Coe
2+2W033: DMB.SYdWR DpAddrsW Coe DMB.SYdWR DpAddrsW Coe
2+2W034: DMB.SYdWR DpAddrsW Coe DMB.SYdWR DpAddrdW Coe
2+2W035: DMB.SYdWR DpAddrdW Coe DMB.SYdWR DpAddrdW Coe
```
We can see `2+2W013: PodWW Coe DMB.SYdWR DpAddrsW Coe` which generate from `Po`, `Coe` `[DMB.SY*** DpAddr]`. One can also run the command with `-v`, where the fully unfold edges will print at very beginning.

Last, Given all the chances above, we unify the paring process cross three `diy*` tools. Previously different `diy*` has different parsing path and the actual code are in different files such as `diy.ml`, `diycross.ml` and `diyone.ml`. Now the main passing function unifies at `parse_expand_relax` and `parse_expand_relaxs` in `relax.ml`. 

Some _future plan_:
- remove the `Ppo` type which is only used in `PPCCompile_gen.ml`, and make the `Ppo` as a special wildcard.
- Parse the `Ast.t` in place. The current parsing is after we flatten the `Ast.t` to `string list list`, so we can hook to the existing parsing functions.

We decide to leave it as _future plan_ for smaller change in this pull request and role out the new syntax as soon.


gen/common/ast.ml
```diff
@@ -0,0 +1,64 @@
+type 'prim t =
+  | One of 'prim
+  | Opt of 'prim t
+  | Multi of 'prim t
+  | Seq of 'prim t list
+  | Choice of 'prim t list
```
fsestini, line 5: Please add an interface file `ast.mli`, and document the meaning of each `Ast.t` constructor in `ast.mli`, since IMO it is not that obvious from the names alone. In particular, it's not immediately clear what `Multi` represents.

I think these operators should also be documented in the respective `-help` messages.

gen/common/ast.ml
```diff
@@ -0,0 +1,64 @@
...
+   backward compatibility. If the top-level sequence already mixes in an
+   explicit `Choice`, keep the sequence structure so that `A|B,C` means
+   `(A|B),C` rather than `A|B|C`. *)
+let seq_as_choice = function
+  | Seq seq as ast when List.exists (function Choice _ -> true | _ -> false) seq
+    -> ast
+  | Seq seq -> Choice seq
+  | ast -> ast
+
+let rec flatten t =
```
fsestini: It would be good to have a comment that explains what `Ast.flatten` does. Can you please add one to the `ast.mli` file.

gen/common/ast.ml
```diff
@@ -0,0 +1,64 @@
...
+  List.map ( fun l ->
+    List.map ( fun r ->
+      f l r
+    ) rhs
+  ) lhs
+  |> List.flatten
+
+let rec to_list = function
+  | One _ as one -> [ one ]
+  | Opt t -> to_list t
```
fsestini: Is this correct? Shouldn't the optionality represented by `Opt` be taken into account here?

gen/common/ast.ml
```diff
@@ -0,0 +1,64 @@
...
+
+let list_cross_product_map f lhs rhs =
+  List.map ( fun l ->
+    List.map ( fun r ->
+      f l r
+    ) rhs
+  ) lhs
+  |> List.flatten
+
+let rec to_list = function
```
fsestini: `Ast.to_list` does not seem to behave like a conventional tree-to-list conversion (which I'd expect to have type `'a t -> 'a list`). Can you please rename it to something more specific and document its contract in `ast.mli`?

gen/common/ast.ml
```diff
@@ -0,0 +1,64 @@
...
+  | Multi multi -> [ multi ]
+
+(* Legacy `diy7` parsing interprets a plain top-level sequence as a choice for
+   backward compatibility. If the top-level sequence already mixes in an
+   explicit `Choice`, keep the sequence structure so that `A|B,C` means
+   `(A|B),C` rather than `A|B|C`. *)
+let seq_as_choice = function
+  | Seq seq as ast when List.exists (function Choice _ -> true | _ -> false) seq
+    -> ast
+  | Seq seq -> Choice seq
```
fsestini: Nit/personal preference:
```suggestion
  | Seq seq as ast ->
    if List.exists (function Choice _ -> true | _ -> false) seq then
      ast
    else Choice seq
```

gen/common/lexUtil.mll
```diff
@@ -15,49 +15,62 @@
 (****************************************************************************)
 
 {
+open Parser
+exception Eof
+(* Track whether the current scope has already seen an operand.
+   Whitespace after an operand is treated as a sequence separator. *)
```
fsestini: The comment says ‘operand’, but the grammar is about relaxation expressions. It's not clear what 'operand' and 'operator' refer to, in this context. Could you instead describe this as ‘a relaxation term has already been seen in the current scope’ and explain that the whitespace that follows is then tokenized as a sequence separator?

gen/common/lexUtil.mll
```diff
@@ -15,49 +15,62 @@
...
+let has_previous_relaxation = ref [false]
```
fsestini: Let's please avoid module-global mutable lexer state. If mutable state must be used, can we please make it local to the `parse` function? That is, the lexing `token` function should take a mutable reference as an argument, instead of using a global one.

gen/common/parser.mly
```diff
@@ -0,0 +1,42 @@
...
+  | lhs, Seq rhs -> Seq (lhs :: rhs)
+  | Seq lhs, rhs -> Seq (lhs @ [rhs])
+  | lhs, rhs -> Seq [lhs ; rhs]
+
+let mk_choice lhs rhs =
+  match lhs,rhs with
+  | Choice lhs, Choice rhs -> Choice (lhs @ rhs)
+  | lhs, Choice rhs -> Choice (lhs :: rhs)
+  | Choice lhs, rhs -> Choice (lhs @ [rhs])
+  | lhs, rhs -> Choice [lhs ; rhs]
```
fsestini: Is the "normalization" done in `mk_seq` and `mk_choice` really necessary? Wouldn't, for instance,`Choice [Choice lhs; Choice rhs]` be treated, semantically, as equivalent to `Choice [lhs; rhs]` anyway, by downstream functions?

gen/common/parser.mly
```diff
@@ -0,0 +1,42 @@
...
+
+%}
+
+%token <string> RELAXATION
+%token LEFT_SQUIRE RIGHT_SQUIRE
```
fsestini:
```suggestion
%token LEFT_SQUARE RIGHT_SQUARE
```

gen/common/parser.mly
```diff
@@ -0,0 +1,42 @@
...
+main:
+  | r = relax EOF { r }
+  | r = relax COMMA EOF { r }
+
+relax:
+  | r = RELAXATION { One r }
+  | opt = relax OPTION { Opt opt }
+  | lhs = relax CHOICE_BAR rhs = relax { mk_choice lhs rhs }
+  | lhs = relax COMMA rhs = relax { mk_seq lhs rhs }
+  | LEFT_SQUIRE r = relax RIGHT_SQUIRE { Multi r }
```
fsestini: The operator precedence between `OPTION`, `CHOICE_BAR`, and `COMMA` should be documented and tested, because it is implicit in Menhir declarations.

gen/tests/diy7-check.t
```diff
@@ -0,0 +1,4 @@
+An explicit comma stays a sequence after a choice in `diy7`
```
fsestini: Since this PR introduces new parser/AST logic, I’d prefer the corresponding test suite to mainly comprise of unit-style OCaml tests that verify parsing/expansion directly, with only a few cram tests reserved for a small number of end-to-end CLI checks. We also need more tests, covering various combinations of operators for all `diy*` tools. Each module in `gen/diy.ml`, `gen/diyone.ml` and `gen/diycross.ml` should probably expose their respective parsing logic as single functions, so that it's easy to invoke them from the test scripts. [Here](https://github.com/herd/herdtools7/blob/master/tools/uoaim/test/test_defs.ml) is an example of pure OCaml test.

gen/atomize.ml
```diff
@@ -87,21 +88,24 @@ module Make (F:Fence.S)(A:Atom.S) =
         done with End_of_file -> ()
 
       let zyva_argv es =
-        let es = List.map E.parse_edge es in
-        let es = atomize es in
-        printf "%s\n" (pp_edges es)
+        E.parse_edges es
+        |> atomize
+        |> pp_edges
+        |> printf "%s\n"
```
fsestini: Personal preference, but I actually find the previous format more readable.

gen/atomize.ml
```diff
@@ -87,21 +88,24 @@ module Make (F:Fence.S)(A:Atom.S) =
...
 
 let pp_es = ref []
 
 let () =
   Util.parse_cmdline
     opts
-    (fun x -> pp_es := x :: !pp_es)
+    (fun x ->
+      let segment = String.trim x in
+      if segment <> "" then pp_es := segment :: !pp_es)
```
fsestini, line 106: Can you explain why this `segment <> ""` check is now necessary?

gen/diy.ml
```diff
@@ -33,18 +32,67 @@ module Make(C:Builder.S)(O:DiyConfig) = struct
 
 open C.E
 open C.R
+open Ast
```
fsestini: `Ast` only seems needed in `parse_fence`; I’d prefer a local `let open Ast` in there rather than a module-global open.

gen/diy.ml
```diff
@@ -33,18 +32,67 @@ module Make(C:Builder.S)(O:DiyConfig) = struct
...
       List.fold_right
-        (fun s k -> C.E.parse_fence s::k)
+        (fun s k ->
+          match s with
+          | One s -> C.E.parse_fence s::k
+          | _ -> assert false)
         fs k
+  | Opt _ -> assert false
+  | Choice _ -> assert false
+  | Multi _ -> assert false
```
fsestini: I don't think `parse_fence` should `assert false` on the input `s` here. `s` comes from `O.cumul` and there's nothing in principle preventing `O.cumul` from being a `Choice` or a `Multi`. In other words, `parse_fence` is asserting false on inputs that are seemingly perfectly valid.

If some `-cumul ...` inputs are to be considered invalid, then the `O.cumul` argument should be parsed and sanitized into a valid shape _before_ being passed as an input to the `Make` functor. Moreover, invalid shapes of `O.cumul` should be properly caught and reported to the user earlier. For example, this shouldn't happen:
```
$ dune exec diy7 -- -arch AArch64 -cumul '[DMB.SY|ISB]'
Fatal error: exception File "gen/diy.ml", line 59, characters 16-22: Assertion failed
```

gen/diy.ml
```diff
@@ -55,9 +103,16 @@ let parse_fences fs = List.fold_right parse_fence fs []
     let min_relax = !Config.min_relax
 
     let prefix =
-      match List.map parse_relaxs O.prefix with
-      | [] -> [[]] (* No prefix <=> one empty prefix *)
-      | pss -> pss
+      List.map ( fun segment ->
+        parse_input_ast segment
+      ) O.prefix
+      |> fun e -> Ast.Choice e
```
fsestini: can you add a comment explainin why we do `fun e -> Ast.Choice e`?

gen/diy.ml
```diff
@@ -76,41 +131,13 @@ let parse_fences fs = List.fold_right parse_fence fs []
...
-        List.fold_left
-          (var_relax V.varatom_one) [] relax_list
-
   let gen lr ls rl n =
-    let lr = to_relax_list lr
-    and ls = to_relax_list ls
-    and rl = to_relax_list rl in
+    let parse_argument_opt argument =
+      Option.map parse_argument argument
+      |> Option.value ~default:[] in
```
fsestini: How about
```suggestion
    let parse_argument_opt = function
      | Some argument -> parse_argument argument
      | None -> []
    in
```

gen/diy.ml
```diff
@@ -188,11 +199,9 @@ let exec_conf s =
   ()
 
 let split_cands xs =
-  match
-    List.concat (List.map LexUtil.split xs)
-  with
-  | [] -> None
-  | _::_ as xs -> Some xs
+  if xs = [] then None
+  else (Some (String.concat " " xs))
```
fsestini: I don't quite follow why we need this `split_cands`. AFAICT this function takes pre-parsed command-line pieces and turns them back into one string that is then parsed later again. Could we not parse the relax and safe lists here directly?

gen/diycross.ml
```diff
@@ -53,25 +45,30 @@ module Make (Config:Config) (M:Builder.S) =
...
+
+    let parse_segment segment =
+      M.R.parse_ast segment |> Ast.seq_as_choice
 
-    let expand_edge es = M.E.expand_edges es Misc.cons
-    let expand_edges ess =
-      List.flatten (List.map (fun e -> expand_edge e []) ess)
+    let parse_edges input =
+      List.map parse_segment input
+      (* Wrap the parsed segments in a top-level sequence before expansion. *)
```
fsestini: can you add a comment to explain why we need this?

gen/diyone.ml
```diff
@@ -123,11 +123,14 @@ module Make(O:Config) (M:Builder.S) =
...
-          |> List.map M.R.parse_relax
-          |> List.map M.R.edges_of
-          |> List.flatten in
+        let es =
+          M.R.parse_sequence_ast pp_rs
+          |> M.R.parse_expand_relaxs ~ppo:M.ppo
+          |> ( function
+            | [x] -> x
+            | _ ->
+              Warn.user_error "`diyone7` only accepts exactly one input cycle." )
```
fsestini: Might be personal preference again, but to be honest I don't find this use of `|> function ...` or `|> fun ...` very readable, compared to a plain old let binding. The `|>` operator is useful to chain a sequence of functions without having to introduce explicit binders, so using `|>` and then introducing binders anyway via `fun`/`function` seem to defeat the purpose and makes the data flow harder to follow IMO.

gen/diyone.ml
```diff
@@ -166,7 +169,7 @@ module Make(O:Config) (M:Builder.S) =
               do_rec in
             D.all gen
         | _ -> dump name_opt es
-      end with Misc.Fatal msg ->
+      end with Misc.Fatal msg | Misc.UserError msg ->
```
fsestini, line 172: Perhaps user errors should be reported differently from fatal errors.

gen/relax.ml
```diff
@@ -284,14 +254,7 @@ and type edge = E.edge
...
 
-        open LexUtil
-
-        let parse_relax = function
-          | One r -> do_parse_relax r
-          | Seq [] -> Warn.fatal "Empty relaxation"
-          | Seq es -> ERS (List.map E.parse_edge es)
-
-        let parse_relaxs = List.map parse_relax
+        open Ast
```
fsestini: `open Ast` is not needed here.

gen/relax.ml
```diff
@@ -39,22 +39,21 @@ module type S = sig
   val pp_relax : relax -> string
   val pp_relax_list : relax list -> string
   val edges_of : relax -> edge list
+  val edges_ofs : relax list -> edge list
```
fsestini: How about
```suggestion
  val edges_of_list : relax list -> edge list
```

gen/relax.ml
```diff
@@ -304,6 +267,13 @@ and type edge = E.edge
           F.fold_some_fences
             (fun f k -> er (E.Fenced (f,sd,Dir d1,Dir d2))::k)
 
+        let relax_list_to_choice relax_list =
+          List.map (fun relax -> Ast.One relax) relax_list
+            |> ( function
+              | [] -> assert false
+              | [relax] -> relax
+              | relax_list -> Ast.Choice relax_list )
```
fsestini: As above: I think this would be better written as a pattern match.
```suggestion
          match relax_list with
              | [] -> assert false
              | [relax] -> Ast.One relax
              | relax_list -> Ast.Choice (List.map (fun relax -> Ast.One relax) relax_list)
```

gen/relax.ml
```diff
@@ -347,26 +321,60 @@ and type edge = E.edge
...
-          let rs = List.flatten rs in
-          rs
+        let parse_ast s =
+          Lexing.from_string s
+          |> LexUtil.parse Parser.main
+
+        let parse_sequence_ast segments =
+          Ast.Seq (List.map parse_ast segments)
+
+        let remove_invalid_relaxes relaxes =
```
fsestini, line 337: Can you explain why `remove_invalid_relaxes` is needed only now? My understanding is that the new AST-based expansion logic can end up generating some sequences with invalid adjacent edge combinations. Is that the reason for using `remove_invalid_relaxes` here? Could you please add a short comment explaining this?

gen/relax.ml
```diff
@@ -347,26 +321,60 @@ and type edge = E.edge
...
-          | Seq [] -> Warn.fatal "Empty relaxation"
-          | Seq es -> [ERS (List.map E.parse_edge es)]
-
-        let expand_relax_macros lus =
-          let rs = List.map expand_relax_macro lus in
-          let rs = List.flatten rs in
-          rs
+        let parse_ast s =
+          Lexing.from_string s
+          |> LexUtil.parse Parser.main
```
fsestini: `parse_ast` should catch and report menhir errors in a more informative way. Currently we get this:
```
$ dune exec diyone7 -- -arch AArch64 '['
Fatal error: exception Parser.MenhirBasics.Error
```

gen/norm.ml
```diff
@@ -71,16 +71,19 @@ module Make(Co:Config)(F:Fence.S)(A:Atom.S) = struct
...
+        |> ( function
+          | [x] -> x
+          | _ ->
+            Warn.user_error "`norm7` only accepts exactly one input cycle." )
+        |> R.edges_of in
       let base,es,_ = Norm.normalise_family (E.resolve_edges es) in
       let name =  N.mk_name base ?scope:None es in
       Printf.printf "%s: %s\n" name (E.pp_edges es)
-    with Misc.Fatal msg ->
+    with Misc.Fatal msg | Misc.UserError msg ->
```
fsestini: perhaps user errors should be reported differently from fatal errors.

fsestini (commented):
Thanks for this @ShaleXIONG . I've left a few comments in the code, and I have two overarching concerns:
- The lexer uses global mutable state, in a way that I think is quite error prone. I'm also not convinced it's really necessary. At the very least, I'd like to see it replaced with _local_ state. See the comments in the code.
- This PR introduces a new parser for existing syntax, and new user-facing syntax for relaxations. The parsing semantics of both the old and new syntax is not trivial (for example, white space is interpreted in a context-dependent way, and `,` has different meaning in `diy7` vs `diycross7`). For this reason, I think we need significantly more tests than what the PR currently offers. Again, see the comments for additional details.

Happy to have a chat online/offline about any of these points.

gen/common/ast.ml
ShaleXIONG: Done

gen/common/ast.ml
ShaleXIONG: That is a good point. I think the `flatten` here is actually the conventional of `to_list`. I will rename the current `to_list` to `get_sequence_list`. The latter function is actually only use in two places so let me rework a bit to hopefully  remove it.

gen/common/parser.mly
ShaleXIONG: Done

gen/diy.ml
ShaleXIONG: I will need to explore more here. 
I made the very conservative choice to preserve exact behaviour rather than expand the behaviour on `-cumul`. However I think I should expand the behaviour uniformly.

gen/relax.ml
ShaleXIONG: This is a performance only improvement for BOTH previous parser and the new parser, hence I put in a separate individual commit. Without it, invalid relaxation will only be typed check at the last point after go through all the unnecessary step in the searching algorithm.

I will add more comment here.

gen/common/lexUtil.mll
ShaleXIONG: Ah, you are right, let me update it.

gen/common/lexUtil.mll
ShaleXIONG: I update the comment:

The lexer consumes optional blanks around explicit syntax such as `[`, `]`, `,`, `|`, and `?`, because, for backward compatibility, standalone blanks are interpreted as `COMMA` after an concreate relax.

gen/common/parser.mly
ShaleXIONG: How does type work here since the `Ast.t` has`Choice of Ast.t list`. I think you mean `Choice [A; B; C]` is equivalent to `Choice [A; Choice [B; C] ]` by downstream functions?

I think we can simplify to be a binary tree version, e.g.  `Choice of (Ast.t * Ast.t)`. I will call the `flatten` (or better function name `to_list`) which flattens the tree to lists.

gen/common/parser.mly
ShaleXIONG: I see. I will make the order explicit by more rules and comments. Will also add more test to `diy7-check.t`

gen/tests/diy7-check.t
ShaleXIONG: I will write a small frontend in `gen/tests` to specifically test the parser and lexer. Also will remove the parser tests here indirectly use `diy7`

gen/common/parser.mly
fsestini: > How does type work here since the `Ast.t` has`Choice of Ast.t list`. I think you mean `Choice [A; B; C]` is equivalent to `Choice [A; Choice [B; C] ]` by downstream functions?

Yes, sorry, I think I meant `Choice [Choice lhs; Choice rhs] ~ Choice (lhs @ rhs)` there. Essentially the first pattern match clause of `mk_choice`. This was just an example of a pair of expressions that ought to be treated equivalently by downstream functions.

> I think we can simplify to be a binary tree version, e.g. `Choice of (Ast.t * Ast.t)`. I will call the `flatten` (or better function name `to_list`) which flattens the tree to lists.

I think the type/constructors are fine as they are. My question is specifically about whether we actually need to normalise/simplify expressions via `mk_seq`/`mk_choice`.

ShaleXIONG: I see what you mean. In short there is no difference semantically between `Choice [A; Choice [B; C] ]` and `Choice [A; B; C]` for example. Here it is a simple structure clean up by construction.

Do you know what is the convention here?

gen/common/parser.mly
ShaleXIONG: Following your other comment on tests directly on parser. I will add those tests using the new small `parser_frontend.ml`.

fsestini: You don't actually need to run this `parser_frontend` program within the cram tests. You can just write `parser_frontend.ml` as a regular OCaml script producing some output, and have `dune` diff that output with some expected results.

See https://dune.readthedocs.io/en/stable/tests.html#custom-tests

See how the `tests` stanza is declared for `uoaim`: https://github.com/herd/herdtools7/blob/master/tools/uoaim/test/dune

In your case, you can:
* Move the input test cases listed in `parser-check.t` into a plain list inside `parser_frontend.ml`
* In `parser_frontend.ml`, do a `List.iter` over those input test cases and print output to stdout. The printing part should already be there in your current `parser_frontend.ml`
* Add an `.expected` file: from the project root, run `touch gen/tests/parser_frontend.expected`
* Run tests, then update the `.expected` file with `dune promote`. 
* Remove the `parser-check.t` cram test. Perhaps also rename `parser_frontent.ml` to something else (like `test_parser.ml`), since it wouldn't be a frontend anymore.

ShaleXIONG: I just updated. Thanks for the guide !

gen/atomize.ml
ShaleXIONG: Will change back.

gen/atomize.ml
ShaleXIONG: Since our "Ast.t" does not have `empty`, so we pre-filter out whitespace-only input such as `"  "` is necessary. In previous parser, it will returns an error.

gen/diy.ml
ShaleXIONG: Will move it into the function.

gen/diy.ml
ShaleXIONG: Will add more comment.

Parse each `-prefix` argument separately, then combine them as one top-level choice. Thus `-prefix A -prefix B` is interpreted as `-prefix [A|B]`.

gen/diy.ml
ShaleXIONG: Done.

gen/diyone.ml
ShaleXIONG: Let me change to explicit match.

gen/diycross.ml
ShaleXIONG: Will do:

Each command-line argument in `input` denotes one segment of the diycross cycle. Parse those arguments separately, then rebuild one op-level `Seq` before expansion. For example `diycross7 A 'B|C' D` is treated as `A,(B|C),D`.

gen/relax.ml
ShaleXIONG: Done.

gen/relax.ml
ShaleXIONG: Let make it more explicit `edges_of_relax_list`

gen/relax.ml
ShaleXIONG: Done.

gen/diy.ml
ShaleXIONG: Let me rework here to parse individually and manipulate in `Ast.t` level.

In short I manually concatenate, because `-safe ***` can have two consecutive inputs `-safe [A B] [C D]` and `-safe [E F] [G H]` where the `Config.safe` is a list of `["[A B] [C D]"; "[E F] [G H]"]`. I concatenate the entire list by `" "` so to create the equivalent input one larger`-safe [A B] [C D] [E F] [G H]`.

gen/norm.ml
ShaleXIONG: This two errors are indistinguishable at the moment in `diy`.  Another ref location is: line 401 in function `call_rec_base` in `alt.ml` which is used in the searching algorithm in `diy7`.

Let me print different messages at least.

gen/diy.ml
ShaleXIONG: I changed it to use the new parser followed by sanity check.

gen/relax.ml
ShaleXIONG: Done.

gen/diyone.ml
ShaleXIONG: Done.

gen/common/ast.ml
ShaleXIONG: Update I have remove the old `to_list` function which is not necessary anymore due to other change. The old `flatten` is renamed to `to_list`, which is actually the conventional name.

gen/common/ast.ml
ShaleXIONG: Function remove.

gen/common/ast.ml
ShaleXIONG: Function remove.

gen/common/ast.ml
ShaleXIONG: The multi represents the square bracket `[...]`. Will put comment there.

I added parser explanation at  beginning of the `-help`.

ShaleXIONG:
> Thanks for this @ShaleXIONG . I've left a few comments in the code, and I have two overarching concerns:
> 
> * The lexer uses global mutable state, in a way that I think is quite error prone. I'm also not convinced it's really necessary. At the very least, I'd like to see it replaced with _local_ state. See the comments in the code.
> * This PR introduces a new parser for existing syntax, and new user-facing syntax for relaxations. The parsing semantics of both the old and new syntax is not trivial (for example, white space is interpreted in a context-dependent way, and `,` has different meaning in `diy7` vs `diycross7`). For this reason, I think we need significantly more tests than what the PR currently offers. Again, see the comments for additional details.
> 
> Happy to have a chat online/offline about any of these points.

@fsestini  I have addressed all your comments in several new commits. Can you have a look? Once we settle, I will merge and rewrite the history.

gen/common/lexUtil.mll
```diff
@@ -15,49 +15,55 @@
 (****************************************************************************)
 
 {
+open Parser
+exception Eof
```
fsestini: It seems this line is superfluous.

gen/common/lexUtil.mll
fsestini: Sorry, but the comment still mentions 'operand', so my point about it not being clear what 'operand' means still applies.
```suggestion
(* Track whether the lexer is currently inside a square bracket pair and has consumed a relaxation string.
   This is because whitespace after a relaxation within brackets is treated as a sequence separator. *)
```

gen/common/parser.mly
fsestini: Sorry, I should have already replied to this, but for some reason my message didn't get sent.

Unless downstream function depend on these expressions being simplified, I would simply not normalise/simply here and just do `let mk_seq x y = Seq [x; y]`.

gen/diy.ml
```diff
@@ -311,34 +296,35 @@ let () =
...
+    | _ -> (* The common path to generate tests *)
+      M.go !Config.size reject relax safe;
     exit 0
   with
-  | Misc.Fatal msg | Misc.UserError msg->
+  | Misc.Fatal msg ->
     eprintf "%s: Fatal error: %s\n" Config.prog msg ;
     exit 2
+  | Misc.UserError msg->
+    eprintf "%s: User error: %s\n" Config.prog msg ;
```
fsestini:
```suggestion
    eprintf "%s: %s\n" Config.prog msg ;
```

gen/tests/diy7-check.t
fsestini: I don’t think this has been fully addressed quite yet. The new `gen/tests/test_parser.ml` script is a great start, but it only tests the generic parser/AST layer. It checks `Parser.main` + AST expansion, but not the actual tool-specific parsing semantics.

As I mentioned in my comment, I think we should also have OCaml tests for the parsing paths used by the different `diy*` tools, since they do not all interpret the parsed AST in the same way. In particular, I would like to see direct OCaml tests for:
  - `diy7`’s top-level sequence-as-choice behavior, and semantics of multiple `-relax` and `-safe` options
  - `diyone7`’s requirement that the input expand to exactly one cycle
  - `diycross7`’s interpretation of "comma `,` as choice", and semantics of multi-argument combinations like `diycross7 A 'B|C' D`
  - possibly some additional tests covering:
      - fence and `-cumul` parsing
      - the newly-introduced invalid-relax filter

Again, it might be the case that `gen/diy.ml`, `gen/diyone.ml`, and `gen/diycross.ml` need a bit of extra work so that the corresponding parsing logic can be exposed through small functions that can be tested directly from OCaml tests.

gen/diycross.ml
```diff
@@ -53,25 +45,33 @@ module Make (Config:Config) (M:Builder.S) =
...
+      M.R.parse_ast segment |> Ast.seq_as_choice
 
-    let expand_edge es = M.E.expand_edges es Misc.cons
-    let expand_edges ess =
-      List.flatten (List.map (fun e -> expand_edge e []) ess)
+    let parse_edges input =
+      (* Each command-line argument in `input` denotes one segment of the
+         `diycross` cycle. Parse those arguments separately, then rebuild one
+         top-level `Seq` before expansion. For example, `diycross7 A 'B|C' D` is
+         treated as `A,(B|C),D`. *)
```
fsestini, line 62: I find this comment a bit confusing, because in `diycross7` the comma `,` means choice. Therefore, given the expression `A,(B|C),D` here, it's not clear if `,` means "choice" or "sequence".
Could we perhaps use a different, non-ambiguous symbol to pretty-print AST-level sequence?

gen/diycross.ml
```diff
@@ -53,25 +45,33 @@ module Make (Config:Config) (M:Builder.S) =
...
           let module Fold = struct
             type atom = M.E.atom
             let fold f k = M.E.fold_atomo_list atoms f k
           end in
           let module V = VarAtomic.Make(M.E)(Fold)  in
-          List.map V.varatom_es
+          V.varatom_es
+
+    let parse_segment segment =
+      M.R.parse_ast segment |> Ast.seq_as_choice
```
fsestini: The new comment added to `Ast.seq_as_choice` would make one think that this function is only relevant to diy7, however here we are also using it for diycross7.

gen/common/ast.mli
```diff
@@ -0,0 +1,24 @@
...
+  | Choice of 'prim t list
+
+val bind : ('a -> 'b t) -> 'a t -> 'b t
+val pp : ('a -> string) -> 'a t -> string
+
+(* Legacy `diy7` parsing interprets a plain top-level sequence as a choice
+   for backward compatibility. If the top-level sequence already mixes in an
+   explicit `Choice`, keep the sequence structure so that `A|B,C` means
+   `(A|B),C` rather than `A|B|C`. *)
+val seq_as_choice : 'a t -> 'a t
```
fsestini: I get why this function is needed, but would you be willing to consider whether `seq_as_choice` can be replaced with tool-specific parser entry points instead? In other words, rather than always parsing into one generic AST shape and then rewriting that AST with `seq_as_choice` depending on which `diy*` tool is going to consume it, how about encoding those top-level distinctions directly in the parser entry points for the respective tools, while still sharing most of the underlying grammar? It seems to me that this would make the semantics easier to follow, and would avoid introducing tool-specific meaning later via an AST rewrite. It might also allow us to get rid of the `Seq` constructor entirely (because now different per-tool parsers can just interpret whitespace as either `Choice` or `Multi`, without the need for an intermediate `Seq` operator).

gen/common/ast.ml
fsestini: Would you mind documenting _in the `ast.mli` interface file_ the meaning of _each `Ast.t` constructor_?

gen/common/ast.ml
fsestini: I see that you've removed the previous `to_list` function, however you have also renamed `flatten` to `to_list`, and much of the same criticism I had for the previous `to_list` implementation still apply to the current one, thus I feel my comment is still unaddressed.

I'm not sure I agree that `to_list` is a conventional name for the function that is currently `Ast.to_list`.
This is not really a generic “to list” conversion, but rather an expansion of the AST into the set of concrete sequences it denotes. I think the function should be renamed accordingly. Perhaps `expand_sequences` or just `expand`.

gen/common/ast.mli
```diff
@@ -0,0 +1,24 @@
...
+
+(* Flatten the AST into the list of concrete sequences.
+   - `One` and `Multi` directly upwraps
+   - `Opt` contributes the either `'a list` or empty,
+   - `Choice` unions alternatives,
+      for example Choice [A;B] becomes [[A];[B]]
+   - `Seq` computes the cross-product in order.
+      for example Seq [A;B] becomes [[A*B]] *)
```
fsestini: Some comments:
* The comments says "flatten" but it doesn't feel like this function is "flattening", but rather "enumerating" or "expanding".
* `upwraps` might be a typo of "unwraps".
* `[[A*B]]` is not valid OCaml, and it's not super clear what is represents. Perhaps a concrete example is more apt to explain `Seq`.
* 
```suggestion
(** Enumerate all concrete sequences represented by an AST expression.
     The result is a list of alternatives, where each alternative is one fully
     expanded sequence of primitive elements.
     Alternatives are introduces by `Opt` and `Choice` operators.

   Semantics:
     - `One x` expands to `[[x]]`
     - `Opt t` expands to the alternatives of `t`, plus the empty sequence
     - `Choice ts` concatenates the alternatives of each branch
     - `Seq ts` computes the ordered cross-product of its children.
        For example: `Seq [Choice [One x; One y]; One z]` expands to `[[x; z]; [y; z]]`
     - `Multi t` is grouping only; it expands exactly like `t` *)
```

gen/common/config.ml
```diff
@@ -141,6 +141,20 @@ let parse_cumul = function
   | "true" -> All
   | s -> Set s
 
+let parser_syntax_doc =
+  "Parser syntax: whitespace or ',' for sequence, '|' for choice, '?' for optional, and '[...]' for grouping, for examples:\n\
+    - 'A B' means the sequence A,B.\n\
+    - 'A|B,C' and '[A|B] C' both mean the choice between A and B, then the sequence with C.\n\
+    - '[A,B]?' means either the group '[A,B]' or the empty '[]'.\n\
+    - 'A|B,C|[D,E]?' parses as '(A|B),(C|([D,E]?))'.\n\
+   Depending on the tool and context, a sequence may be interpreted either as a 'followed-by' relation between relaxations or as a choice between inputs."
```
fsestini: I find the new "Parser syntax" help text rather parser-/grammar-centric, and IMO not that helpful from a user perspective.  I also think this documentation text should be tailored _per tool_, and should describe the accepted input forms in user-facing terms rather than in terms of "parser syntax" (which users shouldn't really be concerned with) or "sequences". 
The term "sequence" in particular doesn't seem to be well-defined from the user's point of view: is the "sequence" `A,B` denoting a composite relaxation? It is some other kind of list? Is it a choice/alternative between two relaxations? I'm not super convinced that this help message provides sufficient answer to these questions.

In particular, users need to understand things like:
  - how to write a composite relaxation
  - how and when to write alternative candidate relaxations with `|`
  - what `?` applies to and what is its semantics
  - how top-level whitespace and `,` is interpreted in each `diy*` tool

As written, the current text feels too generic, and the list of parsing examples does not seem very informative. I think the examples should be written in the form of concrete commands `diy7 -relax "A B...` that show how these operators behave concretely.

gen/common/edge.ml
```diff
@@ -198,6 +198,16 @@ and module RMW = A.RMW = struct
...
 
+  let pre_parse_string s =
+    String.trim s
+    |> (fun s -> Lexing.from_string s)
+    |> LexUtil.parse Parser.main
+    |> Ast.to_list
+    |> ( function
+      | [x] -> x
+      | _ ->
+        Warn.user_error "only accepts exactly one input cycle." )
```
fsestini: Similar to my previous comment: perhaps we could also replace this uses of `|> fun -> ...` and `|> function ...` with a plain old pattern match. If you agree, could you also go through the rest of the PR to apply the same fix to similar instances?

gen/common/ast.mli
```diff
@@ -0,0 +1,24 @@
+type 'prim t =
+  | One of 'prim
+  | Opt of 'prim t
+  | Multi of 'prim t
+  | Seq of 'prim t list
+  | Choice of 'prim t list
+
+val bind : ('a -> 'b t) -> 'a t -> 'b t
```
fsestini: Nitpick, but conventionally, monadic bind has signature `'a t -> ('a -> 'b t) -> 'b t` (i.e. the "continuation" function is the second parameter). What you wrote is usually named `flat_map` (or `concat_map` in lists).

gen/diy.ml
```diff
@@ -34,17 +33,64 @@ module Make(C:Builder.S)(O:DiyConfig) = struct
...
+  (* Parse the `-cumul` argument, which should be a list of individual fences. *)
+  let parse_fences input_fences =
+    let ast = parse_argument_ast input_fences in
+    (* Note that `parse_fence` might fail *)
+    let fences = Ast.bind ( fun input -> Ast.One(C.E.parse_fence input) ) ast
+    |> Ast.to_list in
+    (* Each expanded alternative must contain exactly one fence. *)
+    if List.for_all ( function [_] -> true | _ -> false ) fences then
+      List.flatten fences
+    else Warn.user_error "%s is not a list of fence" input_fences
```
fsestini, line 48: I don’t quite follow the current implementation of `parse_fences`. If `-cumul` is only meant to accept a simple list of whitespace-separated fences, this feels rather roundabout: we first parse it through the general AST/parser machinery, then expand it, and then reject anything that does not collapse back to singleton fences. Would it not be simpler and clearer to not go through the whole AST and instead use a dedicated parser for `-cumul`, which only accepts this much smaller syntax? That would make both the code and the user-facing semantics easier to understand.

For instance, if the accepted syntax is really just "a list of fence names separated by whitespace", then how about:

```
let split_words s =
    Str.split (Str.regexp "[ \t\r\n]+") (String.trim s)

let parse_fences input =
    split_words input
    |> List.map C.E.parse_fence
```

EDIT: oh I see, I think the point is to allow choice operators `|` to be used in `-cumul`?

gen/common/ast.ml
ShaleXIONG: Ahh sorry I miss this and let me update it.

gen/common/ast.ml
ShaleXIONG: Let do with `expand`.

gen/common/lexUtil.mll
ShaleXIONG: Remove. It was previously used to raised error but in the end I do no think I need it.

gen/common/lexUtil.mll
ShaleXIONG: I updated it further since it is a stack:
The stack records, for each scope, whether the lexer has consumed a relaxation string. Scopes are either top-level or inside square brackets. This is because whitespace after a relaxation in the same scope is treated as a sequence separator.

gen/common/ast.mli
ShaleXIONG: I apply the change and add a more involved concrete example:
```
 A concrete example is
   `Seq [Choice [One x; One y]; Opt (Multi (Seq [One z; One k]))]`,
   which expands to `[[x]; [x; z; k]; [y]; [y; z; k]]`, where:
   - `[x]` comes from `One x` followed by the empty alternative of `Opt`.
   - `[x; z; k]` comes from `One x` followed by `Multi (Seq [One z; One k])`.
   - `[y]` comes from `One y` followed by the empty alternative of `Opt`.
   - `[y; z; k]` comes from `One y` followed by `Multi (Seq [One z; One k])`.
```

gen/common/ast.mli
ShaleXIONG: Ahh, I will change to the more convention type of monadic bind.

gen/diy.ml
ShaleXIONG: You are correct. Partially we want the parsing code to be in one place so we do not need to maintain another parsing related function. Partially we will allow the use of `|` syntax.

gen/common/parser.mly
ShaleXIONG: That will make things very different because we sometime do top-level manipulation to the `Ast.t`.
For example a top level rewrite from `Seq[A; B; C]` to `Choice [A;B;C]` is different from `Seq [A; Seq [B;C]]` to `Choice[A;Seq[B;C]]`.

gen/diy.ml
ShaleXIONG: Done. Also made the consistent change to `norm.ml`.

gen/common/edge.ml
ShaleXIONG: Will do !

gen/common/parser.mly
fsestini: > That will make things very different because we sometime do top-level manipulation to the `Ast.t`. For example a top level rewrite from `Seq[A; B; C]` to `Choice [A;B;C]` is different from `Seq [A; Seq [B;C]]` to `Choice[A;Seq[B;C]]`.

That is precisely what I meant with "unless downstream function depend on these expressions being simplified". It seems you are saying that indeed, downstream uses of `Ast.t` do rely on the assumption that the input has been normalised/simplified. In that case, I think the current setup is too brittle: `mk_seq` and `mk_choice` ensure that `Ast.t` expressions coming out of the parser are normalised, but nothing is stopping other parts of the code from creating `Ast.t` trees that are _not_ normalised, or to modify existing trees in a way that breaks the normalisation.

I think my preferred solution out of this is to just not do "top-level manipulation" and remove the "sequence as choice" ambiguity by using different, dedicated parsers for different tools (see other comment [here](https://github.com/herd/herdtools7/pull/1685#discussion_r3086630516)). But alternatively, we might want to consider turning `Ast.t` into some kind of abstract type, so that external code cannot inadvertently construct non-normalised trees.

gen/common/ast.mli
ShaleXIONG: Let me explore the idea directly baking into the top level constructor in the lexing phrase.
I think it should be possible to remove either `Multi` or `Seq`. Maybe keep `Seq` is bette idea: (1) more conventional name and (2) the type works better for our needs `Seq of t list`, which we do need a list.

gen/diycross.ml
ShaleXIONG: seq_as_choice has been removed without problem. We can move the functionality to lexer

gen/diycross.ml
ShaleXIONG: I will unify the pretty-print by print `[...]` rather than `(...)` I think I mean the canonical form here. I will make it more clear in the comments

gen/common/ast.mli
ShaleXIONG: Just update, with a `is_backwards_compatible`, default is `true`. It enables directly parsing the top level `,` and whitespace to `Choice`. Then I removed `Multi`. Separate I change the `pp` so the `pp` result is a valid input syntax too but in canonical form.

gen/common/config.ml
ShaleXIONG: I will update based on different tools and give example in different tool after the common parser wording.

For `diyone7` and `diycross7` the explanation is at the top. For `diy7` the explanation is at the argument level since only `-safe`, `-relax` etc accept relaxation input.

gen/common/parser.mly
ShaleXIONG: I have changed the lexer so `seq_as_choice` is removed. I changed the mk_seq and mk_choice and introduce a normalise function in ast.

gen/tests/diy7-check.t
ShaleXIONG: - `diy7` is tested in the existing regression tests, where all the old behaviour is preserved. I can add a few equivalence check on the new syntax and old syntax.
- `diyone7`, I have added one test previously; I will add more tests on invalid cycles
- `diycross7` similar, the regression tests check all the previous behaviour is preserved. I will add a few equivalence check on the new syntax in the regression tests
- I add test for `-cumul` and removal of invalid relax list via `diy7 -filter-check`. The `-cumul` basically affects the filter. The latter test for removal of invalid relax list can be replace for `diy7 -unfold-only`, which is a future pull request I will open.

ShaleXIONG:
@fsestini Hi, I just updated all the comments can you have a look, thanks !

fsestini:
Summary of offline discussion to move this PR forward. @ShaleXIONG  let me know if anything is incorrect or missing.

- We should aim to simplify the lexer. In particular, I don't think it should include branching logic for "top-level" vs "backward-compatible" modes, or emit dedicated "top-level" tokens. IMO, it should be responsibility of the parser to determine top-level and tool-specific semantics.

The following points expand and further clarify my [previous comment](https://github.com/herd/herdtools7/pull/1685#discussion_r3086517498), which I think it's still not fully addressed yet.

- The parser should expose dedicated entry points for the different tool-level semantics (`diy7`, `diyone7`, `diycross7`, etc.). These entry points can share common grammar pieces, but the tools do have genuinely different parsing semantics, and I think that should be reflected more clearly in the grammar and parsers.

- These tool-specific parsing functions should be tested with dedicated OCaml unit tests, rather than mainly through broad end-to-end tests as proposed by the PR. The current approach validates parser behavior indirectly via integration/regression tests (e.g. `diycross-syntax` in `Makefile`), which I think are not very effective to test parsing behaviour specifically, since:

  - they don't have sufficient failure locality: a test failure could originate from a bug in any of: parsing, expansion, cycle generation, pretty-printing, herd7, etc., or the test infrastructure itself. Thus it may be very difficult to trace a test failure to the corresponding parser bug, or to determine that such failure is caused by a parsing bug in the first place.
  - conversely, because these tests cover a long pipeline of steps at once, a parser bug could end up being masked by later stages and not result in any detectable test failure.

  IMO parsing should be tested directly because this PR touches on subtle, tool-specific parsing semantics, and those are more effectively verified by testing the parsing functions in isolation, without interference from later steps.

- On of the main points of difference between tools is how they interpret the `,` symbol. E.g. it has the meaning of "disjunctive choice" in diycross7 and "sequence" in diyone7. This PR affects this behaviour, so the test suite should aim to provide enough confidence that backwards compatibility is not broken. IMO we are not quite there yet, for example there are no tests checking how diyone7 interprets `,` (which is different from how diycross7 interprets it). We should take a second look at the test suite and add any missing cases. Moreover, we should double check the outcome of the tests against `diy*` tools pre-PR.

- We should remove the `grep` step from the `-filter-check` test cases, as the test is more clear without it.

ShaleXIONG:
> Summary of offline discussion to move this PR forward. @ShaleXIONG let me know if anything is incorrect or missing.
> 
> * We should aim to simplify the lexer. In particular, I don't think it should include branching logic for "top-level" vs "backward-compatible" modes, or emit dedicated "top-level" tokens. IMO, it should be responsibility of the parser to determine top-level and tool-specific semantics.
> 
> The following points expand and further clarify my [previous comment](https://github.com/herd/herdtools7/pull/1685#discussion_r3086517498), which I think it's still not fully addressed yet.
> 
> * The parser should expose dedicated entry points for the different tool-level semantics (`diy7`, `diyone7`, `diycross7`, etc.). These entry points can share common grammar pieces, but the tools do have genuinely different parsing semantics, and I think that should be reflected more clearly in the grammar and parsers.
> * These tool-specific parsing functions should be tested with dedicated OCaml unit tests, rather than mainly through broad end-to-end tests as proposed by the PR. The current approach validates parser behavior indirectly via integration/regression tests (e.g. `diycross-syntax` in `Makefile`), which I think are not very effective to test parsing behaviour specifically, since:
>   
>   * they don't have sufficient failure locality: a test failure could originate from a bug in any of: parsing, expansion, cycle generation, pretty-printing, herd7, etc., or the test infrastructure itself. Thus it may be very difficult to trace a test failure to the corresponding parser bug, or to determine that such failure is caused by a parsing bug in the first place.
>   * conversely, because these tests cover a long pipeline of steps at once, a parser bug could end up being masked by later stages and not result in any detectable test failure.
>   
>   IMO parsing should be tested directly because this PR touches on subtle, tool-specific parsing semantics, and those are more effectively verified by testing the parsing functions in isolation, without interference from later steps.
> * On of the main points of difference between tools is how they interpret the `,` symbol. E.g. it has the meaning of "disjunctive choice" in diycross7 and "sequence" in diyone7. This PR affects this behaviour, so the test suite should aim to provide enough confidence that backwards compatibility is not broken. IMO we are not quite there yet, for example there are no tests checking how diyone7 interprets `,` (which is different from how diycross7 interprets it). We should take a second look at the test suite and add any missing cases. Moreover, we should double check the outcome of the tests against `diy*` tools pre-PR.
> * We should remove the `grep` step from the `-filter-check` test cases, as the test is more clear without it.

@fsestini. Lexer is now simplified and complexity has been moved to parser. The parser now have three entry point:
- `main`, parsing as expected without special manipulation for `diyone7`
- `main_top_level_choice`, parsing the top level `,` and whitespace as `Choice`.
- `cumul`, where all `,` whitespace and `|` are treated as `Choice`. This is for `diy7 -cumul {INPUT}`. The parser matches previous behaviours where the first level of square bracket is simply ignored, while nest square bracket is forbidden. It is better to expand the behaviour to allowed nested square bracket and ignore them in the parser.

The three parsers now have dedicated test cases. We also add extra test case for `remove_invalid_relaxes`. This means all the `-filter-check` test segments have been removed since it is unnecessary.

gen/alt.ml
```diff
@@ -690,9 +688,6 @@ module Make(C:Builder.S)
 
     let parse_input ~relax ~safe ~reject =
       let r_nempty = Misc.consp relax in
-      let relax = expand_relaxs C.ppo relax
-      and safe = expand_relaxs C.ppo safe
-      and reject = expand_relaxs C.ppo reject in
       if Misc.nilp relax then if r_nempty then begin
         Warn.fatal "relaxations provided in relaxlist could not be used to generate cycles"
       end ;
```
fsestini: Not sure I understand the point of this if statement here.

Since the line `let relax = expand_relaxs C.ppo relax` is now removed, then `Misc.nilp relax` and `Misc.consp relax` above are computed on the same exact list `relax`, so this `if` condition is never true, and the error is never reached.

However, there are cases where I think this error should be displayed, but isn't. For example, if we run this:

```
diy7 -arch AArch64 -relax '[Rfe, Rfe]'
```

it generates the same tests as `diy7 -arch AArch64`, which I think indicates that `[Rfe, Rfe]` has been filtered out as invalid. But in that case, shouldn't I be getting an error notifying that the `[Rfe, Rfe]` input was invalid?

gen/common/config.ml
```diff
@@ -141,6 +141,28 @@ let parse_cumul = function
   | "true" -> All
   | s -> Set s
 
+let parser_syntax_doc =
+  "For the input cycle or cycle segments: '[...]' groups expressions, '|' means choice, and '?' makes the preceding expression optional.\n\
+   Inside '[...]', both whitespace and ',' mean sequence, so '[A B,C]' and '[A,B C]' all have canonical form '[A,B,C]'.\n\
+   At top level, plain whitespace and ',' may denote either choice or sequence, depending on the context."
```
fsestini: Thanks for splitting the help message by tool.

Since the help text is now specialized per diy* tool, can we avoid this rather vague "depending on the context" line here and instead document the actual meaning of `,` and whitespace for each tool within their respective help messages?

gen/common/config.ml
```diff
@@ -141,6 +141,28 @@ let parse_cumul = function
   | "true" -> All
   | s -> Set s
 
+let parser_syntax_doc =
+  "For the input cycle or cycle segments: '[...]' groups expressions, '|' means choice, and '?' makes the preceding expression optional.\n\
+   Inside '[...]', both whitespace and ',' mean sequence, so '[A B,C]' and '[A,B C]' all have canonical form '[A,B,C]'.\n\
```
fsestini: > Inside '[...]', both whitespace and ',' mean sequence

I think this is not true for `-cumul` as implemented.

fsestini (approved):
Thanks @ShaleXIONG . There's an issue with the way we report invalid relaxes that I think needs to be addressed (+ a couple of minor comments about the help messages), but otherwise I think this is fine overall. I'll approve, but please do take a look at those three comments I left before merging.

gen/common/config.ml
ShaleXIONG: ahh yes. Let me update here.

gen/alt.ml
ShaleXIONG: I will move the `if` check to `diy.ml` side. It means to check, as you suggested here, when user input `relax` but none of the `relax` input is valid after parsing.

gen/common/config.ml
ShaleXIONG: I will add more parsing information to `cumul`.

ShaleXIONG:
> Thanks @ShaleXIONG . There's an issue with the way we report invalid relaxes that I think needs to be addressed (+ a couple of minor comments about the help messages), but otherwise I think this is fine overall. I'll approve, but please do take a look at those three comments I left before merging.

Thanks for all the great review. I just rebased to upstream, rewrote the history and will merge after CI.

PR #1719, opened 2026-02-17
[herd] More implicit transitive relations in the Cat interpreter

maranget:
We notice that `(...|r+|...)+` is `(...|r|...)+` and thus also that irrefexivity of  `(...|r+|...)+` is equivalent to acyclicity  of `(...|r|...)`
    
The process may be interesting because we also identify some of the recursive definitions that are equivalent to a transitive closure and represent them as (implicit) transitive closure, thereby saving some transitive closure operations.

However, to avoid the repetive evaluation of bound implicit transitive closures to explicit ones, we rely on an additionnal environment for implicitly   transitive relations, introducing minor extra costs.

 In the end, improvements are barely noticeable, but the interpreter should gain as regards performance robustness.

### DIFF
--- a/lib/ASTUtils.ml
+++ b/lib/ASTUtils.ml
@@ -202,3 +202,28 @@ let free_body xs e =
     (fun r p0 -> remove_pat0 p0 r)
     (free e)
     xs
+
+
+let rec as_plus_args saw_seq x = function
+  | [] -> [],saw_seq
+  | e::es ->
+      if StringSet.mem x (free e) then
+        match e with
+        | Op (_,Seq,[Var (_,x1);Var (_,x2)]) ->
+            if String.equal x1 x && String.equal x2 x then
+              as_plus_args true x es
+            else raise Exit
+        | _ -> raise Exit
+      else
+        let es,saw_seq = as_plus_args saw_seq x es in
+        e::es,saw_seq
+        
+let as_plus = function
+  | [_,Pvar (Some x),Op (_,Union,es)] ->
+      begin
+        try
+          let es,saw_seq = as_plus_args false x es in
+          if saw_seq then Some (x,es) else None
+        with Exit -> None
+      end
+  | _ -> None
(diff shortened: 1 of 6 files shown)


lib/ASTUtils.ml
```diff
@@ -202,3 +202,28 @@ let free_body xs e =
...
+      else
+        let es,saw_seq = as_plus_args saw_seq x es in
+        e::es,saw_seq
+        
+let as_plus = function
+  | [_,Pvar (Some x),Op (_,Union,es)] ->
+      begin
+        try
+          let es,saw_seq = as_plus_args false x es in
+          if saw_seq then Some (x,es) else None
```
HadrienRenaud, line 226: Could we have a warning that the `rec` flag is useless in that case?

HadrienRenaud, line 226: Why not reconstructing the `Union` here? This way all the AST manipulations are done in `ASTUtils`

lib/interpreter.ml
```diff
@@ -1703,6 +1725,28 @@ module Make
...
             else eval accept_implicit env ifnot
 
+(*
+ * Evaluation of `(e1|..|en)+`
+ * Hence, `e1`, ...,`en` must be relations.
+ *)
+      and eval_union_plus env es =
+        try
+          let rs =
+            List.fold_right
```
HadrienRenaud: why not `List.fold_left`?

lib/ASTUtils.ml
maranget: I'd rather leave this to another PR.

maranget: If the recursive definition is equivalent to the transitive closure (`+`) of union of `es`, then the optimising function `eval_union_plus` (in `lib/interpreter.ml`) is applied to `es`.

lib/interpreter.ml
maranget: Done, thanks.

maranget:
Superseded by PR #1851.

PR #1727, opened 2026-02-23
[litmus] Enable check_dic_idc for PreSi and Kvm modes

diaolo01:
This PR enables PreSi and Kvm modes to check whether a core has DIC/IDC implemented. It also runs tests with `DIC=0 & IDC=0` as part of `make-aarch64-litmus`, while only compiling tests that require either DIC or IDC. 

=====

This has been identified as part of #1716. This fix has been tested against that patch and the desired behaviour was observed on an M2:

```
Fatal error: hardware does not support the CacheType feature IDC=1, DIC=1

EXIT: STATUS=1
```

The missing call was not flagged by the compiler because GCC/Clang do not warn about unused `static inline` functions by default.

### DIFF
--- a/Makefile.aarch64
+++ b/Makefile.aarch64
@@ -149,7 +149,13 @@ litmus-aarch64-run:: litmus-cata-aarch64-ifetch-test-kvm
 
 litmus-cata-aarch64-ifetch-test-kvm: litmus-aarch64-dep
 	mkdir $(KUT_DIR_AARCH64)/kvm-unit-tests/t
-	if $(RUN_TESTS); then NORUN=UDF+2FH; else NORUN=NO; fi ; \
+	if $(RUN_TESTS); then                                         \
+	  NORUN=UDF+2FH ;                                             \
+	  TESTS=catalogue/aarch64-ifetch/tests/@dic0-idc0 ;          \
+	else                                                         \
+	  NORUN=NO ;                                                 \
+	  TESTS=catalogue/aarch64-ifetch/tests/@all ;                \
+	fi ;                                                         \
 	$(LITMUS)					         \
 		-set-libdir $(LITMUS_LIB_DIR)	                 \
 		-o $(KUT_DIR_AARCH64)/kvm-unit-tests/t	         \
@@ -157,7 +163,7 @@ litmus-cata-aarch64-ifetch-test-kvm: litmus-aarch64-dep
                 -nonames $${NORUN}                               \
                 -driver C -ascall true                           \
                 -outnames $(KUT_NAMES)                           \
-		catalogue/aarch64-ifetch/tests/@all
+		$${TESTS}
 	cd $(KUT_DIR_AARCH64)/kvm-unit-tests/t; make $(SILENTOPT) -j $(J)
 	if $(RUN_TESTS); then \
           ( cd $(KUT_DIR_AARCH64)/kvm-unit-tests && sh t/run.sh ) | \
--- a/catalogue/aarch64-ifetch/tests/@dic0-idc0
+++ b/catalogue/aarch64-ifetch/tests/@dic0-idc0
@@ -1,0 +1,8 @@
+DIC0-IDC0/@all
+MP+rel+blr.litmus
+MP.FF+dc.cvau-dmb.ish+dsb.ish-ic.ivau-dsb.ish-rfiINSTNOP.litmus
+MP.FF+dc.cvau-dsb.ish-ic.ivau-dsb.ish+dmb.ish+rfiINST.litmus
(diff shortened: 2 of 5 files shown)


litmus/preSi.ml
```diff
@@ -2171,6 +2171,20 @@ module Make
...
+        if do_self then begin
+          let cache_type = CacheType.get test.T.info in
+          let needs_dic, needs_idc =
+            let open CacheType in
+            match cache_type with
+            | None -> (fun _ -> false), (fun _ -> false)
+            | Some cache_type -> cache_type.dic, cache_type.idc in
+          begin match forall_procs test needs_dic, forall_procs test needs_idc with
+          | Some dic, Some idc ->
+              O.fi "check_dic_idc(%d, %d);" (if dic then 1 else 0)
```
relokin: It seems like we are checking DIC/IDC on one core and we assume that it's the same for all cores? Is that a good assumption?

If yes can you please add a comment here?

diaolo01: Actually, this code was checking per core rather than a single core. The Arm ARM states: "All PEs in the same Inner Shareable shareability domain must have a common value of this field.".
I moved the check at the test level to match what is done for hardware updates. I have included a comment to capture the requirement above.

Makefile.aarch64
```diff
@@ -149,20 +149,30 @@ litmus-aarch64-run:: litmus-cata-aarch64-ifetch-test-kvm
 
 litmus-cata-aarch64-ifetch-test-kvm: litmus-aarch64-dep
 	mkdir $(KUT_DIR_AARCH64)/kvm-unit-tests/t
-	if $(RUN_TESTS); then NORUN=UDF+2FH; else NORUN=NO; fi ; \
```
relokin, line 152: Couldn't you just do something along the lines of:
```suggestion
	if $(RUN_TESTS); then                                         \
          NORUN=UDF+2FH                                         \
          TESTS=catalogue/aarch64-ifetch/tests/@dic0-idc0 \
    else                                                                        \
         NORUN=NP                                                      \
         TESTS=catalogue/aarch64-ifetch/tests/@all  \
    fi                                                                             \  
```
and then leave the recipe common for both?

catalogue/aarch64-ifetch/tests/@dic0-idc0
```diff
@@ -0,0 +1,9 @@
+DIC0-IDC0/@all
+MP+rel+blr.litmus
+MP.FF+dc.cvau-dmb.ish+dsb.ish-ic.ivau-dsb.ish-rfiINSTNOP.litmus
+MP.FF+dc.cvau-dsb.ish-ic.ivau-dsb.ish+dmb.ish+rfiINST.litmus
+MP.FF+dc.cvau-dsb.ish-ic.ivau-dsb.ish+po.litmus
+MP.FF+dc.cvau-dsb.ish-ic.vau-dsb.ish+dmb.ish+rfiINST.litmus
```
relokin: This tests doesn't exist

relokin (commented):
Two comments, otherwise this PR looks good to me.

Makefile.aarch64
diaolo01: This would result in the DIC1/IDC1 tests not being compiled when `RUN_TESTS=TRUE`, which I don't think is acceptable.

relokin: well with `RUN_TESTS=true`, we compile the DIC0-IDC0 tests twice.

TBF, right now, for the regressions we process these sets of tests:
* DIC0-IDC0/@all
* DIC0-IDC1/@all
* DIC1-IDC1/@all
But these lists contain precisely the same tests, the only difference is the CacheType they use. So I think we are wasting time.

diaolo01: I checked out of curiosity and found that the three suites are mostly the same with `CacheType` changes, but they are not identical. 

DIC0-IDC0 has three extra tests compared to the other two:
- `MP.RF+dc.cvau-dmb.ish+ctrl-ic.ivau-dsb.ish-isb.litmus`
- `MP.RF+dc.cvau-dmb.ish+pickctrl-ic.ivau-isb.litmus`
- `MP.RF+dmb.st+pickaddr-cachesyncisb.litmus`

DIC0-IDC1 has a test that isn't present in the other two suites: `IDC1.MP.FF+dsb.ish-ic.ivau-dsb.ish+po.litmus`.

Since the suites aren't strictly identical, shall we still proceed with only compiling and running the DIC0-IDC0 set? If so, should we add `MP.FF+dsb.ish-ic.ivau-dsb.ish+po.litmus` for DIC0-IDC0?

relokin (approved):
LGTM if you rebase on top of master we can merge this.

PR #1728, opened 2026-02-23
[herd-www] Add .opam file specifying dependencies

TiberiuBucur:
This PR addresses a mismatch between zarith versions >=1.13, required by herdtools7.opam, and zarith_stubs_js, required by herd-www. The oldest version of zarith_stubs_js that does not conflict with zarith 1.13 is v.0.16.1. This version requires Ocaml >=4.14.0.
This .opam file specifies these requirements, giving the user the ability to automatically install the right versions of the packages through `opam install ($OPAM_FILE_PATH) --deps-only` for herd-www, such that these will not conflict with the versions required by herdtools7.opam in the root of the repository.

### DIFF
--- a/herd-www/herd-www.opam
+++ b/herd-www/herd-www.opam
@@ -1,0 +1,18 @@
+opam-version: "2.0"
+synopsis: "Dependencies for herd-www (web interface for herdtools7)"
+maintainer: "Luc Maranget <Luc.Maranget@inria.fr>"
+authors: [
+  "Jade Alglave <j.alglave@ucl.ac.uk>"
+  "Luc Maranget <Luc.Maranget@inria.fr>"
+]
+homepage: "https://developer.arm.com/herd7"
+bug-reports: "http://github.com/herd/herdtools7/issues/"
+license: "CECILL-B"
+depends: [
+  "ocaml" {>= "4.14.0"}
+  "dune"
+  "menhir"
+  "js_of_ocaml" {= "5.4.0"}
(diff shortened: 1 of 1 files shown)


ShaleXIONG (approved):
LGTM

fsestini (commented):
A couple of comments:

1. Let's update CI workflows to use this opam file to solve and install dependencies. Let's also update the `README.md` in `herd-www` to inform the reader of the possibility to do `opam install . .. --deps-only` from within `herd-www/` to install dependencies. (I think `opam install . --deps-only` alone is not sufficient because `herd-www` depends on sources from the core `herdtools` package, which has further dependencies.)

2. Since `herd-www` builds against sources from the core `herdtools` package, it may be worth extending `herd-www.opam` to explicitly include the subset of dependencies from `herdtools7.opam` that are required to build `herd-www`. This would make `herd-www.opam` a more complete and declarative description of the dependencies needed to build `herd-www`, without relying indirectly on another opam file.

EDIT: I've double-checked in an empty opam switch and indeed just running `opam install . --deps-only` inside `herd-www` will miss `zarith`, which is required to build herdtools' `lib/`. Since it's a single dependency and we are already requiring `zarith_stubs_js` anyway, I think we should go with point (2) above and just add `zarith` to `herd-www.opam`.

TiberiuBucur:
Closing this as the .opam file has been added as of [this PR](https://github.com/herd/herdtools7/pull/1933).

PR #1735, opened 2026-03-02
[herd] More efficient computation of atomic load X stores pairs

maranget:
We take inspiration from the efficient computation of register read-from (see PR #1704): atomic stores are ordered according to (extended by iico) program order and an atomic load is paired with the closest atomic write that follows it.

Preliminary to PR #1733.

### DIFF
--- a/herd/mem.ml
+++ b/herd/mem.ml
@@ -901,12 +901,10 @@ struct
 
   let find_first_after e set =
     find_first_opt (fun e' -> is_before_strict e e') set
-  [@@warning "-32"]
+
 end
 
-let map_loc_find loc m =
-  try U.LocEnv.find loc m
-  with Not_found -> []
+let map_loc_find loc m = U.LocEnv.safe_find [] loc m
 
 let match_reg_events add_eq es csn =
   let loc_loads_stores = U.collect_reg_loads_stores es in
@@ -1827,40 +1825,74 @@ let get_rf_value test read =
         | _,_ -> k)
         rfm E.EventRel.empty
 
-(* Reconstruct load/store atomic pairs *)
+(*
+ * Reconstruct load/store atomic pairs,
+ *   By definition such a pair exists when the
+ * store precedes the load in generalised program order
+ * (_i.e._ the union of program order and of iico), and
+ * that there is no atomic effect to the same location
+ * in-between (w.r.t generalised po) the load and the store.
+ *   Computation proceeds as follows:
+ *   First, atomic events are grouped first by thread
+ * and then by location. Then, to each atomic load, we
+ * associate the closest generalised po successor store,
+ * by using a set of stores ordered by generalised po.
+ * We additionally check that no atomic load exists
+ * between the load and store. Notice that it is not possible
+ * to use a set of all atomic events (by a given thread and
+ * with a given location) ordered by po because some atomic loads
+ * may be unrelated.
+ *   Finally, such atomic pairs can be spurious, that is not performed
+ * by a specific thread. In that case, pairs are given
+ * simply by the intra causality data relation.
+ *)
 
     let make_atomic_load_store es =
-      let all = E.atomics_of es.E.events in
-      let atms = U.collect_atomics es in
-      U.LocEnv.fold
-        (fun _loc atms k ->
-          let atms =
-            List.filter
-              (fun e -> not (E.is_load e && E.is_store e))
-              atms in (* get rid of C RMW *)
-          let rs,ws = List.partition E.is_load atms in
-          List.fold_left
-            (fun k r ->
-              let exp = E.is_explicit r in
-              List.fold_left
-                (fun k w ->
-                  if
-                    S.atomic_pair_allowed r w &&
-                    U.is_before_strict es r w &&
-                    E.is_explicit w = exp &&
-                    not
-                      (E.EventSet.exists
-                         (fun e ->
-                           E.is_explicit e = exp &&
-                           U.is_before_strict es r e &&
-                           U.is_before_strict es e w)
-                         all)
-                  then E.EventRel.add (r,w) k
-                  else k)
-                k ws)
-            k rs)
-        atms E.EventRel.empty
-
+      let atms,spurious = U.collect_atomics es in
+      let module StoreSet = EvtSetByPo(struct let es = es end) in
+      let make_atomic_pairs es k =
+        let rs,ws = List.partition E.is_load es in
+        let ws = StoreSet.of_list ws
+        and intervening_read er ew =
+          List.exists
+            (fun e ->
+              StoreSet.is_before_strict er e
+              && StoreSet.is_before_strict e ew)
+            rs in
+        List.fold_left
+          (fun k er ->
+             match StoreSet.find_first_after er ws with
+             | Some ew ->
+                if
+                  S.atomic_pair_allowed er ew
+                  && not (intervening_read er ew)
+                 then
+                   E.EventRel.add (er,ew) k
+                 else k
+             | None -> k)
+          k rs in
+      let r1 =
+        List.map
+          (fun (_,m) ->
+           U.LocEnv.fold
+             (fun _loc es k ->
+                let exps,nexps = List.partition E.is_explicit es in
+                make_atomic_pairs exps @@ make_atomic_pairs nexps k)
+             m E.EventRel.empty)
+        atms |> E.EventRel.unions
+      and r2 =
+        let iico = es.E.intra_causality_data in
+        List.map
+          (fun e ->
+             if E.is_load e then
+               match
+                 E.EventRel.succs iico e |> E.EventSet.as_singleton
+               with
+               | None -> assert false (* spurious updates are by pairs *)
+               | Some w -> E.EventRel.singleton (e,w)
+             else E.EventRel.empty)
+          spurious |> E.EventRel.unions in
+      E.EventRel.union r1 r2
 
 (* Retrieve last store from rfmap *)
     let get_max_store _test _es rfm loc =
@@ -1955,12 +1987,9 @@ let get_rf_value test read =
           if C.debug.Debug_herd.mem then begin
             eprintf "Observed locs: {%s}\n" (pp_locations observed_locs)
           end ;
-          U.LocEnv.fold
-            (fun loc ws k ->
-              if keep_observed_loc loc observed_locs then
-                U.LocEnv.add loc ws k
-              else k)
-            loc_stores U.LocEnv.empty
+          U.LocEnv.filter
+            (fun loc _ ->  keep_observed_loc loc observed_locs)
+            loc_stores
         else loc_stores in
 
       let possible_finals =
--- a/herd/memUtils.ml
+++ b/herd/memUtils.ml
@@ -387,8 +387,25 @@ let lift_proc_info i evts =
   and collect_stores es = collect_by_loc es E.is_store
   and collect_loads_non_spec es = collect_by_loc es (fun e -> E.is_load e && not_speculated es e)
   and collect_stores_non_spec es = collect_by_loc es (fun e -> E.is_store e && not_speculated es e)
-  and collect_atomics es = collect_by_loc es E.is_atomic
   and collect_reg_loads_stores es = collect_by_loc2 es E.is_reg_load_any E.is_reg_store_any
+
+  let accumulate_loc_proc proc loc e =
+    IntMap.update proc @@
+      function
+      | None -> Some (LocEnv.singleton loc [e])
+      | Some m -> Some (LocEnv.accumulate loc e m)
+
+  let collect_atomics es =
+    let m,sp =
+      E.EventSet.fold
+        (fun e (m,sp as k) ->
+           if E.is_atomic e then
+             match E.proc_of e with
+             | None -> if E.is_spurious e then (m, e::sp) else k
+             | Some proc -> accumulate_loc_proc proc (get_loc e) e m, sp
+           else k)
+        es.E.events (IntMap.empty,[]) in
+    IntMap.bindings m,sp
 
   let partition_events es =
     let env =
--- a/herd/memUtils.mli
+++ b/herd/memUtils.mli
@@ -70,7 +70,7 @@ module Make : functor (S: SemExtra.S) -> sig
    relation *)
   val make_fr : S.concrete -> S.event_rel -> S.event_rel
 
-(* Mapping from locations *)
+(* Mapping from locations, a.k.a. location maps *)
   module LocEnv : MyMap.S with type key = S.location
 
 (* Collect various events, indexed by location *)
@@ -85,7 +85,19 @@ module Make : functor (S: SemExtra.S) -> sig
   val collect_stores : S.event_structure -> S.event list LocEnv.t
   val collect_stores_non_spec : S.event_structure -> S.event list LocEnv.t
   val collect_loads_non_spec : S.event_structure -> S.event list LocEnv.t
-  val collect_atomics : S.event_structure -> S.event list LocEnv.t
+
+  (*
+   * Collect atomic effects indexed by threads and by locations.
+   * When given an event structure as argument, the function
+   * returns a pair [(maps,evts)], where:
+   *  + [maps] is a list of location maps, one per thread.
+   *    The values of this map are the atomic effects
+        of the given thread.
+   *  + [evts] is the list of spurious (atomic) effects.
+   *)
+
+  val collect_atomics :
+    S.event_structure -> (Proc.t * S.event list LocEnv.t) list * S.event list
 
 (* Partition by location *)
   val partition_events : S.event_set -> S.event_set list


herd/mem.ml
```diff
@@ -803,7 +815,7 @@ let match_reg_events add_eq es csn =
       (* Add the corresponding store for each load *)
       List.fold_left
         (fun k load ->
-          let f e = is_before_strict e load in
+          let f e = StoreSet.is_before_strict e load in
           let rf =
             match StoreSet.find_last_opt f stores with
```
HadrienRenaud: Would it make sense to add a function:
```ocaml
val find_last_before : E.event -> StoreSet.t -> E.event option

(* ... *)

let find_last_before e set = find_last_opt (fun e' -> is_before_strict e' e) set
```

herd/mem.ml
```diff
@@ -1672,37 +1684,43 @@ let get_rf_value test read =
...
-                           U.is_before_strict es r e &&
-                           U.is_before_strict es e w)
-                         all)
-                  then E.EventRel.add (r,w) k
-                  else k)
-                k ws)
-            k rs)
-        atms E.EventRel.empty
-
+      let atms,spurious = U.collect_atomics es in
```
maranget, line 1851: I am unsure here. The new code computes atomic load X write pairs by associating each atomic load to the closest atomic store that follows it in po-order and that _accesses the same location_. I am not sure at all that this "access to the to same location" condition is correct.

herd/mem.ml
maranget: You mean add this function to the StoreSet module?

HadrienRenaud: Yes!

herd/mem.ml
maranget: The current code proceeds by considering locations and I now think this is correct:
  1. Atomic pairs are computed _after_ rf matching. As a consequence locations at determined.
  2. There cannot be an atomic  load X store pair accessing different locations (the store must fail, at least for herd)
 Beside in the case of `LDXP` X `STXP` pair, the `STXP` instruction  performs two po-unrelated stores, which we cannot distinguish without considering their locations.

herd/mem.ml
maranget: Why not, but this function will have one call site only, as `make_atomic_load_store` uses `StoreSet.find_first_opt` in place of `Store.find_last_opt`. But we spare one reference to `StoreSet` (twice, if we also define `StoreSet.find_first_after`):
```OCaml
let find_first_after e set = find_first_opt (fun e' -> is_before_strict e e')
```

maranget: I'll follow your suggestion (twice).

herd/memUtils.ml
```diff
@@ -387,9 +387,31 @@ let lift_proc_info i evts =
...
+             | Some proc ->
+                 IntMap.update proc
+                   (fun old ->
+                      let m =
+                        match old with
+                        | None -> LocEnv.empty
+                        | Some m -> m in
+                      Some (LocEnv.accumulate (get_loc e) e m))
+                   m,sp
+           else k)
```
HadrienRenaud, line 406: Why not extract the accumulation function?

```ocaml
let accumulate_loc_proc proc loc e =
  IntMap.update proc @@ function
  | None -> Some (LocEnv.singleton loc e)
  | Some le -> Some (LocEnv.accumulate loc e le)
```

Then

```suggestion
           if E.is_atomic e then
             match E.proc_of e with
             | None -> if E.is_spurious e then (m, e::sp) else k
             | Some proc -> (accumulate_loc_proc proc (get_loc e) e m, sp)
           else k)
```

herd/mem.ml
```diff
@@ -1797,12 +1817,12 @@ let get_rf_value test read =
...
-          U.LocEnv.fold
-            (fun loc ws k ->
+          U.LocEnv.filter_map
+            (fun loc ws ->
               if keep_observed_loc loc observed_locs then
-                U.LocEnv.add loc ws k
-              else k)
-            loc_stores U.LocEnv.empty
+                Some ws
+              else None)
```
HadrienRenaud: I believe this is directly `U.LocEnv.filter`:


```suggestion
          U.LocEnv.filter (fun loc _ -> keep_observed_loc loc observed_locs)
```

herd/mem.ml
```diff
@@ -1672,37 +1687,42 @@ let get_rf_value test read =
...
+        let rs,ws = List.partition E.is_load es in
+        let stores = StoreSet.of_list ws in
+        List.fold_left
+          (fun k load ->
+             match StoreSet.find_first_after load stores with
+             | None -> k (* No matching store (e.g. final load reserve) *)
+             | Some store ->
+                 if S.atomic_pair_allowed load store then
+                   E.EventRel.add (load,store) k
+                 else k)
```
HadrienRenaud: What if you have 2 CAS in a row at the same location, the first fails and the second succeeds. Do you have an atomic pair between the read in the first CAS and the write in the second?

If I understand correctly the previous implementation removed this case by checking that there was no atomic event at all between the pairs.
Maybe we can do this by putting all the atomic events in a `storeset`, hoping that they are all ordered, and then check that the first event after is indeed a store.

HadrienRenaud (commented):
This looks good to me, thanks @maranget. There is an important point about interleaving reads in a comment that I think needs to be addressed before this is merged. Maybe a justification of why this is not a problem is enough.

herd/memUtils.ml
maranget: Done, much clearer, thanks.

herd/mem.ml
maranget: Nice catch.

herd/mem.ml
maranget: Oups, you are right. We cannot put all atomic events in a storeset because some of them are unrelated by generalised po (consider load or store pairs, or mixed-size mode). It remains doable by considering atomic events that access the same location, as this PR does for stores.

In fact this is a small modification of the code above : replace `storeset` by a set of all atomic effects that access this location and additionally  check that the next po-atomic is a store, as you suggest.

HadrienRenaud: Maybe worth adding a test that checks this?

maranget: I have pushed the modified code.

maranget:
The complete test `make test-all` fails on `./herd/tests/instructions/AArch64.vmsa+mte/A016.litmus`...

maranget:
In some cases, such as combining vmsa and mte, there exist (atomic) read effects that are unrelated by generalised po. As a consequence it is not possible to build a set of effects that rely on generalised po.

I have reverted to finding the closest po-after write (and not closest effect) and added a naive non-intervening load test. As writes are  po-ordered. It works.

herd/mem.ml
```diff
@@ -1801,40 +1816,74 @@ let get_rf_value test read =
...
-
+      let atms,spurious = U.collect_atomics es in
+      let module StoreSet = EvtSetByPo(struct let es = es end) in
+      let make_atomic_pairs es k =
+        let rs,ws = List.partition E.is_load es in
+        let ws = StoreSet.of_list ws
+        and intervening_read er ew =
+          List.exists
+            (fun e ->
+              StoreSet.is_before_strict er e
```
HadrienRenaud, line 1859: At this point, shouldn't we construct `is_before_strict` outside of the storeset?

herd/mem.ml
```diff
@@ -864,28 +864,45 @@ and get_written e = match E.written_of e with
...
 
-let map_loc_find loc m =
-  try U.LocEnv.find loc m
-  with Not_found -> []
+let map_loc_find loc m = U.LocEnv.safe_find [] loc m
+
+(* Event set ordered by to (generalised) program order *)
+
+module
+  EvtSetByPo
```
HadrienRenaud: Question: should this belong in `memUtils`?

herd/memUtils.ml
```diff
@@ -387,9 +387,25 @@ let lift_proc_info i evts =
   and collect_stores es = collect_by_loc es E.is_store
   and collect_loads_non_spec es = collect_by_loc es (fun e -> E.is_load e && not_speculated es e)
   and collect_stores_non_spec es = collect_by_loc es (fun e -> E.is_store e && not_speculated es e)
-  and collect_atomics es = collect_by_loc es E.is_atomic
   and collect_reg_loads_stores es = collect_by_loc2 es E.is_reg_load_any E.is_reg_store_any
 
+  let accumulate_loc_proc proc loc e =
+  IntMap.update proc @@ function
```
HadrienRenaud: nit: indentation is wrong

herd/mem.ml
```diff
@@ -1801,40 +1816,74 @@ let get_rf_value test read =
...
+      let make_atomic_pairs es k =
+        let rs,ws = List.partition E.is_load es in
+        let ws = StoreSet.of_list ws
+        and intervening_read er ew =
+          List.exists
+            (fun e ->
+              StoreSet.is_before_strict er e
+              && StoreSet.is_before_strict e ew)
+            rs in
+        List.fold_left
```
HadrienRenaud, line 1862: do we know if each load is totally ordered wrt the stores? If so, we could do 2 passes:
1. Compute for each load the first store after, and store in the map from store to reads
2. For each store, only check the intervening reads for the reads that are before the next store

Something like:
```ocaml
let intervening is_before_strict es e1 e2 =
  List.exists (fun e -> is_before_strict e1 e && is_before_strict e e2) es

let make_atomic_pairs es k =
  let rs, ws = List.partition E.is_load es in
  let store_to_reads_map =
    List.fold_left
      (fun map er ->
         match StoreSet.find_first_after er ws with
         | Some ew -> E.EventMap.accumulate ew er map
         | None -> map)
      E.EventMap.empty rs
  in
  E.EventMap.fold
    (fun ew ers k ->
       List.fold_left
         (fun k er ->
            if
              S.atomic_pair_allowed er ew
              && not (intervening StoreSet.is_before_strict ers er ew)
            then E.EventRel.add (er, ew) k
            else k)
         ers k)
    store_to_reads_map k
```

Maybe this hypothesis could be checked in a linear way too ...

HadrienRenaud (commented):
Thanks @maranget, this looks very good. Sorry I am coming with other big questions

Should we check that the order is total? I think it would only need a linear pass on the store set. We could probably put this in `StoreSet` actually:

```ocaml
let check_total set =
  if not (is_empty set) then
    Set.fold
      (fun e' e ->
         if e' != e (* min element *) then
           assert (is_before_strict e e');
         e')
      set (min_elt set)
    |> ignore

let of_list li =
  let set = of_list li in
  let () = check_total set in
  set
```

herd/mem.ml
maranget: I plan to replace `is_before_strict` by a simple program order check, see below.

herd/mem.ml
maranget: That's a possibility, I was a bit affraid of the signature but there should be no difficulty,

herd/memUtils.ml
maranget: Thanks.

herd/mem.ml
maranget: Loads are not totally ordered with stores, here is a counter-example on a simplified A016.litmus test:
```
AArch64 A016
variant=memtag,precise,store-only,vmsa
TTHM=P0:HD
{
 [PTE(x)]=(oa:PA(x),db:0,dbm:1,attrs:(TaggedNormal));
 0:X9=x:red;
}
 P0         ;
MOV W0,#1   ;
STR W1,[X9] ;
exists fault(P0,x) /\ PTE(x)=(oa:PA(x),db:1,dbm:1,attrs:(TaggedNormal))
```
One of the executions is : 
A016

Effects b: and c: are unrelated. That said, effect c: probably does not need to be an atomic read, as no write will ever be paired with it.

herd/mem.ml
HadrienRenaud: Nice one

herd/mem.ml
HadrienRenaud: Ah indeed. Thanks

herd/mem.ml
maranget: Done.

herd/mem.ml
maranget: See below... Not very conclusive.

herd/memUtils.mli
```diff
@@ -25,6 +25,32 @@ module Make : functor (S: SemExtra.S) -> sig
...
 (* po union iico_data union iico_control *)
   val is_before_strict : S.event_structure -> S.event -> S.event -> bool
+
+(*
+ * Set of events ordered by  is_before_strict.
+ * Notice it is an error to add unrelated events
+ *  to this set.
+ *)
+  module
+    EvtSetByPo :
```
fsestini: FWIW, I would have kept this in `mem.ml` for now. `EvtSetByPo`'s underlying assumptions seem fairly coupled to its usage within `mem.ml`, as it relies on a non‑total `compare`, which make sense in the specific context of `mem.ml` but seem riskier/easier to miss if exposed as a shared utility.

maranget: Ok, I can revert this easily.

maranget:
Commit cb2e0cd077b4ab5c6fcf98353035a1591f797f5d changes some of iico_data into iico_order, which may be more appropriate. Hopefully this change is of aesthetic nature. I am not sure we should do it...

herd/memUtils.ml
HadrienRenaud: I think you dropped the fix on this one

herd/mem.ml
```diff
@@ -1827,40 +1825,74 @@ let get_rf_value test read =
...
+          (fun k er ->
+             match StoreSet.find_first_after er ws with
+             | Some ew ->
+                if
+                  S.atomic_pair_allowed er ew
+                  && not (intervening_read er ew)
+                 then
+                   E.EventRel.add (er,ew) k
+                 else k
```
HadrienRenaud: Indentation is slightly off:

```suggestion
                if
                  S.atomic_pair_allowed er ew
                  && not (intervening_read er ew)
                then E.EventRel.add (er,ew) k
                else k
```

HadrienRenaud (approved):
This looks very good, thanks @maranget

herd/mem.ml
relokin: I am not sure this is correct. Luc, I think you had observed this test on hardware and we also verified that it is architecturally allowed.
```
AArch64 CoRW1+podxx
{
 0:X0=x; 0:X2=y;
}
 P0              ;
 LDXR W1,[X0]    ;
 MOV W3,#1       ;
 STXR W4,W3,[X2] ;
exists([y]=1)
```
This is ofc not modelled correctly currently, but we will need to fix at some point.

relokin (commented):
Sorry to reopen this but I wanted my question to have the right context.

I think there are two assumptions in this PR which might not be quite right.

If I understand correctly, this PR makes the assumption that there cannot be unordered atomic pairs, and I think this following test demonstrates the need to handle atomic pairs which are in the same instruction and not ordered in the extended po:
```
AArch64 STR-TaggedNormal-HA
Variant=vmsa,mte,sync
TTHM=HA
{
 [PTE(x)]=(af:0,attrs:(TaggedNormal));
 0:X0=x:green;
}
 P0          ;
 MOV W1,#1   ;
L0:          ;
 STR W1,[X0] ;
exists([x]=0)

(* forall([x]=1 /\ [PTE(x)]=(oa:PA(x),af:1,attrs:(TaggedNormal))) *)
```
For this test there are two concurrent almost identical address translations, one for the data and one for tag. Both of them can potentially update the access flag. Which ever loses the race will not perform an update. Prior to this change, this test was handled correctly.

I am happy to relax the restriction in `herd/mem.ml` but before I start writing code, I wanted to check if this would break something else.

Then there is one more concern which I have raised inline.

cc: @maranget @HadrienRenaud

HadrienRenaud:
> If I understand correctly, this PR makes the assumption that there cannot be unordered atomic pairs, and I think this following test demonstrates the need to handle atomic pairs which are in the same instruction and not ordered in the extended po:

I think more precisely this PR assumes that there cannot be two atomic stores to the same location in a single instruction. As your test demonstrates, this assumption is incorrect.
You also point out that the assumption that each load/store pair is on a single memory location is incorrect.

This PR also assumes that in each load/store pair, the load is extended-po-before the store, is that correct?

Is there a good reason why `STR-TaggedNormal-HA` is not in the regression test suite?

> I am happy to relax the restriction in `herd/mem.ml` but before I start writing code, I wanted to check if this would break something else.

I'm wondering if it would be worth having both algorithms possible? For example, in the general case (where there are no non-ordered writes to the same location), we can use the `EvtSetByPo`, but if while constructing we detect two non-ordered writes we fall back on the previous algorithm?

relokin:
Thanks for looking into this @HadrienRenaud!

> This PR also assumes that in each load/store pair, the load is extended-po-before the store, is that correct?

I don't have any counter example.

> Is there a good reason why `STR-TaggedNormal-HA` is not in the regression test suite?

I was trying to upstream `STR-TaggedNormal-HA` and other fixes/changes when I found this :)

> > I am happy to relax the restriction in `herd/mem.ml` but before I start writing code, I wanted to check if this would break something else.
> 
> I'm wondering if it would be worth having both algorithms possible? For example, in the general case (where there are no non-ordered writes to the same location), we can use the `EvtSetByPo`, but if while constructing we detect two non-ordered writes we fall back on the previous algorithm?

I think it should be possible to say that unless `-variant vmsa` there can only be one atomic operation to the same location generated by a single instruction. I am not sure how we would do this for exclusives.

maranget:
>>[Nikos]For this test there are two concurrent almost identical address translations, one for the data and one for tag. Both of them can potentially update the access flag. Which ever loses the race will not perform an update. Prior to this change, this test was handled correctly.

>[Hadrien]I think more precisely this PR assumes that there cannot be two atomic stores to the same location in a single instruction. As your test demonstrates, this assumption is incorrect.

I do not completlly understand. Is this execution with two atomic stores to the same location a valid one? If no maybe we can discard it early?

maranget:
Here is the problematic execution:
A

I am wondering: aren't all non-explicit atomic pairs related by iico_data? So maybe it is sufficiient to proceed as we do for spurious updates?

maranget:
I am trying to define the rmw relation for non-explicit accesses as a restricton of `iico_data`. `make test` apparently succeeds.

maranget:
My attempt: commit 83f34710eb6f90759a55e2430df2e7145e5088ac.

maranget:
Complete PR, PR #1876.

PR #1235, opened 2025-03-20
PAC with collisions

RemyCiterin:
This PR extend the previous PR I did for Pointer Authentication Code to add the support for hash collisions, for example a program like this one:
```
AArch64 Simple Collision
Variant=pac,fpac
{ 0:X0=x }
P0                  ;
  PACDZA X0 ;
  AUTDZB X0;
exists ( ~Fault(P0) )
```
may run without fault because of a hash collision between `pacda(x,0)` and `pacdb(x,0)`. But this program:

```
AArch64 Collision contradiction
Variant=pac,fpac
{ 0:X0=x; 0:X1=x;; }
P0;
  PACDZA X1   ;
  PACDZB X0   ;
  CMP X0,X1   ;
  B.EQ finish ;
  AUTDZA X0   ;
  MOV X2,#1   ;
finish:       ;
exists (~Fault(P0) /\ 0:X2=1)
```

can't return without fault and `0:X2=1` because otherwise their is a contradiction about the same collision. And herd will return (ideally with some debug flags because this output is not possible in litmus):

```
Test Collision Allowed
States 2
0:X2=0;  ~Fault(P0); pacda(x, 0x0)=pacdb(x, 0x0);
0:X2=0; Fault(P0,PacCheck:DA);
No
Witnesses
Positive: 0 Negative: 2
Condition exists (not (fault(P0)) /\ 0:X2=1)
Observation Collision Never 0 2
Time Collision 0.01
Hash=e743d111643250ebd8d3bb2a4c991cf9
```

This PR is till not 100% finish, I still have some update to do, as example: ensure that the pretty printing by default is the same between `herd` and `litmus` like I did for the version without collisions, moving the docs from `pac_tests`...

PR #1736, opened 2026-03-02
PAC with collision

diaolo01:
This PR continues #1235. It uses Predicates for handling PAC collisions. The PAC solver is rebuilt using the Predicates to preserve collision history across multiple `AUT*` instructions.


lib/AArch64Op.ml
```diff
@@ -112,11 +170,45 @@ module
...
+      raise (Constraint (Ne (PAC.canonical,PAC.canonical),one))
+
+    let assumeNoCollision c1 c2 =
+      let open Constant in
+      let one = Concrete S.one in
+      match collision c1 c2 with
+      | Some (p1,p2) ->
+          raise (Constraint (Ne (p1,p2),one))
+      | None ->
+          let instr_eq i1 i2 = Misc.polymorphic_compare i1 i2 = 0 in
```
HadrienRenaud: Can we justify the use of polymorphic_compare here? Why can't we use the correct instruction equal? If it doesn't happen, can we simply `assert false`, or warn?

lib/AArch64Op.ml
```diff
@@ -26,13 +26,24 @@ type 'op1 unop =
...
+  | AddErrorCode of PAC.key (* Add a PAC error code in a virtual address *)
   | Extra1 of 'op1
 
 type 'op binop =
   | Extra of 'op
+  (* Add a PAC field to a virtual address using a modifier and a key *)
   | AddPAC of bool * PAC.key
+  (* Test if two variabels may be equals modulo the equality of two PAC fields *)
+  | CollisionPossible
+  (* Assume the presence of a hash collision *)
```
HadrienRenaud: Can we describe what the result of `AssumeCollision` on 2 addresses would be?

lib/AArch64Op.ml
```diff
@@ -172,12 +264,17 @@ module
...
       | Symbolic (Virtual a) when not (PAC.is_canonical a.pac) ->
           None
-      | Symbolic (Virtual ({pac; offset; _} as v)) ->
+      | Symbolic (Virtual ({name; offset; pac; _} as v)) ->
         let modifier = pp_cst true modifier in
-        let pac = PAC.add key modifier offset pac in
+        let name =
+            match name with
+            | Data s -> s
+            | Label (_, lbl) -> lbl in
```
HadrienRenaud: Maybe worth refactoring out in a separate function?

herd/arch_herd.mli
```diff
@@ -37,8 +37,49 @@ module type S =
     (* Those register are to be initialised to the default value explicitly *)
     val reg_defaults : reg list
 
+(* Return if an equality of two syntactically different constants may be equals
+ * modulo a predicate. This may represent a hash or random generator collision *)
```
HadrienRenaud: nit: indentation is wrong here

herd/ARMArch_herd.ml
```diff
@@ -178,4 +182,18 @@ module Make (C:Arch_herd.Config) (V:Value.S) =
...
+
+    let empty_solver = ()
+
+    let pp_solver_state _ = ""
+
+    let add_predicate _ _ = Some ()
+
+    let normalize cst _ = cst
+
+    let compare_solver_state _ _ = 0
```
HadrienRenaud: Can we have a module `NoPAC` or `NoSolverState` that we could include everywhere except `AArch64`?

herd/constraints.ml
```diff
@@ -359,4 +317,140 @@ module Make (C:Config) (A : Arch_herd.S) :
           | ForallStates p ->
               pp_as_kind c ^ ": "^ pp_prop p
 
+        module Mixed (SZ : ByteSize.S) = struct
+          module AM = A.Mixed(SZ)
+
+          type 'a monad = A.solver_state -> ('a * A.solver_state) list
+          let (let*) x f = fun st ->
+            List.concat (List.map (fun (a,s) -> f a s) (x st))
```
HadrienRenaud: Why not use concat_map? If needed, put a stub in `Misc`

lib/AArch64Value.ml
```diff
@@ -13,14 +13,14 @@
...
 (****************************************************************************)
-
 module Make (C : sig
   val is_morello : bool
 end) : Value.AArch64 = struct
   module AArch64I = AArch64Instr.Make(C)(AArch64Instr.IdTr)
   module AArch64Cst = SymbConstant.Make (Int64Scalar) (AArch64PteVal) (AArch64AddrReg) (AArch64I)
   module NoCst = SymbConstant.Make (Int64Scalar) (PteVal.No) (AddrReg.No) (AArch64I)
   module NoArchOp = ArchOp.No(NoCst)
+  module AArch64Op' = AArch64Op
```
HadrienRenaud: Why this change?

lib/symbValue.ml
```diff
@@ -87,17 +87,17 @@ module
   let pp_v =  do_pp Cst.pp_v
   let pp_v_old =  do_pp Cst.pp_v_old
 
+  type arch_pred = ArchOp.predicate
+  exception Constraint of arch_pred * v
```
HadrienRenaud: Why redefine this? Is there somewhere where this is raised with a `Var`?

herd/AArch64Arch_herd.ml
```diff
@@ -557,4 +561,29 @@ module Make (C:Arch_herd.Config)(V:Value.AArch64) =
...
+      | Some (p1,p2) -> Some (AArch64Op.Eq (p1,p2))
+      | None -> None
+
+    type solver_state = PAC.solver_state
+
+    let empty_solver = PAC.empty_solver
+
+    let pp_solver_state st = PAC.pp_solver st
+
+    let normalize cst st = Constant.normalize cst st
```
HadrienRenaud: Why put this here if the only thing that is called is `Constant.normalize`?

herd/AArch64Sem.ml
```diff
@@ -127,6 +136,66 @@ module Make
...
+         hash collision of two PAC fields, this version is optimised to first
+         check if a collision is possible, so it generate less branches if the
+         operands are known, but it can be the oposite otherwise *)
+      let collision_wrapper op c1 c2 =
+        if pac then
+          M.op (Op.ArchOp AArch64Op.CollisionPossible) c1 c2 >>= fun c ->
+          M.choiceT c (
+            (* The only difference between `c1`and `c2` is their PAC fields *)
+            M.altT (
+              M.op (Op.ArchOp AArch64Op.AssumeCollision) c1 c2 >>= fun _ ->
```
HadrienRenaud: If I don't do anything with the result, why not register it simply as a predicate?

herd/AArch64Sem.ml
```diff
@@ -127,6 +136,66 @@ module Make
       let sxtw_op = sxt_op MachSize.Word
       and uxtw_op = uxt_op MachSize.Word
 
+      (* Wrap an operation by checking if the operands can be equals due to a
+         hash collision of two PAC fields, this version is optimised to first
+         check if a collision is possible, so it generate less branches if the
+         operands are known, but it can be the oposite otherwise *)
+      let collision_wrapper op c1 c2 =
+        if pac then
+          M.op (Op.ArchOp AArch64Op.CollisionPossible) c1 c2 >>= fun c ->
```
HadrienRenaud: Why not merge `collision_wrapper` and `collision_wrapper_unknown` with the following code that should always work:

```suggestion
          let collision_possible =
            if V.is_var_determined c1 && V.is_var_determined v2 then
              M.op (Op.ArchOp AArch64Op.CollisionPossible) c1 c2
            else M.unitT true
          in
          collision_possible >>= fun c ->
```

herd/AArch64Sem.ml
```diff
@@ -924,7 +995,33 @@ module Make
...
+      (* Return if a PAC field is canonical *)
+      let check_canonical a =
+        M.op1 (Op.ArchOp1 AArch64Op.MakeCanonical) a >>= fun a_canon ->
+        is_eq a a_canon
+
+(* Check if the virtual address is out of range, raise a MMU:Translation fault
+   overwise. An address may be out of range if it contains a non-canonical PAC
+   field, this functions also make the address syntactically canonical in case of
+   a hash collision to construct the corect read-from map, if the address is not
+   virutal, it just return `mok ma` *)
```
HadrienRenaud: typo at virtual

herd/AArch64Sem.ml
```diff
@@ -1003,19 +1100,21 @@ module Make
...
-                else if ha then
-                  an_xpte an,AArch64Explicit.NExp AArch64Explicit.AF
-                else
-                  (* Ordinary non-explicit access *)
-                  an_pte an,AArch64.nexp_annot in
-              mextract_whole_pte_val
-                an nexp a_pte (E.IdSome ii) >>== fun pte_v ->
-              (mextract_pte_vals pte_v) >>= fun ipte ->
-              M.unitT ((pte_v,ipte),a_pte)
+              let kont = (fun _ -> M.op1 Op.PTELoc a_virt >>= fun a_pte ->
```
HadrienRenaud:
```suggestion
              let kont _ =
                M.op1 Op.PTELoc a_virt >>= fun a_pte ->
```

herd/AArch64Sem.ml
```diff
@@ -1003,19 +1100,21 @@ module Make
...
+                      an_xpte an,AArch64Explicit.NExp AArch64Explicit.AF
+                    else
+                      (* Ordinary non-explicit access *)
+                      an_pte an,AArch64.nexp_annot in
+                  mextract_whole_pte_val
+                    an nexp a_pte (E.IdSome ii) >>== fun pte_v ->
+                  (mextract_pte_vals pte_v) >>= fun ipte ->
+                  M.unitT ((pte_v,ipte),a_pte)
+                ) in
+              if pac then M.bind_ctrldata ma kont else ma >>= kont
```
HadrienRenaud:
```suggestion
              (if pac then M.bind_ctrldata else (>>=)) ma @@ fun _ ->
              M.op1 Op.PTELoc a_virt >>= fun a_pte ->
              let an,nexp =
                if hd then (* Atomic accesses, tagged with updated bits *)
                  an_xpte an, AArch64Explicit.NExp AArch64Explicit.AFDB
                else if ha then
                  an_xpte an, AArch64Explicit.NExp AArch64Explicit.AF
                else
                  (* Ordinary non-explicit access *)
                  an_pte an,AArch64.nexp_annot in
              mextract_whole_pte_val an nexp a_pte (E.IdSome ii) >>== fun pte_v ->
              (mextract_pte_vals pte_v) >>= fun ipte ->
              M.unitT ((pte_v,ipte),a_pte)
```

herd/AArch64Sem.ml
```diff
@@ -1415,6 +1520,25 @@ module Make
                  mop ac ma >>= M.ignore >>= B.next1T
         )
 
+(* Check that the Pointer Authentication Code of a virtual address is canonical
```
HadrienRenaud, line 1580: nit: indentation wrong

herd/constraints.ml
```diff
@@ -359,4 +317,140 @@ module Make (C:Config) (A : Arch_herd.S) :
...
+              match A.eq_satisfiable c1 c2 with
+              | Some pred -> begin
+                match A.add_predicate pred st with
+                | None -> contradiction st
+                | Some st -> pure () st
+              end
+              | None ->
+                  if V.equal x y
+                  then pure () st
+                  else contradiction st
```
HadrienRenaud: Extract this in a separate function?

herd/valconstraint.ml
```diff
@@ -440,21 +479,30 @@ and type state = A.state =
...
+        (* Check if the equality of the two assignation imply an architecture
+           specific constraint *)
+        match A.eq_satisfiable cst cst' with
+        | Some pred ->
+            sol,Predicate pred::cns
+        | None ->
+            if V.Cst.eq cst cst' then sol,cns
+            else raise Contradiction
+      with
+      | Not_found -> (V.Solution.add x cst sol,cns)
```
HadrienRenaud: Slightly improved so no need for double insert in solution, but it might be a bit more complicated so up to you to take it or not.

Maybe worth trying the performance with `time make vmsa-test` or `time make test-asl-vmsa test-aarch64-asl cata-test-asl`

```suggestion
    (** [add_sol x cst (sol, csn)] updates [sol] to map to [cst], raising [Contradiction] in case of conflicting solutions, and adding a predicate to [csn] if needed. *)
    let add_sol =
      let exception FoundPredicate of predicate in
      let unify_solution cst cst' =
        if V.Cst.eq cst cst' then Some cst' else
          match A.eq_satisfiable cst cst' with
          | Some pred -> raise (FoundPredicate pred)
          | None -> raise Contradication
      in
      fun x cst (sol, csn) ->
        try
          let sol =
            V.Solution.update x
              (function None -> Some cst | Some cst' -> unify_solution cst cst')                  
              m
          in
          (sol, csn)
        with FoundPredicate pred -> (sol, Predicate pred :: csn)
```

lib/AArch64Op.ml
```diff
@@ -112,11 +170,45 @@ module
       | PteVal p -> Some (PteVal (op p))
       | _ -> None
 
-    (* Check that the PAC field of a virtual address is canonical *)
-    let checkCanonical =
-      let open Constant in function
+    let assumeContradiction () =
+      let open Constant in
+      let one = Concrete S.one in
+      raise (Constraint (Ne (PAC.canonical,PAC.canonical),one))
```
HadrienRenaud: Why not return `2`? this would probably kill the execution without needing a `Constraint`

Makefile
```diff
@@ -600,6 +599,18 @@ mte-test:
 		$(REGRESSION_TEST_MODE)
 	@ echo "herd7 catalogue aarch64-MTE tests: OK"
 
+pac-test:
```
HadrienRenaud: Is it normal that this is not run in CI? Shouldn't we add it to `make test` or `make test-all` (depending on how long it needs to run)?

herd/arch_herd.mli
```diff
@@ -37,8 +37,49 @@ module type S =
...
+ * collisions).
+ *
+ * The `add_predicate` function return the new solver state if the new
+ * constraint is satisfiable in the current environment. And `None` otherwise.
+ *
+ * In addition the compare function must only compare the part of the solver
+ * state we print in `pp_solver_state`, as example the PAC solver of `AArch64`
+ * only show it's equalities.
+ *)
+    type solver_state
```
HadrienRenaud: I wonder if this name is confusing because it could be confused with the state of the non-arch-specific solver in `valconstraint.ml`. A few suggestions to change this:

- calling it `arch_solver_state` (more pertinent in the other modules, but it's good to share the same name accross modules)
- Wrapping the type and the related functions in a module `ArchSolver`

HadrienRenaud (commented):
A lot of work into this, nice one. I have a lot of small esthetics comments, feel free to choose the ones you want to keep.

Do you think there are some performance difference? For example on `make vmsa-test`? How long does `make pac-test` takes to run?

The main question I have is the following. If I understand correctly, right now the way to add a predicate is the following:
1. Use an operation that use a collision wrapper
2. This splits the execution and adds an equation `Assign (fresh_var (), Binop (AssumeCollision, v1, v2))` (resp. `AssumeNoCollision` on the other execution) (skipping the case where both values are constants)
3. When in `valconstraint.ml` `v1` and `v2` are reduced to constants `c1 = {x with pac = p1}` and `c2 = {x with pac = p2}`, `Constant.collision` returns `Some (p1, p2)`, and thus `AArch64Op.assumeCollision` raises a `Constraint (Eq (p1, p2))` (resp `Neq`)
4. `valconstraint.ml` then pushes on the equation list `Predicate (Eq (p1, p2))`
5. `valconstraint.ml` calls the PAC solver at the end of its run with all the predicates accumulated.

Would it possible to skip the `binop` phase and directly push the predicates in `AArch64Sem`? That would change:
1. similar
2. This splits the execution and adds the *predicates* `AssumeCollision (v1, v2)` (resp. `AssumeNoCollision`)
3. `valconstraint` doesn't touch the predicates
4. `valconstraint` doesn't touch the predicates
5. similar as before, but with `AssumeNoCollision` instead of `Neq` and `AssumeCollision` instead of `Eq`. Potentially we need to add a step of substitution inside predicates to convert `AssumeCollision (V.Var s1, V.Var s2)` into `AssumeCollision (V.Val c1, V.Val c2)`, but this is a very standard step.

The main advantage with this is that it would remove the need for the `Constraint` exceptions, and probably simplify the interaction between the PAC solver and the non-pac solver.

Please note that I don't know how `litmus` nor `herd/constraints.ml` work and so I can't really review this part of the code. I also haven't reviewed the litmus tests.

herd/constraints.ml
```diff
@@ -359,4 +317,140 @@ module Make (C:Config) (A : Arch_herd.S) :
           | ForallStates p ->
               pp_as_kind c ^ ": "^ pp_prop p
 
+        module Mixed (SZ : ByteSize.S) = struct
+          module AM = A.Mixed(SZ)
+
+          type 'a monad = A.solver_state -> ('a * A.solver_state) list
```
fsestini: Perhaps I would rename this to something more descriptive (e.g., `type 'a solver = ...`) to avoid the very generic `monad` name.

herd/constraints.ml
```diff
@@ -359,4 +317,140 @@ module Make (C:Config) (A : Arch_herd.S) :
...
+          let (let*) x f = fun st ->
+            List.concat (List.map (fun (a,s) -> f a s) (x st))
+          let pure : 'a -> 'a monad = fun x st -> [x,st]
+          let contradiction : 'a monad = fun _ -> []
+          let test_cond c = if c then pure () else contradiction
+          let alt xs = fun st -> List.concat (List.map (fun x -> x st) xs)
+          let rec iter : unit monad list -> unit monad = function
```
fsestini: The name `iter` here may be confused with the established `iter` idiom (e.g. `List.iter`, `Array.iter`, etc.). Maybe rename to `sequence` or similar.

herd/constraints.ml
```diff
@@ -359,4 +317,140 @@ module Make (C:Config) (A : Arch_herd.S) :
...
+            if is_eq then add_equality x y else add_inequality x y
+
+          module SolverSet = MySet.Make(struct
+            type t = A.solver_state
+            let compare = A.compare_solver_state
+          end)
+
+          let do_check_prop solver look_type look_val flts =
+            (* Return the list of solver states that satisfy `(if sign then p
+             * else Not p)`. This implementation is ineficient because it
```
fsestini: Could you elaborate on the expected performance impact here? For example: is the inefficiency likely to show up in real-world litmus tests, and have you been able to benchmark the current approach vs the suggested optimization? Since `SolverSet.of_list` already compares solver states at the end, it'd be helpful to understand what performance improvements are expected from moving comparisons into `do_rec`.

herd/constraints.ml
```diff
@@ -359,4 +317,140 @@ module Make (C:Config) (A : Arch_herd.S) :
...
+              | Implies (p1,p2) ->
+                  do_rec sign (Or [Not p1; p2]) in
+            fun p ->
+              try
+                (* The solver states must be uniques in the outputs *)
+                let solver_set sign =
+                  SolverSet.of_list (List.map snd (do_rec sign p solver)) in
+                List.map (fun s -> true,s) (SolverSet.elements (solver_set true)) @
+                List.map (fun s -> false,s) (SolverSet.elements (solver_set false))
+              with A.LocUndetermined -> assert false
```
fsestini: I don't quite see the value in catching `LocUndetermined` only to `assert false` right after. If `LocUndetermined` is raised, the program will crash in any case, so I would consider just letting `LocUndetermined` bubble up uncaught, for a more informative crash message/stack trace.

herd/constraints.ml
```diff
@@ -359,4 +317,140 @@ module Make (C:Config) (A : Arch_herd.S) :
...
+                (AM.look_in_state senv state)
+                tenv rloc in
+            do_check_prop solver (A.look_rloc_type tenv) look_val flts p
+
+          let check_prop_rlocs p tenv (state,flts,solver) =
+            let look_val rloc =
+              AM.look_in_state_rlocs state rloc in
+            match do_check_prop solver (A.look_rloc_type tenv) look_val flts p with
+            | [result,_] -> result
+            | _ -> Warn.user_error "check_prop_rlocs return multiple solutions"
```
fsestini: This seems more like an error due to an internal invariant violation rather than bad user input. `Warn.fatal` might be more appropriate (and similarly for other uses of `Warn.user_error` in this PR).

herd/constraints.ml
```diff
@@ -359,4 +317,140 @@ module Make (C:Config) (A : Arch_herd.S) :
...
+              A.FaultSet.map (A.map_fault (V.map_const (fun cst ->
+                A.normalize cst st))) flts
+            in [flts,st]
+
+          let add_predicate is_eq x y =
+            if is_eq then add_equality x y else add_inequality x y
+
+          module SolverSet = MySet.Make(struct
+            type t = A.solver_state
+            let compare = A.compare_solver_state
```
fsestini: I'm a bit concerned about using `compare_solver_state` here. It's declared “imprecise” in `Arch_herd.S`, and the implementation in `lib/PAC.ml` ignores inequalities, so presumably it isn't a proper total order, contrary to what `MySet` would expect. See also my related comment in `herd/arch_herd.mli`.

herd/arch_herd.mli
```diff
@@ -37,8 +37,49 @@ module type S =
...
+    val eq_satisfiable : V.Cst.v -> V.Cst.v -> V.arch_pred option
+
+(* Functions to interact with a constraint solver, this constraint solver must
+ * support at least the computation of equalities/inequalities (modulo the
+ * theory represented by the solver), plus the resolution of a set of predicates
+ * given by the architecture. As example for AArch64, the solver must sopport
+ * the equalities/disequalities of Pointer Authentication Codes (hash
+ * collisions).
+ *
+ * The `add_predicate` function return the new solver state if the new
```
fsestini: Minor: could move this comment closer to the actual function signature of `add_predicate`.

herd/arch_herd.mli
```diff
@@ -37,8 +37,49 @@ module type S =
...
+     * `pp_solver_state empty_solver = ""` otherwise the final result may be
+     * invalid if "-debug pred-solver" is not set *)
+    val pp_solver_state : solver_state -> string
+
+    (* The comparison is used to store the final state in a Map before pretty
+     * printing, but not before doing any resolution of predicates. So the
+     * `compare` function may be imprecise. As example if we don't show the
+     * inequalities at pretty-printing, then the comparison may return that
+     * two solver are equals even if they contains different inequalities. *)
+    val compare_solver_state : solver_state -> solver_state -> int
```
fsestini: The comment says `compare_solver_state` is used to prepare `solver_state` terms for pretty‑printing, but `herd/constraints.ml` uses it to build a `SolverSet` during constraint checking, and `ArchExtra_herd` uses it to implement comparison for `final_state` (line 1025). So it looks like the comparison affects logical behavior, not just pretty-printing. Also, the comment mentions “inequalities” even though `arch_pred` is abstract here, therefore carries no notion of (in)equality.

More generally, a public comparison function like `compare_solver_state` making assumptions about its callers is a leaky abstraction, especially since comparison functions are very often used with `Map`/`Set` structures, which expect a proper total order.

Similarly, `pp_solver_state` shouldn't be constrained by debug behaviour or assumptions about its callers. The current contract (as stated in the comment) couples the pretty‑printer's behaviour to debugging policy, but I think these should be separate concerns.

If the intent of `compare_solver_state` and `pp_solver_state` is to be used for the very specific case of pretty-printing _a subset_ of the `solver_state` (for example, for debugging purposes), I think that should be made more evident.
Some options would be:
  * expose separate functions that make this intent very clear in their name (e.g., `debug_solver_state`/`compare_for_pp`, etc.) and keep `pp_solver_state` general and `compare_solver_state` semantically correct (or drop them entirely), or
  * if comparison is not needed besides debug-printing, remove `pp_solver_state` and `compare_solver_state` entirely and just expose an aptly-named `debug_solver_states : solver_state list -> string` which hides the comparison logic as an implementation detail, or
  * make this behaviour part of the type-checked contract of the signature, by exposing a separate "pretty-printing-specific" solver_state, and forcing the comparison and pretty-printing functions to _only_ act on terms of that type. For example:
    ```
    type finalized_solver_state
    val render_solver_state : solver_state -> finalized_solver_state option
    val pp_solver_state : finalized_solver_state -> string
    val compare_finalized_solver_state : finalized_solver_state -> finalized_solver_state -> int
    ```

herd/top_herd.ml
```diff
@@ -532,6 +537,9 @@ module Make(O:Config)(M:XXXMem.S) =
 (**********)
         let tr_out = tr_out test in
         printf "States %i\n" nfinals ;
+        let finals =
+          if O.debug.Debug_herd.pred_solver then finals
+          else A.StateSet.map (fun (x,y,_) -> (x,y,A.empty_solver)) finals in
```
fsestini: AFAICT, this code fakes "turning debug off" by replacing the solver with an `empty_solver`, relying on implicit knowledge about the implementation of `do_dump_final_state` and `pp_solver_state`.
This leaks `do_dump_final_state`'s implementation details through the supposedly abstract interface of `ArchExtra_herd.S` and makes `pp_solver_state`'s output depend on caller behaviour, which is not great.
If this call site needs to be able to selectively disable solver printing in `do_dump_final_state`, a better approach would be to make this controllable via a parameter of `do_dump_final_state`. For example, `do_dump_final_state` could be made to accept an explicit "solver printer" function as a parameter, so callers can control solver printing without having to make assumptions about the function internals.
  ```
  let pp_solver_st = if O.debug.Debug_herd.pred_solver then Some A.pp_solver_state else None in
  A.do_dump_final_state test.Test_herd.type_env test.Test_herd.ffaults tr_out pp_solver_st st
  ```

herd/arch_herd.mli
```diff
@@ -37,8 +37,49 @@ module type S =
...
+(* Return if an equality of two syntactically different constants may be equals
+ * modulo a predicate. This may represent a hash or random generator collision *)
+    val eq_satisfiable : V.Cst.v -> V.Cst.v -> V.arch_pred option
+
+(* Functions to interact with a constraint solver, this constraint solver must
+ * support at least the computation of equalities/inequalities (modulo the
+ * theory represented by the solver), plus the resolution of a set of predicates
+ * given by the architecture. As example for AArch64, the solver must sopport
+ * the equalities/disequalities of Pointer Authentication Codes (hash
+ * collisions).
```
fsestini: Can you elaborate on the reason why this collection of “functions to interact with a constraint solver” needs to live in `Arch_herd.S`, since they don't really represent general architectural properties (and in fact only AArch64 implements/uses them), nor they make use of types defined in `Arch_herd.S`. The functions only interact with types declared by `Value.S` (namely `arch_pred` and `Cst.v`), so if they had to be bundled with an existing signature, I'd imagine `Value.S` would be a better place instead of `Arch_herd.S`.
Frankly, for the same reasons, I would also consider factoring these solver operations out altogether, into a separate, dedicated `ConstraintSolver` signature/module that is then passed as parameter to functors that need it (e.g., `Constraints.Make`).

herd/parseTest.ml
```diff
@@ -77,6 +77,42 @@ module Top (TopConf:RunTest.Config) = struct
...
+          Warn.user_error "\"fpac\" variant require \"pauth2\" variant" ;
+
+        if Conf.variant Variant.ConstPacField && not pauth2 then
+          Warn.user_error "\"const-pac-field\" variant require \"pauth2\" variant" ;
+
+        if pauth1 && pauth2 then begin
+          let msg =
+            "\"pauth1\" and \"pauth2\" variants are incompatible, " ^
+            "\"pauth1\" is suppose to represent \"FEAT_PAuth\" without " ^
+            "\"FEAT_PAuth2\", but \"pauth2\" is suppose to represent " ^
```
fsestini:
```suggestion
            "\"pauth1\" is supposed to represent \"FEAT_PAuth\" without " ^
            "\"FEAT_PAuth2\", but \"pauth2\" is supposed to represent " ^
```

herd/valconstraint.ml
```diff
@@ -440,21 +479,30 @@ and type state = A.state =
...
 (* merge of solutions, with consistency check *)
-    let add_sol x cst =
-      V.Solution.update x @@ function
-      | None -> Some cst
-      | Some cst' ->
-        if V.Cst.eq cst cst' then Some cst' else raise Contradiction
+    let add_sol x cst (sol,cns) =
+      try
+        let cst' = V.Solution.find x sol in
+        (* Check if the equality of the two assignation imply an architecture
```
fsestini: Typo I think? (unless "assignation" makes sense in this domain-specific context)
```suggestion
        (* Check if the equality of the two assignments imply an architecture
```

lib/AArch64Op.ml
```diff
@@ -75,22 +90,65 @@ module
...
+      | AddErrorCode k -> Printf.sprintf "ErrorCode:%s" (PAC.pp_upper_key k)
 
     type scalar = S.t
     type pteval = AArch64PteVal.t
     type addrreg = AArch64AddrReg.t
     type instr = AArch64Base.instruction
     type cst = (scalar,pteval,addrreg,instr) Constant.t
 
+    type nonrec predicate = predicate
+    exception Constraint of predicate * cst
```
fsestini: I understand this `Constraint` exception is already used in similar ways in other parts of the codebase, so this is more of a personal curiosity to understand the logic better: what is the rationale for using an exception here, instead of a regular return type (perhaps a monad with short-circuiting behaviour, a la [ `Result.t` ](https://ocaml.org/manual/5.4/api/Result.html))?

lib/AArch64Op.ml
```diff
@@ -75,22 +90,65 @@ module
...
+        match PAC.get_name p1,PAC.get_name p2 with
+        | Some name, _ | _, Some name ->
+            Printf.sprintf "%s != %s" (PAC.pp p1 name 0) (PAC.pp p2 name 0)
+        | _, _ ->
+            "false"
+      end
+
+    let inverse_predicate = function
+      | Eq (p1,p2) -> Ne (p1,p2)
+      | Ne (p1,p2) -> Ne (p1,p2)
```
fsestini: Shouldn't `inverse_predicate` map `Ne` to `Eq` here?

lib/PAC.ml
```diff
@@ -1,61 +1,103 @@
...
+  | DA -> "DA"
+  | DB -> "DB"
 
-let parse_key : string -> key = function
-  | "da" | "DA" -> DA
+let pp_lower_key = function
+  | IA -> "ia"
+  | IB -> "ib"
+  | DA -> "da"
+  | DB -> "db"
```
fsestini: Minor simplification:
```suggestion
let pp_lower_key k = String.lowercase_ascii (pp_upper_key k)
```

lib/PAC.ml
```diff
@@ -1,61 +1,103 @@
...
       end
   | r -> r
 
+let compare_signature_err p1 p2 =
+  match String.compare p1.err_name p2.err_name with
+  | 0 ->
+      begin match (p1.is_key_a,p2.is_key_a) with
+      | true,true | false,false -> 0
+      | true,false -> -1
+      | false,true -> 1
```
fsestini: Consider whether `Bool.compare` could be used instead, to compare boolean values.

lib/PAC.ml
```diff
@@ -69,20 +111,206 @@ let is_canonical pac =
...
 let equal = PacSet.equal
 
-let pp pac s =
-  let rec aux = function
-    | x :: xs -> pp_signature x (aux xs)
-    | [] -> s
+let get_name p =
+  match PacSet.choose_opt p with
+  | Some s -> Some (signature_name s)
+  | None -> None
```
fsestini: Minor suggestion for brevity:
```suggestion
let get_name p = Option.map signature_name (PacSet.choose_opt p)
```

lib/PAC.ml
```diff
@@ -69,20 +111,206 @@ let is_canonical pac =
...
+          offset
   in
-  aux (PacSet.elements pac)
+  do_rec offset (PacSet.elements pac)
+
+let pp_litmus pac s =
+  let do_rec = function
+    | [] -> s
+    | Ok ok :: [] ->
+        assert (ok.offset == 0);
```
fsestini: If invalid offset fields can be a consequence of incorrent input from client code or user error (as opposed to purely bugs that "shouldn't happen"), consider `Invalid_argument`/`Warn.user_error` here instead of asserts.

lib/PAC.ml
```diff
@@ -69,20 +111,206 @@ let is_canonical pac =
...
+  | Some pivot ->
+    let def = PacSet.remove pivot new_eq in
+
+    (* simplify all the current equations with the new one *)
+    let simplified_equalities = PacMap.map
+      (fun x -> simplify x (PacMap.singleton pivot def))
+      state.equalities
+    in
 
+    (* simplify all the current inequations with the new one *)
```
fsestini:
```suggestion
    (* simplify all the current inequalities with the new one *)
```

herd/debug_herd.ml
```diff
@@ -87,6 +90,7 @@ let none =
 let parse t tag = match tag with
   | "solver" -> Some { t with solver = t.solver+1; }
   | "solver+" -> Some { t with solver = t.solver+2; }
+  | "pred-solver" -> Some { t with pred_solver = true; }
```
fsestini: `tags` a few lines above contains `pred_solver` but the parser here accepts `pred-solver`. Either of them should be changed to be made consistent with the other (I suggest changing `tags` to contain `pred-solver`).

lib/symbValue.ml
```diff
@@ -993,11 +999,17 @@ module
...
+              match ArchOp.do_op o c1 c2 with
+              | Some c -> Val c
+              | None ->
+                  Warn.user_error "Illegal operation %s on %s and %s"
+                    (ArchOp.pp_op o) (pp_v v1) (pp_v v2)
+            with
+            | ArchOp.Constraint (pred,cst1) ->
+                raise (Constraint (pred,Val cst1))
+            | exn ->
+                raise exn))
```
fsestini: Re-raising `exn` here is not needed.
```suggestion
                       ))
```

fsestini (commented):
Thanks for the work you put into this. I'm not too familiar with the PAC feature, or the logic around constraints, so please bear with me if some of my comments on those aspects don't quite make sense.

I do have a more general concern however: this PR introduces new abstract functions/types in `Arch_herd.S` (for the constraint solver), but the abstraction feels fragile in both directions: implementations assume specific caller behavior (e.g., `pp_solver_state` must return "" for `empty_solver` because `top_herd` replaces solver states when debugging is disabled; similarly for `compare_solver_state`), and callers assume specific implementation details (e.g., `top_herd` relies on `do_dump_final_state` always including solver printing via `pp_solver_state`).
This creates a brittle contract that's not enforced by the types and makes reading the code harder (you can't understand local choices, like the "empty solver_state" trick at `top_herd.ml:542`, without knowing implementation details of distant call sites) and later refactors more prone to break things inadvertently.
My suggestion would be to either make those assumptions more explicit in the API, or keep the abstraction honest by removing caller‑specific behavior from implementations and implementation‑specific assumptions from call sites. I appreciate this is quite vague, so please refer to the in-code comments.

lib/AArch64Op.ml
diaolo01: @HadrienRenaud raised similar feedback. I have now removed this `Constraint` indirection and capture predicates directly from `AArch64Sem.ml`.

lib/AArch64Op.ml
diaolo01: No, the name was poorly chosen. What this function does in `add_inequality` in `constraints.ml` is enforce two values don't collide. I have updated the name.

lib/AArch64Op.ml
diaolo01: This is no longer relevant after the refactoring to avoid `Constraint`.

lib/AArch64Value.ml
diaolo01: I've removed it.

lib/symbValue.ml
diaolo01: No longer applies after the refactor without `Constraint` indirection.

lib/symbValue.ml
diaolo01: Thanks. I've also updated the similar function above.

herd/ARMArch_herd.ml
diaolo01: Done

herd/AArch64Arch_herd.ml
diaolo01: `normalize` is called in `constraints.ml` which is architecture agnostic. If I try to call `Constant.normalize` then instead of requiring `A.solver_state` it asks for `PAC.solver_state`. For architectures other than AArch64, `solver_state` is `unit`.

herd/valconstraint.ml
diaolo01: Thanks, the time for `make test.pac` went from 1.2 s to 0.6 s.

Makefile
diaolo01: Good catch, I've now added it in the CI.

herd/arch_herd.mli
diaolo01: Thank you - I've updated `compare_solver_state` to give the total order (including inequalities) and introduced `finalized_solver_state` and `compare_finalized_solver_state` for printing purposes. 

I have also removed the hack that used the `empty_solver`.

herd/arch_herd.mli
diaolo01: Agreed. These functions didn't make sense in Arch_herd.S. I have moved them in a dedicated ConstraintSolver module and each architecture now provides a CS implementation. Now the solver is separated from the architecture interface.

herd/AArch64Sem.ml
diaolo01: Thank you for the suggestion, but I am going to keep the current `kont` style for readability.

herd/constraints.ml
diaolo01: It looks like this comment was a self-note from Remy. I have not measured this or seen it cause significant performance degradation, so I have opted for just removing the comment.

herd/AArch64Sem.ml
diaolo01: Thank you. I'm going to keep the current `collision_wrapper`/`collision_wrapper_unknown` logic for now. I don't want to change semantics without having a good understanding of the trade-offs.

herd/eventsMonad.ml
```diff
@@ -1717,6 +1727,14 @@ Monad type:
 
     let add =  op Op.Add
 
+    let assume_collision v1 v2 =
+      make_one_monad ()
+      [VC.Predicate (A.CS.pred_assume_collision v1 v2)] E.empty_event_structure
+
+    let assume_no_collision v1 v2 =
+      make_one_monad ()
+      [VC.Predicate (A.CS.pred_assume_no_collision v1 v2)] E.empty_event_structure
```
HadrienRenaud: Instead of adding those for all architectures, you could add them only in `AArch64Sem` by using `Monad.restrict`

herd/mem.ml
```diff
@@ -968,15 +969,22 @@ let get_rf_value test read =
...
+          | V.Val c1, V.Val c2 ->
+              begin match A.CS.eq_satisfiable c1 c2 with
+              | Some _ -> VC.Assign (v1, VC.Atom v2) :: eqs
+              | None -> raise Contradiction
+              end
+          | _ -> raise Contradiction
+      else
+        VC.Assign (v2, VC.Atom v1) :: eqs
+    else if V.equal v1 v2 then eqs
+    else VC.Assign (v1, VC.Atom v2) :: eqs
```
HadrienRenaud: The `raise Contradiction` at the end of the match block should be an `assert false`, but at this point why not rewrite it more clearly like:

```suggestion
let add_eq v1 v2 eqs =
  let assign v1 v2 eqs = VC.Assign (v1, VC.Atom v2) :: eqs in
  match (v1, v2) with
  | (V.Val c1, V.Val c2) -> (
      if V.Cst.eq c1 c2 then eqs
      else
        match A.CS.eq_satisfiable c1 c2 with
        | Some _ -> assign v1 v2 eqs
        | None -> raise Contradiction)
  | (V.Val _, V.Var _) -> assign v1 v2 eqs
  | (V.Var _, V.Val _) -> assign v2 v1 eqs
  | (V.Var csym1, V.Var csym2) ->
      if V.equal_csym csym1 csym2 then eqs
      else assign v1 v2 eqs
```

herd/eventsMonad.ml
diaolo01: It makes more sense to add them only in `AArch64Sem`, however there are already aarch64 specific functions in `eventsMonad`, so this wouldn't be unique.
I’ve moved the collision assumptions into `AArch64Sem` using `Monad.restrict` as suggested.

HadrienRenaud: Thank you! There are indeed, but we should try and limit them as much as possible (`monad.mli` is already hard enough to understand...)

herd/mem.ml
```diff
@@ -968,15 +969,20 @@ let get_rf_value test read =
...
-      else if V.equal v1 v2 then eqs
-      else VC.Assign (v1, VC.Atom v2)::eqs
+let add_eq v1 v2 eqs =
+  let assign v1 v2 eqs = VC.Assign (v1, VC.Atom v2) :: eqs in
+  match (v1, v2) with
+  | (V.Val c1, V.Val c2) -> (
+      if V.Cst.eq c1 c2 then eqs
+      else
+        match A.CS.eq_satisfiable c1 c2 with
+        | Some _ -> assign v1 v2 eqs
```
HadrienRenaud: Actually now that I think about it, why delay using the predicate? Why keep the equation as is here?

```suggestion
        | Some pred -> Predicate pred :: eqs
```

herd/mem.ml
```diff
@@ -1788,11 +1793,15 @@ let get_rf_value test read =
     module CM = S.Cons.Mixed(C)
 
 (* Internal filter *)
-    let check_filter test fsc = match test.Test_herd.filter with
+    let check_filter solver test fsc = match test.Test_herd.filter with
     | None -> true
     | Some p ->
-        not C.check_filter ||
-          CM.check_prop p (S.type_env test) (S.size_env test) fsc
+        (* TODO: List.hd is probably an unsound approximation *)
```
artkhyzha, line 1797: Do we have any more info on what the comment meant?

artkhyzha (commented):
Thanks to the contributors and reviewers -- this is a massive amount of work.

I looked through the PR, mostly changes to herd. Changes to the semantics seemed very sensible. Two small things stood out to me:
* changes in what is now `check_ptw_kont`; my understanding is that this has to do with the check of the VA range rather than with collisions;
* generally, I wonder if there is a methodology for writing code with fewer continuations: the ones added in this PR are not too hard to follow, but personally I find them hard to read.

Superficially, the code changes look OK. I have not looked in depth into `PAC.{ml,mli}` and `constraintSolver.{ml,mli}`.

The PR looks mature. Do we know of parts of it that need extra attention? Otherwise, should we consider this one as ready for merging? (Subject to resolving the current merge conflict.)

lib/misc.ml
```diff
@@ -172,6 +172,7 @@ let to_c_name =
 let find_opt = List.find_opt
 let filter_map = List.filter_map
 let split_on_char = String.split_on_char
+let concat_map f xs = List.concat (List.map f xs)
```
fsestini: This should now be available from the stdlib after rebasing on master.

herd/AArch64Sem.ml
```diff
@@ -135,6 +143,74 @@ module Make
...
+          assume_no_collision c1 c2 >>= fun _ -> M.unitT ()
+        else
+          M.neqT c1 c2
+
+      (* Wrap exclusive-or to check for a hash collision of two PAC fields *)
+      let exclusive_or c1 c2 =
+        collision_wrapper (M.op Op.Xor) c1 c2
+
+      (* Wrap substraction to check for a hash collision of two PAC fields *)
+      let substraction c1 c2 =
```
artkhyzha:
```suggestion
      let subtraction c1 c2 =
```
A typo in the function name.

herd/mem.ml
diaolo01: I'm afraid I don't have more context than what the code suggests.

diaolo01:
Thanks @artkhyzha. 

On `check_ptw_kont`, it is also my understanding that it is introducing VA-range checks around the PT walk, rather than having to do with the collision.

On the continuation generally, I agree that the style can be hard to read, but I don't plan to refactor that in this PR.

I am not aware of any other blocker beyond the merge conflict, which I have resolved. @fsestini and I worked together on `constraintSolver.{ml,mli}` and he provided in depth review - thank you. It used to be embedded in `arch_herd.ml`, until @fsestini suggested to make it into a module. Extra attention to `PAC.{ml,mli}` would be welcome, but otherwise I think this PR is ready.

herd/valconstraint.ml
```diff
@@ -367,6 +403,18 @@ and type state = A.state =
...
 
+    let check_equality v w k =
+      match v,w with
+      | V.Val c1,V.Val c2 -> begin
+        match A.CS.eq_satisfiable c1 c2 with
+        | Some pred ->
+            Predicate pred::k
+        | None ->
+            if V.compare v w = 0 then k
+            else raise Contradiction
```
fsestini, line 415: Most "concrete vs concrete" equality checks now go through `eq_satisfiable`, so syntactically unequal constants can still be accepted under an architecture predicate such as a PAC collision. However, `Hint.hint_solve_one` down below in this file still appears to use the previous strictly-syntactic check.

Could you explain why the `Hint` module does not need to use arch-specific predicates? For example, as far as I understand `herd/mem.ml` uses this `Hint` module to cheaply prune possible store/load matches before full solving, and the comment on `hint_solve_one` says that `NoSolns` from the hint should imply full `solve` would also reject. Is that guarantee still valid under PAC?

herd/top_herd.ml
```diff
@@ -532,13 +537,23 @@ module Make(O:Config)(M:XXXMem.S) =
...
+               tr_out pp_solver_st st) in
+        begin match pp_solver_st with
+        | Some _ ->
+            A.StateSet.pp stdout "" pp_final finals
+        | None ->
+            let finals_no_solver =
+              A.StateSet.fold
+                (fun st -> A.StateSetNoSolver.add st)
+                finals A.StateSetNoSolver.empty in
+            A.StateSetNoSolver.pp stdout "" pp_final finals_no_solver
```
fsestini: Above, we compute `nfinals` using `StateSet.cardinal finals`, where `StateSet` includes the solver state in its comparison, but when `-debug pred-solver` is disabled it prints final states through `StateSetNoSolver`, which ignores the solver state. This means herd could be listing fewer final states than what claimed by the `States %i` line, depending on the value of `-debug pred-solver`. For example, `A09.litmus.expected` shows output with a line saying `States 2` but just one visible final-state line being printed.

I think this behaviour might be a bit confusing from a UX perspective, but I wonder if we have considered this and concluded it's fine. What is the argument against always printing states including the solver state? On a related note, would it make sense to also have regression tests covering the `-debug pred-solver` printing mode?

herd/mem.ml
```diff
@@ -1746,17 +1752,16 @@ let get_rf_value test read =
...
+                    | VC.Maybe (sol,cs,solver) ->
+                        (* Time to complete rfmap *)
+                        let rfm = add_mems rss wss rfm in
+                        let rfm = add_mem tag_loads tag_stores rfm in
+                        (* And to make everything concrete *)
+                        let es = E.simplify_vars_in_event_structure sol es
+                        and rfm = S.simplify_vars_in_rfmap sol rfm in
+                        kont es rfm cs solver res
                 with
-                | Contradiction -> res  (* can be raised by add_mem_eqs *)
```
fsestini, line 1759: Could you explain why this `Contradiction -> res` catch branch is no longer needed here?

As far as I can see, before this change, a `Contradiction` from one iteration of the inner `fold_cross` would end up discarding just that inner fold iteration alone, but now it seems the contradiction is caught only by the outer fold_cross, therefore discarding the entire inner fold. Are these two behaviours equivalent in the executions they end up computing? Do we have/need a litmus test in the test suite that demonstrates this?

lib/constant.mli
```diff
@@ -128,6 +129,9 @@ val collision :
     ('scalar, 'pte, 'addrreg, 'instr) t ->
       (PAC.t * PAC.t) option
 
+val normalize : ('scalar, 'pte, 'addrreg, 'instr) t ->
+  PAC.solver_state -> ('scalar, 'pte, 'addrreg, 'instr) t
+
```
fsestini, line 134: I notice that now this `Constant` module includes PAC-specific functions like `collision` and `normalize`, and the type `symbolic_data` also now carries a `pac` field. My understanding is that `Constant` is meant to model architecture-agnostic constant values, whereas PAC is a AArch64-specific feature. Is this correct? If so, then I wonder if it could have been possible to model PAC-enabled symbolic data in a way that does not end up leaking AArch64-specific concerns into architecture-agnostic modules like Constant.

I realise this PR is not uniquely responsible for all these changes, so my comment is to suggest opportunities for future refinements, rather than a change request.

herd/valconstraint.ml
```diff
@@ -58,19 +59,26 @@ module type S = sig
   val pp_eq : cnstrnt -> string
   val pp_cnstrnts : cnstrnt list -> string
 
+  type solver_state
```
fsestini: Just a thought: the `Valconstraint` module itself already implements constraint solving logic, and the local `Hint` submodule also has its own `solver_state` type. This PR now adds another `solver_state` type. A reader of this code might wonder what is the difference between all these different solvers and why we need all of them.

I understand this new `solver_state` type is representing the architecture-specific predicate solver state. Perhaps it would make sense to rename it to `arch_solver_state` to disambiguate?

herd/archExtra_herd.ml
```diff
@@ -185,13 +187,15 @@ module type S = sig
...
+  type final_state = rstate * FaultSet.t * I.CS.arch_solver_state
   val do_dump_final_state :
     type_env -> FaultAtomSet.t ->
-    (string -> string) -> final_state -> string
+    (string -> string) -> (I.CS.finalized_solver_state -> string) option ->
+    final_state -> string
 
   (* Set of final states *)
   module StateSet : MySet.S with type elt = final_state
+  module StateSetNoSolver : MySet.S with type elt = final_state
```
fsestini: This `StateSetNoSolver` module is exposed from the generic architecture interface of `ArchExtra_herd.S`, but from the interface alone it is not clear what architectural role the module `StateSetNoSolver` fulfills. It seems to exist only so that `top_herd.ml` can suppress solver-only duplicates when `-debug pred-solver` is off, but that is output policy rather than an architectural property.

In a follow-up PR, I would consider moving `StateSetNoSolver` out of `archExtra_herd.ml` and into a local declaration inside `top_herd.ml`, where its relevance would be clearer.

herd/AArch64Sem.ml
```diff
@@ -978,7 +1056,34 @@ module Make
...
+      let check_canonical a =
+        M.op1 (Op.ArchOp1 AArch64Op.MakeCanonical) a >>= fun a_canon ->
+        is_eq a a_canon
+
+(* Check if the virtual address is out of range, raise a MMU:Translation fault
+   overwise. An address may be out of range if it contains a non-canonical PAC
+   field, this functions also make the address syntactically canonical in case of
+   a hash collision to construct the corect read-from map, if the address is not
+   virtual, it just return `mok ma` *)
+      let check_va_range mok ma mfault ii domain =
```
fsestini: This function has a generic-sounding name, but the implementation seems fairly, if not completely, PAC-specific to me. Indeed it seems this function is only used in PAC-enabled code. Perhaps I would rename it to something like `check_pac_va_range` that makes its PAC-specific nature more evident.

fsestini (commented):
I've had a second look at the PR, spending more time on trying to understand the solver-specific logic.

Overall, I think non-PAC behaviour shouldn't be impacted by this PR (although I do wonder about `herd/mem.ml:1759. See below). I am not fully convinced about all of the PAC-specific paths, though, and I have a few questions there, mostly to clarify the intended invariants and whether some edge cases are covered. I’m still not fully familiar with all parts of herd affected by this PR, so apologies if some of my questions are obvious or based on misunderstandings.

Apart from the comments asking for clarification, my other notes are mostly suggestions for follow-up PRs and do not need to block this one. Overall, I think this can be merged, with the caveat that we should address some of the rough edges in follow-up work.

diaolo01:
@relokin I investigated `check_ptw_kont`, and it does not introduce a new continuation in the semantic/monadic sense. It is just a helper function, used as the success branch passed to `check_va_range`.

I agree the `*_kont` name is misleading here, so I have renamed it to `do_check_ptw` to make it clearer that this is just the body of the PTW after the optional VA-range check.

herd/valconstraint.ml
diaolo01: Thanks @fsestini. This is something that was missed during a rebase: the PAC work was done before the Hint optimisation, and `hint_solve_one` kept the older check when this PR was rebased. It needs to be updated in line with `check_equality` so that PAC equalities are not pruned too early.

herd/top_herd.ml
diaolo01: I agree that the current UX is not acceptable. If States 2 is reported then two states need to be printed. I updated the code to always print the solver state and checked the impact. States that were already counted separately are now displayed with solver-state as follows:
```
0:X1=42;  ~Fault(P0,MMU:Translation); pacda(x,0x35)=x;
0:X1=53; Fault(P0,x,MMU:Translation);
```
This results in an update in the expected files, but just as an output update rather than changing results.

herd/mem.ml
diaolo01: I don't have a good explanation why that was removed. On re-reading the code, the two behaviours do not look equivalent, so I've added the inner `Contradiction -> res` back.
I have not added a dedicated test for this as I don't have a minimal test that would exercise this path, but restoring the old behaviour should be safe to do.

lib/constant.mli
diaolo01: I think the intent of `constant.ml` is to be architecture agnostic, however with `PTE, PTE2, TLB`, even before PAC this wasn't totally true.

Agreed that this should be a clean-up rather than a bug fix for this PR.

herd/valconstraint.ml
diaolo01: Good point. I've rename it to `arch_solver_state`.

herd/AArch64Sem.ml
diaolo01: It is a PAC specific function, so I've proceeded with the renaming.

herd/archExtra_herd.ml
diaolo01: This naturally gets resolved by removing the pretty printer for hiding solver-only duplicates. Now the solver-distinct final states are always printed and I removed the pred-solver mode, this is no longer needed, so I removed it as well.

herd/valconstraint.ml
fsestini: Understood, thanks @diaolo01 . If you don't mind, could you add a `(* TODO: ... *)` comment about this, near the relevant `Hint` function, as a reminder for us that more work is needed. Thanks

fsestini (approved):
Thanks a lot of addressing my comments @diaolo01 . 

The reviews have surfaced some TODOs/things that need to be ironed out, but I think these can be addressed in later PRs and don't need to block this one. I think this is ready to merge now.

To summarise some of the outstanding issues for future work:
- `check_filter` in `herd/mem.ml` calls `check_prop` but then discards all but the first item of the returned list. The code includes a comment noting that this lossy use of `check_prop` might be "unsound". We should investigate to what extent the current code is really unsound and if anything needs to be changed in `check_prop` or elsewhere.
- Only tangentially related to this PR, but at some point we should consider refactoring `constant.ml` to restore its interface to be more architecture-agnostic (as I think it was its original intent), and find a better home for the various arch-specific elements that have crept in over time.

herd/valconstraint.ml
diaolo01: I already updated the code in the last revision for the `Hint` function to include the PAC check, which is why I marked the comment as resolved.
@HadrienRenaud, do you mind taking a look at this change, please?

fsestini: Oh sorry, I missed that. Thanks.

lib/AArch64Value.ml
relokin, line 1: I am nitpicking but unnecessary white space changes like this make this PR looks a bit overwhelming to review.

lib/archOp.ml
relokin, line 1: ditto

lib/CapabilityValue.ml
relokin, line 1: unnecessary white space changes.

lib/symbValue.ml
relokin, line 1: except from the removal of equalityPossible the rest of the diff is noise, can we undo the white space changes please?

lib/value.mli
relokin, line 1: ditto

herd/AArch64Sem.ml
relokin: I agree with Hadrien on this, the existing code is unnecessarily creating two different functions. Can we please make this simplification?

herd/eventsMonad.ml
relokin, line 1: I am not sure I understand this change, why do we need to change the semantics for CAS?

herd/archExtra_herd.ml
```diff
@@ -185,10 +187,11 @@ module type S = sig
   val rstate_filter : (rlocation -> bool) -> rstate -> rstate
   val debug_rstate : rstate -> string
 
-  type final_state = rstate * FaultSet.t
+  type final_state = rstate * FaultSet.t * I.CS.arch_solver_state
   val do_dump_final_state :
     type_env -> FaultAtomSet.t ->
-    (string -> string) -> final_state -> string
+    (string -> string) ->
```
relokin: unnecessary whitespace change.

herd/constraints.ml
```diff
@@ -359,4 +317,131 @@ module Make (C:Config) (A : Arch_herd.S) :
           | ForallStates p ->
               pp_as_kind c ^ ": "^ pp_prop p
 
+        module Mixed (SZ : ByteSize.S) = struct
```
relokin: Moving this module and also moving it makes the diff difficult to review, can you please move it back to where it was, so that we can focus on the actual changes?

relokin (commented):
Some comments, I know most of my comments are about Remy's code but if you could address them that'd be great.
Can you please also cherry-pick this commit https://github.com/relokin/herdtools7/commit/7b4c2abb3ffbb3f6b7defe16e3d35fd9e2741c4a

herd/eventsMonad.ml
diaolo01: It's unclear where that change was pulled from, but with the latest rebase, it is no longer part of this PR.

fsestini: I thought that change was due to the fact that:
- now [`casp`](https://github.com/diaolo01/herdtools7/blob/eaee8543afa12e773fb5d2db17812770fd974ec3/herd/AArch64Sem.ml#L2390) uses [`is_eq`](https://github.com/diaolo01/herdtools7/blob/eaee8543afa12e773fb5d2db17812770fd974ec3/herd/AArch64Sem.ml#L177) as the `rne` argument passed to `aarch64_cas_no`
- under PAC, `is_eq` can now produce multiple alternatives with different PAC collision/no-collision assumptions
- `aarch64_cas_no` was thus changed to enumerate all such alternatives instead of assuming the returned list of alternatives is a singleton

Am I misunderstanding? Were you able to verify that PAC-enabled CAS/CASP scenarios are not affected by reverting this change?

relokin: The only interesting behaviour about CAS in relation to PAC as far as I understand is illustrated by this litmus test:
```
AArch64 A22
Variant=pauth2
{
  void *x=y;
  0:X0=y; 
  0:X1=z;
  0:X2=x;
}
P0             ;
PACDZA X0      ;
CAS X0,X1,[X2] ;
exists ([x]=z /\ 0:X0=y)
```
What's interesting here is that there is a small chance that `pacda(y,0x0)=y`. When that happens the comparison is successful and CAS updates the location x with the address of z. This is what herd gives after Diana's latest update:

```
Test CAS-pac Allowed
States 2
0:X0=y; [x]=y;
0:X0=y; [x]=z; pacda(y,0x0)=y;
Ok
Witnesses
Positive: 2 Negative: 2
```
Now I find this use of CAS very unlikely, and it makes me doubt whether we would ever need it. But in any case, we're not adding a whole lot of complexity to support this.

But maybe I am missing something because the code in `aarch64_cas_no` was meant to implement something else?

BTW @diaolo01 this is a test that we might want to add to `herd/tests`.

relokin: @fsestini, I think I see the problem now. This test:
```
AArch64 CASP-pac
Variant=pauth2
{
  int64_t x[2]={y,z};
  0:X0=y; 0:X3=z; 
  0:X1=z;
  0:X2=x;
}
P0              ;
PACDZA X0       ;
CASP X0,X3,X1,X4,[X2] ;
exists (x={z,0} /\ 0:X0=y)
```
fails without the change.

I am on the fence about these changes. I would argue that this test is not that important, and I am not sure it's worth the extra complexity in eventsMonad.ml. What do you think?

relokin: I think I am missing something. I was trying to understand why `casp` is special and we don't have to do something similar for `cas`. I still haven't figured it out, but if I revert the changes to `herd/eventMonad.ml` and the changes to the `casp` function, then the CASP-pac test works as I would expect it to work.

relokin: Here is the patch that illustrates this: https://github.com/relokin/herdtools7/commit/a3ad4e0b63052e25d2039f7b85c48843055124be

artkhyzha: I can't see the changes to `eventsMonad.ml` anymore. Were they substantial?

I agree the test above is a bit niche. It could appear as a part of an MP shape though. Say, if P0 creates a data dependency into CAS to write a signed address into memory, and P1 reads the address, unsigns it and creates an address dependency with it.

If we were to consider not changing the semantics of CAS to account for collisions, it would be good to have a principled explanation for that and a safety check informing the user they are trying something unsupported. For instance, we could contemplate a principle that, for the memory ordering purposes, we are not interested in tests that write addresses into memory, and instead we use arithmetic operations as we usually do.

fsestini: > I think I am missing something. I was trying to understand why `casp` is special and we don't have to do something similar for `cas`. I still haven't figured it out, but if I revert the changes to `herd/eventMonad.ml` and the changes to the `casp` function, then the CASP-pac test works as I would expect it to work.

Uhm, I wonder if the reason this test gives the expected result is because it exercises the "successful CASP" path, whereas the point of contention is regarding the `neqp` predicate within the `casp` function, which as far as I can tell only applies to _unsuccessful_ CASP.

Consider that same test where instead we filter for executions where CASP fails.

```
AArch64 CASP-pac-unsuccessful
Variant=pauth2
{
  int64_t x[2]={y,z};
  0:X0=y; 0:X3=z;
  0:X1=z;
  0:X2=x;
}
P0              ;
PACDZA X0       ;
CASP X0,X3,X1,X4,[X2] ;
exists (x={y,z} /\ 0:X0=pacda(y,0))
```

herd before the latest changes reports this as forbidden.

```
$ dune exec herd7 -- -set-libdir ./herd/libdir casp.litmus
Test CASP-pac-unsuccessful Allowed
States 2
0:X0=y; x={y,z};
0:X0=y; x={z,0}; pacda(y,0x0)=y;
No
Witnesses
Positive: 0 Negative: 4
```

This seems correct to me AFAICT. The condition `x={y,z}` forces to only consider CASP failures, therefore `{pacda(y,0), z} != {y, z}`, which in turn implies `pacda(y,0) != y`. In other words `0:X0=pacda(y,0)` cannot hold if `x={y,z}` also holds.

However, if we change `neqp` to use `M.op Op.Eq` directly, the test becomes allowed:

```
$ dune exec herd7 -- -set-libdir ./herd/libdir casp.litmus
Test CASP-pac-unsuccessful Allowed
States 2
0:X0=y; x={y,z};
0:X0=y; x={z,0}; pacda(y,0x0)=y;
Ok
Witnesses
Positive: 2 Negative: 2
```

The set of final states being reported by herd is the same in both cases, which is a bit confusing. In any case, I _suspect_ the issue here is that the simpler equality comparison `M.op Op.Eq` does not reason about PAC-collision assumptions, thus it won't keep record of the fact that `{pacda(y,0), z} != {y, z}` is equivalent to `pacda(y,0) != y`, i.e. that `pacda(y,0) != y` is a necessary condition for any "CASP has failed" branch. However, the `pacda(y,0) != y` assumption is precisely what's needed, in this scenario, to make the test forbidden.
Let me know if this makes sense. This is according to my understanding of the CASP semantics from AArch64Sem.ml, which I now read in detail for the first time.

> I am on the fence about these changes. I would argue that this test is not that important, and I am not sure it's worth the extra complexity in eventsMonad.ml. What do you think?

I think it's perfectly fine if we decide not to support PAC-aware semantics for some instructions, if it's not worth it.
However, in that case we should make the boundary between what's "supported" and what's "not supported" very clear, both to users and to us developers. I think I would like to avoid the situation where herd silently applies a mix of PAC-aware and non-PAC-aware semantics to an instruction based on the particular litmus test. Perhaps we could consider having herd reject with an early error any tests with CAS/CASP altogether, if PAC is enabled.

relokin: Thanks @fsestini! That makes sense and I agree CASP-pac-unsuccessful should be forbidden.

> I can't see the changes to eventsMonad.ml anymore. Were they substantial?

Thanks  @artkhyzha! Here is a patch that reverts the changes to eventsMonad.ml, perhaps they are not substantial, but the overall complexity of CAS/CASP is not insignificant.

I don't think it would be trivial to define what herd supports in terms of PAC. I don't know that all instructions have support for PAC anyway. But for the case of CASP we have the implementation so we might as well merge it.

@diaolo01 can you please resurrect the changes to CASP?

diaolo01: Thanks everyone, I have reinstantiated the changes to CASP in eventsMonad.ml and added the three tests in this discussion.

relokin:
@diaolo01 thank you very much for your work on this. Can I please ask you to rebase one last time and resolve the conflict that is due to the Makefile changes? Then we will merge.

relokin (approved):
Thanks @diaolo01, I am going to merge this PR now.

PR #1768, opened 2026-03-30
[tools] Preserve negated miaou list wrapper

relokin:
Keep negated composite lists wrapped in a dedicated node instead of reusing the original list operator.

Without this, miaou can flatten the negated sub-list into the enclosing conjunction and lose the intended 'it is not the case that all of the following apply' structure.

For example:

  `let rel1 = [A]; rel2; [B \ range([C]; rel; [D])]`

should render as:
```
  - E1 is an A effect.
  - E1 is rel2-before E2.
  - E2 is a B effect.
  - It is not the case that all of the following apply:
    - E2 is a D effect.
    - E3 is rel-before E2.
    - E3 is a C effect.
```
Assisted-by: OpenAI Codex

### DIFF
--- a/tools/miaou.ml
+++ b/tools/miaou.ml
@@ -549,15 +549,15 @@ and cons_seqs (fs:exp list) (es:exp list) =
          and a = tr_evts e1 a
          and b =  tr_evts e1 b in
          IfCond (c,a,b)
-      | App (_,Var (_,"range"), Op (loc2,Seq,es)) ->
+      | App (_,Var (_,"range"),Var (_,"lxsx")) ->
+         pp_evts_id e1 "rangelxsx"
+      | App (_,Var (_,"range"), rel) ->
         let e3 = Next.next () in
-        begin match tr_rel e3 e1 (Op (loc2,Seq,es)) with
+        begin match tr_rel e3 e1 rel with
         | List ({ items; _ } as l) ->
            List { l with items = List.rev items; }
         | _ as i -> i
         end
-      | App (_,Var (_,"range"),Var (_,"lxsx")) ->
-         pp_evts_id e1 "rangelxsx"
       | e ->
          Item (fail (ASTUtils.exp2loc e) "ignoring expression")
 
--- a/tools/miaou.ml
+++ b/tools/miaou.ml
@@ -341,25 +341,12 @@ and cons_seqs (fs:exp list) (es:exp list) =
       and b = tr b in
       DiffPair (a,b)
 
-    let tr_diff tr tr_not tr_id a b =
-      match a,b with
-      | Var (loc1,id1),Op (_,Seq,[c;Var (_,id2);])
-           when String.equal id1 id2
-        ->
-        begin
-          match tr_not c with
-          | None -> do_tr_diff tr a b
-          | Some c ->
-             mk_list Inter [tr_id loc1 id1 ;c]
-        end
-      | _,_ ->
-         begin
-           match tr_not b with
-           | None ->
-              do_tr_diff tr a b
-           | Some c ->
-              mk_list Inter [tr a; c;]
-         end
+    let tr_diff tr tr_not a b =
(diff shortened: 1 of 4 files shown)


fsestini (commented):
Thanks @relokin . I see why this fixes the issue, but the representation of negated lists still feels slightly off to me.

As far as I can tell, the original issue is that `flatten_out` merges nested `List` nodes when their operators match. This patch avoids that by rewriting a negated composite list to `List (Diff, ...)`.

That works in practice, but it means the `op` field is no longer describing the meaning of the node. In the case of `notItem`, the list being produced is not necessarily a difference, but the `op` field will be set to `Diff` regardless; that is because `Diff` is just being used as a sentinel to opt out of flattening. I think that makes the code harder to understand locally, because the reason for setting `Diff` here only becomes clear once the reader knows about this specific bug and how `flatten_out` happens to behave.

Since the intent here is to preserve negated composite lists as their own node, IMO that would be better represented explicitly in `t`, rather than encoded indirectly through the `AST.op2` field.

For example, this could be a dedicated constructor for negated `t` terms:

```
type t =
  ...
  | Negated of t
  ...
```

or a separate `flattenable` flag on the existing `List` constructor that controls whether that list participates in flattening:

```
type t =
  ...
  | List of { op : AST.op2; flattenable : bool; intro_txt : string; sep_txt : string * string; children : t list; }
  ...
```

then functions like `flatten_out` and `rm_dump` can selectively ignore negated/non-flattenable lists:

```
let rec flatten_out = function
  | List ({ op = Inter|Union|Seq ; flattenable = true; intro_txt ; sep_txt ; children } as l) ->
     List { l with children = flatten_op l.op l.children }
  | List l ->
     List { l with children = List.map flatten_out l.children }
  | ...
```

tools/miaou.ml
```diff
@@ -370,7 +364,7 @@ and cons_seqs (fs:exp list) (es:exp list) =
       | Op (_,(Union|Inter as op),es) ->
          tr_op e1 e2 op es
       | Op (_,(Seq as op),es) ->
-         List (op,intro op,sep op,tr_seq e1 e2 es)
+         List { op; intro_txt = intro op; sep_txt = sep op; flattenable = true; items = tr_seq e1 e2 es; }
```
fsestini: This looks like it could fit `mk_list`
```suggestion
         mk_list op (tr_seq e1 e2 es)
```

tools/miaou.ml
```diff
@@ -422,7 +416,7 @@ and cons_seqs (fs:exp list) (es:exp list) =
          let bottom = List.map (tr_rel e3 e2) es in
          let items = top @ [tr_evts e3 evts] @ bottom in
          (* Use Union to prevent flattening into surrounding Inter list, so the intro text survives. *)
-         List (Union,intro op,sep op,items)
+         List { op = Union; intro_txt = intro op; sep_txt = sep op; flattenable = true; items; }
```
fsestini: The comment above this line reads "Use Union to prevent flattening into surrounding Inter list, so the intro text survives."

Probably out of scope here, this the current PR is mainly about negation. But I think it would be good, for a later PR, to investigate whether the new `flattenable` flag for lists can be used as a slightly more principled way to prevent unwanted flattening for this `intervening` case as well, without the need to use `Union` for it.

fsestini (approved):
Thanks @relokin for addressing my concerns. I left a couple of minor non-blocking thoughts/comments, but this LGTM already.

tools/miaou.ml
relokin: Let me check!

tools/miaou.ml
relokin: ack!

fsestini:
Thanks @relokin , feel free to squash/rebase/merge at any time!

relokin:
Thanks @fsestini!

PR #1777, opened 2026-04-07
[herd] Event Register Semantics

RaghavRoy145:
Adds semantics for the event register instructions: `WFE, SEV, SEVL `

#### Summary
The three instructions interact with event registers as follows:

SEVL: Send Event Local. Sets the current PE's event register to 1.
SEV: Send Event. Sets every PE's event register to 1 (broadcast wake-up).
WFE: Wait For Event. If the event register is 1, clear it and fall through. If it is 0, wait until some other PE issues SEV before clearing and continuing.

The event register is modelled as a per-thread symbolic memory location. A new `Constant.EventReg` of `string symbolic` kind is introduced, initialised to 0 at test start in `test_herd.ml.` An `Access.EVREG` access kind is added so reads/writes to these locations can be distinguished from ordinary memory.

SEVL: a single `write_loc` of 1 to the current thread's event register.
SEV: iterates over `test.start_points,` emitting one `write_loc` 1 per thread's event register in parallel.
WFE: a bounded loop (`wfe_loop 3`) non-deterministically branches on the value read.

#### Tests
Not added

### DIFF
--- a/herd/AArch64Arch_herd.ml
+++ b/herd/AArch64Arch_herd.ml
@@ -32,6 +32,9 @@ module Make (C:Arch_herd.Config)(V:Value.AArch64) =
 
     let is_kvm = C.variant Variant.VMSA
 
+    let is_wfe = function
+      | I_WFE -> true
+      | _ -> false
     let pp_barrier_short = pp_barrier
     let reject_mixed = true
 
@@ -103,6 +106,7 @@ module Make (C:Arch_herd.Config)(V:Value.AArch64) =
     | I_SWP _| I_SWPBH _| I_SXTW _| I_TLBI _ | I_AT _ | I_UBFM _
     | I_UDF _| I_UNSEAL _ | I_ADDSUBEXT _ | I_ABS _ | I_REV _ | I_EXTR _
     | I_MOPL _ | I_MOP _
+    | I_WFE | I_SEV | I_SEVL
     | I_WHILELT _ | I_WHILELE _ | I_WHILELO _ | I_WHILELS _
     | I_UADDV _
     | I_LD1SP _ | I_LD2SP _ | I_LD3SP _ | I_LD4SP _
@@ -291,7 +295,8 @@ module Make (C:Arch_herd.Config)(V:Value.AArch64) =
       | I_CASBH (v,_,_,_,_) | I_SWPBH (v,_,_,_,_)
       | I_LDOPBH (_,v,_,_,_,_) | I_STOPBH (_,v,_,_,_) ->
           Some (bh_to_sz v)
-      | I_NOP|I_B _|I_BR _|I_BC (_, _)|I_CBZ (_, _, _)
+      | I_NOP | I_WFE | I_SEV | I_SEVL
+      | I_B _|I_BR _|I_BC (_, _)|I_CBZ (_, _, _)
       | I_CBNZ (_, _, _)|I_BL _|I_BLR _|I_RET _|I_ERET| I_SVC _ | I_LDAR (_, _, _, _)
       | I_TBNZ(_,_,_,_) | I_TBZ (_,_,_,_) | I_MOVZ (_,_,_,_) | I_MOVK(_,_,_,_)
       | I_MOVN _
@@ -352,7 +357,8 @@ module Make (C:Arch_herd.Config)(V:Value.AArch64) =
       | I_FENCE _
       | I_IC _|I_DC _|I_TLBI _ | I_AT _
       | I_NOP|I_TBZ _|I_TBNZ _
-      | I_ERET | I_SVC _  | I_UDF _
+      | I_WFE | I_SEV | I_SEVL
+      | I_ERET | I_SVC _ | I_UDF _
       | I_ST1SP _ | I_ST2SP _ | I_ST3SP _ | I_ST4SP _
       | I_ST1SPT _ | I_CTERM _
         -> [] (* For -variant self only ? *)
@@ -465,7 +471,8 @@ module Make (C:Arch_herd.Config)(V:Value.AArch64) =
       | I_LDARBH (bh,(XX|AX),_,_) -> MachSize.Ld (bh_to_sz bh)
       | I_STXR _|I_STXRBH _ | I_STXP _ -> MachSize.St
       | I_LDAR (_, (AA|AQ), _, _)|I_LDARBH (_, (AA|AQ), _, _)
-      | I_NOP|I_B _|I_BR _|I_BC _|I_CBZ _|I_CBNZ _
+      | I_NOP | I_WFE | I_SEV | I_SEVL
+      |I_B _|I_BR _|I_BC _|I_CBZ _|I_CBNZ _
       | I_TBNZ _|I_TBZ _|I_BL _|I_BLR _|I_RET _|I_ERET | I_SVC _
       | I_UBFM _ | I_SBFM _
       | I_LDR _|I_LDRSW _|I_LDRS _|I_LDUR _|I_LD1 _|I_LDAP1 _
--- a/herd/AArch64Sem.ml
+++ b/herd/AArch64Sem.ml
@@ -636,6 +636,16 @@ module Make
         if mixed then fun _sz -> write_reg
         else write_reg_sz
 
+(* Create a symbolic event register for a given proc id `proc_id` *)
+      let ev_loc proc_id =
+        A.Location_global (V.Val (Constant.mk_ev_reg (Printf.sprintf "%d" proc_id)))
+
+(* Write to the event register of a given proc id `proc_id`, a given value `v` *)
+      let set_ev proc_id v ii =
+        M.write_loc
+        (mk_write quad Annot.N aexp Access.VIR (V.intToV v))
+        (ev_loc proc_id) ii
+
 (* Emit commit event *)
       let commit_bcc ii = M.mk_singleton_es (Act.Commit (Act.Bcc,None)) ii
       and commit_pred_txt txt ii =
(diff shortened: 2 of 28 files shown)


herd/AArch64Arch_herd.ml
```diff
@@ -81,7 +81,8 @@ module Make (C:Arch_herd.Config)(V:Value.AArch64) =
     *   Arm ARM B2.2.5 "Concurrent modification and execution of instructions"
     *)
     let is_cmodx_restricted_instruction = function
-    | I_B _| I_BL _| I_CBNZ _| I_CBZ _| I_FENCE ISB | I_NOP | I_TBNZ _| I_TBZ _ | I_SVC _
+    | I_B _| I_BL _| I_CBNZ _| I_CBZ _| I_FENCE ISB | I_NOP | I_WFE | I_SEV | I_SEVL | I_TBNZ _
```
diaolo01: From the Arm ARM B2.2.5:

> Concurrent modification and execution of instructions can lead to the resulting instruction behaving like any possible sequence of instructions executable at the same Exception level. However, the behavior is more restricted if both the instruction before and the instruction after modification are one of B, B.cond, BL, BRK, CB<cc>, CBB<cc>, CBH<cc>, CBNZ, CBZ, HVC, ISB, NOP, SMC, SVC, TBNZ, TBZ, TRCIT, or UDF.

WFE, SEV, SEVL should be true.

herd/AArch64Sem.ml
```diff
@@ -3644,12 +3644,48 @@ Arguments:
       let nextSet = B.nextSetT
       (* And now, just forget about >>! *)
       let [@warning "-32"](>>!) (_:unit) (_:unit) = ()
+      let ev_loc proc_id =
```
diaolo01: This feels out of place, it might be better suited around where `Basic read, from register` functions are.

herd/AArch64Sem.ml
```diff
@@ -3644,12 +3644,48 @@ Arguments:
...
            !(M.mk_singleton_es (Act.NoAction) ii)
+        (* Event Register Instructions *)
+        | I_WFE ->
+            let loc = ev_loc ii.A.proc in
+            let read_ev = M.read_loc Port.No (mk_read quad Annot.N aexp) loc ii in
+            let write_zero = M.write_loc (mk_write quad Annot.N aexp Access.EVREG (V.intToV 0)) loc ii in
+            let rec wfe_loop n =
+            if n <= 0 then M.zeroT
+            else
+              M.altT
```
diaolo01, line 4113: Since the architectural behaviour is to read the Event Register, then either clear or continue waiting depending on that value, I think this should be restructured so the choice is made from the read's result.

herd/AArch64Sem.ml
```diff
@@ -3644,12 +3644,48 @@ Arguments:
...
       let do_build_semantics test inst ii =
         let open AArch64Base in
         match inst with
         | I_NOP ->(* Instructions nop and branch below do not generate events, use a placeholder *)
            !(M.mk_singleton_es (Act.NoAction) ii)
+        (* Event Register Instructions *)
+        | I_WFE ->
+            let loc = ev_loc ii.A.proc in
+            let read_ev = M.read_loc Port.No (mk_read quad Annot.N aexp) loc ii in
+            let write_zero = M.write_loc (mk_write quad Annot.N aexp Access.EVREG (V.intToV 0)) loc ii in
```
diaolo01: This and the writes in SEV+SEVL would benefit from a set_ev helper function.

herd/archExtra_herd.ml
```diff
@@ -858,7 +861,7 @@ module Make(C:Config) (I:I) : S with module I = I
               Warn.user_error
                 "No default value defined for location %s\n"
                 (pp_location loc)
-          | Location_global (I.V.Val (Symbolic (Virtual _|Physical _)))
+          | Location_global (I.V.Val (Symbolic (Virtual _|Physical _|EventReg _)))
```
diaolo01: Since Event Register is a conceptual register and shouldn't be available to be accessed, I don't think this function should result in `reg_default_value`. I'd be curious what this functions is used for.

herd/machAction.ml
```diff
@@ -99,6 +99,7 @@ end = struct
     | A.Location_reg _ -> REG
     | A.Location_global (V.Val (Symbolic (Virtual _))|V.Var _)
       -> Access.VIR
+    | A.Location_global (V.Val (Symbolic (EventReg _))) -> Access.EVREG
```
diaolo01: I'm not convinced EVREG needs a special access kind. What value does it bring from modelling it as a normal register access?

lib/AArch64Base.ml
```diff
@@ -504,6 +505,7 @@ let pp_xreg r = match r with
 | Internal i -> Printf.sprintf "i%i" i
 | PState p -> PSTATE.pp p
 | ResAddr -> "Res"
+| EventReg -> "Event"
```
diaolo01: EventRegister would be closer to the architecture.

diaolo01 (commented):
Thanks for this!
I'm confused by EventReg being modelled as a symbolic location and as a register. Architecturally, it is register, but software cannot access it through direct reads/writes. My current preference would be to model it as register state, while making sure it is not reachable through the ordinary architectural register access paths such as MRS/MSR.

diaolo01:
I also found in the Arm ARM that: "The Event Register for the PE executing the Exception return instruction is set." This needs to be reflected in the ERET semantics.

RaghavRoy145:
> Thanks for this! I'm confused by EventReg being modelled as a symbolic location and as a register. Architecturally, it is register, but software cannot access it through direct reads/writes. My current preference would be to model it as register state, while making sure it is not reachable through the ordinary architectural register access paths such as MRS/MSR.

Thanks for the reviews!

Registers as they are currenlty implemented don't seem to allow writes from remote threads, and I get the following warning when I try this:

`Ambiguous po for register 1:Event: (eeid=g action=W1:Eventq=1) and (eeid=i action=W1:Eventq=0) are not ordered.`

A work-around for this is perhaps to have the local thread (to which the event register being written to belongs) write to its own register after SEV reads a 0 on the remote thread's event register, which doesn't sound faithful to the implementation where SEV acts like a broadcast write.

So for now I have implemented this as a global memory location. Unless it is worthwhile to create a type of a register that allows remote writes.

It could also be that I am missing something and using registers in a way where SEV can work correctly is already possible.

herd/AArch64Sem.ml
RaghavRoy145: I did try a choiceT approach, and the problem I had was the edges between WFE events depend on which combinator we use (choiceT vs altT):

`altT` with `seq_mem` (current implementation): produces `iico_order` between successive reads and between the final read and write. This is seems correct since the data from the read doesn't flow to the next read or to the write (which always writes 0).

choiceT: produces `iico_data` throughout, since the value must flow through the condition check.

Which edges are preferred?

herd/machAction.ml
RaghavRoy145: The idea was just to separate the access from VIR accesses. It does not make any difference functionally, you're right, I'll change this to use Access.VIR for event-register accesses.

lib/AArch64Base.ml
RaghavRoy145: Since we are modelling this as a global mem location, I will remove instances of this.

herd/archExtra_herd.ml
RaghavRoy145: The event register should be initialised to 0, however I have handled this [in test_herd](https://github.com/herd/herdtools7/pull/1777/changes#diff-3faa336cc7ec303c93b29513bb5780fbb5f65d74b8cd7612f40528e0f20f3ea8R264) so there is no need for this, you're right. I will instead replace the branch for EventReg to return a warning:  "No default value defined for location %s\n"

herd/AArch64Arch_herd.ml
RaghavRoy145: Thank you for this! I think I just assumed it would behave like a NOP in this context

RaghavRoy145:
> I also found in the Arm ARM that: "The Event Register for the PE executing the Exception return instruction is set." This needs to be reflected in the ERET semantics.

This is interesting, this is perhaps an example of a "spurious" update, so the thread is woken up without an SEV, thanks for catching this!

For now I am going to [implement it](https://github.com/herd/herdtools7/pull/1777/changes#diff-f2a12c2f4bc2d0aa25e415c57b77bfe6c912fb2a2b841e23fc03922cdeafb1ecR3732) as a parallel write alongside the ELR read, unsure if this the right way to do it.

```
fun () ->
      (set_ev ii.A.proc 1 ii >>| read_reg_ord AArch64.elr_el1 ii)
      >>= fun ((), v) -> eret_to_addr v
```

diaolo01:
Thanks, that makes sense. I understand that the current registers are not able to take cross threaded writes. However, I don't think modelling Event Registers as symbolic locations is the right abstraction. Architecturally, it's a hidden per-PE register, which is neither ordinary memory nor software accessible register. It feels like this needs to be a separate category of state/location that does not currently exist: something register-like and per-PE, but whose location still includes the target PE so that remote writes are possible.

herd/AArch64Sem.ml
diaolo01: Thanks, this is helpful. I’m fairly convinced the edges we want here are order edges rather than data edges. I’m still uncomfortable with the current altT encoding though, because it duplicates the EVREG read across the two paths. I can’t yet fully justify that objection in terms of the graph relations, but it does not feel like the right abstraction: architecturally this is a single read followed by a guarded choice rather than two separate guarded reads.

diaolo01:
>This is interesting, this is perhaps an example of a "spurious" update, so the thread is woken up without an SEV, thanks for catching this!

I don’t think “spurious” is the right term here. The Arm ARM explicitly lists exception return as a cause that sets the Event Register, so this is an architected update. Modelling it alongside ERET makes sense.

PR #1781, opened 2026-04-08
[github, herd-www] Fix filter paths inside check-build-www runner and herd-www tst target

TiberiuBucur:
I have noticed that the `check-build-www` action is only run as part of a cronjob, and not on changes to some of the relevant directories (`herd-www` and `asllib/asllib-www`, as evidenced by [this PR](https://github.com/herd/herdtools7/pull/1779)).

This fixes the filter path for `asslib-www` and adds `herd-www` as a filter path to this Github runner. In addition, it modifies the tst target inside `herd-www/Makefile` so that it matches the expected tab separator, and renames a variable to a more descriptive choice.


Roman-Manevich (approved):
The change to the github workflow file looks fine.
I don't fully understand the change to `Makefile` but I understand it has been reviewed otherwise.

TiberiuBucur:
Hi @maranget. I was also thinking about this, but I did not want this PR to interfere with anything I did not know enough about. My guess is `build-www` exists because we might want the ability to only trigger the `herd-www` build, without `asllib`.

TiberiuBucur:
Hi @maranget, @relokin. I've had a go at merging the two. Now we have a conditional upload artifact step that can only be triggered on manual runs with a flag set to true. What do you think?

TiberiuBucur:
@ShaleXIONG, do you have any opinion on this? @relokin @maranget is this in a good state or would I need to change something? Thanks :)

fsestini (approved):
I've had a look, and everything seems reasonable.

PR #1788, opened 2026-04-10
[tools] AARCH-25405: Clarification οf the IC-Coherence-before and Instruction-read-ordered-before definitions

TiberiuBucur:
This changes the way `miaou7` transliterates variants into LaTeX. In the case of a conjunction of multiple variants as part of a condition, each variant is now a separate `\item`, which is then translated into exactly one line of markdown. In addition, each variant needs to be uniquely defined in `catdefinitions.tex`. The generic `Variant[1]` and `NotVariant[1]` commands have been removed.

### DIFF
--- a/herd/libdir/catdefinitions.tex
+++ b/herd/libdir/catdefinitions.tex
@@ -1,6 +1,12 @@
 \newcommand{\notthecase}[1]{it is not the case that #1}
-\newcommand{\Variant}[1]{#1 is implemented}
-\newcommand{\NotVariant}[1]{it is not the case that #1 is implemented}
+\newcommand{\ETSTwo}{FEAT_ETS2 is implemented}
+\newcommand{\NotETSTwo}{\notthecase{\ETSTwo{}}}
+\newcommand{\ETSThree}{FEAT_ETS3 is implemented}
+\newcommand{\NotETSThree}{\notthecase{\ETSThree{}}}
+\newcommand{\DIC}{The Effective value of DIC is 1}
+\newcommand{\NotDIC}{The Effective value of DIC is 0}
+\newcommand{\IDC}{The Effective value of IDC is 1}
+\newcommand{\NotIDC}{The Effective value of IDC is 0}
 \newcommand{\flag}[1]{By construction, #1}
 \newcommand{\assert}[1]{By construction, #1}
 %
@@ -42,7 +48,6 @@
 \newcommand{\M}[1]{#1 is a \ME{}}
 \newcommand{\W}[1]{#1 is a \MWE{}}
 \newcommand{\R}[1]{#1 is a \MRE{}}
-\renewcommand{\_}{any Effect}
 \newcommand{\B}[1]{#1 is a Branching Effect}
 \newcommand{\IntrBranching}[1]{the Intrinsic Branching Effect which checks #1}
 \newcommand{\BccBranching}{the Branching Effect of the instruction}
--- a/tools/miaou.ml
+++ b/tools/miaou.ml
@@ -290,24 +290,30 @@ and cons_seqs (fs:exp list) (es:exp list) =
          variant_dnf neg vc1 @ variant_dnf neg vc2
       | OpAnd (vc1,vc2) ->
          let d1 = variant_dnf neg vc1
-         and d2  =variant_dnf neg vc2 in
+         and d2 = variant_dnf neg vc2 in
          List.fold_right
            (fun a1 k ->
              List.fold_right
                (fun a2 k -> (a1@a2)::k) d2 k)
            d1 []
 
+    let pp_atom = function
+    | Pos s -> sprintf "\\%s{}" (MiaouNames.to_variant_csname s)
+    | Neg s -> sprintf "\\Not%s{}" (MiaouNames.to_variant_csname s)
+
     let pp_dnf d =
       List.map
         (fun a ->
-          List.map
-            (function
-             | Pos s -> sprintf "\\Variant{%s}" s
-             | Neg s -> sprintf "\\NotVariant{%s}" s)
-            a |> String.concat " and ")
+          List.map pp_atom a |> String.concat " and ")
         d |> String.concat " or "
 
     let pp_vc vc = variant_dnf false vc |> pp_dnf
+
+    let vc_to_items vc =
+      match variant_dnf false vc with
+      | [conjs] -> List.map (fun a -> Item (pp_atom a)) conjs
+      | [] -> Warn.fatal "variant_dnf returned empty list"
+      | d -> [Item (pp_dnf d)]
 
     let do_pp_rel_id  e1 e2 id =
       sprintf "\\%s{%s}{%s}" id (pp_evt e1) (pp_evt e2)
@@ -375,11 +381,11 @@ and cons_seqs (fs:exp list) (es:exp list) =
       | If (_,VariantCond vc,Konst (_,Empty _),e) ->
          let op = Inter in
          mk_list op
-           [Item (pp_vc (OpNot vc)); tr_rel e1 e2 e;]
+           (vc_to_items (OpNot vc) @ [tr_rel e1 e2 e;])
       | If (_,VariantCond vc,e,Konst (_,Empty _)) ->
          let op = Inter in
          mk_list op
-           [Item (pp_vc vc); tr_rel e1 e2 e;]
+           (vc_to_items vc @ [tr_rel e1 e2 e;])
       | If (_,VariantCond vc,a,b) ->
          let c = pp_vc vc
          and a = tr_rel e1 e2 a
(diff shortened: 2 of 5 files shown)


relokin:
Thanks for this Tiberiu.

I think it would be nice to avoid the post processing step and I am wondering whether we should actually remove the macros \Variant{#1} and \NotVariant{#1} altogether.

If miaou would produce a different macro for each variant we would be able to handle that specific macro in the catdefinitions.tex file correctly. We need to support:
* \Variant{ETS2} which we should change to \ETS2{} and then \ETS2{} would expand to "FEAT_ETS2 is implemented"
* \Variant{ETS3} which we should change to \ETS3{} and then \ETS3{} would expand to "FEAT_ETS3 is implemented"
* \Variant{DIC} which we should change to \DIC{} and then \DIC{} would expand to "The Effective value of DIC is 1"
* \NotVariant{DIC} which we should change to \NotDIC{} and then \NotDIC{} would expand to "The Effective value of DIC is 1"
* \Variant{IDC} which we should change to \IDC{} and then \IDC{} would expand to "The Effective value of IDC is 1"
* \NotVariant{IDC} which we should change to \NotIDC{} and then \NotIDC{} would expand to "The Effective value of IDC is 1"

@fsestini please chime in if you think there is a better solution.

fsestini:
Unless there’s a particular reason for preferring post-processing the markdown, I would agree with @relokin. Moreover, having dedicated macros in `catdefinitions.tex` would incidentally also benefit `uoaim7`, since that tool (at least currently) relies on `catdefinitions.tex` to match prose to corresponding cat symbols.

TiberiuBucur:
Thanks for your comments @relokin @fsestini. I made the change you requested, with a small tweak: unfortunately non-alphabetic characters are not supported in the name of new commands, so I transliterated the digits in letters. I converted the PR to ready to review, since the ticket has been communicated.
The CI seems to be failing due to an issue that is not apparent to me. I have not modified the required Ocaml version, nor did I use code that requires a newer version, so perhaps it is a hiccup with the runner.

herd/libdir/catdefinitions.tex
```diff
@@ -1,6 +1,12 @@
 \newcommand{\notthecase}[1]{it is not the case that #1}
-\newcommand{\Variant}[1]{#1 is implemented}
-\newcommand{\NotVariant}[1]{it is not the case that #1 is implemented}
+\newcommand{\ETSTwo}{FEAT\string_ETS2 is implemented}
+\newcommand{\NotETSTwo}{it is not the case that FEAT\string_ETS2 is implemented}
+\newcommand{\ETSThree}{FEAT\string_ETS3 is implemented}
+\newcommand{\NotETSThree}{it is not the case that FEAT\string_ETS3 is implemented}
```
fsestini: I'm assuming these commands should result in printing `FEAT_ETS2` or similar.

I don't know to what extent the markdown/pandoc post-processing is affected, but `\string_` does not seem to print an actual underscore when run through `pdflatex` on my local TeX Live installation. `\textunderscore{}`, however, does.

TiberiuBucur: Thanks @fsestini. After I conducted some experiments, I can say the following:

- While it is true that `\string` is not recognised by `pdflatex`, unfortunately `\textunderscore{}` is not recognised by `pandoc`.
- A solution that works for both engines is something like `\let\LiteralUnderscore\_`, which would then be used as `FEAT\LiteralUnderscore ETS2`. This to me seems like a bit of over-engineering.
- This occurrence is not the only thing that `pdflatex` seems to struggle with from the current `catdefinitions` dictionary. Another example is `\item \expandafter{\MakeUppercase\po{E\textsubscript{1}}{E\textsubscript{3}}}` in the expansion of `DSB-ob`, which `pdflatex` converts into "E1 E3", as opposed to the expected output (produced by pandoc), which is "E1 appears in program order before E3".

While the solution using the `\let` directive should work across most real TeX engines, I am not sure it is still 100% cross-compatible with any tool one might use to interpret their LaTeX. Given the current usage of `miaou7`, do you think going for something (close to a) cross-compatible solution is valuable? Imo, suggesting (maybe in a README file) that `pandoc` gives the best results for interpreting the generated `tex` sounds reasonable. @relokin, do you have any opinions?

tools/miaou.ml
```diff
@@ -283,25 +283,31 @@ and cons_seqs (fs:exp list) (es:exp list) =
...
-         and d2  =variant_dnf neg vc2 in
+         and d2 = variant_dnf neg vc2 in
          List.fold_right
            (fun a1 k ->
              List.fold_right
                (fun a2 k -> (a1@a2)::k) d2 k)
            d1 []
 
+    let pp_atom = function
+    | Pos s -> sprintf "\\%s{}" (MiaouNames.to_variant_csname s)
```
ShaleXIONG, line 301: Does this change break backwards compatibility? since you are now mapping `s` directly to `\s` rather than `\Variant` wrap?

Codex also flag this change might not compatible with `uoaim7` tool, @fsestini can you confirm whether it is true?

fsestini: `uoaim7` does not support parsing variants at all right now, so this change would not affect it.

TiberiuBucur: The only variants we are using are the ones @relokin mentioned above, which have been added separately in `catdefinitions.tex`. So no, it does not break backwards compatibility.

herd/libdir/catdefinitions.tex
relokin: Thank you both, is there any particular reason we can't use `\_`? I saw there is a macro defined with this name further down in catdefinitions.tex and that creates problems but I think we should remove the macro:
`\renewcommand{\_}{any Effect}`

fsestini: `\_` is an alias of `\textunderscore` AFAIK, so as far as building with pdflatex is concerned, I'd be happy if we can use `\_`. @TiberiuBucur could you please look into whether `\_` would work with pandoc and if the current `\renewcommand{\_}` macro can be removed?

fsestini:
@TiberiuBucur FYI: PR #1768, which is now merged, introduced some regression tests for miaou. It would be great if you could at some point rebase on that and update the tests to reflects your changes, as well as add any test cases for any variants that your PR affects but that aren't represented in the test suite yet.

TiberiuBucur:
Thanks @fsestini. I rebased on top of the latest master and updated the test outputs to match my changes. They already cover every place that should be affected by this PR. If you and @ShaleXIONG could re-review, that would be great. Thanks

herd/libdir/catdefinitions.tex
TiberiuBucur: Thanks @fsestini @relokin. I checked and miaou does not generate literal underscores which can be interpreted as a command. In fact the universal effect case, which the `\renewcommand{\_}{any Effect}` seemed to refer to, is handled by `anyevent`. I went ahead and removed it and simplified the definitions of the other variants. `pandoc` is happy with them.

herd/libdir/catdefinitions.tex
```diff
@@ -1,6 +1,12 @@
 \newcommand{\notthecase}[1]{it is not the case that #1}
-\newcommand{\Variant}[1]{#1 is implemented}
-\newcommand{\NotVariant}[1]{it is not the case that #1 is implemented}
+\newcommand{\ETSTwo}{FEAT_ETS2 is implemented}
+\newcommand{\NotETSTwo}{it is not the case that FEAT_ETS2 is implemented}
```
relokin:
```suggestion
\newcommand{\NotETSTwo}{\notthecase{\ETSTwo{}}
```

herd/libdir/catdefinitions.tex
```diff
@@ -1,6 +1,12 @@
...
+\newcommand{\ETSThree}{FEAT_ETS3 is implemented}
+\newcommand{\NotETSThree}{it is not the case that FEAT_ETS3 is implemented}
```
relokin:
```suggestion
\newcommand{\NotETSThree}{\notthecase{\ETSThree{}}
```

herd/libdir/catdefinitions.tex
relokin: or actually this unused, should we remove?

herd/libdir/catdefinitions.tex
relokin: or actually this unused, should we remove?

relokin (commented):
LGTM, just two minor commnents, and if Filippo is happy with this we can merge.

herd/libdir/catdefinitions.tex
TiberiuBucur: My opinion was we should keep it because it could theoretically be used in the future, and someone might not think of adding it to `catdefinitions.tex`, because it is not a new variant, since the affirmative case is treated (as opposed to a variant which was not mentioned at all).

fsestini (approved):
LGTM

herd/libdir/catdefinitions.tex
fsestini: > someone might not think of adding it to catdefinitions.tex

Out of scope for this PR, but I think a more principled defence against this scenario would be to have automated tests checking `miaou7`-generated macros against `catdefinitions.tex`.

TiberiuBucur: What would fall into `miaou7`-generated macros? Are you referring to the testing infrastructure @relokin developed?

fsestini: That would be a part of it, yes. Roughly, I'm thinking of a program that 1) parses all macros used in the regression tests (or generates an equivalent of them internally), then 2) parses `catdefinitions.tex`, and then 3) verifies that all macros mentioned in the former are actually defined in the latter.

Relatedly I think it would also make testing and interoperation with `uoaim7` easier if our source of truth w.r.t. supported macros were not `catdefinitions.tex` but rather some ocaml/json datastructure that can be compiled to latex but that is also easily machine-readable by other programs. Haven't thought too hard about how feasible this would be, but worth investigating at some point IMO.

PR #1789, opened 2026-04-10
[gen] Introduce `async` and `sync` variants in `diy`

ShaleXIONG:
This pull adds generator-side support for the `async` and `sync` variants and updates the MTE `diy` regression target.

- add `async` and `sync` to the generator variant set so it matches the variant already supported by `herd7`
- update fault-check generation to omit label and variable details when `async` is enabled 
- pass `-variant async` in `diy-test-mte` target in the CI and refresh the affected `gen/tests/AArch64.MTE/*.expected` logs, because those logs in CI originally are done in `async` mode.

ref #1786

### DIFF
--- a/gen/top_gen.ml
+++ b/gen/top_gen.ml
@@ -672,6 +672,7 @@ let max_set = IntSet.max_elt
     else []
 
   let do_memtag = O.variant Variant_gen.MemTag
+  let do_async = O.variant Variant_gen.Async
   let do_morello = O.variant Variant_gen.Morello
   let do_kvm = Variant_gen.is_kvm O.variant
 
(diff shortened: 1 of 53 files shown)


gen/top_gen.ml
relokin: Shale maybe I am complete wrong but is there a reason why you can't check here that we're in async mode and thefore the label and the location shouldn't be recorded?

gen/top_gen.ml
```diff
@@ -812,8 +813,8 @@ let max_set = IntSet.max_elt
           (* TODO: the `if-else` pattern on flags is not a good idea as it may short circuit *)
            if O.variant Variant_gen.NoFault then
              F.FaultAtomSet.empty,F.FaultAtomSet.empty
-           else if do_memtag || do_kvm || do_morello then
-             List.fold_left
+            else if do_memtag || do_kvm || do_morello then begin
```
relokin: nit: indentation

ShaleXIONG: Will change this.

gen/top_gen.ml
ShaleXIONG: Ah yes you are right! I just updated in a new commit and will rewrite the history once we settle.

relokin (approved):
LGTM thanks Shale!

PR #1790, opened 2026-04-13
[catalogue] Add missing label in LB+posat+amo.cas-no

relokin:


relokin:
@diaolo01 can you please have a look at this PR?

diaolo01:
Hi Nikos
The L1 label needs to be after the STG. This is because the postcondition requires the fixing of the tag only if the CAS has already progressed. Placing L1 before STG removes the cycle. With L1 after STG, we maintan the cycle that made the pre‑[PR](https://github.com/herd/herdtools7/pull/1771) outcome allowed and the post‑[PR](https://github.com/herd/herdtools7/pull/1771) outcome forbidden.

relokin:
> The L1 label needs to be after the STG. This is because the postcondition requires the fixing of the tag only if the CAS has already progressed. Placing L1 before STG removes the cycle. With L1 after STG, we maintan the cycle that made the pre‑https://github.com/herd/herdtools7/pull/1771 outcome allowed and the post‑https://github.com/herd/herdtools7/pull/1771 outcome forbidden.

I am not sure I follow, if I am not mistaken the fixed `LB+posat+amo.cas-no` is forbidden with master and allowed if I revert https://github.com/herd/herdtools7/pull/1771.

We could add the label L1 at the end of P1, but I would rather not do that. In this litmus test, the crucial point about the branch is its ordering properties and not its ability to change the control flow.

Alternatively, we should change the test and instead of the control dependency we could make the LDR in P1 a LDAR.

diaolo01:
I see what you mean. I assumed skipping the `STG` was the important part to the cycle, but testing shows it’s the `CBZ` control dependency between the `LDR` and `STG` that matters. Given that, I’m happy with the change.

relokin:
Thank you. Can I double check did you verify that the new test was allowed before your PR and forbidden with master?

diaolo01:
Yes, I can confirm the current shape of the test was allowed before the [PR](https://github.com/herd/herdtools7/pull/1771) and is now forbidden.

relokin:
Thanks!

PR #1791, opened 2026-04-14
[ASL-916] Try to fix build issues

hrutvik:
- Add new `{build,install,uninstall}-aslref` targets to `Makefile`
- Update `asllib/README.mld` with new build instructions and updated disclaimer
- Test new build instructions in `make-test-all` workflow
- Slightly reduce error messages when `dune build`ing without all optional dependencies

### DIFF
--- a/.github/workflows/make-test-all.yml
+++ b/.github/workflows/make-test-all.yml
@@ -61,3 +61,37 @@ jobs:
           opam pin asllib
           opam install .
 
+  aslref:
+    name: Install ASLRef on OCaml v5
+
+    runs-on: ubuntu-latest
+
+    env:
+      DUNE_CACHE: 'enabled-except-user-rules'
+      INSTALL_DIR: ${{ runner.tmp }}/prefix
+
+    steps:
+      - name: Checkout tree
+        uses: actions/checkout@v4
+
+      - name: Set-up OCaml 5
+        uses: ocaml/setup-ocaml@v3
+        with:
+          ocaml-compiler: 5
+          dune-cache: true
+
+      - name: Install dependencies
+        run: |
+          opam exec -- dune build asllib/aslref.opam
+          opam install ./asllib --deps-only --yes
+
+      - name: Create installation directory
+        run: mkdir $INSTALL_DIR
+
+      - name: Build and install ASLRef
+        run: |
+          opam exec -- make build-aslref PREFIX=$INSTALL_DIR
+          opam exec -- make install-aslref PREFIX=$INSTALL_DIR
+
+      - name: Check ASLRef installed
+        run: $INSTALL_DIR/aslref --version
--- a/Makefile
+++ b/Makefile
@@ -47,8 +47,17 @@ build-release: Version.ml
 
 build: check-deps | just-build
 
-install:
+install-herdtools:
 	sh ./dune-install.sh $(PREFIX)
+
+build-aslref:
+	dune build -p aslref --profile $(DUNE_PROFILE)
+
+install-aslref: build-aslref
+	# There are no lib files for aslref so we don't need dune-install.sh
+	dune install aslref --prefix $(PREFIX)
+
+install: install-herdtools install-aslref
 
 uninstall:
 	sh ./dune-uninstall.sh $(PREFIX)
--- a/Makefile
+++ b/Makefile
@@ -47,11 +47,23 @@ build-release: Version.ml
 
 build: check-deps | just-build
 
-install:
+install-herdtools:
 	sh ./dune-install.sh $(PREFIX)
+
+build-aslref:
+	dune build -p aslref --profile $(DUNE_PROFILE)
+
+install-aslref: build-aslref
+	# There are no lib files for aslref so we don't need dune-install.sh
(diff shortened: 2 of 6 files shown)


dune
```diff
@@ -17,6 +17,7 @@
  (name default)
  (deps
   (alias install)
+  (alias asllib/install)
```
HadrienRenaud: I wouldn't do this as `asllib/install` now is part of both `herdtools7` and `aslref` opam packages

.github/workflows/make-test-all.yml
```diff
@@ -61,3 +61,33 @@ jobs:
...
+          opam exec -- dune build asllib/aslref.opam
+          opam install ./asllib --deps-only --yes
+
+      - name: Build and install ASLRef
+        run: |
+          opam exec -- make build-aslref
+          opam exec -- make install-aslref
+
+      - name: Check ASLRef installed
+        run: opam exec -- aslref --version
```
HadrienRenaud: `PREFIX` is set to `$HOME` by default in the `Makefile`, which I don't think is in the `$PATH` of the GitHub runner.

`opam exec -- env` doesn't even show `PREFIX` on my machine.

Suggestion to use an explicit `PREFIX` directory:

```suggestion
      - name: Create installation directory
        run: mkdir $RUNNER_TEMP/prefix

      - name: Build and install ASLRef
        run: |
          opam exec -- make build-aslref PREFIX=$RUNNER_TEMP/prefix
          opam exec -- make install-aslref PREFIX=$RUNNER_TEMP/prefix

      - name: Check ASLRef installed
        run: $RUNNER_TEMP/prefix/aslref --version
```

Other possibility is (which I don't really like because it is less explicit):

```suggestion
      - name: Build and install ASLRef
        run: |
          opam exec -- make build-aslref
          opam exec -- make install-aslref

      - name: Check ASLRef installed
        run: $HOME/aslref --version
```

We could also add `$HOME` to `$PATH` in the runner, but it's even less explicit.

Makefile
```diff
@@ -991,18 +1003,16 @@ clean-asl-pseudocode:
 	@ $(MAKE) -C herd/libdir/asl-pseudocode clean
 
 .PHONY: asldoc
-asldoc: Version.ml
-	@ dune build -j $(J) --profile $(DUNE_PROFILE) $(ASLREF)
+asldoc: Version.ml build-aslref
```
HadrienRenaud: Do we still need `Version.ml` here? I don't think so

.github/workflows/make-test-all.yml
```diff
@@ -61,3 +61,37 @@ jobs:
...
+      - name: Create installation directory
+        run: mkdir $INSTALL_DIR
+
+      - name: Build and install ASLRef
+        run: |
+          opam exec -- make build-aslref PREFIX=$INSTALL_DIR
+          opam exec -- make install-aslref PREFIX=$INSTALL_DIR
+
+      - name: Check ASLRef installed
+        run: opam exec -- $INSTALL_DIR/aslref --version
```
HadrienRenaud: Do we need `opam exec --` here?
```suggestion
        run: $INSTALL_DIR/aslref --version
```

Makefile
```diff
@@ -991,18 +1003,16 @@ clean-asl-pseudocode:
...
 	@ $(MAKE) $(MFLAGS) -C asllib/doc all ASLREF=$(CURDIR)/$(ASLREF)
 
 .PHONY: clean-asldoc
 clean-asldoc:
 	@ $(MAKE) $(MFLAGS) -C asllib/doc clean
 
 .PHONY: type-check-asl
-type-check-asl: Version.ml
+type-check-asl: Version.ml build-aslref
```
HadrienRenaud: idem here:
```suggestion
type-check-asl: build-aslref
```

asllib/README.mld
```diff
@@ -31,22 +30,43 @@ The following steps have been tested on Unix.
...
+
 + Clone herdtools7:
   {@bash[
     $ git clone https://github.com/herd/herdtools7.git
   ]}
 + Install dependencies:
   {@bash[
     $ cd herdtools7
-    $ opam install . --deps-only
+    $ opam exec -- dune build asllib/aslref.opam
```
HadrienRenaud: We don't put `opam exec --` later on. So either we put it everywhere, either we suppose that people have run `eval $(opam env)`. I'm suggesting the former:

```suggestion
    $ dune build asllib/aslref.opam
```

asllib/README.mld
```diff
@@ -31,22 +30,43 @@ The following steps have been tested on Unix.
...
+    $ opam install aslref
+  ]}
+
+To uninstall, simply run
+  {@bash[
+    $ opam remove aslref
+  ]}
+
+{3 From sources}
+
```
HadrienRenaud, line 44: Maybe we need to add a line about running `eval $(opam env)`:

```suggestion
+ Add the `opam` environment to the current shell. For example on bash:
    {@bash[
      $ eval $(opam env)
    ]}
```

HadrienRenaud (approved):
Some small suggestions, but nothing major, it looks good to me, thanks @hrutvik.

PR #1799, opened 2026-04-20
[asl][www] Refactor running of ASLRef

hrutvik:
Introduce a `Runner` module to better delimit functionality for running ASLRef, allowing `aslref.ml` and `jaslref.ml` to each use this module without copying `aslref.ml` into `jaslref.ml`. As suggested by @HadrienRenaud.

### DIFF
--- a/asllib/Runner.ml
+++ b/asllib/Runner.ml
@@ -1,0 +1,203 @@
+(******************************************************************************)
+(*                                ASLRef                                      *)
+(******************************************************************************)
+(*
+ * SPDX-FileCopyrightText: Copyright 2022-2023 Arm Limited and/or its affiliates <open-source-office@arm.com>
+ * SPDX-License-Identifier: BSD-3-Clause
+ *)
+(******************************************************************************)
+(* Disclaimer:                                                                *)
+(* This material covers both ASLv0 (viz, the existing ASL pseudocode language *)
+(* which appears in the Arm Architecture Reference Manual) and ASLv1, a new,  *)
+(* experimental, and as yet unreleased version of ASL.                        *)
+(* This material is work in progress, more precisely at pre-Alpha quality as  *)
+(* per Arm’s quality standards.                                               *)
+(* In particular, this means that it would be premature to base any           *)
+(* production tool development on this material.                              *)
+(* However, any feedback, question, query and feature request would be most   *)
+(* welcome; those can be sent to Arm’s Architecture Formal Team Lead          *)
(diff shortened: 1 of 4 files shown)


asllib/asllib-www/jaslref.ml
```diff
@@ -67,10 +68,9 @@ let run_asl code version exec =
       Js_of_ocaml.Sys_js.update_file ~name:"web-input.asl" ~content:_code
   in
 
-  try run_with argv with
-  | Asllib.Error.ASLException e ->
-      Printf.eprintf "%s\n%!" (Asllib.Error.error_to_string e)
-  | Exit code ->
+  try Runner.run_with argv with
+  | Asllib.Error.ASLException _ -> assert false
```
Roman-Manevich: why hide the exception information?

asllib/Runner.ml
```diff
@@ -0,0 +1,153 @@
+open Typing
```
Roman-Manevich, line 26: A module-level comment would be nice.

Roman-Manevich (approved):
Looks good to me!

asllib/asllib-www/jaslref.ml
```diff
@@ -32,33 +32,34 @@ let run_asl code version exec =
...
+    Runner.
+      {
+        exec = _exec;
+        files =
+          [
+            ( (if String.equal _version "v0" then NormalV0 else NormalV1),
+              "web-input.asl" );
+          ];
+        opn = None;
+        (* … rest of the boolean flags left to their defaults … *)
```
HadrienRenaud: Why not define a `default_args` in `Runner.ml` and then just override the previous 3 here?

asllib/Runner.ml
```diff
@@ -0,0 +1,153 @@
...
+  override_mode : override_mode;
+  no_primitives : bool;
+  no_stdlib : bool;
+  no_stdlib0 : bool;
+  v0_use_split_chunks : bool;
+}
+
+exception Exit of int
+
+let run_with (args : args) : unit =
```
HadrienRenaud, line 79: I think it would be important to say that this function never returns.

HadrienRenaud (approved):
Thanks for this @hrutvik. Just 2 minor comments.

asllib/asllib-www/jaslref.ml
hrutvik: Due to the use of `or_exit` in `Runner.ml`, this line should never be reached - will add a comment

PR #1800, opened 2026-04-20
[ASL] Fix bug where environment updates from IF test expressions were…

solswords:
… reverted.

In the ASLRef implementation, on an IF expression with branches that are both simple expressions (containing no function calls), updates to the environment from the test expression are not kept. Test case:

```
var myglob : integer = 0;
func mytest () => boolean
begin
	myglob = 1;
	return TRUE;
end;
func main () => integer
begin
	var test = if mytest() then 0 else 1;
	println "myglob: ", myglob;
	return test;
end; 
```
prints
```
myglob: 0
```

### DIFF
--- a/asllib/tests/regressions.t/if-test-env-updated.asl
+++ b/asllib/tests/regressions.t/if-test-env-updated.asl
@@ -1,0 +1,12 @@
+var myglob : integer = 0;
+func mytest () => boolean
+begin
+  myglob = 1;
+  return TRUE;
+end;
+func main () => integer
+begin
+  var test = if mytest() then 0 else 1;
+  println "myglob: ", myglob;
+  return test;
+end;
--- a/asllib/tests/regressions.t/if-test-env-updated.asl
+++ b/asllib/tests/regressions.t/if-test-env-updated.asl
@@ -7,6 +7,6 @@ end;
 func main () => integer
 begin
   var test = if mytest() then 0 else 1;
-  println "myglob: ", myglob;
-  return test;
+  assert myglob == 1;
+  return 0;
 end;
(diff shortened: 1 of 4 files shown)


hrutvik (commented):
Please could you add the test exercising this to `asllib/tests/regressions.t`?

HadrienRenaud (approved):
Nicely spotted, thanks!

asllib/tests/regressions.t/if-test-env-updated.asl
```diff
@@ -0,0 +1,12 @@
+var myglob : integer = 0;
+func mytest () => boolean
+begin
+  myglob = 1;
+  return TRUE;
+end;
+func main () => integer
+begin
+  var test = if mytest() then 0 else 1;
+  println "myglob: ", myglob;
```
HadrienRenaud: It's usually easier to test with assertions than with `println` (which I think we're trying to move away from).

```suggestion
  assert myglob == 1;
```

hrutvik (approved):

solswords:
Not sure we should hold up the PR for this, but the code refers to SemanticsRule.ECondSimple which doesn't seem to exist in the document. Should we get rid of it?

hrutvik:
^ question for @Roman-Manevich

Roman-Manevich:
> Not sure we should hold up the PR for this, but the code refers to SemanticsRule.ECondSimple which doesn't seem to exist in the document. Should we get rid of it?

Yes, please remove `SemanticsRule.ECondSimple`. By the way, there's no need to fix the reference, since it actually gets it right.

Roman-Manevich (approved):
Looks good to me!

PR #1803, opened 2026-04-22
[herd] Add rmw to po, except for the model `aarch64.cat`

maranget:
The simulator herd computes the program order by itself.

In the case of hardware semantics, effects from the same instruction were excluded from po. This is wrong for atomic pairs, where this order is important, _e.g._ for the internal check. However, in the case of the `aarch64.cat` model, this situation is handled correctly by inserting `rmw` where appropriate, _i.e._ in the internal check coRW1-Exp.

As the model `aarch64.cat` identifies itself with the `catdep` flag, we add `rmw` to `po` by default, except for `catdep` models.

As a consequence, the meaning of catdep is now as follows:
  1. The model computes the dependencies by itself,
  2. and it considers that the po relation always relates effects generated by different instruction executions.

### DIFF
--- a/herd/machModelChecker.ml
+++ b/herd/machModelChecker.ml
@@ -351,10 +351,19 @@ module Make
           (if O.wide_po then
              E.EventRel.filter_nodes relevant
            else
-             E.EventRel.restrict_rel
-               (fun e1 e2 ->
-                  relevant e1 && relevant e2
-                  && not (E.same_instance e1 e2)))
+             fun po ->
+               let po =
+                 E.EventRel.restrict_rel
+                   (fun e1 e2 ->
+                     relevant e1 && relevant e2
+                     && (not (E.same_instance e1 e2)))
+                   po in
+               if catdep then po
+               else
+                 (* Non-catdep models do not explicit include
+                    rmw into po where appropriate,
+                    for instance in their internal check. *)
+                 E.EventRel.union conc.S.atomic_load_store po)
           conc.S.po in
       let partial_po =
         lazy begin
--- a/herd/tests/instructions/AArch64/A264.litmus
+++ b/herd/tests/instructions/AArch64/A264.litmus
@@ -1,0 +1,9 @@
+AArch64 A264
+{
+0:X2=y;
+1:X2=y;
+}
+ P0          | P1            ;
+ MOV W1,#1   | MOV W3,#2     ;
+ STR W1,[X2] | SWP W3,W1,[X2] ;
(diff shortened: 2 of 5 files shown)


herd/machModelChecker.ml
```diff
@@ -351,10 +351,19 @@ module Make
...
-             E.EventRel.restrict_rel
-               (fun e1 e2 ->
-                  relevant e1 && relevant e2
-                  && not (E.same_instance e1 e2)))
+             fun po ->
+               let po =
+                 E.EventRel.restrict_rel
+                   (fun e1 e2 ->
+                     relevant e1 && relevant e2
+                     && (not (E.same_instance e1 e2)))
```
relokin, line 359: This doesn't affect AArch64 so I am not worried but this means that for other architectures `po` cannot start/end on a register event?

maranget: You are correct. However: 
  + No cat model uses po for registers events, except for asl and catdep models (AArch64). In those cases all events are relevant.
  + Program order for registers is used internally for registers events, mostly for computing the rf-reg relation. To that aim, herd uses the full program order as computed by the po_iico function in memUtils.ml.

It may be troublesome that execution diagrams show po arrows even from/to register events. However those 'po' edes are cosmetic ones that relate the iico-maximal events of one instruction to the iico-minimal events of the next instuction. Consider the test:
```
X86_64 T
{ }
  P0           ;
 movl $1,%eax  ;
 movl $2,%ebx  ;
forall 0:rax=1 /\ 0:rbx=2 /\ 0:rcx=0
```
and the execution diagram:
T

 To show the actual 'po' relation, use `-doshow po`. This operation will add nothing to the diagram above.

maranget: For completeness, here are a X86_test and diagram with memory effects:
```
X86_64 U
{
int x=1;
0:rax=x;
0:rbx=y;
}
 P0               ;
 movl(%rax),%ecx  ;
 movl $2,(%rbx)   ;

forall 0:rcx=1 /\ [y]=2
```
U

The display options of **herd7** are: `show all -showevents all -doshow po -showpo false -unshow iico_data  -withbox true -labelbox true`. As a result, the diagram highlights the `po` relation as perceived by the Cat interpreter.

maranget: Hi @relokin. I'd like to merge this PR, as it is a bug fix, the bug being present in the web interface. Your opinion?

relokin (approved):
Thanks for this Luc!

maranget:
Merged, thanks @relokin.

PR #1804, opened 2026-04-22
[herd] Fix "Post-condition introduces new symbolic location" error.

maranget:
Symbols referred from code are part of symbols defined by a test and can appear in the post-condition.

More specifically, the following test is now accepted again:
```
X86_64 A012
(* Implicit initialsation of int x=0; *)
{ }
  P0         ;
 movl $1,(x) ;
forall [x]=1
```
Note: the error was introduced by PR #1607 to flag missing symbols in post-condition, a very useful check. Unfortunately, the X86 and X86_64 tests generated by **diy7** were rejected. This problem is now fixed.

### DIFF
--- a/herd/test_herd.ml
+++ b/herd/test_herd.ml
@@ -129,6 +129,15 @@ module Make(A:Arch_herd.S) =
             | Some c -> add_constant_symbols acc c) in
       ConstrGen.fold_constr add_atom constr acc
 
+    let locs_in_ins =
+      A.pseudo_fold
+        (A.fold_addrs
+           (fun v acc -> add_constant_symbols acc v))
+
+    let locs_in_code =
+       List.fold_left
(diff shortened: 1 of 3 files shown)


HadrienRenaud:
This is strange that this was not caught in CI. This might be important as similar tests are on the web interface. Are the x86 catalogue tests run in CI?

relokin:
Thanks Luc, the change looks good to me.

diaolo01:
Looks good to me! 
For CI coverage, only the unit tests from `herd/tests/` are run for x86/x86_64.

maranget:
Merged, thanks @relokin, @diaolo01 nd @HadrienRenaud for your approval and comments. PR #1806 add some basic tests by running the X86_64 tests of the catalogue.

PR #1805, opened 2026-04-23
[Makefile] Add pattern rules and simplify the test targets

relokin:
Simplify the makefile but introducing pattern rules for the herd tests. In addition, clean up the target names by introducing the following pattern for test targets:

test.<tool>.<cata/unit>.<suite>

for example:

test.herd.cata.aarch64-pick


Makefile
```diff
@@ -385,313 +244,66 @@ test-aarch64-noasl-mixed:
...
-		-herd-path $(HERD) \
-		-libdir-path ./herd/libdir \
-		-kinds-path catalogue/aarch64-faults/tests/kinds.txt \
-		-shelf-path catalogue/aarch64-faults/shelf.py \
-		-conf-path  catalogue/aarch64-faults/cfgs/asl.cfg \
-		$(REGRESSION_TEST_MODE)
-	@ echo "herd7 catalogue aarch64-faults tests (ASL): OK"
-
-cata-test:: pick-test-mixed
-pick-test-mixed:
```
relokin, line 603: @maranget just a quick question do you know what is the purpose of this test?

relokin:
@maranget @HadrienRenaud this is WIP but I wanted to check with you if you would be ok with such a change in principle. I find that the Makefile has grown a bit too large and it's hard to understand what's going on. Hopefully by introducing some patterns we can make it shorter, easier to extend and easier to follow.

Makefile
```diff
@@ -385,313 +244,66 @@ test-aarch64-noasl-mixed:
...
-test.vmsa+ifetch:
-	@ echo
-	$(HERD_REGRESSION_TEST) \
-		-herd-path $(HERD) \
-		-libdir-path ./herd/libdir \
-		-litmus-dir ./herd/tests/instructions/AArch64.vmsa+ifetch \
-		-conf ./herd/tests/instructions/AArch64.vmsa+ifetch/vmsa+ifetch.cfg \
-		$(REGRESSION_TEST_MODE)
-	@ echo "herd7 AArch64 VMSA+ifetch instructions tests: OK"
+test-all-asl:: test.herd-asl.cata.aarch64-VMSA
```
HadrienRenaud, line 323: Is it on purpose that you've removed the `-verbose` argument?

HadrienRenaud (commented):
I'm generally very pro this.

(I haven't approved because I didn't check that there was no change in the actual commands)

Makefile
relokin: I didn't realise this was intentional. I will restore it unless you think it's not really necessary.

relokin:
> I'm generally very pro this.
> 
> (I haven't approved because I didn't check that there was no change in the actual commands)

Thanks Hadrien! That's absolutely fine, just wanted to check if the direction of the patch is ok. I will do another pass. Perhaps one thing that I wanted to check with you, the ASL test targets are not as uniform as the rest, is the non-uniformity useful and intentional?

HadrienRenaud:
> the ASL test targets are not as uniform as the rest, is the non-uniformity useful and intentional?

I'm not sure what non-uniformity you are talking about. If you are talking about naming conventions, it is probably non-intensional. If you are talking about something else, I'm not sure. Could you expand or give an example?

catalogue/aarch64-ifetch/ci-shelf.py
```diff
@@ -0,0 +1 @@
+shelf.py
```
diaolo01: I notice this shelf covers a much smaller set of tests than before. Is this intended?

diaolo01 (commented):
This cleanup is useful. Generally, looks good to me.

catalogue/aarch64-ifetch/ci-shelf.py
relokin: Does it? I think it's a soft link to shelf.py, so it should have the same set of tests as before.

diaolo01: If I am reading this correctly, previously the makefile used to call `shelf-test.py` and now it calls `shelf.py` via `ci-shelf.py`. shelf has considerably less tests than shelf-test, which was deleted.

Makefile
HadrienRenaud: I think it is necessary, thanks.

catalogue/aarch64-ifetch/ci-shelf.py
relokin: Oh sorry I missed that. Should be fixed now.

Makefile
relokin: Thanks Hadrien!

relokin:
Here is a summary of the changes that this PR brings:
  - make test, make test-local, and make cata-test all got broader because cata-test now includes aarch64-cas and aarch64-ETS2, which used to be test-all-only. The catalague tests for aarch-ETS2, aarch-ETS3 and aarch64-cas are quite fast.
  - make test and make test-local now run AArch64.vmsa+mte instruction tests, whereas previously that suite was test-all-only. These are simple instruction tests.
  - The refactor silently tightens SVE ASL instruction testing. The generic ASL rule adds -checkstates to every test.herd-asl.inst.%, and SVE is still included in test, test-local, and test-all- asl. Before this PR, the dedicated test.asl.sve rule did not pass -checkstates, so those targets now run a stricter test contract.
  -  cata-test and its callers now impose a 16s herd timeout on suites that previously had none
 - previously the aarch64-pick was run with -variant mixed as well (target pick-test-mixed). This PR has removed this regression.

@HadrienRenaud, @fsestini @maranget are these changes reasonable?

HadrienRenaud:
Thanks @relokin for this PR and the very useful summary. I'm in favour of merging.

> cata-test and its callers now impose a 16s herd timeout on suites that previously had none

I'm wondering if we should maybe be a little bit more careful with tests with timeouts. A few suggestions, probably not for this PR (maybe for an issue for the moment?):
1. Logging the timeout tests
2. Having the tooling warn if a test timeout for the first time, in order to prompt the PR writer to justify why the performance regression is worth the PR
3. Running the tests with a bigger timeout in `make test-all`

relokin:
> Thanks @relokin for this PR and the very useful summary. I'm in favour of merging.
>

Thanks for the review Hadrien!
 
> > cata-test and its callers now impose a 16s herd timeout on suites that previously had none
> 
> I'm wondering if we should maybe be a little bit more careful with tests with timeouts. A few suggestions, probably not for this PR (maybe for an issue for the moment?):
> 
>     1. Logging the timeout tests
> 
>     2. Having the tooling warn if a test timeout for the first time, in order to prompt the PR writer to justify why the performance regression is worth the PR
> 
>     3. Running the tests with a bigger timeout in `make test-all`

I agree silent timeouts are not desirable in our CI.
- Tests that we know take unreasonably long to run shouldn't be part of the CI.
- All tests that run as part of the CI either shouldn't have a timeout, or even better register a failure if they timeout.
This way we can have a better control of the performance implications of PRs.

fsestini:
This PR looks reasonable to me. A few comments:

- Running `make test-all` fails at the `test.aarch64assumptions` step on my end. This is probably due to its use of `dirs-and-confs.txt`, which now points to non-existing config files.
- I think `catalogue/aarch64-VMSA/README.md:13` needs to be updated to point to the new kinds file.
- I agree on the timeouts: we should only include fast tests as part of CI, and have timeouts result in a hard failure.

maranget:
Hi @relokin. Nice work. I have had the same problem as @fsestini. 

Additionally I could integrate two more tests in the general `test.herd.inst` pattern, with minor modification of pattern and tools. See commit c529dc5c000c28e34fd6ec05eb16be5445be79c8.

relokin:
Thanks for having a look @fsestini!

> This PR looks reasonable to me. A few comments:
> 
>     * Running `make test-all` fails at the `test.aarch64assumptions` step on my end. This is probably due to its use of `dirs-and-confs.txt`, which now points to non-existing config files.
> 
Good catch, I think this is now fixed.

>     * I think `catalogue/aarch64-VMSA/README.md:13` needs to be updated to point to the new kinds file.
> 

Fixed, thank you!

>     * I agree on the timeouts: we should only include fast tests as part of CI, and have timeouts result in a hard failure.

Let's fix that properly in a separate PR. We need to create a new list for catalogue/aarch64-VMSA and for some ASL tests.

relokin:
> Additionally I could integrate two more tests in the general `test.herd.inst` pattern, with minor modification of pattern and tools. See commit [c529dc5](https://github.com/herd/herdtools7/commit/c529dc5c000c28e34fd6ec05eb16be5445be79c8).

Thanks @maranget  I've added your commit to the PR.

relokin:
@maranget I've made a minor edit to your commit as some cfg files were missing. Are you happy for me to merge this? cc: @HadrienRenaud.

HadrienRenaud (approved):
All good for me, thanks @relokin.

maranget:
I am happy, please merge.

relokin:
Merged thank you all for the reviews and suggestions!

PR #1807, opened 2026-04-23
[tools] Fix regex matching bug in dot2desc

TiberiuBucur:
A bug was discovered when running dot2desc with an `-instr` parameter on dot graphs that contain instructions where the operands are separated by `, ` (mind the whitespace), as opposed to just commas. The parameter matching did not work for the operands after the whitespace. For example, on an execution graph of `PACDA X0, X1`, running `dot2desc7` with `-instr "PACDA Xd,Xn"` yielded proses in which `X0` was replaced by `X~d~`, but `X1` was NOT replaced by `X~n~`. This PR addresses that.

### DIFF
--- a/tools/dot2desc.ml
+++ b/tools/dot2desc.ml
@@ -672,7 +672,7 @@ module DotGraph = struct
     [CAS X0, X1, [X2]] in their label, and [instr] is [CAS Xs, Xt, [Xn]]. The map
     will contain the entries [X0] -> [Xs], [X1] -> [Xt] and [X2] -> [Xn]. *)
   let get_param_maps stmts instr =
-    let regex = Str.regexp {|\([A-Z]+\)\( \([][,a-zA-Z0-9\._]+\)\)?|} in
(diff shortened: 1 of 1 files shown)


tools/dot2desc.ml
```diff
@@ -672,14 +672,15 @@ module DotGraph = struct
     [CAS X0, X1, [X2]] in their label, and [instr] is [CAS Xs, Xt, [Xn]]. The map
     will contain the entries [X0] -> [Xs], [X1] -> [Xt] and [X2] -> [Xn]. *)
   let get_param_maps stmts instr =
-    let regex = Str.regexp {|\([A-Z]+\)\( \([][,a-zA-Z0-9\._]+\)\)?|} in
+    let regex = Str.regexp {|\([A-Z]+\)\( \([][, a-zA-Z0-9\._]+\)\)?$|} in
```
ShaleXIONG, line 675: You really mean to add the whitespace here right?

TiberiuBucur: Yes, that is the bug that the PR addresses.

ShaleXIONG (approved):
Look good to me

PR #1808, opened 2026-04-24
[herd] Fix placement of po edges in dot graphs

fsestini:
This PR fixes a regression introduced by my previous PR #1628

The goal of #1628 was to add information to dot graph nodes reflecting the static order of each effect's instruction. However, it erroneously ended up affecting the way `po` arrows are displayed as well. For this reason, I'm reverting the changes of PR #1628 that pertain to dot graphs rendering. I'm also adding a dedicated regression test.


herd/tests/other/poi-loop.t/run.t
```diff
@@ -0,0 +1,17 @@
+Check that po indices/edges are assigned correctly in the presence of loops (only checks the last (longest) execution graph)
+  $ herd7 -set-libdir ../libdir -dumpes true -o - ./wait-flag.litmus | awk '/digraph G \{/ {block=""} {block = block $0 "\n"} END {printf "%s", block}' | grep -e 'poi' -e 'label="po"'
```
relokin, line 2: Could you make use of splitdot instead?

herd/tests/other/poi-loop.t/wait-flag.litmus
```diff
@@ -0,0 +1,9 @@
+AArch64 wait-flag
+{
+   0:X0=x; 1:X0=x;
+}
+   P0          | P1          ;
+   MOV W1,#1   |L0:          ;
+   STR W1,[X0] | LDR W1,[X0] ;
+               | CBZ W1,L0   ;
+exists(1:X1=1)
```
relokin:
```suggestion
```

relokin (commented):
Two suggestions otherwise the PR looks good to me. Thanks for fixing this!

herd/tests/other/poi-loop.t/wait-flag.litmus
fsestini: Done

herd/tests/other/poi-loop.t/run.t
fsestini: Good question. The current test script uses `herd -o -` to print the dot graphs to stdout and process them. However, it seems splitdot can only consume and produce `.dot` files stored on the filesystem. So we could use splitdot in principle, although to achieve the same result we'd need to do through some temporary files. Personally I prefer the more "stateless" current approach, but I don't mind either way.

relokin: ack!

relokin (approved):
LGTM, thanks Filippo!

PR #1810, opened 2026-04-28
[herd] Improve  filter application

maranget:
Litmus tests may have an optional "filter" proposition that filters outcomes.
    
This PR extends the optimisation of PR #1657 to the filters. Namely, from filter some constraints may be extracted and executions be pruned early.


herd/mem.ml
```diff
@@ -1014,26 +1019,43 @@ let get_rf_value test read =
     let debug_solver = C.debug.Debug_herd.solver > 0
 
     let wanted_final_value test =
+      let add_check loc v m =
+        A.LocMap.update loc
+          (function
+           | None -> Some v
+           | Some v0 ->
+               if A.V.equal v0 v then  Some v
```
HadrienRenaud: nit: no need to rebuild the map in this case

```suggestion
               if A.V.equal v0 v then Some v0
```

HadrienRenaud (approved):
This looks good to me.

The only bit I'm not sure is whether a self-contradicting filter should be considered a user error, maybe more a warning?

herd/mem.ml
maranget: Nice catch, thanks.

maranget:
> The only bit I'm not sure is whether a self-contradicting filter should be considered a user error, maybe more a warning?

I have hesitated a bit. In the end, such a mistake calls for a fix. A warning might be overlooked...

HadrienRenaud:
> A warning might be overlooked...

That makes sense for me. If we need to allow it for some reason, we can always relax it to a simple warning later.

From my point of view, you can merge when you want.

maranget:
Merged, thanks @HadrienRenaud

PR #1811, opened 2026-04-28
[litmus] Add makefile recipe for catalogue/aarch64-faults

diaolo01:
This PR adds a `catalogue/aarch64-faults` recipe to Makefile.aarch64 and skips three tests that cannot currently be run with ltmus7:

  - `LDRredF`: HW supporting MTE is still rare
  - `SVC-pte.exs1eis0`, `SVC-pte.exs1eis1`: the two tests use custom fault handlers, which litmus7 cannot combine with `fault(...)` postconditions. They also execute as infinite loops, which would not work with litmus7.


relokin (approved):
LGTM, thanks @diaolo01 !

PR #1813, opened 2026-04-29
[ci,asl] Move use of runner where it is available

HadrienRenaud:
See documentation here: https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#available-contexts


HadrienRenaud:
Duplicate of #1814.

PR #1816, opened 2026-04-30
[herd] Handle memory locations in speedcheck/filter optimisation

maranget:
With explicit speedcheck setting or filter presence in test, one may specify the final value of some memory location(s). If so, we use this information to prune irrelevant executions earlier. This is an extension to PR #1657 and #1810.

### DIFF
--- a/herd/mem.ml
+++ b/herd/mem.ml
@@ -647,7 +647,7 @@ module Make(C:Config) (S:Sem.Semantics) : S with module S = S	=
             (Label.Set.pp_str ","  Label.pp ii.A.labels) ;
         m ii in
 
-      let  sem_instr =  SM.build_semantics test in
+      let sem_instr =  SM.build_semantics test in
       let rec add_next_instr re_exec fetch_proc proc env seen addr (code_instr : I.t) nexts =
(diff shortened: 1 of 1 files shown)


herd/mem.ml
```diff
@@ -2063,6 +2060,26 @@ let get_rf_value test read =
...
+                   match
+                     List.filter
+                       (fun e ->
+                          match E.written_of e with
+                          | Some v1 -> V.equal v1 v0
+                          | None -> assert false)
+                       evts
+                   with
+                   | [] -> None
+                   | evts -> Some evts)
```
HadrienRenaud: nit: I have a slight preference for simplifying matched expressions

```suggestion
                   let evts = 
                     List.filter
                       (fun e ->
                          match E.written_of e with
                          | Some v1 -> V.equal v1 v0
                          | None -> assert false)
                       evts in
                   match evts with [] -> None | _ -> Some evts)
```

HadrienRenaud (approved):
LGTM, thanks Luc.

It looks simpler than what was in my head, nice one.

herd/mem.ml
maranget: Let us do that.

maranget:

PR #1819, opened 2026-05-06
[all] Bump minimum OCaml version to 4.14

fsestini:
OCaml < 4.11 is [no longer supported](https://github.com/ocaml/opam-repository/pull/29848) on `opam` and won't build with it. See also [this](https://github.com/herd/herdtools7/actions/runs/25429247642/job/74590828906?pr=1812) failing CI run.

I take this as a good opportunity to bump the minimum version of OCaml supported by the project. In light of [this discussion](https://github.com/herd/herdtools7/issues/1783), I'm proposing to bump the minimum to 4.14.0, as a version that is still conservatively < 5 but that also brings significant advantages in terms of stdlib improvements.

### DIFF
--- a/asllib/menhir2bnfc/dune-project
+++ b/asllib/menhir2bnfc/dune-project
@@ -4,4 +4,6 @@
 (cram enable)
 (package
  (name menhir2bnfc)
- (allow_empty))
+ (allow_empty)
+ (depends
+  (menhir (>= 20231231))))
--- a/.github/workflows/check-opam-minimals.yml
+++ b/.github/workflows/check-opam-minimals.yml
@@ -31,10 +31,10 @@ jobs:
       - name: Checkout tree
         uses: actions/checkout@v4
 
-      - name: Set-up OCaml 4.08
+      - name: Set-up OCaml 4.14
         uses: ocaml/setup-ocaml@v3
         with:
-          ocaml-compiler: 4.08.1
+          ocaml-compiler: 4.14.0
       - run: |
           opam install . --deps-only --with-test --solver=builtin-mccs+glpk --criteria="-new"
           opam exec -- make build test install DUNE_PROFILE=dev
--- a/.github/workflows/make-test-all.yml
+++ b/.github/workflows/make-test-all.yml
@@ -23,7 +23,6 @@ jobs:
       fail-fast: false
       matrix:
         ocaml-version:
-          - "4.08.1"
           - "4.14"
           - "5.3"
 
--- a/.github/workflows/make-test-diy.yml
+++ b/.github/workflows/make-test-diy.yml
@@ -24,7 +24,7 @@ jobs:
           - ubuntu-latest
         cfg:
             # Minimal versions: we want to test both ends of the supported space
-          - ocaml-version: "ocaml-base-compiler.4.14.0,dune.3.10.0,zarith.1.13,menhir.20220210,logs.0.7.0"
+          - ocaml-version: "ocaml-base-compiler.4.14.0,dune.3.10.0,zarith.1.13,menhir.20231231,logs.0.7.0"
             job-name: "OCaml v4.14.0"
             dune-cache: "enabled"
 
--- a/.github/workflows/make-test-diy.yml
+++ b/.github/workflows/make-test-diy.yml
@@ -24,8 +24,8 @@ jobs:
           - ubuntu-latest
         cfg:
             # Minimal versions: we want to test both ends of the supported space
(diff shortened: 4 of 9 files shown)


relokin (approved):
LGTM, but please wait for wider consensus before merging!

HadrienRenaud (approved):
I'm very fine with the bump from 4.08 to 4.11, but I am not convinced of the necessity of bumping from 4.11 to 4.14. That said, I don't personally use any version before 4.14 (except for testing) so this change is not bothering me.
I am approving this PR, but please wait (at least) for @maranget's approval before merging.

Note: maybe a follow-up of this PR is to refactor `lib/misc.ml` to use the standard library functions?

asllib/menhir2bnfc/dune-project
```diff
@@ -4,4 +4,6 @@
 (cram enable)
 (package
  (name menhir2bnfc)
- (allow_empty))
+ (allow_empty)
+ (depends
+  (menhir (>= 20231231))))
```
hrutvik, line 9: I think `20230608` should be sufficient here

hrutvik (commented):
I think the following files could also be updated in this PR:
- `README.md`
- `asllib/dune-project`
- `asllib/aslspec/dune-project`

> Note: maybe a follow-up of this PR is to refactor lib/misc.ml to use the standard library functions?

Agreed there should be a follow-up PR - also to add:
- `asllib/ASTUtils.ml`
- `asllib/aslspec/utils.ml`
- `internal/lib/base.ml` (maybe?)

asllib/menhir2bnfc/dune-project
fsestini: Tried that version locally but it failed with a different error:

```
File "asllib/menhir2bnfc/CvtGrammar.ml", line 189, characters 28-45:
189 |       | Some (N n, _, _) -> Nonterminal.equal n nterm
                                  ^^^^^^^^^^^^^^^^^
Error: Unbound value Nonterminal.equal
```

ShaleXIONG:
@TiberiuBucur can you check the website `herd-www` too? I believe it is not part of the CI?

asllib/menhir2bnfc/dune-project
hrutvik: Ah I had not realised, good to know - thank you!

maranget, line 9: Shouln't we modify herdtool7.opam accordingly?

maranget: And also the action file .github/workflows/make-test.yml that contains an explicit reference to an older version of menhir. That said I do not understand why this failure on minimal versions occurs now and not earlier.

HadrienRenaud: Theoretically, `menhir2bnfc` is a different package, so it should be fine to have it on the side.
That being said, in practice, I cannot make it work (see for example HadrienRenaud/herdtools7#13) because of some (strange) workings of dune.

maranget: Setting   20231231 as the minimal version of menhir may fix the CI failure, something we need now.

fsestini: > That said I do not understand why this failure on minimal versions occurs now and not earlier.

That is because earlier, the 'minimal versions' build step used a version of OCaml that was not recent enough to trigger a `menhir2bnfc` build, which is defined to only build on OCaml 4.14 or above.

> Setting 20231231 as the minimal version of menhir may fix the CI failure, something we need now.

I agree that we probably should just set the minimum version to 20231231 for now, and think about how to compartmentalise menhir2bnfc later.

HadrienRenaud: Prototype: HadrienRenaud/herdtools7#14

HadrienRenaud: I don't think so, it needs `Nonterminal.equal` which only appears on 20231231. See working version here: HadrienRenaud/herdtools7#14

fsestini:
Thank you very much for the comments @HadrienRenaud @maranget @hrutvik ! I've bumped versions where needed, and CI is passing. If you're happy with this, we can merge.

asllib/menhir2bnfc/dune-project
maranget: CI passes now, well done.

maranget (commented):
As CI is now restored, I suggest merging ASAP.

maranget (approved):

PR #1821, opened 2026-05-06
[build] Remove dependency from build to install

maranget:
When developing I repetitively use the command
```
% make all install PREFIX=/opt/WIP
```

If some install target in Makefile depends on some build target, as it was the case for aslref, the command above
triggers unnecessary recompilations.

Suppressing this dependency should not be a problem, when one follows the usual practice of separating build and installation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,11 +54,11 @@ install-herdtools:
 build-aslref:
 	dune build -p aslref --profile $(DUNE_PROFILE)
 
-install-aslref: build-aslref
+install-aslref:
 	# There are no lib files for aslref so we don't need dune-install.sh
 	dune install aslref --prefix $(PREFIX)
 
-install: install-herdtools install-aslref
+install: install-herdtools
 
 uninstall:
 	sh ./dune-uninstall.sh $(PREFIX)
--- a/.github/workflows/make-test.yml
+++ b/.github/workflows/make-test.yml
@@ -55,7 +55,7 @@ jobs:
       - name: Install opam dependencies
         run: opam install . --deps-only --yes --with-test
 
-      - run: opam exec -- make build DUNE_PROFILE=dev
+      - run: opam exec -- make all DUNE_PROFILE=dev
 
       - run: opam exec -- make test DUNE_PROFILE=dev
 


maranget:
Hi @hrutvik and @HadrienRenaud  The problem I ran across may be related to the recent PR #1791. I'd value your comments.

maranget:
Hum, problem unsolved, now the CI is failing...

HadrienRenaud:
Hi @maranget, see #1819 for CI failures.

For the content of this PR, I think @hrutvik will be better placed than me to answer.

maranget:
> Hi @maranget, see #1819 for CI failures.

Well the failure of "lastest lts" is not related to PR #1819.

maranget:
Hum PR #1818 removed build-aslref from all, shouldn't we also remove install-aslref from install?

maranget:
> Hum PR #1818 removed build-aslrel from all, shouldn't we also remove install-aslref from install?

Just done that. Apparently it works. What we can't do is building both herdtools7 and aslref and then install them together. Maybe because some *.install files are removed somewhere. To build and install everything;
```
% make all install # build and install herdtools7
% make build-aslref install-aslref
```
Then if we try `make all` again, there are a lot of recompilations....

HadrienRenaud:
> Then if we try make all again, there are a lot of recompilations....

But are you going to do this workflow often? I'm wondering if that's not good enough

maranget:
> > Then if we try make all again, there are a lot of recompilations....
> 
> But are you going to do this workflow often? I'm wondering if that's not good enough

The workflow `make all install PREFIX=/opt/WIP`, modify something, `make all install PREFIX=/opt/WIP`, is my usual way of working. I have no idea wether  it harms or not not to recompile aslref for me. I guess it is not.
 Anyway, if you can help, I'd like to understand:

+ What is the meaning of `make all` ? If this means build and install everything why PR #1818 ?
+ What is the meaning of `make install`? In any case, I expect it to be synchrornised with `make all`.
+ `dune` apprently removes some `*.install` files when one compiles or (install ?) some packages. Do we understand this? Can we avoid it?

maranget:
CI failing again...

Makefile
HadrienRenaud: Maybe:
```suggestion
install: install-herdtools
```

maranget: Thanks, same player shoot again.

maranget:
CI succeeds. I'd like to merge, @HadrienRenaud?

maranget:
Merged, considering that you approval of the last modification is an approval... Thanks @HadrienRenaud.

PR #1825, opened 2026-05-07
[asl][asl reference] Fix name-mangling for return-identifiers and overridden functions

Roman-Manevich:
* Fixed a bug in generating identifiers for the write effects for values returned by a subprogram call. The values generated by `SemanticsRule.WriteFolder` were arbitrary fresh identiiers, whereas at the call site (`SemanticsRule.MatchFuncRes`) they were expected to be named `return-0`,...,`return-k`.
* The unique identifier generated for a possibly overloaded subprogram was changed to be any fresh identifier, with a comment recommending that it reflects the original name of the subprogram.


HadrienRenaud (approved):
This is good, thanks.

Maybe the commit name/PR title can get more meaningful, suggesting to change it for example to: `[asl][doc] Fix name-mangling for return-identifiers and overridden functions`

hrutvik (approved):

PR #1829, opened 2026-05-08
[asl reference] removed unused entries and files

Roman-Manevich:


.gitignore
```diff
@@ -12,9 +12,6 @@ catIncludes.ml
 herdtools7.install
 jerd.byte
 jerd.js
-asllib/doc/ASLSemanticsLines.tex
```
hrutvik, line 15: Should we consider a separate `.gitignore` in `asllib`?

hrutvik (approved):

.gitignore
Roman-Manevich: Done.

PR #1830, opened 2026-05-08
[all] Clean-up utils following update to ocaml 4.14

fsestini:
Replacing various locally-defined utility functions with equivalents from the stdlib, following https://github.com/herd/herdtools7/pull/1819


HadrienRenaud (approved):
Looks good to me! Thanks

PR #1831, opened 2026-05-09
[aslspec] parameterized types

Roman-Manevich:
Extended aslspec types to allow specifying a single optional parameter.
For example
```
ast clistone[[x]] { "a non-empty comma-separated list" } =
  | x
  | (x, Tok_comma, clistone[[x]])
;
```
This is needed to support generic types needed to represent parse trees, for example, comma-separated lists of an arbitrary type.

I acknowledge that the syntax is a bit weird and requires math layout to space out brackets to avoid clashing.
Also, the type system has not been fully adapted so some legal specifications might be rejected, like the one added to this PR. This is postponed to a later PR.

### DIFF
--- a/asllib/aslspec/AST.ml
+++ b/asllib/aslspec/AST.ml
@@ -211,6 +211,10 @@ module Term = struct
     | Label of { loc : source_location; label : string }
         (** Either a set containing the single value named by the given string
             or a reference to a type with the given name. *)
+    | Parameter of { loc : source_location; name : string }
+        (** A named type parameter. The parser constructs an instance of
+            [Label], which is later substituted by an instance of [Parameter] in
+            [SubstituteTypeParameters]. *)
     | TypeOperator of {
         loc : source_location;
         op : type_operator;
@@ -218,6 +222,22 @@ module Term = struct
       }
         (** A set containing all types formed by applying the type operator [op]
             to the type given by [term]. *)
+    | ParamType of {
+        loc : source_location;
+        typename : string;
+        term : opt_named_type_term;
+      }
+        (** A set containing all types formed by instantiating the parameterized
+            type [typename] with the parameter [term]. For example, if we have
+            {[
+              typedef Tree[[T]] =
+              | Leaf(value: T)
+              | Node(left: Tree[[T]], right: Tree[[T]])
+
+              relation walk_tree(x: Tree[[Z]]) -> list0(Z) = ...
+            ]}
+            then {[ Tree[[Z]] ]} is represented by a [ParamType] with [typename] =
+            "Tree" and [term] represented by the type parameter [Z]. *)
     | Tuple of {
         loc : source_location;
         label_opt : string option;
@@ -263,7 +283,9 @@ module Term = struct
   let loc_of term =
     match term with
     | Label { loc }
+    | Parameter { loc }
     | TypeOperator { loc }
+    | ParamType { loc }
     | Tuple { loc }
     | Record { loc }
     | Function { loc }
@@ -277,6 +299,8 @@ module Term = struct
   (** [make_type_operation loc op term] constructs a type term in which [op] is
       applied to [term] at source location [loc]. *)
   let make_type_operation loc op term = TypeOperator { loc; op; term }
+
+  let make_param_type loc typename term = ParamType { loc; typename; term }
 
   (** [make_tuple loc args] constructs an unlabelled tuple with arguments [args]
       at source location [loc]. *)
@@ -544,6 +568,7 @@ module Type : sig
   type t = {
     loc : source_location;
     name : string;
+    param_opt : string option;
     type_kind : Term.type_kind;
     variants : TypeVariant.t list;
     att : Attributes.t;
@@ -551,10 +576,11 @@ module Type : sig
 
   val make :
     source_location ->
-    Term.type_kind ->
-    string ->
-    TypeVariant.t list ->
-    attribute_pairs ->
+    type_kind:Term.type_kind ->
+    name:string ->
+    param_opt:string option ->
+    variants:TypeVariant.t list ->
+    attribute_pairs:attribute_pairs ->
     t
 
   val attributes_to_list : t -> attribute_pairs
@@ -568,12 +594,13 @@ end = struct
   type t = {
     loc : source_location;
     name : string;
+    param_opt : string option;
     type_kind : Term.type_kind;
     variants : TypeVariant.t list;
     att : Attributes.t;
   }
 
-  let make loc type_kind name variants attribute_pairs =
+  let make loc ~type_kind ~name ~param_opt ~variants ~attribute_pairs =
     (* The [type_kind] of the variants is the [type_kind] given above. *)
     let variants_with_parent_type_kind =
       List.map
@@ -586,6 +613,7 @@ end = struct
       loc;
       type_kind;
       name;
+      param_opt;
       variants = variants_with_parent_type_kind;
       att = Attributes.of_list attribute_pairs;
     }
--- a/asllib/aslspec/PP.ml
+++ b/asllib/aslspec/PP.ml
@@ -54,10 +54,13 @@ let rec pp_type_term fmt =
   let open Term in
   function
   | Label { label } -> pp_print_string fmt label
+  | Parameter { name } -> pp_print_string fmt name
   | TypeOperator { op; term } ->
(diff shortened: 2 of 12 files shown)


asllib/aslspec/AST.ml
```diff
@@ -568,12 +583,13 @@ end = struct
...
     loc : source_location;
     name : string;
+    param_opt : string option;
     type_kind : Term.type_kind;
     variants : TypeVariant.t list;
     att : Attributes.t;
   }
 
-  let make loc type_kind name variants attribute_pairs =
+  let make loc type_kind name param_opt variants attribute_pairs =
```
hrutvik: nit: this is now quite a lot of arguments - is it worth naming them or similar?

asllib/aslspec/spec.ml
```diff
@@ -915,7 +924,13 @@ module Check = struct
             Error.undefined_reference missing_location id elem_name)
         ids_referenced_by_elem
     in
-    List.iter (check_no_undefined_ids_in_elem id_to_defining_node) ast
+    List.iter
+      (fun elem ->
+        let id_to_defining_node =
+          symbol_table_for_id id_to_defining_node (elem_name elem)
+        in
```
hrutvik: Sorry, not sure I follow this change - why has this become necessary?

asllib/aslspec/AST.ml
```diff
@@ -211,13 +211,22 @@ module Term = struct
...
       }
         (** A set containing all types formed by applying the type operator [op]
             to the type given by [term]. *)
+    | ParamType of {
+        loc : source_location;
+        typename : string;
+        term : opt_named_type_term;
+      }
+        (** A set containing all types formed by instantiating the parameterized
+            type [typename] with the parameter [term]. *)
```
hrutvik: In some ways this is an unexpanded type substitution it seems? It seems as if expanding the substitution makes the specification easier to type-check, but you want to keep the unexpanded version for rendering output. Is it worth exploring a way to have both co-exist?

asllib/aslspec/spec.ml
Roman-Manevich: I think it temporarily solved a bug, but after some code changes it became unnecessary and I forgot to remove. Removing it now. By the way, `symbol_table_for_id` will go away soon and relation parameters will be replaced as done for parameterized types, so the whole code will become simpler.

asllib/aslspec/AST.ml
Roman-Manevich: I'm not sure I follow this. An unexpanded type term can also exist in relation inputs unless by "output" you mean the PDF. (It can also be used to define type variants.)
I'm not sure what keeping both expanded and unexpanded means. In both cases, the constructor would contain a type term.

hrutvik (approved):

hrutvik:
Looks good to me, but please wait for @HadrienRenaud's review too

asllib/aslspec/AST.ml
```diff
@@ -211,13 +211,22 @@ module Term = struct
     | Label of { loc : source_location; label : string }
         (** Either a set containing the single value named by the given string
             or a reference to a type with the given name. *)
+    | Parameter of { loc : source_location; name : string }
+        (** A named type parameter. *)
```
HadrienRenaud: Can you add an example? ~~And if it's parsed like this or just a product of the type system?~~ Ah just read through, you need to say that it is constructed from `Label`s in the `SubstituteTypeParameters` module of `spec.ml`

asllib/aslspec/PP.ml
```diff
@@ -219,11 +222,16 @@ let type_kind_to_string =
...
+let pp_type_definition fmt
+    ({ Type.name; param_opt; type_kind; variants } as def) =
   let eq_str = if Utils.list_is_empty variants then "" else " = " in
-  fprintf fmt "%s %s@.%a@.%s@,@[<v>  %a@]@,;"
+  fprintf fmt "%s %s%a@.%a@.%s@,@[<v>  %a@]@,;"
     (type_kind_to_string type_kind)
-    name pp_attribute_key_values
+    name pp_type_parameter param_opt pp_attribute_key_values
     (Type.attributes_to_list def)
     eq_str pp_variants_with_attributes variants
```
HadrienRenaud: Up to you but at this point I would just open `Type` and use `def.name` or `def.variants` all the time

asllib/aslspec/spec.ml
```diff
@@ -593,7 +598,8 @@ module ResolveRules = struct
           let args = List.map arg_of args in
           Expr.make_opt_labelled_tuple (Term.loc_of term) label_opt args
       | ( None,
-          ( Term.Label _ | Term.Record _ | Term.TypeOperator _ | Term.Function _
+          ( Term.Label _ | Term.Parameter _ | Term.Record _
+          | Term.TypeOperator _ | Term.ParamType _ | Term.Function _
           | Term.ConstantsSet _ ) ) ->
```
HadrienRenaud: Does this work?
```suggestion
          Term.( Label _ | Parameter _ | Record _ | TypeOperator _
            | ParamType _ | Function _ | ConstantsSet _ )
        ) ->
```

asllib/aslspec/spec.ml
```diff
@@ -359,7 +361,10 @@ let symbol_table_for_id id_to_defining_node id =
   | Some (Node_Relation { Relation.parameters; loc }) ->
       List.fold_left
         (fun curr_table param ->
-          let type_for_param = Type.make loc TypeKind_Generic param [] [] in
+          let type_for_param =
+            Type.make loc ~type_kind:TypeKind_Generic ~name:param
+              ~param_opt:None ~variants:[] ~attribute_pairs:[]
```
HadrienRenaud, line 366: Maybe worth considering making the arguments `param_opt`, `variants` and `attribute_pairs` optional on `Type.make`?

asllib/aslspec/spec.ml
```diff
@@ -861,7 +869,9 @@ module Check = struct
     let open Term in
     function
     | Label { label } -> [ label ]
+    | Parameter _ -> []
```
HadrienRenaud, line 872: So a parameter doesn't reference its name? Maybe this can be justified in a comment

asllib/aslspec/AST.ml
HadrienRenaud: Could you add examples? For example I think it'd be good to say that those are parsed (with `term = None`), and the result of a substitution (with `term = Some t'`)

HadrienRenaud (commented):
Thank you @Roman-Manevich for this. I'm not going to block this merging but I'm wondering about the following:

> Also, the type system has not been fully adapted so some legal specifications might be rejected, like the one added to this PR. This is postponed to a later PR.

Is there a good reason not to add the type-system to support this? Why merge this right now and not wait until the type-system is ready to merge?

asllib/aslspec/spec.ml
Roman-Manevich: since `Type.make` is only ever used once, I think it's better to keep things explicit.

asllib/aslspec/spec.ml
Roman-Manevich: Because it's a variable. I'll add a comment.

asllib/aslspec/spec.ml
HadrienRenaud: ok

asllib/aslspec/PP.ml
Roman-Manevich: I'm not quite sure what you mean by "open `Type`". Do you mean an OCaml open at the top level? Somehwere else?

HadrienRenaud: Something like this:
```ocaml
let pp_type_definition fmt def =
  let open Type in
  let eq_str = if Utils.list_is_empty def.variants then "" else " = " in
  fprintf fmt "%s %s%a@.%a@.%s@,@[<v>  %a@]@,;"
    (type_kind_to_string def.type_kind)
    def.name pp_type_parameter def.param_opt pp_attribute_key_values
    (Type.attributes_to_list def)
    eq_str pp_variants_with_attributes def.variants
```

asllib/aslspec/AST.ml
Roman-Manevich: The optional here is just the name of the parameter term, which is just for documentation and isn't usually used. Substitution replaces a parameter variable by a given term.
I added an example.

PR #1833, opened 2026-05-12
[herd] Decouple analysis results from CLI output

fsestini:
This PR is a step toward making `herd7` reusable from OCaml code, as outlined in https://github.com/herd/herdtools7/issues/1782. It primarily addresses point 2 of that issue, namely:

> Gradually decouple outputs from stdout/files, so callers can consume results directly as OCaml values instead of scraping CLI output or temporary files.

The main change is to split the functionality of the `Top_herd` module into:

- a result-producing core, which returns structured OCaml values; and
- a CLI layer, which preserves the existing stdout/dot-file behaviour.

The user-facing `herd7` CLI is intended to remain unchanged.

## What Changed

Before this PR, `Top_herd` mixed three responsibilities in the same control flow:

- generating candidate executions and computing analysis results;
- formatting analysis results and selected executions for output; and
- managing CLI output resources such as stdout dot blocks, output directories, temporary dot files, and viewers.

This PR separates those responsibilities. `Top_herd.Make(...).run` is now the result-producing part: it generates executions, checks them against the model, and returns structured OCaml values rather than printing directly. `Top_herd.Printer` is the formatting layer: it knows how to turn those structured values into strings. The new `Cli` module implements the remaining CLI-specific resource management: directories, temporary files, channels, etc.

The intended split is that the first two pieces can be reused from library-style callers (hence why they are both accessed through `Top_herd`), while file and directory management stays in a CLI-specific place and won't be exposed.

One detail of this split is that selected executions are returned through an iterator-shaped value, rather than as a list. This is to preserve the previous memory behaviour where executions are processed one by one and not retained in memory.

Note for reviewers: most of this refactor consists of taking pieces of the old `Top_herd` module and moving them to more specific places. The diff might therefore look noisier than the actual conceptual change, as most of the code that is shown being added/removed in the diff is simply being moved around with little to no change. Overall, I tried my best to keep the diff as tight as possible.

Concretely, this PR proposes the following changes:

### New module `Top_herd.TestResult`

The main types introduced here are `stats` and `execution`. `stats` represents the overall results of a litmus test simulation, i.e. the final states, candidate counts, witnesses, flags, etc. `execution` carries data for a single execution.

Calling `Top_herd.Make(M).run` now takes a `test : M.S.test` as input, and returns a triple:

- A `M.S.test`, which is an updated version of the input
- A list `M.S.event_structure list` of event structures (pre solver)
- An iterator `(execution -> unit) -> stats` that iterates over executions and produces a `stats` summary at the end

The iterator follows the existing CLI selection policy: it visits the executions selected by options such as `show` and `nshow`, not every candidate execution. This keeps the new API close to the behaviour currently exposed by the CLI. A more fine-grained iterator over candidate executions can be added later if needed.

### Updated `ParseTest`/`RunTest`

The "run" functions in `ParseTest`/`RunTest` have been updated so parsing and running return structured outcomes, rather than print out those outcomes themselves.

Outcomes are returned as a first-class module. This seems needed, because the type of analysis outcomes is indexed by semantics `S : SemExtra.S`, and moreover the input test determines which `SemExtra.S`/`XXXMem.S` instance herd7 constructs. Since the exact semantics module is only known after parsing and dispatching on the test architecture and variants, it has to be part of the returned value somehow.

### New module `Cli`

The new `Cli` module contains the parts of the old `Top_herd` implementation that are specific to the CLI. Most of this code is intentionally copied almost directly from the old `Top_herd` module.

### Changed how `showsome` is computed within `ParseTest`

Before this refactor, `showsome` was determined inside `ParseTest` (in part) from the `outputdir` parameter. As `ParseTest` is now moving toward being a more generic API, I thought a very CLI-specific parameter like `outputdir` would not be appropriate. The new `collect_graph_data` parameter for `ParseTest` is an attempt to preserve the old behaviour using a less CLI-specific parameter. Having said this, I do still find it a bit awkward to use a module parameter for controlling optimization behaviour, so it would be great if we can later on find a way to remove this while preserving the optimizations as much as possible.

### Tests

Added cram tests under `herd/tests/other/output.t` to ensure that CLI behaviours touched by the refactor are preserved.


herd/parseTest.ml
```diff
@@ -119,58 +125,58 @@ module Top (TopConf:RunTest.Config) = struct
...
+  let from_string ~filename ~contents env =
+    do_from_string env ~filename ~contents
+
   let from_file name env =
-(* START NOTWWW *)
-(* Interval timer will be stopped just before output, see top_herd *)
-    Itimer.start name TopConf.timeout ;
-(* END NOTWWW *)
-    let start_time = Sys.time () in
-    Misc.input_protect (do_from_file start_time env name) name
```
TiberiuBucur, line 211: Is there any concern in dropping the timing data recording?

herd/parseTest.ml
```diff
@@ -96,10 +104,8 @@ module Top (TopConf:RunTest.Config) = struct
         let bell_model_info = Conf.bell_model_info
         let model = model
         let showsome =
-          begin match Conf.outputdir with
-          | PrettyConf.StdoutOutput | PrettyConf.Outputdir _ -> true
-          | _ -> false
-          end || Misc.is_some Conf.PC.view || Conf.variant Variant.MemTag
+          Conf.collect_graph_data
+              || Misc.is_some Conf.PC.view || Conf.variant Variant.MemTag
```
TiberiuBucur, line 152: I am personally fine with having this flag, but I was curious if you measured a change in performance with and without the showsome flag guard for the code that builds the `show` member of the state in `interpreter.ml`. How big of an optimisation are we looking at, compared to recording every relation and filtering at the end?

TiberiuBucur (commented):
I haven't gotten around to reviewing everything just yet, but so far I like the implementation. I left two minor comments so far. In addition, I'm having trouble running `make install` after the project builds (using `make`). Apparently the `_build/default/herdtools7.install` is missing.

herd/parseTest.ml
fsestini: Time recording is still performed as before, it just has been moved out of `ParseTest` and into the new `Cli` module.

The rationale is that I think timing is a property of a particular client’s consumption of the result, not of the result itself. Since `ParseTest` is now moving towards exposing a reusable library API whose results are consumed through an iterator in various different ways (in CLI, concurrently, in tests, etc.), the library itself should not impose any particular timing policy onto callers.
In herd7, the CLI the component that knows what operations need to be timed, so now timing is handled there. Other consumers of this API will have the choice to set up their own timing policy, or to not measure time at all (like we do in our test suite).

Having said this, you've reminded me that I should make the `time` parameter of `Top_herd.Printer.pp_stats` optional, in case callers want to print out run stats without the 'Time' bit!

herd/parseTest.ml
fsestini: Thanks, this is a good point. The short answer is that I have not looked deeply into this yet, but I agree it is a question worth asking.

> compared to recording every relation and filtering at the end?

I am interpreting this as: what would happen if the interpreter always constructed the `show` state, and we only decided later which parts of it to use.

One thing worth noting is that `show` is currently represented lazily via `Lazy.t`. So, even with `showsome = true`, we are not necessarily eagerly computing every shown relation while the interpreter runs. What we may still be doing is allocating thunks needed to compute the shown relations at a later stage. So my current understanding is that the likely cost of setting `showsome = true` unconditionally is less "eagerly compute every relation" and more "allocate and retain more show-related data, possibly increasing memory footprint and GC pressure".

I did a rough comparison on this test:

```
AArch64 wait-flag1
{
   0:X0=x; 1:X0=x;
   1:X2=y;
}
   MOV W1,#1   |L0:                ;
               | MOV W3,#1         ;
               | LDADD W3,WZR,[X2] ;
   STR W1,[X0] | LDR W1,[X0]       ;
               | CBZ W1,L0         ;
exists(1:X1=1)
```

with

```sh
$ /usr/bin/time -l ./_build/default/herd/herd.exe -set-libdir ./herd/libdir ./wait-flag1.litmus -unroll 3
```

For this particular test, I did not see a measurable difference in runtime or peak memory between the normal build and a build with `showsome =
true` unconditionally. On average:

```
Standard:

        5.74 real         3.52 user         1.20 sys
            17907712  maximum resident set size
            11616664  peak memory footprint

With showsome = true:

        5.71 real         3.53 user         1.19 sys
            17874944  maximum resident set size
            11600304  peak memory footprint
```

That said, this is not a definitive benchmark. I chose this test because it is heavier on herd7 than others, but nevertheless, it might not be stressing the `show` machinery particularly well. (@HadrienRenaud  I wonder if you might know of any litmus tests that could serve as a good benchmark for this.) I also tried `time make test` and did not see a visible performance difference there either.

Bottom line is, I don't feel I have enough evidence one way or the other, right now. My reason for keeping the flag is to keep the refactor conservative w.r.t. runtime behaviour, but I agree this is worth measuring more systematically at some point.

HadrienRenaud: Hi for a bigger test you can try say `MP+dmb.sy+addrR-pos-addr+W+PTE`, or `Marc03`, both in `catalogue/aarch64-VMSA/tests`. `make vmsa-test` might be a good aggregation score, but execute those tests with `-speedcheck true -optace true` which speeds up considerably the execution. Note that `Marc03` is not in `make vmsa-test`.

herd/parseTest.mli
```diff
@@ -17,7 +17,13 @@
 (** Run a test from source file, dispatch on tests architecture *)
 
 module Top :
-  functor (C : RunTest.Config) ->
+  functor (C : sig
+    include RunTest.Config
+    val collect_graph_data : bool
+    (** Whether callers intend to consume execution graph data.
+        Used as optimization. *)
```
HadrienRenaud, line 24: Could this be added to `RunTest.Config`? It would be cleaner here

fsestini:
Hi @maranget @HadrienRenaud , would you mind taking a look and letting me know if you have any comments? Thanks

herd/tests/other/output.t/run.t
```diff
@@ -0,0 +1,311 @@
...
+  1:X0=1; 1:X2=0;
+  1:X0=1; 1:X2=1;
+  Ok
+  Witnesses
+  Positive: 1 Negative: 3
+  Condition exists (1:X0=1 /\ 1:X2=0)
+  Observation MP Sometimes 1 3
+  Hash=211d5b298572012a0869d4ded6a40b7f
+  
+  show MP file
```
HadrienRenaud, line 304: I find it clearer with
```suggestion
show MP file
```

herd/tests/other/output.t/run.t
```diff
@@ -0,0 +1,311 @@
...
```
HadrienRenaud, line 166:
```suggestion
show MP file
```

herd/runTest.ml
```diff
@@ -36,51 +36,46 @@ module type Config = sig
...
   DirtyBit.t option ->
-  float (* start time *) ->
-  string (* file name *) ->
-  in_channel (* source channel *) ->
+  filename:string option (* file name *) ->
+  contents:string (* litmus test contents *) ->
   TestHash.env ->
   Splitter.result ->
-  TestHash.env
+  TestHash.env * (module Outcome) option
```
HadrienRenaud, line 53: I find modules as values non-intuitive and it's difficult to know if they have any performance implications (my previous use of them had).

Could we replace this by making the `runfun` type parametric in this `'output` type which would be instantiated by `Make` to `(M.S.test * Top_herd.TestResult.Make(M.S).t)`?
If this is too hard to handle in here, maybe a solution could be to make `TestResult` standalone or to move it in `SemExtra`?

herd/top_herd.mli
```diff
@@ -0,0 +1,124 @@
...
+end
+
+module type Config = sig
+  include CommonConfig
+  val byte : MachSize.sz
+  val dirty : DirtyBit.t option
+end
+
+module TestResult : sig
+  type 'stset stats
```
HadrienRenaud, line 38: Could you add a line documenting what is `'stset` supposed to stand for here?

HadrienRenaud (commented):
Hi Filippo, thanks for this good work, and apologies for the delay.

I have a few comments, the main one being that I'm a bit uncomfortable using TestOutput like this and maybe I'd be happier if it was included in `SemExtra`.

Could you please add to the extensive PR documentation (thanks) the results of a few performance tests? For example, that you don't find any difference in performance for `make test` between this branch and `master`. I'll approve once this is done.

herd/tests/other/output.t/run.t
fsestini: That line is part of the expected stdout output from herd7 that this test is checking, so it needs to be indented that way due to how dune cram tests work (non-indented lines are interpreted as comments and ignored).

HadrienRenaud: Ah sorry indeed I thought this was an explanation of the following command

herd/runTest.ml
fsestini: Generally I agree first-class modules are an advanced feature and should be used sparingly. I think what you are suggesting might work for `RunTest`, where `Make` is semantics-aware as it takes `M`/`S` as parameters. In other words, at the level of `RunTest.Make`, it is statically known which istance of `SemExtra.S` is being used.

The difficulty is with the functions exposed by `ParseTest`, which as things stand is the main entry point into herd. Those functions take a raw litmus test (in the form of string or file path) and return the result of simulating the test. At that level, there is no statically fixed instance of `SemExtra.S`, because the semantics used to execute each input test is dynamically constructed by `from_string/from_file` based on the input. However, we also want those functions to return structured test results, and those results necessarily depend on a specific instance of `SemExtra.S`.

I haven't found a solution to this that doesn't involve existentials in some form or another. Happy to explore alternatives, however, if you have any.

On the performance concern: `time make test` did not reveal any significant differences:

```
Pre-PR:  157.26s user 34.50s system 413% cpu 46.377 total
Post-PR: 155.97s user 33.64s system 413% cpu 45.840 total
```

I'm not aware of first-class modules being particularly unpredictable performance-wise. In this PR, they are only constructed once per test run, at the end of each run, so I wouldn't expect any noticeable impact. However I can run more specific benchmarks if you have any particular concern in mind.

HadrienRenaud: That is good enough a reason for me. Thanks.

And thanks for the performance regressions, I'm gonna approve the PR.

HadrienRenaud (approved):

herd/tests/other/output.t/MP-forall.litmus
```diff
@@ -0,0 +1,17 @@
+AArch64 MP-forall
+"PodWW Rfe PodRR Fre"
+Cycle=Rfe PodRR Fre PodWW
+Generator=diycross7 (version 7.54+01(dev))
+Prefetch=0:x=F,0:y=W,1:y=F,1:x=T
+Com=Rf Fr
+Orig=PodWW Rfe PodRR Fre
```
relokin: This is not generated with diyone is it? Should we remove the metadata?

herd/tests/other/output.t/MP-notexists.litmus
```diff
@@ -0,0 +1,17 @@
+AArch64 MP-notexists
+"PodWW Rfe PodRR Fre"
+Cycle=Rfe PodRR Fre PodWW
+Generator=diycross7 (version 7.54+01(dev))
+Prefetch=0:x=F,0:y=W,1:y=F,1:x=T
+Com=Rf Fr
+Orig=PodWW Rfe PodRR Fre
```
relokin: ditto

herd/tests/other/output.t/MP.litmus
```diff
@@ -0,0 +1,18 @@
+AArch64 MP
+"PodWW Rfe PodRR Fre"
+Cycle=Rfe PodRR Fre PodWW
+Generator=diycross7 (version 7.54+01(dev))
+Prefetch=0:x=F,0:y=W,1:y=F,1:x=T
+Com=Rf Fr
+Orig=PodWW Rfe PodRR Fre
```
relokin: ditto

herd/tests/other/output.t/run.t
```diff
@@ -0,0 +1,311 @@
+Check that duplicate tests with the same hash are only reported once.
+
+  $ herd7 -set-libdir ../libdir MP.litmus MP.litmus | grep '^Test '
+  Test MP Allowed
+
+Check correct handling of outputdir, stdout, and suffix.
+
+  $ rm -rf out-graph
```
relokin: Does it make sense to place that in the temp directory instead? Starting off with an `rm -rf` and assuming that out-graph is not used is not too much but also using a temp directory is not a big change if you use mktemp.

herd/cli.ml
```diff
@@ -0,0 +1,195 @@
...
+    close_dot ochan ;
+    let do_show () =
+(* Show if something to show *)
+      begin match ochan with
+      | Some (_,fname) when shown > 0 ->
+          let module SH = Show.Make(S.O.PC) in
+          if O.PC.debug then Printf.eprintf "show %s file\n" fname ;
+          SH.show_file fname
+      | Some _|None -> ()
+      end ;
```
relokin, line 136: This is not in the scope of this change, but I think the logic that shows the pdf and removes the temp dot file is not great and doesn't work on MacOS.

herd/parseTest.mli
fsestini: I’m not sure `RunTest.Config` is cleaner semantically. It would work syntactically, but `collect_graph_data` is not used by `RunTest.Make`. It is only used in `ParseTest.Top` to compute `showsome`.
Putting it in `RunTest.Config` would make every config that includes `RunTest.Config` transitively carry a field that `RunTest` itself doesn't consume, which feels a bit leaky to me.

If you prefer, we could avoid the inline signature by introducing a small `ParseTest.Config` signature.

herd/tests/other/output.t/run.t
fsestini: Ah cool, no worries!

herd/cli.ml
```diff
@@ -0,0 +1,195 @@
...
+    | Restrict.CondOne when TR.positive c <> count_prop () -> do_show ()
+    | _ ->
+(* Header *)
+      if not O.badexecs && TR.has_bad_execs ~badflag:O.badflag c then ()
+      else
+(* Stop interval timer *)
+        Itimer.stop O.timeout ;
+(* Now output *)
+        let time = Sys.time () -. start_time in
+        Format.printf "%a\n" (fun fmt () -> PP.pp_stats ~time test c fmt) ();
```
relokin, line 157: Again not in scope for this PR but I am a bit surprised by the control flow here. What happens to pp_stats and other print statements for other cases of restrict?

herd/top_herd.mli
fsestini: Yep, thanks, hopefully https://github.com/herd/herdtools7/pull/1833/commits/b4e05942a5a219f9620ef8a7d4938611243ea5dc does the trick

herd/top_herd.ml
```diff
@@ -43,6 +36,228 @@ module type Config = sig
...
+            name
+            pp_model
+        else
+          sprintf "Test %s%s%s" name
+            (match pp_model with
+            | "" -> ""
+            | _ -> sprintf ", %s" pp_model)
+            pp_flag
+      end
+    in
```
relokin: is it very hard to move this into dump_legend?

relokin (commented):
Thanks for putting this together. I like the direction of this change.
I've left some comments, some of which might not be in the scope of this PR so please feel free to ignore.

maranget:
Hi @fsestini. I am just starting to look at the PR.  I noticed that `make -C herd-www` fails. Is the fix simple?

fsestini:
> Hi @fsestini. I am just starting to look at the PR. I noticed that `make -C herd-www` fails. Is the fix simple?

Hi @maranget , thanks for noticing. Indeed, herd-www needs a small patch as a consequence of this. I was planning on it but then somehow completely forgot about it. Let me take a look, it should be quick.

herd/tests/other/output.t/run.t
fsestini: Fair. AFAIU dune cram tests are run in a sandboxed location, so that `rm -rf` _should_ carry limited risk, however such behaviour is not immediately obvious to the reader, and bugs in dune could always happen, so I agree. I'll look into an alternative.

fsestini:
> > Hi @fsestini. I am just starting to look at the PR. I noticed that `make -C herd-www` fails. Is the fix simple?
> 
> Hi @maranget , thanks for noticing. Indeed, herd-www needs a small patch as a consequence of this. I was planning on it but then somehow completely forgot about it. Let me take a look, it should be quick.

This should now be covered by the last few commits.

In summary, I've added a web-specific module `WebRun` for `herd-www` which simply invokes the new library-style `ParseTest` API and prints it out in the form expected by the existing JS code.

For now, `WebRun` deliberately preserves the stdout-based protocol that was already in use by `jerd`: test run stats and DOT graph output are printed to stdout, then `jerd_driver.js` intercepts them, parses them, and displays them. This keeps the web app working with minimal change. In a follow-up PR, we should revisit `jerd.ml` and `jerd_driver.js` to handle the test results received from herd as first-class values, instead of having the OCaml side print a stream that JavaScript then parses back into structured data.

Note that `WebRun` duplicates a few lines of the branching logic around `O.restrict` to match `cli.ml`. This is intentional for now. I do not fully understand what that branch means, at least not well enough to be able to refactor it confidently, so I would rather leave the web and CLI behaviours explicitly in sync for now.

This refactoring revealed a couple of configuration bindings in `jerd.ml` that were declared but unused. I've removed them both, but do shout out if you don't agree:

- I've removed the `suffix` config, as it was unused there and it doesn't seem relevant to the web app anyway (since `suffix` is used by the herd CLI on DOT files saved to disk, but we don't do that in the web app).

- I've also removed the `timeout` config, as this config was _not_ wired into the web app logic even pre-PR, and the logic to install timeout handlers was actually explicitly excluded from web builds via the `notwww.awk` script. We could revisit this in a follow-up PR if we decide we want timeout support in the web application.

herd/tests/other/output.t/run.t
fsestini: Update: that `out-graph` directory is only used in one place in the whole `run.t`, and each execution of `run.t` takes place in its own sandbox, so that `rm` command is effectively a no-op and isn't actually needed. I've removed it.

relokin: Thanks for checking Filippo!

herd/top_herd.ml
fsestini: Addressed by ef575db36501ed1b2c106101ed0c712ec7a0f0da

herd/cli.ml
fsestini: Completely agree. Will open a ticket and plan a follow-up PR after this is merged.

herd/top_herd.ml
```diff
@@ -233,8 +333,10 @@ module Make(O:Config)(M:XXXMem.S) =
       S.E.EventSet.subset loads obs
 
 (* Called by model simulator in case of success *)
-    let model_kont solver ochan test do_restrict cstr =
+    let model_kont solver f test do_restrict =
```
maranget: A slightly mode explicit name for `f`:
```suggestion
    let model_kont solver do_show test do_restrict =
```

herd/top_herd.ml
```diff
@@ -278,52 +380,22 @@ module Make(O:Config)(M:XXXMem.S) =
...
+              let exec =
+                {
+                  TestResult.concrete = conc;
+                  passes_check = ok;
+                  flags;
+                  sets = set_pp;
+                  rels = vbpp;
+                }
+              in
+              f exec
```
maranget: See above
```suggestion
              do_show exec
```

herd/cli.ml
```diff
@@ -0,0 +1,187 @@
...
+          begin try
+            let f,chan = Filename.open_temp_file "herd" ".dot" in
+            Some (chan,f)
+          with  Sys_error msg ->
+            Warn.warn_always "Cannot create temporary file: %s" msg ;
+            None
+          end
+         | None -> None
+       end
+    | PrettyConf.StdoutOutput ->
```
maranget, line 51: I am not sure we still need this case except, perhaps, for debugging the web interface?

herd/cli.ml
fsestini: My understanding is that `-restrict` will cause the printing of final stats to be suppressed on those branches. As you noted, this is the case pre-PR as well as post-PR. I've added some cases to the cram test to illustrate some of the behaviours exhibited by various values of `-restrict`.

From the code, it looks like `-restrict observed` is the only variant that actually filters candidate executions during simulation. Then, if the filtering results in 0 candidate executions, the final stdout stats are suppressed. I find this a bit surprising, since with `-restrict none` herd will print the normal stats block even when there are 0 executions.

The other variants `-restrict nonambiguous` and `-restrict condone` do not seem to filter candidate executions at all. Instead, they just suppress the final stdout stats if their characterising condition is not satisfied.

I agree this may be worth revisiting in a follow-up PR. I do not have the full historical context for these options, so I do not want to claim the current behaviour is incorrect. But from a UX point of view, I do find it somewhat non-obvious that an option named `-restrict` sometimes filters executions and sometimes only suppresses output. It might be worth splitting these functionalities into different CLI options that make this separation of concern more explicit. On the stats printing, I think it might be clearer to always print stats, and perhaps print additional diagnostic lines such as "executions are ambiguous" or "no observed executions" when the relevant conditions are met.

herd-www/webRun.ml
```diff
@@ -0,0 +1,85 @@
...
+        within_dot fname (fun () ->
+          let dump_graph exec = Printer.dump_exec_graph M.model test exec stdout in
+          result.TR.exec_iter dump_graph)
+      in
+      let nfinals = M.S.A.StateSet.cardinal (TR.states c) in
+      let module TRS = TR.Make (M.S) in
+      match O.restrict with
+      | Restrict.Observed when TR.candidates c = 0 -> ()
+      | Restrict.NonAmbiguous when TR.candidates c <> nfinals -> ()
+      | Restrict.CondOne when TR.positive c <> TRS.count_prop ~byte:O.byte test c -> ()
```
maranget: Those cases are quite obscure: the general idea is not to output anything when some conditions are not met. For instance with `-restrict nonambiguous` when different executions yield the same final state. I guess the web interface can ignore this subtlety.

herd/cli.ml
fsestini: Ah fair. Happy to remove this, though maybe a good candidate for a follow-up PR.

herd/top_herd.ml
fsestini: Good point, I've improved the name with f57109eb0030f3f6e49c95dc4e896e2f15c99e83
I've used `emit_exec` rather than the suggested `do_show` as technically this code provides to callers a way to iterate over executions but it's up to those callers whether they show them or not.

herd/cli.ml
maranget: Agreed.

maranget:
I am impressed by this work.  I am in favor of merging it ASAP.

herd-www/webRun.ml
fsestini: I followed your suggestion and removed the logic that suppresses stdout in herd-www with https://github.com/herd/herdtools7/pull/1833/commits/a49be0b2efd632371f13f41579d927dc4d96a74b, which in any case preserves the ability to filter executions via `-restrict observed`. Normal herd7 CLI behaviour is unaffected.

fsestini:
Thanks for the reviews @maranget @HadrienRenaud . Once I get an ACK from @relokin that all his comments have been addressed, I'll clean up the history and merge.

herd/top_herd.ml
relokin: that's great, thanks Filippo!

herd/cli.ml
relokin: sounds good to me.

relokin (approved):
Thanks for all the changes Filippo!

fsestini:
Merged, thanks @relokin and everybody!

PR #1834, opened 2026-05-12
AARCH-25405: Add href to Effective value

TiberiuBucur:
Builds on top of [this PR](https://github.com/herd/herdtools7/pull/1788) and adds links to the section in the Arm ARM describing the "Effective Value", and the `CTR_EL0` register, which contains the `DIC` and `IDC` bits. This is for better clarity and accessibility when navigating the Arm ARM.


TiberiuBucur:
Note: Future changes might be needed to support similar links for other terms in the English prose. In case it's needed, we can discuss a potentially more robust strategy at that moment

ShaleXIONG (approved):

PR #1835, opened 2026-05-14
[gen] add `@before` and `@after` predicates for relaxations in `diy7`

ShaleXIONG:
Add `diy7`-only predicate syntax for relax inputs:
  - `@before(...)`
  - `@after(...)`

The parser entry points are split so predicate syntax is accepted only by `diy7`. The shared `main`/`diyone7`, `diycross7`, and `cumul` parsers reject predicate syntax.

During relax expansion, invalid predicate placement is filtered before search:
  - `@before(...)` is only valid at the start of a relaxation sequence
  - `@after(...)` is only valid at the end of a relaxation sequence
  - predicates in the middle of a concrete composition are discarded

During `diy7` cycle search, boundary predicates are checked against neighbouring concrete edges. Once checked, predicates are stripped before passing candidate cycles to the normal test generator.

Added tests for:
  - `diy7` parser acceptance of `@before(...)` and `@after(...)`, while other parsers reject of predicate syntax
  - filtering invalid predicate placement to `[]`
  - In `diy7` cycle generation, the predicates `@before`and `@after` merge as expected
  
# Example
```
diy7 -arch AArch64 -cycleonly true -size 4 -relax '[@before(Po) PodRW Rfe @after(Po)]' -safe Po
```
This constrains the generated boundary edges to match the predicates. `Po` will match `@before(Po)` and `after(Po)` during the searching.
```
Generator produced 2 tests
Relaxations tested: {[@before(PosRR),PodRW,Rfe,@after(PosRR)]} {[@before(PodRR),PodRW,Rfe,@after(PodRR)]}
LB000: PosRR PodRW Rfe PosRR PodRW Rfe
LB001: PodRR PodRW Rfe PodRR PodRW Rfe
```
Another example contains composite relax:
```
diy7 -arch AArch64 -cycleonly true -size 4 -relax '[PodRW Rfe @after([PodRW Rfe])]'
```
In this case, the only relaxation will be match with themselves, i.e., the leading `PodRW Rfe` matches the trailing `@after([PodRW Rfe])`.
```
Generator produced 3 tests
Relaxations tested: {[PodRW,Rfe,@after(PodRW),@after(Rfe)]}
LB000: PodRW Rfe PodRW Rfe
3.LB000: PodRW Rfe PodRW Rfe PodRW Rfe
4.LB000: PodRW Rfe PodRW Rfe PodRW Rfe PodRW Rfe
```

### DIFF
--- a/gen/alt.ml
+++ b/gen/alt.ml
@@ -26,9 +26,9 @@ module type AltConfig = sig
   val max_relax : int
   val min_relax : int
   val choice : check
-  type relax
-  val prefix : relax list list
+  val prefix : string list
   val variant : Variant_gen.t -> bool
+  val varatom : string list
   type fence
   val cumul : fence list Config.cumul
   val wildcard : bool
@@ -270,10 +270,20 @@ struct
 end
 
 module Make(C:Builder.S)
-    (O:AltConfig with type relax = C.R.relax and type fence = C.A.fence) :
+    (O:AltConfig with type fence = C.A.fence) :
     sig
-      val gen : ?relax:C.R.relax list -> ?safe:C.R.relax list -> ?reject:C.R.relax list -> int -> unit
-      val filter_check: relax:C.R.relax list -> safe:C.R.relax list -> C.E.edge list -> C.E.edge list -> bool
+      type predicate_relax
+
+      val plain : predicate_relax -> C.R.relax
+      val lift : C.R.relax -> predicate_relax
+      val gen : ?relax:predicate_relax list -> ?safe:predicate_relax list -> ?reject:predicate_relax list -> int -> unit
+      val parse_argument : string -> predicate_relax list
+      val parse_input :
+        relax:string list -> safe:string list ->
+        reject:string list -> predicate_relax list * predicate_relax list * predicate_relax list
+      val remove_invalid_relaxes : predicate_relax list -> predicate_relax list
+      val pp_predicate_relax_list : predicate_relax list -> string
+      val filter_check: safe:predicate_relax list -> predicate_relax -> predicate_relax -> bool
     end
 
     =
@@ -284,6 +294,27 @@ module Make(C:Builder.S)
     open C.E
     open C.R
 
+    type edge_predicate =
+      | Before
+      | After
+
+    type predicate_edge = {
+      plain : C.E.edge;
+      pred : edge_predicate option;
+    }
+
+    type predicate_relax = predicate_edge list
+
+    let parse_predicate = function
+      | "before" -> Before
+      | "after" -> After
+      | s -> Warn.user_error "predicate %s is not supported." s
+
+    let predicate_edge plain = { plain; pred = None }
+
+    let plain relax = List.map (fun e -> e.plain) relax
+    let with_plain relax = plain relax, relax
+
     let dbg = false
 
     let is_int e = match get_ie e with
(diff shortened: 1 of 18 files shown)


gen/alt.ml
fsestini, line 1: This file is difficult to review in its current form because the predicate-related change is mixed with quite a few changes that seem unrelated at first sight.

For example:

- Several existing local names are renamed even when their role has not changed, for example the `call_rec_base` parameter `f0` is renamed to `test_generator`.
- Many comments explaining almost every individual line, to an excessive degree in my view.
- Some existing helpers seem to have been renamed and rewritten rather than minimally adapted. For example, the old `c_minprocs_es`, `c_minprocs_suff`, `c_minint_es`, and `c_minint` helpers are replaced by `procedure_count`, `max_edges_in_procedure`, etc. I completely see the value in refreshing the names and implementation of some of these functions, however I don't think this PR is the right time for that.
- Some of the changes look potentially semantic, not merely cosmetic. For example, inside `call_rec_base`, the old code used `let n = n - sz r`, whereas the new code uses `let n = n - 1`. It's not immediately clear to me whether this is a simple refactor or an important change in the logic.

Could you please rework the changes in this file so that the diff is as small and targeted as possible? In case some of these seemingly cosmetic changes are actually necessary for the functionality of the PR, could you please elaborate on those.

Concretely, I would prefer if we can keep these to a minimum:

- renaming of existing functions or parameters
- whitespace/indentation changes
- added/removed `begin`/`end`
- line-by-line comments unless there's some particular complex sequence that needs extra discussion

gen/alt.ml
```diff
@@ -243,10 +243,12 @@ struct
        - whether the boundary edges are compatible via `Edge.can_precede`
        - whether the mode-specific rule holds *)
     let can_precede safes po_safe xs ys =
-      let e1 = last_non_insert xs in
-      let e2 = hd_non_insert ys in
-      C.E.can_precede e1 e2
-      && match e1.edge,e2.edge with
+      if xs = [] || ys = [] then false
```
fsestini: Could you explain why these new empty-list checks are needed now?

gen/alt.ml
```diff
@@ -307,48 +306,123 @@ module Make(C:Builder.S)
...
-        match e.C.E.edge,get_ie e with
-        | (Back _|Leave _),_
-        | _,Int -> c
-        | _,Ext -> c + 1
-        | _,UnspecCom -> assert false
-      ) c
-
-    let c_minprocs_suff c =
-      List.fold_left ( fun c (_,es) -> c_minprocs_es c es) c
+    (* Count generated procedures from external communication edges.
```
fsestini: What is a "procedure" in this context?

gen/common/edge.ml
```diff
@@ -72,7 +76,7 @@ module type S = sig
   val compute_rmw : RMW.rmw -> value -> value -> value
   val is_valid_rmw : RMW.rmw list -> bool
 
-  type edge = { edge: tedge;  a1:atom option; a2: atom option; }
+  type edge = { edge: tedge;  a1:atom option; a2: atom option; pred : edge_predicate option }
```
fsestini: I am not completely convinced adding this `pred : edge_predicate option` to the common `edge` type is the right abstraction, though admittedly I don't have a strong alternative to propose at the moment. I can see the implementation reason for this choice, since the field is used in `alt.ml`. However, from the perspective of someone reading the declaration of `edge` in `edge.ml`, I think this `pred` field might be slightly confusing.

First of all, it's not immediately clear what it means for an edge to have a predicate field set. What does that field indicate? How is an edge "with a predicate" different from one "without"?

Secondly, AFAIU these predicates are supposed to be a diy7-specific mechanism. However, here predicates are being added to the common `edge` type (used by all `gen/` tools) as if they were an intrinsic property of edge in general. Indeed if we look at how the `pred` field is used in `edge.ml`, it is pretty much always set `pred = None` or ignored. I'm concerned that we might be leaking `diy7`-specific concepts into the common `edge` abstraction.

Curious if you have thought about using a different, diy7-specific predicate-aware relaxation type for this, instead of the common `edge` type? (That would mean the cycle search code would have to be adapted to operate on that representation instead.)

gen/common/edge.ml
```diff
@@ -451,6 +473,9 @@ let fold_tedges f r =
 
   let get_access_atom = A.get_access_atom
 
+  let equal_edge_atoms lhs rhs =
+    lhs.edge = rhs.edge && lhs.a1 = rhs.a1 && lhs.a2 = rhs.a2
```
fsestini: Let's not use the builtin polymorphic equality please. Let's use dedicated `equal` predicates or similar.

gen/diy.ml
```diff
@@ -34,14 +34,14 @@ open C.E
...
-    String.trim input |> parse_ast Parser.main_top_level_choice
+    String.trim input |> parse_ast Parser.diy7
 
   (* Parse the `-cumul` argument, which should be a list of individual fences. *)
   let parse_fences input_fences =
     let ast = String.trim input_fences |> parse_ast Parser.cumul in
     (* Note that `parse_fence` might fail *)
     let fences = Ast.bind ast ( fun input -> Ast.One(C.E.parse_fence input) )
-    |> Ast.expand in
+    |> Ast.expand ( fun _ -> Warn.user_error "cumul input: %s contains predicate." input_fences ) in
```
fsestini: I find `Warn.user_error` here a bit misleading. This `ast` is parsed with `Parser.cumul`, which grammar does not accept predicates. So if the user writes a predicate in the `-cumul` input, parsing should already fail before `Ast.expand` is reached. In other words, this `fun _ -> Warn.user_error ...` function should be unreachable under the current parser implementation, and it only exists only because `Ast.expand` requires it.

Could we use something that makes this aspect clearer, such as `failwith`/`assert false`/`Warn.fatal`, rather than `Warn.user_error`? A `user_error` here implicitly suggests that this is part of normal input validation and that this line _could_ be reached under bad user input, however that is not the case.

gen/relax.ml
```diff
@@ -342,11 +342,30 @@ and type edge = E.edge
                     for_all_adjacent_concrete_edge predicate (rhs :: list)
                 | false, false ->
                     for_all_adjacent_concrete_edge predicate list in
+          let leading_before_trailing_after_predicate list =
```
fsestini: Could you clarify what this does?

fsestini (commented):
Thanks @ShaleXIONG . I had a first look at the PR. I'm still trying to digest it, so most of the comments I left are questions/clarifications. In particular, I think the changes in `alt.ml` are a bit hard to review as they are, and there should be things we can do to make the diff more readable. See the comment on that file for more details.

gen/common/edge.ml
ShaleXIONG: For the question of meaning of predicate, there is no unifying meaning of predicates in general. We currently have `before` and `after`. We might add more based on need/requirements. They are ad-hoc extension of the expressibility and directly bind to code level for sure. I am trying to avoid introducing a full language here. Also I will prefer if something can be done without predicate.

For the changing of `type edge`, I do not have good solution neither tbh, as the `diy7` searching algorithm is based on `type edge`. I made the minimum change to the code here. Alternative idea might be keep the predicate in the `Ast` level, however this mean we need to rework the entire `alt.ml`, i.e., searching algorithm to take an `Ast` as input rather than `edge list list`.

gen/common/edge.ml
ShaleXIONG: I have pipe through several `equal_*` function in different places so to remove this polymorphic equality check. Will keep this extra change in a separate commit in the history.

gen/alt.ml
ShaleXIONG: Let me separate some semantics equivalent change from the history as I did this a while ago to .

To answer some of your questions immediately:
-`c_minprocs_es`,`c_minprocs_suff`, `c_minint_es` and `c_minint` are semantic equivalent change. As you suggested here, it is for better naming and also I try to rewrite with standard library function such as `fold_left`.
- `n = n - sz r` is a bug fix or clarification. The search algorithm will stop after certain `size` default `6`. An input relax, including composite relax such as `[PosWR Fre]` counts as one. However, if the input only contain  annotations such as `A`, it does not count. This is unnecessary behaviour and also there is no use case we explicitly do not want to count individual annotation `A`.

gen/alt.ml
ShaleXIONG: It is the `P0`, `P1` and so on in the generated tests. The function counts the external communication edges which maps to the numbers of threads/procedures.

fsestini: I think the comments should then refer to "processor" rather than "procedure" in that case.

gen/alt.ml
fsestini: That makes sense. I think these two are changes worth making, however I'd say they aren't very relevant to the predicate feature being introduced here. Their concerns are also different enough to warrant their own dedicated PR.

I would therefore suggest submitting these two changes (semantics-preserving refactoring and `n = n - sz r` bug fix) as _two_ distinct PRs _separate_ from the current one. Same treatment can be given to other changes in `alt.ml` that are not related to predicates.

gen/alt.ml
ShaleXIONG: I was over cautious here. The potentially problem is that `last_non_insert` and `hd_non_insert` trigger `assert false` if `xs` or `ys` is empty.

Let rewrite the history on this file and separate out some semantics equivalent change into separate PR here.

gen/relax.ml
ShaleXIONG: It check: `before` predicates must form a leading prefix and `after` predicates must form a trailing suffix. I added comment above this function now.

gen/diy.ml
ShaleXIONG: It is my mistake. I will change it to `Warn.fatal`

gen/alt.ml
ShaleXIONG: Ah you are right. This change will likely go to separate PR but I will ensure it reflects in the new  PR.

gen/common/edge.ml
fsestini: > directly bind to code level

Not sure I follow. Can you clarify what you mean by this?

> Alternative idea might be keep the predicate in the `Ast` level, however this mean we need to rework the entire `alt.ml`, i.e., searching algorithm to take an `Ast` as input rather than `edge list list`.

That would be a slight improvement, but you'd still be tying the search algorithm to a concrete data structure, one that just happens to be different from `edge list list`. As you correctly noted, this dependence of the search algo on a specific concrete data type is precisely what makes it non-modular and harder to extend, something that has affected the choices taken in this PR. So perhaps we could instead focus on making `alt.ml` more modular.

Could you have a look and estimate how hard it would be to refactor `alt.ml` into a functor, so that the cycle search algorithm is expressed parametrically w.r.t. an abstract type and abstract operations that are passed as a parameter? Something along the lines of how [this](https://github.com/herd/herdtools7/blob/926588440df84f3ba9b6f690d5afef4513cee1ef/gen/cat2config/normalization.ml#L86) normalisation algo operates on abstract types passed as functor parameters.

ShaleXIONG: Yes you are right the searching algorithm will still largely base on the concrete type of `edge` (If assume one particular architecture backend). Yet we can do something local to `alt.ml` such as the entire input will be a `Ast.t`, then internal we can have some data type such as `type edge_pred = Edge of edge | Predicate of pred*edge`. The current `Alt` is actually a functor, taking an argument `C:Builder.S` which corresponds to different architecture backend. The may interface is `gen` which take `relax list` aka equivalent `edge list list` as the inputs.

I will have a look of possible refactoring here further. Given how intertwine of all the module and functor in `diy*`, it will be a sizeable re-organisation of code. Yet I do not see any blocker and we possibly want to do in a separate PR.

fsestini: > The current `Alt` is actually a functor, taking an argument `C:Builder.S` which corresponds to different architecture backend.

You're right. What I'm suggesting is to functorise the search algorithm part so that the cycle data structure is provided as a parameter. Perhaps this entails extending an existing functor.

> I will have a look of possible refactoring here further. Given how intertwine of all the module and functor in `diy*`, it will be a sizeable re-organisation of code. Yet I do not see any blocker and we possibly want to do in a separate PR.

Thanks! Also, happy to discuss this further on a dedicated GH issue.

gen/alt.ml
ShaleXIONG: I have rewrite the history which should be much simpler now. Will open separate PRs later.

gen/common/AArch64Arch_gen.ml
```diff
@@ -949,6 +933,13 @@ type csel = OkCsel|NoCsel
 
 type dp = D.dp * csel
 
+let equal_csel c1 c2 = match c1,c2 with
+| OkCsel,OkCsel
+| NoCsel,NoCsel -> true
+| (OkCsel|NoCsel),_ -> false
+
+let equal_dp (d1,c1) (d2,c2) = D.equal_dp d1 d2 && equal_csel c1 c2
+
```
fsestini, line 959: Hi @ShaleXIONG , could you please open a separate PR for all these new `equal_*` predicates? Not that they are much hard to review, but they do create extra noise in the diff, and they aren't directly tied to predicates anyway.

ShaleXIONG: Certainly! I create #1913, once it is merged I will rebased here so the history is clear.

PR #1837, opened 2026-05-14
[aslspec] strengthened subsumption test for recursive types

Roman-Manevich:
The existing subsumption test was very conservative when it came to nested terms of recursive types, allowing to expand the types at most ones. The new test allows expanding types based on the depth of the input terms.

### DIFF
--- a/asllib/aslspec/spec.ml
+++ b/asllib/aslspec/spec.ml
@@ -1165,12 +1165,47 @@ module Check = struct
           true
       | _ -> false
 
-    (** [subsumed id_to_defining_node expanded_types sub super] conservatively
-        tests whether all values in the domain of [sub] are also in the domain
-        of [super]. Labels that represent types may be expanded, that is,
-        replaced by their list of type variants, using [id_to_defining_node]. To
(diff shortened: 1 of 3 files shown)


asllib/aslspec/spec.ml
```diff
@@ -1165,12 +1165,47 @@ module Check = struct
...
+      match StringMap.find_opt typename expanded_types with
+      | Some count -> count
+      | None -> 0
+
+    (** [increment_expansion_count typename expanded_types] increments the
+        expansion count for [typename] in [expanded_types]. *)
+    let increment_expansion_count typename expanded_types =
+      StringMap.add typename
+        (expansion_count typename expanded_types + 1)
+        expanded_types
```
HadrienRenaud:
```suggestion
      StringMap.update typename
        (function Some count -> Some (count + 1) | None -> Some 1)
        expanded_types
```

HadrienRenaud (approved):
LGTM, I'm not really grasping the theoretical consequences of this but if the only problem was removing infinite recursion in the type-checker, this sounds like a good way

PR #1839, opened 2026-05-16
[aslspec] improvements to layouts: checks, tests, and documentation

Roman-Manevich:
Added missing checks to layout (`math_layout` attributes) of expressions and rules.
Also, fixed the checks to type variants.
Added positive and negative tests.
Add documentation.

This PR was assisted by AI.

### DIFF
--- a/asllib/aslspec/aslspec.md
+++ b/asllib/aslspec/aslspec.md
@@ -5,9 +5,9 @@ to be included in the ASL Reference.
 aslspec allows specifying the following kinds of elements:
 - Types
 - Constants
-- Relation signatures
+- Relations with or without inference rules
+- Operators
 - Rendering elements
-- Inference rules
 
 # Type definitions
 Types definitions define mathematical types (like Booleans) and AST node types.
@@ -110,7 +110,7 @@ constant empty_tenv {
 # Relation definitions
 Relations define the type, or *signature* of a relation symbol.
 
-FOr example:
+For example:
 ```
 relation annotate_expr(tenv: static_envs, e: expr) -> (t: ty, new_e: expr, ses: powerset(TSideEffect)) | type_error
 {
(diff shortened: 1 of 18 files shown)


asllib/aslspec/aslspec.md
```diff
@@ -143,22 +297,215 @@ They also allow rendering types with just a subset of their variants.
...
 
 Type renders do not support any attributes.
 
+# Inference rule rendering definitions
+Rule rendering definitions select which inference rules are emitted as LaTeX
+macros. The simplest form renders all expanded rules for a relation under the
+same name:
+
+```
+render rule choose;
```
hrutvik: nit: `choose` as an example name is perhaps confusing (e.g. could be confused with choosing which relations to render)

asllib/aslspec/aslspec.md
```diff
@@ -143,22 +297,215 @@ They also allow rendering types with just a subset of their variants.
...
+
+## Layouts for type terms
+
+For type terms, `_` is always valid. Any more specific layout follows the type
+constructor:
+
+- Atomic type terms such as `Num` or `Bool` accept only the default layouts `_`,
+  `(_)`, and `[_]`.
+- Tuple and labelled tuple terms use one layout cell per component:
+  `Pair(Num, (Bool, Num))` can use `(_, (_, _))`. The shorthand `(_)` or `[_]`
```
hrutvik, line 377: Does the shorthand `[_]` mean that the `(Bool, Num)` is nested vertically internally too? I.e. is it equivalent to `[_, [_,_]]`? Perhaps this can be clarified once and for all above where the shorthands are introduced

asllib/aslspec/spec.ml
```diff
@@ -804,65 +804,281 @@ let filter_rule_for_path { Relation.name; rule_opt; loc } path_str =
...
+            (LayoutUtils.unspecified_for_elements fields)
+      | ( ConstantsSet { loc; labels },
+          (Horizontal [ Unspecified ] | Vertical [ Unspecified ]) ) ->
+          check_term_list
+            (List.map (fun _ -> Label { loc; label = "" }) labels)
+            (LayoutUtils.unspecified_for_elements labels)
+      | ( Function { from_type = _, from_term; to_type = _, to_term; _ },
+          (Horizontal [ Unspecified ] | Vertical [ Unspecified ]) ) ->
+          check_term from_term Unspecified;
+          check_term to_term Unspecified
```
hrutvik: This all seems to be handling the "default" cases - is there no way of streamlining these or factoring them out?

asllib/aslspec/spec.ml
```diff
@@ -804,65 +804,281 @@ let filter_rule_for_path { Relation.name; rule_opt; loc } path_str =
...
-          check_layout (Label { loc; label = name }) (math_layout_for_node node)
-      | Node_TypeVariant { TypeVariant.term } ->
-          check_layout term (math_layout_for_node node)
-      | Node_Relation def ->
-          check_layout (relation_to_tuple def) (math_layout_for_node node)
-      | Node_RecordField { term } ->
-          check_layout term (math_layout_for_node node)
-    in
-    iter_defined_nodes spec check_math_layout_for_definition_node
+      | Tuple { args }, (Horizontal [ Unspecified ] | Vertical [ Unspecified ])
```
hrutvik: Does the raw `Unspecified` case not need handling? _edit: I see this is at the bottom - suggest moving up here with all other default cases_

asllib/aslspec/spec.ml
```diff
@@ -804,65 +804,281 @@ let filter_rule_for_path { Relation.name; rule_opt; loc } path_str =
...
+      | ( Record { fields },
+          (Horizontal [ Unspecified ] | Vertical [ Unspecified ]) ) ->
+          check_term_list
+            (List.map (fun { Term.term } -> term) fields)
+            (LayoutUtils.unspecified_for_elements fields)
+      | ( ConstantsSet { loc; labels },
+          (Horizontal [ Unspecified ] | Vertical [ Unspecified ]) ) ->
+          check_term_list
+            (List.map (fun _ -> Label { loc; label = "" }) labels)
+            (LayoutUtils.unspecified_for_elements labels)
```
hrutvik: I'm a bit confused - this seems to check a list of dummy labels against a list of `Unspecified` (repeated elsewhere). Is a check on the length not sufficient?

asllib/aslspec/spec.ml
```diff
@@ -804,65 +804,281 @@ let filter_rule_for_path { Relation.name; rule_opt; loc } path_str =
...
+      | ConstantsSet { loc; labels }, (Horizontal cells | Vertical cells) ->
+          check_term_list
+            (List.map (fun _ -> Label { loc; label = "" }) labels)
+            cells
+      | ( Function { from_type = _, from_term; to_type = _, to_term; _ },
+          (Horizontal cells | Vertical cells) ) ->
+          if List.length cells <> 2 then
+            Error.bad_layout term layout ~consistent_layout
+          else check_term from_term (List.nth cells 0);
+          check_term to_term (List.nth cells 1)
```
```suggestion
          (Horizontal [from_layout; to_layout] | Vertical [from_layout; to_layout]) ) ->
          check_term from_term from_layout;
          check_term to_term to_layour
      | Function _, _ -> Error.bad_layout term layout ~consistent_layout
```

asllib/aslspec/spec.ml
```diff
@@ -804,65 +804,281 @@ let filter_rule_for_path { Relation.name; rule_opt; loc } path_str =
...
+    let rec check_expr_list spec exprs layout ~expr_for_error =
+      let consistent_layout =
+        Horizontal (LayoutUtils.unspecified_for_elements exprs)
+      in
+      match layout with
+      | Unspecified -> ()
+      (* List renderers treat a single default cell as "use the default layout
+         for every element", regardless of list length. For example, both
+         "r(a, b) { (_) }" and "r(a, b) { [_] }" are valid. *)
+      | Horizontal [ Unspecified ] | Vertical [ Unspecified ] ->
```
hrutvik: Throughout this file - could we treat `Horizontal [ Unspecified ]` and `Vertical [ Unspecified ]` as `Unspecified for checking purposes in various cases? In other words, what does the additional checking provide in each case?

asllib/aslspec/spec.ml
```diff
@@ -804,65 +804,281 @@ let filter_rule_for_path { Relation.name; rule_opt; loc } path_str =
...
+            argument-list check would allow empty compound layouts, but only the
+            default layout is valid: "op { _ }" and "op { (_) }" are valid,
+            while "op { () }" and "op { [] }" are not. *)
+          match layout with
+          | Unspecified | Horizontal [ Unspecified ] | Vertical [ Unspecified ]
+            ->
+              ()
+          | Horizontal _ | Vertical _ ->
+              Error.bad_expr_layout expr layout ~consistent_layout:Unspecified)
+      | [ _ ], [ arg ] when not operator.Relation.is_variadic ->
```
hrutvik: Not sure I follow the `is_variadic` check here

asllib/aslspec/spec.ml
```diff
@@ -804,65 +804,281 @@ let filter_rule_for_path { Relation.name; rule_opt; loc } path_str =
...
+          match layout with
+          | Unspecified -> ()
+          | Horizontal [ Unspecified ] | Vertical [ Unspecified ] -> ()
+          | Horizontal [ _index_layout; body_layout ]
+          | Vertical [ _index_layout; body_layout ] ->
+              check_expr spec body body_layout
+          | Horizontal _ | Vertical _ ->
+              Error.bad_expr_layout expr layout
+                ~consistent_layout:
+                  (Horizontal [ Unspecified; layout_for_expr body ]))
```
hrutvik: In general, I find this less readable than it could be - what about something like:
```suggestion
          if is_default_layout layout then () else
          match layout with
          | Horizontal [ _index_layout; body_layout ]
          | Vertical [ _index_layout; body_layout ] ->
              check_expr spec body body_layout
          | _ ->
              Error.bad_expr_layout expr layout
                ~consistent_layout:
                  (Horizontal [ Unspecified; layout_for_expr body ]))
```

asllib/aslspec/aslspec.md
Roman-Manevich: changed to `if_then_else`

asllib/aslspec/aslspec.md
Roman-Manevich: No, `[_]` and `(_)` only expand to the immediate subexpressions. I will clarify this.

asllib/aslspec/spec.ml
Roman-Manevich: Right. I replaced these constructor-specific cases with a single pattern covering all default layouts, with a comment explaining that default layouts are valid for every term.

asllib/aslspec/spec.ml
Roman-Manevich: Right. `Unspecified` is now handled by the same pattern as the other default layouts.

asllib/aslspec/spec.ml
Roman-Manevich: A length check alone is not sufficient: each constant is atomic, so its corresponding cell must also be a default layout. Otherwise, something like ((_, _), _) would pass for a two-label constants set. The dummy labels obscured that, so I refactored the code to make both checks explicit.

asllib/aslspec/spec.ml
Roman-Manevich: accepted. thanks.

asllib/aslspec/spec.ml
Roman-Manevich: The check distinguishes non-variadic unary operators, which pass their layout directly to their argument, from variadic operators, whose arguments are laid out as a list even when there is only one. I rewrote the code to make this clearer and added comments and examples.

asllib/aslspec/spec.ml
Roman-Manevich: Agreed, thanks.

asllib/aslspec/spec.ml
Roman-Manevich: Right. I removed them and now treat the three default layouts equivalently, except for record updates, whose renderer does not accept the single-cell shorthand.

asllib/aslspec/spec.ml
```diff
@@ -804,65 +804,270 @@ let filter_rule_for_path { Relation.name; rule_opt; loc } path_str =
...
+  (** [check spec] checks that all math layout attributes in [spec] match the
+      structure of their respective AST nodes, raising a [SpecError] if any
+      check fails. *)
+  module CheckLayout : sig
+    val check : t -> unit
+  end = struct
+    let is_default_layout = function
+      | Unspecified | Horizontal [ Unspecified ] | Vertical [ Unspecified ] ->
+          true
+      | Horizontal _ | Vertical _ -> false
```
hrutvik, line 816: nit:
```suggestion
      | _ -> false
```

asllib/aslspec/spec.ml
```diff
@@ -804,65 +804,270 @@ let filter_rule_for_path { Relation.name; rule_opt; loc } path_str =
...
-    | ( Function { from_type = _, from_term; to_type = _, to_term; _ },
-        (Horizontal cells | Vertical cells) ) ->
-        if List.length cells <> 2 then
+        else List.iter2 check_term terms cells
+      in
+      match (term, layout) with
+      (* Default layouts are valid for every term. *)
+      | _, (Unspecified | Horizontal [ Unspecified ] | Vertical [ Unspecified ])
+        ->
+          ()
```
hrutvik, line 834:
```suggestion
      (* Default layouts are valid for every term. *)
      if is_default_layout layout then () else
      match (term, layout) with
```

asllib/aslspec/spec.ml
```diff
@@ -804,65 +804,270 @@ let filter_rule_for_path { Relation.name; rule_opt; loc } path_str =
...
+    let rec check_expr_list spec exprs layout ~expr_for_error =
+      let consistent_layout =
+        Horizontal (LayoutUtils.unspecified_for_elements exprs)
+      in
+      match layout with
+      (* List renderers treat a single default cell as "use the default layout
+         for every element", regardless of list length. For example, both
+         "r(a, b) { (_) }" and "r(a, b) { [_] }" are valid. *)
+      | Unspecified | Horizontal [ Unspecified ] | Vertical [ Unspecified ] ->
+          ()
```
hrutvik, line 866:
```suggestion
      (* List renderers treat a single default cell as "use the default layout
         for every element", regardless of list length. For example, both
         "r(a, b) { (_) }" and "r(a, b) { [_] }" are valid. *)
      if is_default_layout layout then () else
      match layout with
```

asllib/aslspec/spec.ml
```diff
@@ -804,65 +804,270 @@ let filter_rule_for_path { Relation.name; rule_opt; loc } path_str =
...
+          if
+            (* Indexed judgments render as an index binder paired with the body.
+             The binder itself has no expression layout, so either the default
+             layout or a two-cell pair is valid:
+             "INDEX(i, xs: body) { (_, body_layout) }". *)
+            is_default_layout layout
+          then ()
+          else
+            match layout with
+            | Horizontal [ _index_layout; body_layout ]
```
hrutvik: Should we be checking that `_index_layout` is default (based on the comment above: if a binder has no expression layout, should it only be permitted to be default)?

asllib/aslspec/spec.ml
```diff
@@ -804,65 +804,270 @@ let filter_rule_for_path { Relation.name; rule_opt; loc } path_str =
...
+      function
+      | Judgment ({ expr } as judgment) ->
+          check_expr spec expr (Rule.judgment_layout judgment)
+      | Cases cases ->
+          List.iter
+            (fun { elements; _ } ->
+              List.iter (check_rule_element spec) elements)
+            cases
+
+    (** [check_math_layout spec] checks that the math layouts for all defining
```
```suggestion
    (** [check spec] checks that the math layouts for all defining
```

hrutvik (approved):

asllib/aslspec/spec.ml
psafont: I don't know if it's actually the case; but If this variant is likely to be extended it's worth keeping the more explicit branches to force an error at this location when this happens.

Roman-Manevich: Right. I prefer exhaustively enumerating the variants. We might actually extend it to support breaking lists at given positions with something like `[ _, _ @ _, _, _]`.

asllib/aslspec/spec.ml
Roman-Manevich: The problem with this is that the patterns below become non-exhaustive and the compiler complains.

asllib/aslspec/spec.ml
Roman-Manevich: same as above - the compiler complains about non-exhaustive pattern matching.

asllib/aslspec/spec.ml
hrutvik: This is easily fixed with a catch-all case, but OK

asllib/aslspec/spec.ml
hrutvik: As above

PR #1841, opened 2026-05-20
[aslref][asl reference] fixed incorrect error message + example

Roman-Manevich:
This PR refines the error message for cases where a function/getter is used where a procedure/setter is expected (or vice versa). Specifically, a new error constructor is added and used to provide a more accurate description of the error.
* The `MismatchedReturnValue` error is replaced by `MismatchedCallType`.
* In `Interpreter.ml` the safety-net check that `main` does not return multiple values now replaces `MismatchedReturnValue` with the more fitting `BadArity` error.
* No changes are made to the ASL Reference, since it does not distinguish between the finding no call candidates and finding call candidates with the wrong call type - they all lead to the same error code `TE_BC`.

Here is how the error message changes:
1. The error message for `$ aslref --no-exec TypingRule.SCall.bad.asl` is changed from
`ASL Type error: No subprogram declaration matches the invocation: zero().` to `ASL Static error: Mismatched call type for subprogram 'zero': expected a procedure and found a function.`
2. The error message for `$ aslref TypingRule.AnnotateRetTy.bad.asl` is changed from
`ASL Type error: No subprogram declaration matches the invocation: flip(bits(64)).` to `ASL Static error: Mismatched call type for subprogram 'nop': expected a function and found a procedure.`

### DIFF
--- a/asllib/Typing.ml
+++ b/asllib/Typing.ml
@@ -314,22 +314,50 @@ module FunctionRenaming (C : ANNOTATE_CONFIG) = struct
       try IMap.find name env.global.overloaded_subprograms
       with Not_found -> undefined_identifier ~loc name
     in
-    let get_func_sig name' =
+    (* From a specification perspective, [filter_by_call_type] can be
+       ignored. It is used below to distinguish between different
+       error message, which in the specification are mapped to the same
+       error code [TE_BC]. *)
+    let get_func_sig ~filter_by_call_type name' =
       match IMap.find_opt name' env.global.subprograms with
       | Some (func_sig, ses)
-        when has_arg_clash env caller_arg_types func_sig.args
-             && call_type_matches func_sig call_type ->
-          Some (name', func_sig, ses)
+        when has_arg_clash env caller_arg_types func_sig.args ->
+          if (not filter_by_call_type) || call_type_matches func_sig call_type
+          then Some (name', func_sig, ses)
+          else None
       | _ -> None
     in
-    let matching_renamings = set_filter_map get_func_sig renaming_set in
+    let matching_renamings =
+      set_filter_map (get_func_sig ~filter_by_call_type:true) renaming_set
+    in
     match matching_renamings with
     | [ (name', func_sig, ses) ] -> (
         match version with
         | V0 ->
             (deduce_eqs env caller_arg_types func_sig.args, name', func_sig, ses)
         | V1 -> ([], name', func_sig, ses) |: TypingRule.SubprogramForName)
-    | [] -> fatal_from ~loc (Error.NoCallCandidate (name, caller_arg_types))
+    | [] -> (
+        (* If there are no candidates while filtering with [call_type] then
+           either there are no subprograms that match the signature,
+           or there are subprograms that match the signature but they have the
+           wrong call type. For example, a function used for a call statement,
+           or a procedure used in an expression. *)
+        let mismatched_renamings =
+          set_filter_map (get_func_sig ~filter_by_call_type:false) renaming_set
+        in
+        match mismatched_renamings with
+        | [] -> fatal_from ~loc (Error.NoCallCandidate (name, caller_arg_types))
+        | (_, func_sig, _) :: _ ->
+            (* All of the matches in [mismatched_renamings] have the wrong call type,
+               but we report only the first one found. *)
+            fatal_from ~loc
+              (Error.MismatchedCallType
+                 {
+                   error_handling_time = Static;
+                   subprogram_name = name;
+                   expected_call_type = call_type;
+                   found_call_type = func_sig.subprogram_type;
+                 }))
     | _ :: _ ->
         (* If more than one candidate exists, the candidate signature should clash,
            which is detected when typechecking the corresponding declarations. *)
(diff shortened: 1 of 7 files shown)


asllib/tests/ASLTypingReference.t/run.t
```diff
@@ -479,7 +479,13 @@ ASL Typing Tests / annotating types:
   File TypingRule.SCall.bad.asl, line 11, characters 4 to 11:
       zero();
       ^^^^^^^
-  ASL Type error: No subprogram declaration matches the invocation: zero().
+  ASL Static error: Mismatched use of return value from call to 'zero'.
+  [1]
+  $ aslref --no-exec TypingRule.ECall.bad.asl
+  File TypingRule.ECall.bad.asl, line 9, characters 8 to 13:
+      - = nop(); // Illegal: nop is a procedure, therefore no value to consume.
```
HadrienRenaud, line 487: Can you put in the description of the PR the previous output on this test please? and maybe even on `TypingRule.SCall.bad.asl`?

HadrienRenaud (commented):
Thank you for this Roman. I'm wondering how much harder it would be to make it specific for the 2 cases we're interested in:
- if a function is called as a procedure
- if a procedure is called as a function
- maybe there are other interesting cases, but maybe they can be filled later when need arises, and left generic for now.

Maybe we could extend the `MismatchReturnedValue` error constructor with the called call-type and the declared call-type?(or create a new one)

asllib/error.ml
```diff
@@ -36,6 +36,12 @@ type error_desc =
   | TypeInferenceNeeded
   | UndefinedIdentifier of error_handling_time * identifier
   | MismatchedReturnValue of error_handling_time * string
```
hrutvik: I'm wondering if this could be folded in to `MismatchedCallType`

asllib/Typing.ml
```diff
@@ -314,22 +314,50 @@ module FunctionRenaming (C : ANNOTATE_CONFIG) = struct
...
       with Not_found -> undefined_identifier ~loc name
     in
-    let get_func_sig name' =
+    let get_func_sig ~filter_by_call_type name' =
       match IMap.find_opt name' env.global.subprograms with
       | Some (func_sig, ses)
         when has_arg_clash env caller_arg_types func_sig.args
-             && call_type_matches func_sig call_type ->
+             && ((not filter_by_call_type)
+                || call_type_matches func_sig call_type) ->
```
hrutvik: nit: would the new case be more readable as a separate line in the pattern match? Perhaps easier to correlate with the reference too

Roman-Manevich: I don't understand. Do you mean visually in a separate line or as a separate case?

hrutvik: A separate case, i.e. something like
```
 when has_arg_clash env caller_arg_types func_sig.args && filter_by_call_type && call_type_matches ... -> ...
 | ... when as_arg_clash env caller_arg_types func_sig.args && not filter_by_call_type -> ...
```
Alternatively passing the `filter_by_call_type` flag to the `call_type_matches` function or similar

Roman-Manevich: that doesn't seems to work so I've refactored the `filter_by_call_type_bit` into the case.

asllib/Typing.ml
```diff
@@ -314,22 +314,50 @@ module FunctionRenaming (C : ANNOTATE_CONFIG) = struct
...
+        in
+        match mismatched_renamings with
+        | [] -> fatal_from ~loc (Error.NoCallCandidate (name, caller_arg_types))
+        | _ ->
+            (* All of the matches in [mismatched_renamings] have the wrong call type,
+               but we report only the first one found. *)
+            let found_call_type =
+              List.hd mismatched_renamings |> fun (_, func_sig, _) ->
+              func_sig.subprogram_type
+            in
```
HadrienRenaud: A little more dense:
```suggestion
        | (_, { subprogram_type = found_call_type; _ }, _) :: _ ->
            (* All of the matches in [mismatched_renamings] have the wrong call type,
               but we report only the first one found. *)
```

asllib/error.ml
```diff
@@ -378,10 +383,27 @@ module PPrint = struct
...
         pp_err (error_handling_time_to_string t) "Undefined identifier:@ '%s'" s
-    | MismatchedReturnValue (t, s) ->
+    | MismatchedCallType
+        {
+          error_handling_time = t;
+          subprogram_name = s;
+          expected_call_type;
+          found_call_type;
+        } ->
+        let call_type_description call_type =
```
HadrienRenaud, line 393: Maybe worth putting this in `PP.ml`?

asllib/Typing.ml
```diff
@@ -314,22 +314,50 @@ module FunctionRenaming (C : ANNOTATE_CONFIG) = struct
       try IMap.find name env.global.overloaded_subprograms
       with Not_found -> undefined_identifier ~loc name
     in
-    let get_func_sig name' =
+    let get_func_sig ~filter_by_call_type name' =
```
HadrienRenaud, line 321: Maybe worth documenting that for documentation purposes we can always assume that `~filter_by_call` is always `false`?

HadrienRenaud (commented):
2 minor comments but it looks good to me, thanks @Roman-Manevich 

(I can't speak for Hrutvik's comments, maybe good to wait for his review?)

asllib/error.ml
HadrienRenaud: Actually thinking about it, probably not.

asllib/Typing.ml
Roman-Manevich: I added an explanation.

hrutvik (approved):

PR #1842, opened 2026-05-21
[litmus] Bug fix for CI

diaolo01:
The change introduced by pr #1811 makes the CI fail. This was missed due to the litmus CI not being triggered when modifying the Makefile.aarch64 and the local machine supporting FEAT_EXS, whereas the machine running the CI doesn't.
This PR fixes the failure.


relokin:
Thanks Diana!

PR #1180, opened 2025-02-18
Introduce instruction label for memory access and generate (PTE) fault check to labels.

ShaleXIONG:
Introduce instruction label for memory access and generate (PTE) fault check to labels in `Diyone7`.

Most key changes are in file `gen/cycle.ml` `gen/final.ml` and `gen/AArch64Compile_gen.ml`. We extend the `node` (`cycle.ml`) to track if the associated event, e.g., memory load and store, needs to (1) check faults, and/or (2) check the effects, (for example the loaded value). This generate better post-conditions.
- If there is PTE value change in the cycle, the next memory access will be labelled and check fault or not fault, and
- If there is a (plain) value change, the next memory load will checked the loaded value (Consistent with the previous behaviour but rule out redundant register value check in the litmus tests purely for fault check).

 Minor code quality improvement:
- Replace a few manually fold and iteration over lists by standard library call of `List.fold_left`, `List.iter`, `List.map` etc. This are most in `gen/top_gen.ml` and `gen/alt.ml`.
- Add a few comments in some key functions purely for readability.

PR #12, opened 2019-02-20
lib: PPC: Add 'codevar' support

fbq:
Signed-off-by: Boqun Feng <boqun.feng@gmail.com>

PR #1843, opened 2026-05-27
 [gen] Add `store-only` variant

ShaleXIONG:
This adds a `StoreOnly` generator variant for `memtag` variant and emits it to `herd7` metadata as `store-only`. The parser accepts both `StoreOnly` and `store-only`, and validation requires it to be used together with `memtag`.

For multi-instruction RMWs, store-only carries the `memtag` fault check from the read event to the write event. In particular, on `Aarch64` `LxSx` instruction's generated fault label moves from the load-exclusive instruction to the store-exclusive instruction, for example, `diyone7 -arch AArch64 -variant memtag,store-only PodRW T Rfe LxSx PodWW Rfe`
```
AArch64 LB+popt+rmw-po
Variant=memtag store-only
Com=Rf Rf
Orig=PodRWPT RfeTP LxSx PodWW Rfe
"PodRWPT RfeTP LxSx PodWW Rfe"
{
 0:X0=x:green; 0:X2=y:red; 0:X3=y:green;
 1:X0=x:green; 1:X2=y:red;
}
 P0          | P1                   ;
 LDR W1,[X0] | MOV W4,#1            ;
 STG X2,[X3] | Loop00:              ;
             | LDXR W1,[X2]         ;
             | L00: STXR W5,W4,[X2] ; (* <--- the label `L00` moves from `LDXR` to here ` STXR` *)
             | CBNZ W5,Loop00       ;
             | MOV W6,#1            ;
             | STR W6,[X0]          ;

exists ([y]=1 /\ 0:X1=1 /\ 1:X1=0 /\ not (fault(P1:L00,y)))
```

This means:
- In `cycle.ml`, we changed fault-label assignment to be event-aware rather than passing only a direction. The new logic detects multi-instruction RMWs, such as `LxSx`, and under store-only suppresses the read-side fault label while preserving the pending fault-check state for the write event.
- In `AArch64Compile_gen.ml`, we replaced the old “label the first instruction” via `add_label_to_first_instructions` handling for exclusive RMW code with logic that recognises exclusive loads and stores separately via `add_label_to_exclusive_load_and_store`. For `LxSx`, it can now attach the read event’s label to `LDXR`/`LDAXR`, and the write event’s label to `STXR`/`STLXR`/`STXP`.

### DIFF
--- a/gen/AArch64Compile_gen.ml
+++ b/gen/AArch64Compile_gen.ml
@@ -25,8 +25,6 @@ end
 module Make(Cfg:Config) : XXXCompile_gen.S =
   struct
 
-    let do_memtag = Cfg.variant Variant_gen.MemTag
-
 (* Common *)
     let naturalsize = TypBase.get_size Cfg.typ
     module A64 =
@@ -1709,14 +1707,14 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
 (* Access *)
 (**********)
 
-    let add_tag =
-      if do_memtag then
-        fun loc tag -> Code.add_tag loc tag
-      else if do_morello then
-        fun loc _ -> Code.add_capability loc 0
-      else fun loc _ -> loc
+    let add_tag atom loc tag = match atom with
+      | Some (Pte _,_) -> loc
+      | _ ->
+        if do_memtag then Code.add_tag loc tag
+        else if do_morello then Code.add_capability loc 0
+        else loc
 
-    let get_tagged_loc e = add_tag (as_data e.C.loc) e.C.tag
+    let get_tagged_loc e = add_tag e.C.atom (as_data e.C.loc) e.C.tag
 
     let add_label_to_last_instructions e cs =
       match e.C.check_fault with
@@ -1730,21 +1728,35 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
         do_rec cs
       | None -> cs
 
-    (* If there is a fault label in `e`, add the `index`-th label
-       to the first instruction in `cs` *)
-    let add_label_to_first_instructions e cs =
-      match e.C.check_fault with
-      | Some (label_name, _) ->
-        (* find the first non-label instruction *)
-        let rec do_rec cs = match cs with
-          | [] -> assert false (* the `cs` should not be empty *)
-          | instr::rem -> begin match instr with
-            (* skip label or instruction that is already labelled *)
-            | Label(_) -> instr::do_rec rem
-            |_ -> Label(label_name, instr)::rem
-        end in
-        do_rec cs
-      | None -> cs
+    let is_ldxr = function
+      | Instruction (I_LDAR (_, (XX|AX), _, _))
+      | Instruction (I_LDARBH (_, (XX|AX), _, _))
+      | Instruction (I_LDXP _) -> true
+      | _ -> false
+
+    let is_stxr = function
+      | Instruction (I_STXR _)
+      | Instruction (I_STXRBH _)
+      | Instruction (I_STXP _) -> true
+      | _ -> false
+
+    (* The fault label is carried by the read event `er`; in `store-only`,
+       attach it to the following exclusive store instead. *)
+    let add_label_to_exclusive_load_and_store er cs =
+      let add_label e instr =
+        match e.C.check_fault with
+        | Some (label_name, _) -> Label(label_name, instr)
+        | None -> instr in
+      let rec do_rec = function
+        | [] -> assert false (* the `cs` should not be empty *)
+        | (Label(_) as label)::rem -> label :: do_rec rem
+        | instr::rem ->
+            if (not do_store_only && is_ldxr instr)
+               || (do_store_only && is_stxr instr) then
+              (add_label er instr) :: rem
+            else instr :: do_rec rem
+      in
+      do_rec cs
 
     let emit_access st p init e =
     let open WPTE in
@@ -1768,7 +1780,7 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
               (pp_dir d) (C.debug_evt e)
         end
     | Some d,Data loc ->
-        let loc = add_tag loc e.C.tag in
+        let loc = add_tag e.C.atom loc e.C.tag in
         let atom = match e.C.atom with
         | None -> None
         | Some (a,m) -> begin match a with
@@ -1981,7 +1993,6 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
         | W,Some (Neon _,Some _) -> assert false
         end in
         (* Add a label to instructions `cs`, when a fault check is required. *)
-          (* Add a label to instructions `cs`, when a fault check is required. *)
         let cs = add_label_to_last_instructions e cs in
         regs,inits,cs,st
     (* END of emit_access *)
@@ -2012,7 +2023,7 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
       let rW,init,csi,st = U.emit_mov st p init (Value.to_int ew.C.v) in
       let arw = check_arw_lxsx er ew in
       let init,cs,st = XSingle.emit_pair arw p st init rR rW rA ew in
-      let cs = add_label_to_first_instructions er cs in
+      let cs = add_label_to_exclusive_load_and_store er cs in
       rR,init,csi@caddr@cs,st
 
     let emit_exch1 = do_emit_exch1 emit_addr_simple
@@ -2025,7 +2036,7 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
       let arw = check_arw_lxsx er ew in
       let init,cs,st =
         XPair.emit_pair arw p st init (rR1,rR2) (rW1,rW2) rA ew in
-      let cs = add_label_to_first_instructions er cs in
+      let cs = add_label_to_exclusive_load_and_store er cs in
       rR1,init,csi@caddr@cs,st
 
     let emit_exch22 = do_emit_exch22 emit_addr_simple
@@ -2038,7 +2049,7 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
       let module X = ExclusivePair(XLoadPair)(XStore) in
       let init,cs,st =
         X.emit_pair arw p st init (rR1,rR2) rW rA ew in
-      let cs = add_label_to_first_instructions er cs in
+      let cs = add_label_to_exclusive_load_and_store er cs in
       rR1,init,csi@caddr@cs,st
 
     let emit_exch21 = do_emit_exch21 emit_addr_simple
@@ -2052,7 +2063,7 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
       let module X = ExclusivePair(XLoad)(XStorePair) in
       let init,cs,st =
         X.emit_pair arw p st init rR (rW1,rW2) rA ew in
-      let cs = add_label_to_first_instructions er cs in
+      let cs = add_label_to_exclusive_load_and_store er cs in
       rR,init,csi@caddr@cs,st
 
     let emit_exch12 = do_emit_exch12 emit_addr_simple
@@ -2275,7 +2286,7 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
       match e.C.dir,e.C.loc with
       | None,_ -> Warn.fatal "TODO"
       | Some d,Data loc ->
-          let loc = add_tag loc e.C.tag in
+          let loc = add_tag e.C.atom loc e.C.tag in
           let atom = match e.C.atom with
           | None -> None
           | Some (a,m) -> begin match a with
@@ -2573,7 +2584,7 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
             match atom with
             | Some (Tag,None) ->
                 let cs0,st = calc0_gen csel st vdep r2 r1 in
-                let rA,init,st = U.next_init st p init (add_tag loc (Value.to_int e.C.v)) in
+                let rA,init,st = U.next_init st p init (add_tag e.C.atom loc (Value.to_int e.C.v)) in
                 let rB,cB,st = sum_addr st rA r2 in
                 rB,pseudo (cs0@cB),init,st,[]
             | Some (_,Some (sz,_)) ->
@@ -2619,7 +2630,7 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
           let r2,cs2,init,st = match atom with
             | Some(Neon _, None) -> r2,cs2,init,st
             | _ -> r2,cs2@pseudo addi,init,st in
-          let loc = add_tag loc e.C.tag in
+          let loc = add_tag e.C.atom loc e.C.tag in
           begin match atom with
           | None ->
               let init,cs,st =
--- a/gen/common/variant_gen.ml
+++ b/gen/common/variant_gen.ml
@@ -38,6 +38,8 @@ type t =
   | KVM | NoFault
 (* Synchronisation mode *)
   | Sync | Async
+(* Store-only mode *)
+  | StoreOnly
 (* Neon AArch64 extension *)
   | Neon
 (* Scalable Vector extension (AArch64) *)
@@ -52,12 +54,12 @@ let tags =
    "Mixed";"FullMixed";"MixedDisjoint"; "MixedStrictOverlap";
    "Ifetch(Self)"; "MemTag";
    "NoVolatile"; "Morello"; "VMSA(KVM)"; "NoFault";
-   "Sync"; "Async"; "Neon"; "ConstrainedUnpredictable"; ]
+   "Sync"; "Async"; "StoreOnly"; "Neon"; "ConstrainedUnpredictable"; ]
 
 let all_t =
   [ AsAmo ; ConstsInInit ; Mixed ; FullMixed ; MixedDisjoint ; MixedStrictOverlap ;
     Self ; MemTag ; NoVolatile ; Morello ; KVM ; NoFault ;
-    Sync ; Async ; Neon ; SVE ; SME ; ConstrainedUnpredictable ]
+    Sync ; Async ; StoreOnly ; Neon ; SVE ; SME ; ConstrainedUnpredictable ]
 
 let parse tag = match Misc.lowercase tag with
 | "asamo" -> Some AsAmo
@@ -74,6 +76,7 @@ let parse tag = match Misc.lowercase tag with
 | "nofault" -> Some NoFault
 | "sync" -> Some Sync
 | "async" -> Some Async
+| "storeonly" | "store-only" -> Some StoreOnly
 | "neon" -> Some Neon
 | "sve" -> Some SVE
 | "sme" -> Some SME
@@ -95,6 +98,7 @@ let pp = function
   | NoFault -> "NoFault"
   | Sync -> "Sync"
   | Async -> "Async"
+  | StoreOnly -> "StoreOnly"
   | Neon -> "Neon"
   | SVE -> "sve"
   | SME -> "sme"
@@ -110,6 +114,7 @@ let pp_herd_variant = function
   | MemTag -> Some "memtag"
   | Sync -> Some "sync"
   | Async -> Some "async"
+  | StoreOnly -> Some "store-only"
   | Morello -> Some "morello"
   | KVM  -> Some "vmsa"
   | ConstrainedUnpredictable -> Some "ConstrainedUnpredictable"
@@ -118,5 +123,6 @@ let is_mixed v = v Mixed || v FullMixed
 let is_kvm v = v KVM
 
 let validate v =
-  if (v Sync || v Async) && not (v MemTag) then
-    Warn.user_error "variants `Sync` and `Async` must be used with `MemTag`"
+  if (v Sync || v Async || v StoreOnly) && not (v MemTag) then
+    Warn.user_error
+      "variants `Sync`, `Async` and `StoreOnly` require `MemTag`"
--- a/gen/cycle.ml
+++ b/gen/cycle.ml
@@ -157,6 +157,7 @@ module Make (O:Config) (E:Edge.S) :
   let do_sve = O.variant Variant_gen.SVE
   let do_sme = O.variant Variant_gen.SME
   let do_no_fault = O.variant Variant_gen.NoFault
+  let do_store_only = O.variant Variant_gen.StoreOnly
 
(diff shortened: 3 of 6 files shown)


gen/tests/diy-basic-check.t
```diff
@@ -128,6 +128,73 @@ A memtag generation test with `Variant` duplicated in metadata, because of (1) `
    STG X6,[X3]       |                  ;
   
   exists ([tag(y)]=:blue /\ 0:X0=x:green /\ 1:X0=0 /\ not (fault(P1:L00,y)))
+A memtag store-only generation test
+  $ diyone7 -arch AArch64 -variant memtag,store-only DpDatadW T PosWW T Rfe PodRW Rfe T -info "Variant=memtag"
```
relokin: Why do you use
```
-info "Variant=memtag"
```
?

relokin: Also can you please explain how you chose this cycle?

gen/tests/diy-basic-check.t
```diff
@@ -128,6 +128,73 @@ A memtag generation test with `Variant` duplicated in metadata, because of (1) `
...
+   P0                | P1          ;
+   MOV X0,X1         | LDR W0,[X6] ;
+   LDG X0,[X1]       | MOV W2,#1   ;
+   EOR X2,X0,X0      | STR W2,[X1] ;
+   ADD X4,X3,W2,SXTW |             ;
+   STG X4,[X5]       |             ;
+   STG X6,[X3]       |             ;
+  
+  exists ([tag(y)]=:blue /\ 0:X0=x:green /\ 1:X0=0)
+A memtag store-only `LxSx` generation test
```
relokin: Is this a store-only test?

gen/tests/diy-basic-check.t
relokin, line 1: Can you please explain how you chose these cycles/three tests?

gen/AArch64Compile_gen.ml
```diff
@@ -1737,21 +1737,42 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
...
-      | None -> cs
+        | Some (label_name, _) -> Label(label_name, instr)
+        | None -> instr in
+      let rec do_rec met_read cs = match cs,met_read with
+        | [],_ -> assert false (* the `cs` should not be empty *)
+        (* skip existing loop label or instruction that is already labelled *)
+        | (Label(_) as label)::rem,met_read -> label :: do_rec met_read rem
+        | instr::rem,false ->
+            if is_ldxr_instruction instr then
+              (* Label the first `ldxr` *)
```
relokin: Can there be more that one ldxr?

gen/AArch64Compile_gen.ml
```diff
@@ -1737,21 +1737,42 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
...
+              (add_label er instr) :: do_rec true rem
+            else instr :: do_rec false rem
+        | instr::rem,true ->
+            if is_stxr_instruction instr then
+              (* Label the first `stxr *)
```
relokin: Can there be more than one stxr?

gen/AArch64Compile_gen.ml
```diff
@@ -1737,21 +1737,42 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
...
+              (add_label ew instr)::rem
```
relokin: How is this different to adding a label to all LDXR and STXR instructions? I think

gen/AArch64Compile_gen.ml
```diff
@@ -1737,21 +1737,42 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
...
-          | instr::rem -> begin match instr with
-            (* skip label or instruction that is already labelled *)
-            | Label(_) -> instr::do_rec rem
-            |_ -> Label(label_name, instr)::rem
-        end in
-        do_rec cs
-      | None -> cs
+        | Some (label_name, _) -> Label(label_name, instr)
+        | None -> instr in
+      let rec do_rec met_read cs = match cs,met_read with
```
relokin: What is `met_read`? And more generally, I don't understand what is the purpose of this.  Why can't you just unconditionally try to add faults to LDXR and STXR? And, I think there can never be more than one LDXR and one STXR in cs.

gen/cycle.ml
```diff
@@ -571,24 +572,60 @@ module CoSt = struct
...
+      (* In `rmw`, it processes read and write events separately.
+         In general, we only need to assign fault label to the (first) read event.
+         Also, when checking faults, always check against `W` in `rmw`.
+         This avoids label gap or extra label after compilation.
+         However, for multi-instruction `rmw`, when `store-only` is on,
+         only assign the fault label to the write event. *)
+      | true,Some W,true
+      | true,Some R,false -> Some W
+      (* None in `dir` indicates it is (1) `rmw` and (2) suppressing fault label *)
+      | true,_,_ -> None in
```
relokin: IMO, this is very complex without a good reason. I would just do
```suggestion
    let dir = n.evt.rmw,n.evt.dir in
```
And then handle story-only further down in `do_memtag`

gen/tests/diy-basic-check.t
ShaleXIONG: The `-info "Variant=memtag"` is redundant here.

For this particular test, it compares to the existing test just above; the label `L00` on `LDR` is removed. This is because there is not more `STR` after,  so no fault is check here in `store-only` variant.

gen/tests/diy-basic-check.t
ShaleXIONG: This section contains two tests with exact same cycle however one have `store-only`. I updated the section description here to emphasis the label `L00`. In particular, in non-`store-only` case, the label is on `LDXR` while in the `store-only` case, it is on the `STXR`.

gen/tests/diy-basic-check.t
ShaleXIONG: Given your comment, I have update some comments in the tests and also I add one more section. Now we have 3 set of tests:
- `store-only` completely removes the label and associated fault check, because there is no more `STR` afterwards to the same location.
- The newly added section contains two tests that without `store-only`, there are labels for both `STR` and `LDR` respectively, while with `store-only` the label to `LDR` is removed.
- `LxSx` test, this is the most tricky case in technical sense, because it is a form of `rmw` however it compiles to two separate instructions, `LDXR` and `STXR`. In this case, the label is moved from `LDXR` without `store-only` to `STXR` with `store-only`.

gen/AArch64Compile_gen.ml
ShaleXIONG: Yes we can, this is a very special case for `LxSx` with unfold argument `-ua 2` for example. Consider the following two examples: 
- `diyone7 -arch AArch64 -variant memtag,store-only T PosWR LxSx Coi -oneloc -ua 2`
- `diyone7 -arch AArch64 -variant memtag T PosWR LxSx Coi -oneloc -ua 2`

```
memtag                                              | memtag,store-only
===========================================================================================================
AArch64 CoWW+postp-rmw-coipt                        | AArch64 CoWW+postp-rmw-coipt
Variant=memtag                                      | Variant=memtag store-only
Generator=diyone7 (version 7.58+1)                  | Generator=diyone7 (version 7.58+1)
Com=Co                                              | Com=Co
Orig=PosWRTP LxSx CoiPT                             | Orig=PosWRTP LxSx CoiPT
"PosWRTP LxSx CoiPT"                                | "PosWRTP LxSx CoiPT"
{                                                   | {
 0:X0=x:red; 0:X1=x:green; 0:X6=ok0;                |  0:X0=x:red; 0:X1=x:green; 0:X6=ok0;
}                                                   | }
 P0                ;                                |  P0                   ;
 STG X0,[X1]       ;                                |  STG X0,[X1]          ;
 MOV W3,#1         ;                                |  MOV W3,#1            ;
 L00: LDXR W2,[X1] ;                                |  LDXR W2,[X1]         ;
 STXR W4,W3,[X1]   ;                                |  L00: STXR W4,W3,[X1] ; (* `L00` label moves here *)
 CBZ W4,Go00       ;                                |  CBZ W4,Go00          ;
 LDXR W2,[X1]      ;                                |  LDXR W2,[X1]         ;
 STXR W4,W3,[X1]   ;                                |  STXR W4,W3,[X1]      ; (* No more label and fault check *)
 CBNZ W4,End0      ;                                |  CBNZ W4,End0         ;
 Go00:             ;                                |  Go00:                ;
 ADD W5,W5,#1      ;                                |  ADD W5,W5,#1         ;
 End0:             ;                                |  End0:                ;
 STR W5,[X6]       ;                                |  STR W5,[X6]          ;
                                                    |
exists (0:X2=0 /\ [ok0]=1 /\ not (fault(P0:L00,x))) | exists (0:X2=0 /\ [ok0]=1 /\ not (fault(P0:L00,x)))
```
In short, I recall a discussion from #1180 that we were only interested to label and fault-check on the first iteration. Therefore the function here only label the first `STXR` and `LDXR` respectively.

gen/AArch64Compile_gen.ml
ShaleXIONG: I think this is more about design choice when we have unfold argument `-ua 2` for example. Consider `diyone7 -arch AArch64 -variant memtag T PosWR LxSx Coi -oneloc -ua 2`, if we add label to all `LDXR` and `STXR` we will get the following tests:
```
AArch64 CoWW+postp-rmw-coipt
Variant=memtag store-only
Com=Co
Orig=PosWRTP LxSx CoiPT
"PosWRTP LxSx CoiPT"
{
 0:X0=x:red; 0:X1=x:green; 0:X6=ok0;
}
 P0                   ;
 STG X0,[X1]          ;
 MOV W3,#1            ;
 LDXR W2,[X1]         ;
 L00: STXR W4,W3,[X1] ;
 CBZ W4,Go00          ;
 LDXR W2,[X1]         ;
 STXR W4,W3,[X1]      ; (* <- an extra `L00` *)
 CBNZ W4,End0         ;
 Go00:                ;
 ADD W5,W5,#1         ;
 End0:                ;
 STR W5,[X6]          ;

exists (0:X2=0 /\ [ok0]=1 /\ not (fault(P0:L00,x)))
```
I do not think it is correct nor a good idea to label two `STXR` with the same label `L00`. Although we can assign different labels here in different iterations, I am not sure it is necessary as labelling the first iteration is good enough?

gen/AArch64Compile_gen.ml
ShaleXIONG: I have reply above. In short we can have more than one `STXR` with unfold `-ua 2` for example

gen/AArch64Compile_gen.ml
ShaleXIONG: I have replied above that more than one `LDXR` and `STXR` are possible with `-ua 2`

gen/cycle.ml
ShaleXIONG: Let me separate the process for `store-only`.

gen/tests/diy-basic-check.t
```diff
@@ -128,6 +128,114 @@ A memtag generation test with `Variant` duplicated in metadata, because of (1) `
    STG X6,[X3]       |                  ;
   
   exists ([tag(y)]=:blue /\ 0:X0=x:green /\ 1:X0=0 /\ not (fault(P1:L00,y)))
+A memtag store-only comparison test against the version above on label `L00`
+  $ diyone7 -arch AArch64 -variant memtag,store-only DpDatadW T PosWW T Rfe PodRW Rfe T
```
relokin: BTW, I think it is problematic that by default diyone7 will give the same name to the test above and this test.

gen/tests/diy-basic-check.t
```diff
@@ -128,6 +128,114 @@ A memtag generation test with `Variant` duplicated in metadata, because of (1) `
...
+   P0                | P1          ;
+   MOV X0,X1         | LDR W0,[X6] ;
+   LDG X0,[X1]       | MOV W2,#1   ;
+   EOR X2,X0,X0      | STR W2,[X1] ;
+   ADD X4,X3,W2,SXTW |             ;
+   STG X4,[X5]       |             ;
+   STG X6,[X3]       |             ;
+  
+  exists ([tag(y)]=:blue /\ 0:X0=x:green /\ 1:X0=0)
+A memtag store-only `PosRW` comparison test on label `L01`
```
relokin: Is this stale?

gen/cycle.ml
```diff
@@ -571,24 +572,67 @@ module CoSt = struct
   let label_pte_fault dir pte_val =
     Some ( (Label.next_label "L"), (Value.can_fault dir pte_val) )
 
+  let is_multi_instruction_rmw n =
```
relokin:
```suggestion
  let is_lxsx n =
```

gen/AArch64Compile_gen.ml
relokin: Fine but then why do you need this complex logic. Isn't the following sufficient:
* if `store-only` do the recursion until the first STXR, add a label to it and then stop the recursion and return the suffix.
* otherwise  do the recursion until the first LDXR, add a label to it and then stop the recursion and return the suffix.

gen/AArch64Compile_gen.ml
relokin: Thanks but I still don't understand the logic of met_read and the name is not helpful.

gen/AArch64Compile_gen.ml
```diff
@@ -1728,21 +1728,42 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
         do_rec cs
       | None -> cs
 
-    (* If there is a fault label in `e`, add the `index`-th label
-       to the first instruction in `cs` *)
-    let add_label_to_first_instructions e cs =
-      match e.C.check_fault with
-      | Some (label_name, _) ->
+    let is_ldxr_instruction = function
```
relokin:
```suggestion
    let is_ldxr = function
```

gen/AArch64Compile_gen.ml
```diff
@@ -1728,21 +1728,42 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
...
+      | Instruction (I_LDAR (_, (XX|AX), _, _))
+      | Instruction (I_LDARBH (_, (XX|AX), _, _))
+      | Instruction (I_LDXP _) -> true
+      | _ -> false
+
+    let is_stxr_instruction = function
```
relokin:
```suggestion
    let is_stxr = function
```

gen/AArch64Compile_gen.ml
```diff
@@ -1728,21 +1728,42 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
...
+              (* Label the first `ldxr` *)
+              (add_label er instr) :: do_rec true rem
+            else instr :: do_rec false rem
+        | instr::rem,true ->
+            if is_stxr_instruction instr then
+              (* Label the first `stxr *)
+              (add_label ew instr)::rem
+            else instr :: do_rec true rem
+      in
+      do_rec false cs
```
relokin:
```suggestion
                  let rec do_rec = function
                    | [] -> assert false (* the `cs` should not be empty *)
                    | (Label(_) as label)::rem -> label :: do_rec rem
                    | instr::rem when (Option.is_some ew && is_stxr_instruction instr) ->
                        (add_label ew instr)::rem
                    | instr::rem when (Option.is_some er && is_ldxr_instruction inst ->
                        (add_label er instr)::rem
                    | _ ->
                        instr::do_rec rem
                  in
                  do_rec false cs
```

relokin (commented):
I think you have to be more specific about suppressing the fault check for loads. Take this test for example:
```
$> diyone7 -metadata false -arch AArch64 -variant vmsa,memtag,store-only -oneloc  PteV0 TLBI-sync.ISHsWR ISB Fri
AArch64 CoWR0+tlbi-sync.ishsptev0p-[isb]-fripptev0
Variant=memtag store-only vmsa
{
 [x]=1;
 0:X0=PTE(x:red); 0:X1=(oa:PA(x), valid:0); 0:X2=x:red; 0:X4=x;
}
 P0              ;
 STR X1,[X0]     ;
 DSB ISH         ;
 LSR X5,X4,#12   ;
 TLBI VAAE1IS,X5 ;
 DSB ISH         ;
 ISB             ;
 LDR W3,[X2]     ;

exists (true)

```
The LDR in this test needs to be checked for faults.

As a side comment, in the initialization, we shouldnot refer to PTE(x:red) but to PTE(x).

ShaleXIONG:
> I think you have to be more specific about suppressing the fault check for loads. Take this test for example:
> 
> ```
> $> diyone7 -metadata false -arch AArch64 -variant vmsa,memtag,store-only -oneloc  PteV0 TLBI-sync.ISHsWR ISB Fri
> AArch64 CoWR0+tlbi-sync.ishsptev0p-[isb]-fripptev0
> Variant=memtag store-only vmsa
> {
>  [x]=1;
>  0:X0=PTE(x:red); 0:X1=(oa:PA(x), valid:0); 0:X2=x:red; 0:X4=x;
> }
>  P0              ;
>  STR X1,[X0]     ;
>  DSB ISH         ;
>  LSR X5,X4,#12   ;
>  TLBI VAAE1IS,X5 ;
>  DSB ISH         ;
>  ISB             ;
>  LDR W3,[X2]     ;
> 
> exists (true)
> ```
> 
> The LDR in this test needs to be checked for faults.
> 
> As a side comment, in the initialization, we shouldnot refer to PTE(x:red) but to PTE(x).

I will strengthen the `store-only` to exclusive to `memtag`, and disallowed when other variants such as `vmsa` exists.

WIll patch the `x:red` in a separate commit here.

[Update] I have patched the `Pte(x:red)` problem in a separate commit `dc12e09`

gen/AArch64Compile_gen.ml
ShaleXIONG: I just updated.

gen/AArch64Compile_gen.ml
ShaleXIONG: I just updated.

gen/tests/diy-basic-check.t
ShaleXIONG: I opened an issue #1898 for this. I will update it in a separate PR.

gen/common/variant_gen.ml
```diff
@@ -118,5 +123,6 @@ let is_mixed v = v Mixed || v FullMixed
 let is_kvm v = v KVM
 
 let validate v =
-  if (v Sync || v Async) && not (v MemTag) then
-    Warn.user_error "variants `Sync` and `Async` must be used with `MemTag`"
+  if (v Sync || v Async || v StoreOnly) && not (v MemTag) then
+    Warn.user_error
+      "variants `Sync`, `Async` and `StoreOnly` require `MemTag` and cannot be combined with `KVM`"
```
relokin: This is stale, isn't it?
```suggestion
      "variants `Sync`, `Async` and `StoreOnly` require `MemTag`"
```

gen/AArch64Compile_gen.ml
```diff
@@ -1766,7 +1778,9 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
               (pp_dir d) (C.debug_evt e)
         end
     | Some d,Data loc ->
-        let loc = add_tag loc e.C.tag in
+        let loc = match e.C.atom with
+        | Some (Pte _,_) -> loc
+        | _ -> add_tag loc e.C.tag in
```
relokin: Is there a reason why this check is not implemented inside add_tag?

gen/cycle.ml
```diff
@@ -572,12 +573,14 @@ module CoSt = struct
...
 
   (* Helper function returns a fresh label and a boolean for if it should fault,
-     if a fault check is need. Otherwise return `None`. *)
+     if a fault check is needed. Otherwise return `None`. *)
   let fault_update st dir =
     let unset_check_fault st = {st with check_fault = NoDir } in
     let pte_val = get_pte_value st in
     match () with
     | _ when (st.check_fault = NoDir || do_no_fault) -> None,unset_check_fault st
+    | _ when do_store_only && dir = R
```
relokin: I am really struggling with the way `fault_update` is structured. Specifically, I don't understand why you do `match () with`.

But specifically the new code makes the whole logic even more complex. Can you please move this case below the `| _ when kvm` and merge it with `| do_memtag`.

I would also favour a complete rewrite of this along the lines of https://github.com/relokin/herdtools7/commit/a107a8500394fc39d065152c8f8c79365b6494cd

ShaleXIONG: Let me rewrite this function here; I will put in a separate commit.  

Here `match ()`-`when` is basically a `if`-`else if` chain.

gen/AArch64Compile_gen.ml
ShaleXIONG: This existing function `add_tag` is frankly of bad name. It basically appends the tag value (second parameter) such as `green` and `red`, after the location (first parameter) such as `x` and `y`. We can move the check inside `add_tag` but we will need extra parameter to carry the `atom` value.

Let me move those check into `add_tag` in a separate commit and we can decide if we like this change.

ShaleXIONG: I update another wrapper function `get_tagged_loc` which includes the check on `Pte`. This function already take a node event `e` as the input so we get all the information we need.  ref: 5692616d1581dfac1b608a4d4c30d948124028c2

gen/cycle.ml
ShaleXIONG: I just update in a new commit ref: 7c64a528cc5a22830980f75f653933ed87daa848. it is mostly what you proposed but the first match `|_,_ when do_no_fault` will short-circuit all the rest of update/check, rather than a if-else branch just decide the `st` value.

relokin: Thanks!

gen/AArch64Compile_gen.ml
relokin: Thanks, but `add_tag` is called many more times and at this point, there should be no circumstances under which we would want to add a tag to a PTE. That's why I suggested you change add_tag instead.

ShaleXIONG: Just updated. ref: 7e85bf16b0dbc11828cb336cdaac075c19800173. The two functions `add_tag` and `get_tagged_loc` become almost the same. I did not unify them to keep minimum change though. If you think it makes sense to do here, I can quickly patch it too.

relokin (commented):
Thanks Shale, can you please clean up the commit history? I will have a last look and then we can merge.

relokin (approved):
Thanks Shale, please go ahead and merge when you’re ready.

PR #1844, opened 2026-05-29
[herd] Turn the big pattern matching in AArch64Lexer.mll into a Hashtbl

relokin:
In `lib/AArch64Lexer.mll`, to avoid a transition table overflow, part of the lexer was implemented through a big pattern matching statement. As a result of this, `js_of_ocaml` produced rather inefficient JS code and that would cause stack overflows in certain browsers (for example, Firefox on MacOS). This inefficiency is now addressed in the development version of `js_of_ocaml`. However, to ensure that herd-www to work with older versions of `js_of_ocaml`, we convert the big pattern matching to a Hashtbl.


relokin:
For reference this PR is the result of this discussion in https://github.com/ocsigen/js_of_ocaml/issues/2122

maranget:
Personally I liked the old way of using patten matching as those strings are static data. Of course I do not object to the change that improves robustness of the web interface. Wouldn't it be possible to have the hash table implementation for the web version and teh pattern matching version for the standard tools?

relokin:
> Personally I liked the old way of using patten matching as those strings are static data. Of course I do not object to the change that improves robustness of the web interface. Wouldn't it be possible to have the hash table implementation for the web version and teh pattern matching version for the standard tools?

Thanks for this Luc.

I am a little worried that keeping the two in sync will become a challenge, especially given the size of the pattern match/hash table and the absence of any testing whatsoever of the web interface.

I would be more in favour of mandating the next version of js_of_ocaml, if we are not keen to change this code.

maranget:
Let us choose robustness over my bias in favour of pattern-matching. An alternative would be to include both techniques and some equivalence test. But this looks too complicated.

So I approve the PR.

lib/AArch64Lexer.mll
```diff
@@ -30,734 +30,730 @@ module LU = LexUtils.Make(O)
 
 let parse_creg = if O.is_morello then A.parse_creg  else fun _ -> None
 
+let keyword_table =
+  let tbl = Hashtbl.create 1024 in
+  let add names tok =
+    List.iter (fun name -> Hashtbl.replace tbl name tok) names in
```
ShaleXIONG: A small sanity check, in the case of a `name` already exists, probably print a warning or raise an error.

relokin: Thanks Shale, I've added a check for duplicate elements.

relokin:
@fsestini do you have an opinion about this PR?

ShaleXIONG (approved):

lib/AArch64Lexer.mll
```diff
@@ -30,734 +30,749 @@ module LU = LexUtils.Make(O)
...
+  "rev64", REV64;
+  "rev", REV;
+  "extr", EXTR;
+  "rbit", RBIT;
+  "abs", ABS;
+  "cmp", CMP;
+  "tst", TST;
+  (* Those operations are factorized *)
+  "eor", OP A.EOR;
+  "eon", OP A.EOR;
```
maranget: Typo?
```suggestion
  "eon", OP A.EON;
```
(Was already here!)

relokin: Very good catch! Thanks Luc!

lib/AArch64Lexer.mll
```diff
@@ -30,734 +30,749 @@ module LU = LexUtils.Make(O)
...
+  "s1e0r", A.AT.(AT_OP {stages=S1; level=A.E0; rw=R; });
+  "s1e0w", A.AT.(AT_OP {stages=S1; level=A.E0; rw=W; });
+  "s1e1r", A.AT.(AT_OP {stages=S1; level=A.E1; rw=R; });
+  "s1e1w", A.AT.(AT_OP {stages=S1; level=A.E1; rw=W; });
+  (* System registers *)
+  "mrs", MRS;
+  "msr", MSR;
+] 
+
+let keyword_table =
```
fsestini, line 685: Consider moving this `keyword_table` binding (and the keyword list) out of the `Make` functor, so that the hashtable is only allocated once

lib/AArch64Lexer.mll
```diff
@@ -30,734 +30,749 @@ module LU = LexUtils.Make(O)
...
-                                end
-                            end
-                        end
-                    end
-                end
-            end
-        end
-    end
+  if O.debug then Printf.eprintf "Check: '%s'\n" name ;
+  let name_key = String.lowercase_ascii name in
```
fsestini: I think this preliminary lowercase makes it so that the lexer/parser now accepts more strings as valid instructions, compared to before the PR. For example, post-PR lexer would tokenise `"Ldr"` as a proper LDR, whereas pre-PR lexer would have only accepted `"LDR"` or `"ldr"`. Are we ok with this change?

fsestini (commented):
Sorry, I realise I'm late to the party. I left a couple of comments.

Like @maranget , I have a slight preference for the pattern matching approach, and personally I would be in favour of just bumping the minimum version of js_of_ocaml. However I'm also fine with the approach taken by this PR.

lib/AArch64Lexer.mll
relokin: I think we are, I don't see any problem with that.

maranget: If we prefer backward compatibility, which I tend to do, we can omit the preliminary lowercase and replace `Hashtbl.add tbl name tok`  by `Hashtbl.add tbl name tok; Hashtbl.add tbl (Sring.uppercase_ascii name) tok`.

We cannot assume that nobody used `Ldr` as  a label...

relokin: Fair, I didn't think of the case of a label.

relokin: Fixed, thanks Luc!

relokin:
@ShaleXIONG thanks for the review. @maranget, @fsestini I am planning to merge this later today. Let me know if you have any objections or need more time.

fsestini (approved):
LGTM

relokin:
Thanks @fsestini @maranget and @ShaleXIONG for the reviews!

PR #1845, opened 2026-05-29
[gen] Limit value checks to communication-adjacent events

ShaleXIONG:
This change narrows generated final-condition value checks so they focus on events that are adjacent to communication edges. A node is considered adjacent if (1) its incoming edge (`node.prev.edge`) or (2) outgoing edge (`node.edge`) is communication edge, or (3) both the read or write side of an RMW is adjacent to a communication edge.

For reads, `check_value` when building the cycle in `cycle.ml` is now only enabled when the read is adjacent to a communication edge and the location already requires value checking. This avoids emitting final-condition checks for plain program-order reads that do not witness communication. For example: `diyone7 -arch AArch64 PodWW Rfe PodRR PosRR Amo.Swp Coe`, now generates:
```
AArch64 S+po+poR-pos-amo.swp
Com=Rf Co
Orig=PodWW Rfe PodRR PosRR Amo.Swp Coe
"PodWW Rfe PodRR PosRR Amo.Swp Coe"
{
 0:X1=x; 0:X3=y;
 1:X1=x; 1:X3=y;
}
 P0          | P1             ;
 MOV W0,#2   | LDR W0,[X3]    ;
 STR W0,[X1] | LDR W2,[X1]    ; (* <- this read from `PosRR` is no longer checked *)
 MOV W2,#1   | MOV W5,#1      ;
 STR W2,[X3] | SWP W5,W4,[X1] ;

exists ([x]=2 /\ 1:X0=1 /\ 1:X4=0) (* <- drop `1:X2=0` from previous version *)
```

For writes, `coherence` write collection in `cycle.ml` now trims trailing local writes until it reaches either:
- a write adjacent to a communication edge
- an inserted `store`. Note the `store` edge always add an extra write to the beginning of the thread hence it should be considered adjacent to communication edge.
This means the final write checks will ignore local writes inside a thread but only focus on those adjacent to communication edges. For example, `diyone7 -arch AArch64 PosWW PosWW Coi -oneloc -obs oo`
```
AArch64 CoWW+posW-pos-coi
Com=Co
Orig=PosWW PosWW Coi
"PosWW PosWW Coi"
{
 0:X1=x;
}
 P0          ;
 MOV W0,#2   ;
 STR W0,[X1] ;
 MOV W2,#3   ;
 STR W2,[X1] ;
 MOV W3,#1   ;
 STR W3,[X1] ;

exists ([x]=2) (* <-- it was [x]=3 previously, while now it checks 
                                  the write associated with communication edges. *)
```

This technical change also leads to more tests in `diy7`, because `diy7` by default only generates tests with up to 2 writes per locations, while now they are writes only adjacent to communication edges. All the new tests are forbidden.


gen/cycle.ml
```diff
@@ -1182,7 +1196,9 @@ let do_set_read_v init =
       | Some R ->
         (* If the result of this read need to be checked,
            i.e. generating postcondition *)
-        let check_value = Some (CoSt.get_check_value st) in
+        let check_value =
+          Some ( adjacent_with_communication_edge n
```
relokin: I am not sue I understand why we can't just do
```suggestion
          Some ((E.is_com n.prev.edge || E.is_com n..edge)
```

gen/cycle.ml
```diff
@@ -1588,7 +1604,20 @@ let merge_changes n nss =
...
+     `top_gen.ml` to decide coherence/write value check.
+     Especially when there are two or more writes, the last value
+     will be checked. Here we remove the tail which means a local,
+     non-communication write will be ignored. *)
+    List.fold_right ( fun (loc,n) (new_list,have_seen_communication) ->
+      match have_seen_communication,adjacent_with_communication_edge_or_store n with
+      | true,_ | _,true -> (loc,n) :: new_list,true
+      | false, false -> new_list,false
+    ) (do_rec n) ([],false)
+    |> fst
```
relokin:
```suggestion
    do_rec n
```

gen/cycle.ml
relokin, line 1585: Similarly I don't understand why we can't do this?
```suggestion
            (E.is_node m.edge.E.edge && && E.is_com m.prev.edge) || not (pbank m.evt.bank)
```

gen/cycle.ml
ShaleXIONG: There are 4 situations here rather than 2 based on the status of `n`
- if `n` is a simple `read`, its `n.edge` might be communication edge e.g. `read-of-n -> Fre`
- if `n` is a simple `write`, its previous `n.prev.edge` might be communication edge e.g. `Fre -> write-of-n`
- if `n` is a `read` from read-modify-write, its next, `n.next.edge` might be communication edge `read-from-rmw-of-n -> write-from-rmw -> Fre`
- Last, `n` is a `write` from read-modify-write, its prev of prev, `n.prev.prev.edge` might be communication edge, `Rfe -> read-from-rmw -> write-from-rmw-of-n`

In the last two cases, the final litmus tests will have the read-modify-write instruction at the end or the beginning of the process. These two cases are covered by the last two `||` branches in `adjacent_with_communication_edge`

relokin: In this instance `n` is a read, isn't it? I think we need to check if the incoming or outgoing edge is a communication edge or if `n` is the read of a `rmw` - as we said we only really need to check the value of a CAS but we can also live with checking all rmw. Do you agree?

ShaleXIONG: Yes you are right. I think we do not need the last case when `n` is `write` from a `read-modify-write`. Let me update accordingly.

relokin: Can you please update the code? I don't think we need the helper function. We need a comment to explain why we are emitting checks for `rmw`.

ShaleXIONG: Will do.

gen/cycle.ml
```diff
@@ -1182,7 +1184,14 @@ let do_set_read_v init =
...
         (* If the result of this read need to be checked,
            i.e. generating postcondition *)
-        let check_value = Some (CoSt.get_check_value st) in
+        let check_value =
+          Some ( ( E.is_com n.prev.edge
+                 || E.is_com n.edge
+                 (* For an RMW sequence such as
+                    read --CAS--> write --communication edge--> ... , check the
+                    read value so the cycle is formed correctly. *)
+                 || (n.evt.rmw && n.evt.dir = Some R && E.is_com n.next.edge ) )
```
relokin: Checking that `n.evt.dir = Some R` is redundant, you are checking that above. Also why do we need to check that the next edge is a communication edge, didn't we agree that the only way to know that a CAS has been successful you have to check the value of the read?
```suggestion
                 || n.evt.rmw)
```

ShaleXIONG: I see what you mean now. I was thinking we still only focus on CAS for example, on begin or the end of a process. 

Yes you are right here, we need to check the read value to ensure the CAS succeed.

gen/cycle.ml
ShaleXIONG: I have changed back here as you suggested here too.
This means to rule out a particular situation `diyone7 -arch AArch64 PosWW PosWW Coi -oneloc -obs oo`, which previously generates a test as below:
```
AArch64 a
Com=Co
Orig=PosWW PosWW Coi
"PosWW PosWW Coi"
{
 0:X1=x;
}
 P0          ;
 MOV W0,#2   ;
 STR W0,[X1] ;
 MOV W2,#3   ;
 STR W2,[X1] ;
 MOV W3,#1   ;
 STR W3,[X1] ;

exists ([x]=3) (* <- this should be [x]=2 *)
```
Yet this is a very uncommon or event uninteresting litmus tests. So maybe we should just leave it for the current check `[x]=3`. I will need to investigate if there are better way to fix this problem possible another PR?

gen/cycle.ml
ShaleXIONG: Just upstream a new commit as your proposal.

gen/cycle.ml
ShaleXIONG: This will not work for the following non-trivial reasoning. I will try to make it short and precise. It will leads to some allowed tests in the `forbidden.conf`, for example `diyone7 -arch AArch64 PodWW L Rfe DpAddrdW PosWR PosRR Fre`.
```
AArch64 Armv8-ext-forbidden00010398
Generator=diy7 (version 7.58+1)
Com=Rf Fr
Orig=PodWWPL RfeLP DpAddrdW PosWR PosRR Fre
Cycle=Fre PodWWPL RfeLP DpAddrdW PosWR PosRR
Relax=
Safe=Rfe [PosRR,Fre] [PodWW,L] [DpAddrdW,PosWR]
"PodWWPL RfeLP DpAddrdW PosWR PosRR Fre"
{
 0:X1=x; 0:X3=y;
 1:X1=x; 1:X3=y;
}
 P0           | P1                  ;
 MOV W0,#2    | LDR W0,[X3]         ;
 STR W0,[X1]  | EOR W2,W0,W0        ;
 MOV W2,#1    | MOV W4,#1           ;
 STLR W2,[X3] | STR W4,[X1,W2,SXTW] ;
              | LDR W5,[X1]         ;
              | LDR W6,[X1]         ;

exists (1:X0=1 /\ 1:X6=1) (* since the filter you proposing here, we lose a `[x]=2` *)
```
Given your suggestion, we will lose a `[x]=2` check here, which is important for this test. Note that `diy*` generate a written value check to a location based on the following:
(1) There are _at least_ two writes to the same location (excluding the initial value write).
(2) The check value is the last value in the order.
 The write events are collected here in `coherence` function, then the actually more-than-two-write check is in `check_write` in `top_gen.ml`, around line `491`. What you are proposing here reduces the write event in step (1) to only those write adjacent to a communication edge, hence in the test above, no value check is generated to location `x`. Because there is only one write adjacent to a communication edge.

The immediate next thought is, we could weaken the assumption in step (1) to at least _one_ write. However this might also not ideal. Consider the common `SB.litmus`, `diyone7 -arch AArch64 PodWR Fre PodWR Fre`.
```
AArch64 SB
Prefetch=0:x=F,0:y=T,1:y=F,1:x=T
Com=Fr Fr
Orig=PodWR Fre PodWR Fre
"PodWR Fre PodWR Fre"
{
 0:X1=x; 0:X2=y;
 1:X1=x; 1:X2=y;
}
 P0          | P1          ;
 MOV W0,#1   | MOV W0,#1   ;
 STR W0,[X1] | STR W0,[X2] ;
 LDR W3,[X2] | LDR W3,[X1] ;

exists (0:X3=0 /\ 1:X3=0) (* if we weaken two write criteria, we get extra `[x]=1 /\ [y]=1` *)

```
If we weaken the previous assumption on step (1), we will get extra unwanted check `[x]=1 /\ [y]=1`.

The second immediate thought is, maybe we weaken step (1) to one write and also strengthen the write events associated with `Fre` and `Coe`, since the write in `Rfe` is probably less interesting. This will work on the first example I present but the `SB` become the counter example again here.

gen/cycle.ml
```diff
@@ -1165,6 +1165,8 @@ let set_read_pair_v n cell check_value =
   let e = { e with v=v; check_value } in
   n.evt <- e
 
+
+
```
relokin: please remove

gen/cycle.ml
```diff
@@ -1182,7 +1184,14 @@ let do_set_read_v init =
       | Some R ->
         (* If the result of this read need to be checked,
            i.e. generating postcondition *)
-        let check_value = Some (CoSt.get_check_value st) in
+        let check_value =
+          Some ( ( E.is_com n.prev.edge
+                 || E.is_com n.edge
+                 (* For an RMW sequence such as
+                    read --CAS--> write, check the read value
+                    so the CAS succeeds and the cycle is formed correctly. *)
```
relokin: Can we change this to something along the lines of (please fix the indentation): 
```suggestion
                 (* Add a check for the read generated by a CAS to ensure that the CAS was successful *)
```

gen/topUtils.ml
```diff
@@ -109,6 +110,14 @@ module Make : functor (O:Config) -> functor (C:ArchRun.S) ->
     | _ -> StringMap.empty
 
 
+    let check_final_write n =
+      match n.C.C.edge.C.E.edge with
+      | C.E.Store | C.E.Node _ -> false
+      | _ ->
+          C.E.is_com n.C.C.prev.C.C.edge
+          || C.E.is_com n.C.C.edge
+          || n.C.C.evt.C.C.rmw
```
relokin: Why are we adding checks for writes generated by atomics?

gen/common/edge.ml
```diff
@@ -149,6 +149,7 @@ module type S = sig
   val is_ext : edge -> bool
   val is_com : edge -> bool
   val is_fetch : edge -> bool
+  val is_rmw : edge -> bool
```
relokin: This is not used.

gen/top_gen.ml
```diff
@@ -426,25 +426,35 @@ let max_set = IntSet.max_elt
...
-      else cons_one loc v k in
+      match v with
+      | None -> k
+      | Some v ->
+          if (not (StringSet.mem loc atoms) && O.optcond) then k
+          else cons_one loc (Some v) k in
+
+    let last_write_event_adjacent_to_communication vs =
+      List.find_opt (fun (_, _, check) -> check) (List.rev vs)
+      |> Option.map (fun (v, _, _) -> v) in
```
relokin: Can't you do?
```suggestion
      List.find_map (fun (v, _, check) -> if check then Some v else None) (List.rev vs) in
```

relokin (commented):
I think the case of writes needs more work.

Prior to this PR, I get:
```
$> diyone7 -metadata false -arch AArch64  PosWW PodWW Rfe PodRR Fre
AArch64 MP+posW-po+po
{
 0:X1=x; 0:X4=y;
 1:X1=x; 1:X4=y;
}
 P0          | P1          ;
 MOV W0,#1   | LDR W0,[X4] ;
 STR W0,[X1] | LDR W2,[X1] ;
 MOV W2,#2   |             ;
 STR W2,[X1] |             ;
 MOV W3,#1   |             ;
 STR W3,[X4] |             ;

exists ([x]=2 /\ 1:X0=1 /\ 1:X2=0)
```
This post-condition is not ideal and I don't think it should check the value of `[x]`, but it is acceptable.

After this PR:
```
$> diyone7 -metadata false -arch AArch64  PosWW PodWW Rfe PodRR Fre
AArch64 MP+posW-po+po
{
 0:X1=x; 0:X4=y;
 1:X1=x; 1:X4=y;
}
 P0          | P1          ;
 MOV W0,#1   | LDR W0,[X4] ;
 STR W0,[X1] | LDR W2,[X1] ;
 MOV W2,#2   |             ;
 STR W2,[X1] |             ;
 MOV W3,#1   |             ;
 STR W3,[X4] |             ;

exists ([x]=1 /\ 1:X0=1 /\ 1:X2=0)
```
We haven't removed the check on `[x]`. What is worse is that the new check means the post-condition is focusing of the write after write ordering rather than the whole cycle.

gen/cycle.ml
ShaleXIONG: Done !

gen/cycle.ml
ShaleXIONG: Done.

gen/top_gen.ml
ShaleXIONG: Yes. Will update.

gen/common/edge.ml

ShaleXIONG:
> I think the case of writes needs more work.
> 
> Prior to this PR, I get:
> 
> ```
> $> diyone7 -metadata false -arch AArch64  PosWW PodWW Rfe PodRR Fre
> AArch64 MP+posW-po+po
> {
>  0:X1=x; 0:X4=y;
>  1:X1=x; 1:X4=y;
> }
>  P0          | P1          ;
>  MOV W0,#1   | LDR W0,[X4] ;
>  STR W0,[X1] | LDR W2,[X1] ;
>  MOV W2,#2   |             ;
>  STR W2,[X1] |             ;
>  MOV W3,#1   |             ;
>  STR W3,[X4] |             ;
> 
> exists ([x]=2 /\ 1:X0=1 /\ 1:X2=0)
> ```
> 
> This post-condition is not ideal and I don't think it should check the value of `[x]`, but it is acceptable.
> 
> After this PR:
> 
> ```
> $> diyone7 -metadata false -arch AArch64  PosWW PodWW Rfe PodRR Fre
> AArch64 MP+posW-po+po
> {
>  0:X1=x; 0:X4=y;
>  1:X1=x; 1:X4=y;
> }
>  P0          | P1          ;
>  MOV W0,#1   | LDR W0,[X4] ;
>  STR W0,[X1] | LDR W2,[X1] ;
>  MOV W2,#2   |             ;
>  STR W2,[X1] |             ;
>  MOV W3,#1   |             ;
>  STR W3,[X4] |             ;
> 
> exists ([x]=1 /\ 1:X0=1 /\ 1:X2=0)
> ```
> 
> We haven't removed the check on `[x]`. What is worse is that the new check means the post-condition is focusing of the write after write ordering rather than the whole cycle.

I think the best fix here is ignore the write event in `Rfe` and `Fre` edge since the read event already check the value. Let me explore the current `forbidden.conf` whether this change will introduce unexpected allowed tests.

gen/topUtils.ml
ShaleXIONG: In short, the best fix here is to remove both (1) the write check if it is followed by `Rfe` event and (2) remove the universal write check in `rmw` (which is your question here).

let consider `diyone7 -arch AArch64 DpAddrdW Rfe Amo.Cas PodWW Rfe`
```
AArch64 LB+addr+amo.cas-po
Com=Rf Rf
Orig=DpAddrdW Rfe Amo.Cas PodWW Rfe
"DpAddrdW Rfe Amo.Cas PodWW Rfe"
{
 0:X0=x; 0:X4=y;
 1:X0=x; 1:X4=y;
}
 P0                  | P1             ;
 LDR W1,[X0]         | MOV W1,#1      ;
 EOR W2,W1,W1        | MOV W2,#2      ;
 MOV W3,#1           | CAS W1,W2,[X4] ;
 STR W3,[X4,W2,SXTW] | MOV W3,#1      ;
                     | STR W3,[X0]    ;

exists ([y]=2 /\ 0:X1=1 /\ 1:X1=1) (* The current `master`, OR if we keep the universal write check on rmw here *)
exists ([y]=1 /\ 0:X1=1 /\ 1:X1=1) (* If we only remove the universal write check on rmw, which I think is incorrect *)
exists (0:X1=1 /\ 1:X1=1) (* remove write check if it is followed by `Rfe` and remove universal write check in `rmw` *)
```

TiberiuBucur:
Jumping in here as I've spent some time last week familiarising myself with the patch to help with the review process.

After the current changes, the problem highlighted above still persists (at least for me, hopefully I haven't done something wrong):

```
> diyone7 -metadata false -arch AArch64  PosWW PodWW Rfe PodRR Fre
AArch64 MP+posW-po+po
{
 0:X1=x; 0:X4=y;
 1:X1=x; 1:X4=y;
}
 P0          | P1          ;
 MOV W0,#1   | LDR W0,[X4] ;
 STR W0,[X1] | LDR W2,[X1] ;
 MOV W2,#2   |             ;
 STR W2,[X1] |             ;
 MOV W3,#1   |             ;
 STR W3,[X4] |             ;

exists ([x]=1 /\ 1:X0=1 /\ 1:X2=0)
```

To me it seems like the issue is more to do with this write that is checked being on the receiving end of that final `Fre` edge, but also being unobservable in the final state due to the internal visibility requirement:
`irreflexive [Exp & W]; (po & same-loc); [Exp & W]; (ca & int); [Exp & W] as coWW-Exp`
So that one check already invalidates the postcondition, therefore not showcasing the underlying forbidden cycle that is of interest here.

Have I missed anything @ShaleXIONG ? If not, then perhaps another check on top of what you're already doing, to verify the checked write can be last in coherence order for that location, might be enough?

I'm also wondering if it's worth distinguishing further between `PosWW` and `Coi` and only check that location in the case of `Coi`? For refererence, this is what using `Coi` currently generates:
```
>diyone7 -metadata false -arch AArch64  Coi PodWW Rfe PodRR Fre
AArch64 MP+coi-po+po
{
 0:X1=x; 0:X4=y;
 1:X1=x; 1:X4=y;
}
 P0          | P1          ;
 MOV W0,#1   | LDR W0,[X4] ;
 STR W0,[X1] | LDR W2,[X1] ;
 MOV W2,#2   |             ;
 STR W2,[X1] |             ;
 MOV W3,#1   |             ;
 STR W3,[X4] |             ;

exists ([x]=2 /\ 1:X0=1 /\ 1:X2=0)
```

What do you think?

ShaleXIONG:
> Jumping in here as I've spent some time last week familiarising myself with the patch to help with the review process.
> 
> After the current changes, the problem highlighted above still persists (at least for me, hopefully I haven't done something wrong):
> 
> ```
> > diyone7 -metadata false -arch AArch64  PosWW PodWW Rfe PodRR Fre
> AArch64 MP+posW-po+po
> {
>  0:X1=x; 0:X4=y;
>  1:X1=x; 1:X4=y;
> }
>  P0          | P1          ;
>  MOV W0,#1   | LDR W0,[X4] ;
>  STR W0,[X1] | LDR W2,[X1] ;
>  MOV W2,#2   |             ;
>  STR W2,[X1] |             ;
>  MOV W3,#1   |             ;
>  STR W3,[X4] |             ;
> 
> exists ([x]=1 /\ 1:X0=1 /\ 1:X2=0)
> ```
> 
> To me it seems like the issue is more to do with this write that is checked being on the receiving end of that final `Fre` edge, but also being unobservable in the final state due to the internal visibility requirement: `irreflexive [Exp & W]; (po & same-loc); [Exp & W]; (ca & int); [Exp & W] as coWW-Exp` So that one check already invalidates the postcondition, therefore not showcasing the underlying forbidden cycle that is of interest here.
> 
> Have I missed anything @ShaleXIONG ? If not, then perhaps another check on top of what you're already doing, to verify the checked write can be last in coherence order for that location, might be enough?
> 
> I'm also wondering if it's worth distinguishing further between `PosWW` and `Coi` and only check that location in the case of `Coi`? For refererence, this is what using `Coi` currently generates:
> 
> ```
> >diyone7 -metadata false -arch AArch64  Coi PodWW Rfe PodRR Fre
> AArch64 MP+coi-po+po
> {
>  0:X1=x; 0:X4=y;
>  1:X1=x; 1:X4=y;
> }
>  P0          | P1          ;
>  MOV W0,#1   | LDR W0,[X4] ;
>  STR W0,[X1] | LDR W2,[X1] ;
>  MOV W2,#2   |             ;
>  STR W2,[X1] |             ;
>  MOV W3,#1   |             ;
>  STR W3,[X4] |             ;
> 
> exists ([x]=2 /\ 1:X0=1 /\ 1:X2=0)
> ```
> 
> What do you think?

Yes you are right here. I am experiencing on the check on `Fre`. If we remove them all together. I need to ensure all the currently forbidden tests remain forbidden. Regarding the different between `PosWW` and `Coi`. I will check after I find a solution on `Fre` and `Rfe` edges first.

TiberiuBucur:
> Yes you are right here. I am experiencing on the check on Fre. If we remove them all together. I need to ensure all the currently forbidden tests remain forbidden.

Ah, didn't know if the last update was targeting this or not, apologies. :)

ShaleXIONG:
> > Yes you are right here. I am experiencing on the check on Fre. If we remove them all together. I need to ensure all the currently forbidden tests remain forbidden.
> 
> Ah, didn't know if the last update was targeting this or not, apologies. :)

Ah unfortunately no. I am still exploring what I can do. Removing the write check associated to `Fre` gives us some unwanted allowed tests for example:
```
AArch64 Armv8-ext-forbidden00011008
Generator=diy7 (version 7.58+1)
Com=Rf Fr
Orig=DMB.STdWW Rfe DpAddrCseldW ISBsWR Fre
Cycle=Rfe DpAddrCseldW ISBsWR Fre DMB.STdWW
Relax=
Safe=Rfe Fre DMB.STdWW [DpAddrCseldW,ISBsWR]
"DMB.STdWW Rfe DpAddrCseldW ISBsWR Fre"
{
 0:X1=x; 0:X3=y;
 1:X1=x; 1:X3=y;
}
 P0          | P1                  ;
 MOV W0,#2   | LDR W0,[X3]         ;
 STR W0,[X1] | MOV W4,#1           ;
 DMB ST      | CMP W0,#0           ;
 MOV W2,#1   | CSEL W2,W4,W5,EQ    ;
 STR W2,[X3] | AND W2,W2,#2        ;
             | MOV W6,#1           ;
             | STR W6,[X1,W2,SXTW] ;
             | ISB                 ;
             | LDR W7,[X1]         ;

exists (1:X0=1 /\ 1:X7=1)
```

ShaleXIONG:
> > I think the case of writes needs more work.
> > Prior to this PR, I get:
> > ```
> > $> diyone7 -metadata false -arch AArch64  PosWW PodWW Rfe PodRR Fre
> > AArch64 MP+posW-po+po
> > {
> >  0:X1=x; 0:X4=y;
> >  1:X1=x; 1:X4=y;
> > }
> >  P0          | P1          ;
> >  MOV W0,#1   | LDR W0,[X4] ;
> >  STR W0,[X1] | LDR W2,[X1] ;
> >  MOV W2,#2   |             ;
> >  STR W2,[X1] |             ;
> >  MOV W3,#1   |             ;
> >  STR W3,[X4] |             ;
> > 
> > exists ([x]=2 /\ 1:X0=1 /\ 1:X2=0)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This post-condition is not ideal and I don't think it should check the value of `[x]`, but it is acceptable.
> > After this PR:
> > ```
> > $> diyone7 -metadata false -arch AArch64  PosWW PodWW Rfe PodRR Fre
> > AArch64 MP+posW-po+po
> > {
> >  0:X1=x; 0:X4=y;
> >  1:X1=x; 1:X4=y;
> > }
> >  P0          | P1          ;
> >  MOV W0,#1   | LDR W0,[X4] ;
> >  STR W0,[X1] | LDR W2,[X1] ;
> >  MOV W2,#2   |             ;
> >  STR W2,[X1] |             ;
> >  MOV W3,#1   |             ;
> >  STR W3,[X4] |             ;
> > 
> > exists ([x]=1 /\ 1:X0=1 /\ 1:X2=0)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > We haven't removed the check on `[x]`. What is worse is that the new check means the post-condition is focusing of the write after write ordering rather than the whole cycle.
> 
> I think the best fix here is ignore the write event in `Rfe` and `Fre` edge since the read event already check the value. Let me explore the current `forbidden.conf` whether this change will introduce unexpected allowed tests.

To keep record on some of our off-line chat. The challenge here is the outgoing write event in `Fre` edge, or to be more general and precise both `Fre` and `Fri`. In some cases, the write event should not be a possible candidate to generate final written value check such as `[x]=1` in `diyone7 -metadata false -arch AArch64 PosWW PodWW Rfe PodRR Fre`.
```
AArch64 MP+posW-po+po
{
 0:X1=x; 0:X4=y;
 1:X1=x; 1:X4=y;
}
 P0          | P1          ;
 MOV W0,#1   | LDR W0,[X4] ;
 STR W0,[X1] | LDR W2,[X1] ;
 MOV W2,#2   |             ;
 STR W2,[X1] |             ;
 MOV W3,#1   |             ;
 STR W3,[X4] |             ;

exists ([x]=1 /\ 1:X0=1 /\ 1:X2=0)
```
While in other case, the final written value check associated with `Fre` is important to generate expected tests, (often make a test forbidden), such as  `diyone7 -metadata false -arch AArch64 DMB.STdWW Rfe DpAddrCseldW ISBsWR Fre`:
```
AArch64 MP+dmb.st+addrcselW-isbs
{
 0:X1=x; 0:X3=y;
 1:X1=x; 1:X3=y;
}
 P0          | P1                  ;
 MOV W0,#2   | LDR W0,[X3]         ;
 STR W0,[X1] | MOV W4,#1           ;
 DMB ST      | CMP W0,#0           ;
 MOV W2,#1   | CSEL W2,W4,W5,EQ    ;
 STR W2,[X3] | AND W2,W2,#2        ;
             | MOV W6,#1           ;
             | STR W6,[X1,W2,SXTW] ;
             | ISB                 ;
             | LDR W7,[X1]         ;

exists ([x]=2 /\ 1:X0=1 /\ 1:X7=1)
```
where there `[x]=2` is necessary.

Our observation is if the write event, says writing a value `1` to a location `x`, is observable, that is, directly read from or immediate overwrites, this write event associated to the `Fre` should be a candidate to generate a final written value check. To recall, `diy*` will pick the last value amongst all candidates as the final check. The actual implementation is in `check_final_write` and `compute_cos`.

psafont (commented):
The code changes look sensible to me, and match the descriptions in the comments and the commit messages.

I did test the removal of the unnecessary write event post-condition of the example in the comment, which indeed changes between a unpatched and a patched executable.

On the changed litmus tests I did note that some that sometimes observe the postcondition, now do so in more cases. But the observation did not changed from never to sometimes, or sometimes to always.

PR #1847, opened 2026-06-02
[herd,asl] Bundler.py system register processing: add --use-array-variables, --no-register-variable-prefix options

solswords:
Both these options are turned off by default. The --use-array-variables option changes the representation of multislot registers to each use a single array variable rather than one variable per slot; this greatly reduces the size of generated code and is especially critical when processing an AST representation that is many times bigger than the source text. The --no-register-variable-prefix removes the underscore prefix from the generated variable names.

### DIFF
--- a/herd/libdir/asl-pseudocode/Makefile
+++ b/herd/libdir/asl-pseudocode/Makefile
@@ -197,11 +197,17 @@ DO_NOT_GENERATE_VARIABLES_FOR := $\
 
 TO_CLEANUP += all-registers-notice.html all_system_registers.asl $(SYSREGS_NAME).log
 
+# To turn these options on, use on the make command line:
+# USE_ARRAY_VARIABLES=--use-array-variables NO_REGISTER_VARIABLE_PREFIX=--no-register-variable-prefix
+USE_ARRAY_VARIABLES := 
+NO_REGISTER_VARIABLE_PREFIX := 
+
 all_system_registers.asl: $(SYSREGS_NAME)
 	@ cp $^/xhtml/notice.html $(@D)/all-registers-notice.html
 	@ echo "Extracting system registers ASL pseudocode from xml files ..."
 	@ touch $@
 	@ $(BUNDLER_CMD) --make-regs -o $@ --log-file $^.log \
+	        $(NO_REGISTER_VARIABLE_PREFIX) $(USE_ARRAY_VARIABLES) \
 		--allow-optional-fields-for "$(ALLOW_CONDITIONAL_FIELDS_FOR)" \
 		--allow-moving-fields-for "$(ALLOW_MOVING_FIELDS_FOR)" \
 		--allow-included-fields-for "$(ALLOW_INCLUDED_FIELDS_FOR)" \
--- a/herd/libdir/asl-pseudocode/bundler.py
+++ b/herd/libdir/asl-pseudocode/bundler.py
@@ -159,10 +159,12 @@ def header_of_tree(root: Element, o_path: Optional[Path] = None) -> str:
         post_header.append("")
 
     elif root_type == "register":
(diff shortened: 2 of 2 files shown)


herd/libdir/asl-pseudocode/Makefile
```diff
@@ -197,11 +197,19 @@ DO_NOT_GENERATE_VARIABLES_FOR := $\
 
 TO_CLEANUP += all-registers-notice.html all_system_registers.asl $(SYSREGS_NAME).log
 
+# To turn these options on, use on the make command line:
+# USE_ARRAY_VARIABLES=--use-array-variables NO_REGISTER_VARIABLE_PREFIX=--no-register-variable-prefix
+# JOBS_OVERRIDE=-j1
+USE_ARRAY_VARIABLES := 
+NO_REGISTER_VARIABLE_PREFIX := 
+JOBS_OVERRIDE :=
```
HadrienRenaud: Is there a difference between this and `BUNDLER_JOBS`?

HadrienRenaud (approved):
This looks good to me.

Note that the reason why I chose not to implement array backing storage is that this introduces additional synchronisations that I don't think are needed, so use those options at your own risk. For example, if `Xi` are using the same storage array, if the code is `X1() = 3; print(X2())` reading X2 is asl-deps after writing X1.

hrutvik (commented):
I can see that `--use-array-variables` should be non-default, but I wonder: is it worth making `--no-register-variable-prefix` the default and not having this extra flag?

HadrienRenaud:
> I can see that `--use-array-variables` should be non-default, but I wonder: is it worth making `--no-register-variable-prefix` the default and not having this extra flag?

I think this creates type-checking errors: for example, for a register `X`, you'd have:

```
var X: bits(64);

accessor X() <=> bits(64)
  readonly getter
    return X;
  end;

  setter
    X = v;
  end;
end;
```

This shouldn't type-check. I'm fine with letting the option of creating code that does not type-check, but I have more reservations with making it the default.

solswords:
> I think this creates type-checking errors: for example, for a register `X`, you'd have:
> 
> ```
> var X: bits(64);
> 
> accessor X() <=> bits(64)
>   readonly getter
>     return X;
>   end;
> 
>   setter
>     X = v;
>   end;
> end;
> ```
> 
> This shouldn't type-check. I'm fine with letting the option of creating code that does not type-check, but I have more reservations with making it the default.

It does currently type-check. Should it not?

herd/libdir/asl-pseudocode/Makefile
solswords: apparently not; will get rid of it

hrutvik:
If the issue is the clashing name of `X`, I believe it should type-check - functions have a separate namespace to identifiers etc. I believe. But I suppose another question is: why is it useful to remove the prefixed underscores?

solswords:
> If the issue is the clashing name of `X`, I believe it should type-check - functions have a separate namespace to identifiers etc. I believe. But I suppose another question is: why is it useful to remove the prefixed underscores?

In ACL2 proofs we've generally used the convention that the backing variable of a system register is named the same as the accessor name. At first we hand-coded these accessor implementations; now we'd like to use the bundler output, but would prefer not to have to change all the spec functions, theorem statements, etc that use this convention.

HadrienRenaud:
> If the issue is the clashing name of X, I believe it should type-check - functions have a separate namespace to identifiers etc. I believe.

Somehow I had missed this.

I would say that probably it is less a problem now in ASL1, but maybe still good practice to keep different names for variables and functions? So I have a slight preference for not putting it by default if possible, but I wouldn't mind otherwise.

solswords:
Sounds good. Are both of you OK with the patch as-is then?

PR #1851, opened 2026-06-05
[herd] rec as plus

maranget:
An attempt to optimise the cat interpreter: Perform the transformation of recursive definition `let rec r = r1|r2|...|rn|(r;r)` into `let r = (r1|r2|..|rn)+`.

Note: supersedes PR #1719. Preliminary to PR #1856.

### DIFF
--- a/lib/interpreter.ml
+++ b/lib/interpreter.ml
@@ -2644,7 +2644,7 @@ module Make
                       let libfind = O.libfind
                     end) in
                 let (_,_,iprog) =
-                  try P.parse fname
+                  try P.parse ~opt:true fname
                   with Misc.Fatal msg | Misc.UserError msg ->
                     error_not_silent loc "%s" msg  in
                 let stst = st.st in
--- a/herd/AArch64ASLSem.ml
+++ b/herd/AArch64ASLSem.ml
@@ -1602,7 +1602,7 @@ module Make (TopConf : AArch64Sig.Config) (V : Value.AArch64ASL) :
 
         let libfind = ASLConf.libfind
       end) in
-      let fname,m = P.find_parse fname in
+      let fname,m = P.find_parse ~opt:true fname in
       Model.Generic (fname,m)
 
     let is_strict = C.variant Variant.Strict
--- a/herd/getModel.ml
+++ b/herd/getModel.ml
@@ -40,7 +40,7 @@ let parse archcheck arch libfind variant model =
             include LexUtils.Default
             let libfind = libfind
           end) in
-      let fname,ast = P.find_parse fname in
+      let fname,ast = P.find_parse ~opt:true fname in
       Model.Generic (fname,ast)
   | _ -> m in
   if archcheck then check_arch_model arch m
--- a/herd/herd.ml
+++ b/herd/herd.ml
@@ -561,7 +561,7 @@ let model,model_opts = match !model with
 | Some (Model.File fname) ->
     let module P = ParseModel.Make(ParserConfig) in
     begin try
-      let (fname,((b,_,_) as r)) = P.find_parse fname in
+      let (fname,((b,_,_) as r)) = P.find_parse ~opt:true fname in
       Some (Model.Generic (fname,r)),b
     with
     | Misc.Fatal msg -> eprintf "%s: %s\n" prog msg ; exit 2
--- a/lib/ASTUtils.ml
+++ b/lib/ASTUtils.ml
@@ -16,6 +16,8 @@
 
 open AST
 
+let _dbg = false
+
 (* Select "all variables" *)
 let is_var = function
   | Var _ -> true
@@ -202,3 +204,45 @@ let free_body xs e =
     (fun r p0 -> remove_pat0 p0 r)
     (free e)
     xs
+
+
+let rec as_plus_args saw_seq x = function
+  | [] -> [],saw_seq
+  | e::es ->
+      if StringSet.mem x (free e) then
+        match e with
+        | Op (_,Seq,[Var (_,x1);Var (_,x2)]) ->
+            if String.equal x1 x && String.equal x2 x then
+              as_plus_args true x es
+            else raise Exit
+        | _ -> raise Exit
+      else
+        let es,saw_seq = as_plus_args saw_seq x es in
+        e::es,saw_seq
+
+let as_plus = function
+  | [loc_x,(Pvar (Some x) as px),Op (loc_union,Union,es)] ->
+      begin
+        try
+          let es,saw_seq = as_plus_args false x es in
+          if saw_seq then Some ((loc_x,px),(loc_union,es)) else None
+        with Exit -> None
+      end
+  | _ -> None
+
+
+let tr_def d =
+  match d with
+  | Rec (loc,bds,None) ->
+      begin
+        match as_plus bds with
+        | Some ((loc_x,px),(loc_es,es)) ->
+            if _dbg then
+              Printf.eprintf "%a: found implicit transitive closure\n%!"
+                TxtLoc.pp loc ;
+            Let (loc,[loc_x,px,Op1(loc_es,Plus,Op (loc_es,Union,es))])
+        | None -> d
+      end
+  | _ -> d
+
+let rec2plus defs = List.map tr_def defs
--- a/lib/ASTUtils.mli
+++ b/lib/ASTUtils.mli
@@ -30,3 +30,11 @@ val flatten : AST.exp -> AST.exp
 
 (* Get free variables *)
 val free_body : AST.var option list -> AST.exp -> AST.varset
+
+(*
+ * Translate recursive definitions `let rec r = r1 | r2 | rn | r;r`
+ * into non-recursive definition of r as a transitive closure:
+ * `let r = (r1 | r2 | ... | rn)  +
+ *)
+val rec2plus : AST.ins list -> AST.ins list
+
--- a/lib/parseModel.ml
+++ b/lib/parseModel.ml
@@ -63,23 +63,25 @@ module Make(O:Config) = struct
 
   module ML = ModelLexer.Make(O)
 
-  let do_parse fname chan =
+  let do_parse opt fname chan =
     let buff,lexbuf = mk_lexbuf chan in
     lexbuf.lex_curr_p <- { lexbuf.lex_curr_p with pos_fname=fname;};
     let model = GenParserUtils.call_parser "model" lexbuf ML.token
         ModelParser.main in
     let pp = Buffer.contents buff in
     let opts,txt,code = model in
-    (opts,txt,set_text pp code)
+    let code = set_text pp code  in
+    let code = if opt then ASTUtils.rec2plus code else code in
+    (opts,txt,code)
 
-  let find_parse fname =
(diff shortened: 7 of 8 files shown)


maranget:
Wallclock times for `make test J=8` on a MacBook Pro.

| PR   | PR-1 | master
|------|-------|-------
|52.45 | 48.17 | 49.65
|50.07 | 48.32 | 47.62
|50.04 | 48.85 | 48.59
|50.36 | 47.37 | 51.63
|52.75 | 47.35 | 49.37

PR is this PR (three commits) and PR-1 conists in the first two commits of this PR, namely some recursive definitions are replaced by transitive closures. The gain is modest, but is potentially important on some examples while the cost (a very simple analysis performed by commit 5603f483d17e0f31ec60dcf257446a4d00b50095) remains low.

lib/interpreter.ml
```diff
@@ -799,22 +801,28 @@ module Make
...
-    let test2pred env t e v = match t,v with
-    | (Acyclic,(Rel r|TransRel r))
-    | (Irreflexive,TransRel r)  -> E.EventRel.is_acyclic r
+    let test2pred loc env t e v = match t,v with
+    | Acyclic,Rel r ->  E.EventRel.is_acyclic r
+    | Acyclic,TransRel (r,s,_) -> E.EventRel.(is_acyclic @@ union r s)
+    | Irreflexive,TransRel (r,s,_)  ->
+        if _dbg then
+          Printf.eprintf "%a: irreflexive -> transitive!\n%!" TxtLoc.pp loc ;
+        E.EventRel.(is_acyclic @@ union r s)
```
HadrienRenaud: Why not check if the lazy is done?

lib/interpreter.ml
```diff
@@ -1338,25 +1344,28 @@ module Make
...
-            begin match eval accept_implicit env e with
+            begin match eval env e with
             | V.Empty -> V.Empty
             | Unv -> Unv
             | Rel r -> Rel (E.EventRel.inverse r)
-            | TransRel r -> TransRel (E.EventRel.inverse r)
+            | TransRel (r,s,th) ->
+                let open E.EventRel in
+                let r = inverse r and s = inverse s  in
+                TransRel (r,s,lazy (inverse @@ Lazy.force th))
```
HadrienRenaud: Maybe?

```suggestion
                TransRel (r,s, Lazy.map inverse th))
```

lib/interpreter.ml
```diff
@@ -1430,11 +1470,11 @@ module Make
                     (* Notice:
                      *   Left-to-right order is significant for efficiency.
                      * Namely, due to the asymmetry of the relation
-                     * data structure, performing domain restriction first
+                     * data structure, perfo     rming domain restriction first
```
HadrienRenaud: Maybe this diff should be removed?

lib/interpreter.ml
```diff
@@ -1459,21 +1499,23 @@ module Make
...
+                -> assert false
             | Rel r1,Rel r2 -> Rel (E.EventRel.inter r1 r2)
-            | TransRel r1, TransRel r2 ->
-                let r1 = E.EventRel.transitive_closure r1
-                and r2 = E.EventRel.transitive_closure r2 in
-                Rel (E.EventRel.inter r1 r2)
-            | (Rel r, TransRel tr)
-            | (TransRel tr, Rel r) ->
+            | TransRel (_,_,r1),TransRel (_,_,r2) ->
+                Rel (E.EventRel.inter (Lazy.force r1) (Lazy.force r2))
```
HadrienRenaud: Could we restrict the transitive closures here to the domain of each? I'm not really sure because it would loose some sharing by recomputing a transitive closure

```suggestion
            | TransRel (t1, s1, r1) as v1, TransRel (t2, s2, r2) as v2 ->
                let w1 =
                  if Lazy.is_val r1 then Lazy.force r1 else
                    E.EventRel.union s1 @@
                    E.EventRel.transitive_closure_from (domain v2) t1
                and w2 =
                  if Lazy.is_val r2 then Lazy.force r2 else
                    E.EventRel.union s2 @@
                    E.EventRel.transitive_closure_from (domain v1) t2
                in
                Rel (E.EventRel.inter w1 w2)
```

lib/interpreter.ml
```diff
@@ -1459,21 +1499,23 @@ module Make
...
+            | (TransRel (t2,s2,_), Rel r1) ->
                 let r =
-                  E.EventRel.restrict_rel
-                    (fun e1 e2 -> E.EventRel.exists_path (e1,e2) tr)
-                    r in
+                  let open E.EventRel in
+                  restrict_rel
+                    (fun e1 e2 ->
+                       let p = e1,e2 in
+                       exists_path p t2 || mem p s2)
```
HadrienRenaud: Short path before?

```suggestion
                       mem p s2 || exists_path p t2
```

lib/interpreter.ml
```diff
@@ -1684,44 +1728,46 @@ module Make
...
+        | Rel m1,TransRel (r2,s2,_) ->
             Rel
               E.EventRel.
                 (restrict_rel
-                   (fun e1 e2 -> not (exists_path (e1,e2) m2)) m1)
-        | TransRel r1,Rel r2 ->
-            Rel E.EventRel.(diff (transitive_closure r1) r2)
+                   (fun e1 e2 ->
+                      let p = e1,e2 in
+                      not (exists_path (e1,e2) r2||mem p s2)) m1)
```
HadrienRenaud:
```suggestion
                      not (mem p s2 || exists_path p r2)) m1)
```

lib/interpreter.ml
```diff
@@ -692,7 +694,7 @@ module Make
 
     let as_rel ks = function
       | Rel r -> r
-      | TransRel tr -> E.EventRel.transitive_closure tr
+      | TransRel (_,_,r) -> Lazy.force r
```
HadrienRenaud: Maybe less explicit but maybe more elegant? I don't know

```suggestion
      | TransRel (_,_, lazy r) -> r
```

(there are a lot of those in this file, I'm not going to comment them all but you get the idea)

lib/interpreter.ml
```diff
@@ -1381,32 +1390,63 @@ module Make
...
+            begin
+              try
+                let vs = (* Union being AC, order is irrelevant *)
+                  List.fold_left
+                    (fun vs e ->
+                       match eval_loc env e with
+                       | (_,V.Empty) -> vs
+                       | (_,Unv) -> raise Exit
+                       | (loc,v) -> (loc,tag2set v)::vs)
+                    [] es in
```
HadrienRenaud: Slightly simpler? Also preserve order (which is irrelevant as you mentioned)

```suggestion
                  List.filter_map
                    (fun e ->
                       match eval_loc env e with
                       | (_,V.Empty) -> None
                       | (_,Unv) -> raise Exit
                       | (loc,v) -> Some (loc,tag2set v))
                    es in
```

lib/interpreter.ml
```diff
@@ -1381,32 +1390,63 @@ module Make
...
+                        let rs,trs =
+                          List.fold_left
+                            (fun (rs,trs as k) v ->
+                               match v with
+                               | TransRel (t,s,_) ->
+                                   let ts,ss = trs in
+                                   rs,(t::ts,s::ss)
+                               | Rel r ->  r::rs,trs
+                               | Empty -> k
+                               | Unv -> raise Exit
```
HadrienRenaud: Those 2 cases are not possible right? `Empty` is filtered out above and `Unv` raised `Exit` above

lib/interpreter.ml
```diff
@@ -1459,21 +1499,23 @@ module Make
...
            eval_diff env loc e1 e2
         | Op (loc,Inter,[e1;e2;]) -> (* Binary notation kept in parser *)
-            let loc1,v1 = eval_loc true env e1
-            and loc2,v2 = eval_loc true env e2 in
+            let loc1,v1 = eval_loc env e1
+            and loc2,v2 = eval_loc env e2 in
             begin match tag2set v1,tag2set v2 with
-            | (V.Tag _,_)|(_,V.Tag _) -> assert false
+              | (V.Tag _,_)|(_,V.Tag _)
+                -> assert false
```
HadrienRenaud: Undo this change?

HadrienRenaud (commented):
Thanks @maranget for this.

I approve gladly the first 2 commits.

For the last commit, even though the performance improvement are not as expected, I think it is cleaner than what we were doing before and the removal of `accept_implicit` in the interpreter is maybe worth the very slight performance drawback.

Maybe it would be interesting to see if there are specific tests that are driving this performance drawback? Maybe there is a size of the graph at which it becomes interesting and so we could introduce a heuristic to activate or not this optimisation

lib/interpreter.ml
maranget: Good point, I fix.

lib/interpreter.ml
maranget: Ok, available since 4.13.

lib/interpreter.ml
maranget: Indeed, thanks.

lib/interpreter.ml
maranget: I'd rather avoid recomputations.

lib/interpreter.ml
maranget: Ok

lib/interpreter.ml
maranget: I'll keep those Lazy.force explicit calls. I am more comfortable with them than with pattern matching, as the place where thunk reduction is performed is clearer.

lib/interpreter.ml
maranget: Not much simpler and allocate extra (transient) memory.  I stick to the orignal code.

lib/interpreter.ml
maranget: Nice catch. I fix.

lib/interpreter.ml
HadrienRenaud: Makes sense ok

lib/interpreter.ml
HadrienRenaud: Can you explain a bit what the allocated transient memory is? I'm not familiar with this, I thought that with tail-mod-cons it was not a problem that the function was not tail recursive

lib/interpreter.ml
maranget: Ok

lib/interpreter.ml
maranget: Transient memory is the Some consructor.

maranget:
I will now move the last commit to a new PR...

lib/interpreter.ml
HadrienRenaud: Ah indeed, ok

maranget:
The PR is now reduced to its first two commits. Wallclock measures  on a 12 core workstation confirm the slight advantage of this PR over master.
```
make test  J=12
```

|   PR  | master |
|-------|--------|
| 69.93 | 71.31  |
| 69.88 | 70.59  |
| 68.64 | 70.97  |

HadrienRenaud (approved):
Looks good to me, thanks @maranget

maranget:
Merged thanks @HadrienRenaud.

PR #1853, opened 2026-06-10
[gen] Simplify macro and edge parsing lookup logic

ShaleXIONG:
This PR cleans up the first step of macro handling in the generator.

We added many tests for macros and wildcards, which safeguarded the following main changes:
- Remove the old `PPO` constructor but now represent `type relax` directly as an `edge list`. The `PPO` become a macro unfold.
- Parse edge names by longest prefix, then parse trailing annotations, for example `PosWRLA` will be parsed in three steps as `PosWR` `L` and `A`. This massively reduce the size of the lookup table for parsing. Also the old lookup table is split into two tables for `edge` and `annotation` respectively.
- Keep legacy macros such as `AllRW` and alias such as `CtrldW` available through the macro table.

This keeps the macro parsing/unfolding behavior explicit and table-driven, while preparing the later wildcard and alias migration work for a follow-up branch.

### DIFF
--- a/gen/alt.ml
+++ b/gen/alt.ml
@@ -204,7 +204,7 @@ struct
           let r =
             not
               (List.exists
-                 (fun es -> C.R.Set.mem (C.R.ERS es) safes)
+                 (fun es -> C.R.Set.mem es safes)
                  cs) in
           r
       | _,_ -> true
@@ -307,8 +307,11 @@ module Make(C:Builder.S)
              |> String.concat list_list_sep )
         |> String.concat list_sep
 
-    let edges_of_relax_list rs =
-      List.map (fun r -> (r, edges_of r)) rs
+    (* Pair each relax with the working edge list used by the generator.
+       The first component preserves the original relax for reporting/filtering;
+       the second component may be modified while building candidate cycles. *)
+    let relaxs_with_work_edges rs =
+      List.map (fun r -> (r, r)) rs
 
 (* Functional for recursive call of generators *)
 
@@ -360,7 +363,7 @@ module Make(C:Builder.S)
           O.prefix
       end
 
-    let prefixes = List.map edges_of_relax_list O.prefix
+    let prefixes = List.map relaxs_with_work_edges O.prefix
 
     let rec mk_can_prefix = function
       | [] -> (fun _ _ -> true)
@@ -429,7 +432,7 @@ module Make(C:Builder.S)
           let d2 =
             List.fold_right
               (fun (r,_) k -> match r with
-              | ERS [{edge=Po (sd,e1,e2); _}] -> SdDir2Set.add (sd,e1,e2) k
+              | [{edge=Po (sd,e1,e2); _}] -> SdDir2Set.add (sd,e1,e2) k
               | _ -> k)
               rs SdDir2Set.empty in
           if dbg then
@@ -447,8 +450,8 @@ module Make(C:Builder.S)
 
     let zyva prefix aset relax safe reject n f =
 (*      let safes = C.R.Set.of_list safe in *)
-      let relax = edges_of_relax_list relax in
-      let safe = edges_of_relax_list safe in
+      let relax = relaxs_with_work_edges relax in
+      let safe = relaxs_with_work_edges safe in
       let po_safe = extract_po safe in
 
       (* ********************************** *)
@@ -682,7 +685,6 @@ module Make(C:Builder.S)
       let sset = C.R.Set.of_list safe in
       let rset = C.R.Set.of_list relax in
       let aset = C.R.Set.union sset rset in
-      let rej = List.map (fun a -> edges_of a) rej in
       D.all
         ~check:(last_minute rej)
         (fun f ->
@@ -734,7 +736,7 @@ module Make(C:Builder.S)
       fold_sd_dir2 (fun sd d1 d2 -> C.A.fold_cumul_fences (fun fe -> f fe sd d1 d2))
     let fold_cum f =  fold_cumul_fences f
 
-    let er e = ERS [plain_edge e]
+    let er e = [plain_edge e]
     let safe =
       let k = [] in
       let k = fold_ie (fun ie k -> er (Ws ie)::er (Fr ie)::k) k in k
@@ -767,6 +769,6 @@ module Make(C:Builder.S)
     let filter_check ~relax ~safe lhs rhs =
       let safe,_,_ = parse_input ~relax ~safe ~reject:[] in
       let safe_set = C.R.Set.of_list safe in
-      let po_safe = edges_of_relax_list safe |> extract_po in
+      let po_safe = relaxs_with_work_edges safe |> extract_po in
       FilterImpl.can_precede safe_set po_safe lhs rhs
   end
--- a/gen/common/edge.ml
+++ b/gen/common/edge.ml
@@ -77,11 +77,7 @@ module type S = sig
   val plain_edge : tedge -> edge
 
   val fold_atomo : (atom option -> 'a -> 'a) -> 'a -> 'a
-  val fold_mixed : (atom option -> 'a -> 'a) -> 'a -> 'a
   val fold_atomo_list : atom option list -> (atom option -> 'a -> 'a) -> 'a -> 'a
-
-  val fold_edges : (edge -> 'a -> 'a) -> 'a -> 'a
-  val iter_edges : (edge -> unit) -> unit
 
   val pp_tedge : tedge -> string
   val pp_atom_option : atom option -> string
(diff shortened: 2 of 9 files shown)


gen/tests/diy-basic-check.t
TiberiuBucur, line 1: I had a look at _some_ of these tests and understood them, but I cannot tell if _all_ of them are "architecturally correct" in some sense. My understanding is that these are regression tests verifying that the pre-existing behaviour is preserved.

gen/alt.ml
```diff
@@ -308,7 +308,7 @@ module Make(C:Builder.S)
         |> String.concat list_sep
 
     let edges_of_relax_list rs =
-      List.map (fun r -> (r, edges_of r)) rs
+      List.map (fun r -> (r, r)) rs
```
TiberiuBucur: The name and purpose of this function now after the relaxation is purely a list of edges are a bit confusing. I'm wondering if it can be renamed or even removed, with the callers adapting their code to fit their purposes.

gen/relax.ml
```diff
@@ -325,6 +275,98 @@ and type edge = E.edge
...
+              "someWR", someW Diff R;
+              "someWW", someW Diff W;
+            ]
+
+        let expand_relaxs rs =
+          let expand_relax r = E.expand_edges r Misc.cons in
+          List.fold_right expand_relax rs []
+
+        let relax_to_sequence relax = match relax with
+        | [] -> assert false
```
TiberiuBucur: May I suggest a `Warn.fatal` with an informative error message here? I believe I saw `assert false` in other places too, but I forgot where exactly.

gen/relax.ml
```diff
@@ -325,6 +275,98 @@ and type edge = E.edge
...
+        | [edge] -> Ast.One edge
+        | edges -> Ast.Seq (List.map (fun edge -> Ast.One edge) edges)
+
+        let relax_list_to_choice relax_list =
+          let ast_relax_list =
+            List.map relax_to_sequence relax_list in
+          match ast_relax_list with
+              | [] -> assert false
+              | [relax] -> relax
+              | relax_list -> Ast.Choice relax_list
```
TiberiuBucur: As far as I can see indentation is normally aligned with the `match`.

```suggestion
          match ast_relax_list with
          | [] -> assert false
          | [relax] -> relax
          | relax_list -> Ast.Choice relax_list
```

gen/common/edge.ml
```diff
@@ -333,25 +310,21 @@ and module RMW = A.RMW = struct
...
+      (pp_tedge e.edge) (pp_atom_option e.a1) (pp_atom_option e.a2)
 
-  let do_pp_edge compat pp_atom_functor e =
-    let annotation = pp_atom_functor e.edge e.a1 e.a2 in
+  let pp_edge_compat compat e =
     let edge = match e.edge with
-    | Id -> ""
-    | _ -> do_pp_tedge compat e.edge in
+      | Id -> ""
+      | _ -> pp_tedge_compat compat e.edge in
```
TiberiuBucur:
```suggestion
    | Id -> ""
    | _ -> pp_tedge_compat compat e.edge in
```

gen/common/edge.ml
```diff
@@ -661,9 +567,71 @@ let fold_tedges f r =
...
+     - For a composite input such as `PosWWLA`, return [`L`; `PosWW`; `A`].
+     - Otherwise, return `None`. *)
+  let internal_parse_edge input =
+    match lookup_edge_prefix input with
+    | None -> parse_annotation input
+    | Some (edge, "") -> Some [edge]
+    (* If there is a suffix, it must parse as two annotations. *)
+    | Some (edge, suffix) ->
+        match parse_annotation_suffix suffix with
+        | Some annotations -> Some (annotations @ [edge])
```
TiberiuBucur: Perhaps you can build the edge right here, instead of having a separate `parse_edge` function below only for this case. The first two cases in the pattern match already return one edge, only this one returns a list of three.

gen/common/edge.ml
```diff
@@ -661,9 +567,71 @@ let fold_tedges f r =
...
+    in
+    do_rec (String.length string)
+
+  let lookup_edge_prefix string =
+    fold_prefixes (fun prefix -> Hashtbl.find_opt t prefix) string
+
+  let lookup_atom_prefix string =
+    fold_prefixes (fun prefix -> Hashtbl.find_opt ta prefix) string
+
+  let annotation_edge a = { a1=a; a2=a; edge=Id }
```
TiberiuBucur, line 592: Not critical for this PR, but I'm wondering if there is value in a bigger refactoring that would split annotations into their own module. Right now I feel like we have "true" edges and annotations represented as "fake" edges. What do you think?

gen/tests/diy-basic-check.t
ShaleXIONG: Yes you are right here. I rely on those tests to ensure any change preserve the previous behaviours.

gen/alt.ml
ShaleXIONG: Let me change the name, it was use to keep the original input, the first protection, and updated version, the second projection. The latter will be modified during `diy7` searching algorithm.

gen/relax.ml
ShaleXIONG: That is a better idea.

gen/relax.ml
ShaleXIONG: Done

gen/common/edge.ml

gen/common/edge.ml
ShaleXIONG: That is a really good comment. I will agree with you, it will be much better we have separate `edge` and `annotation` definitions and then meets at the definition of `relax`. I think it will be large and fundamental change. I will investigate after this PR.

gen/common/edge.ml
ShaleXIONG: Let me update and simplify here.

gen/relax.ml
```diff
@@ -325,6 +275,98 @@ and type edge = E.edge
...
+              "someWR", someW Diff R;
+              "someWW", someW Diff W;
+            ]
+
+        let expand_relaxs rs =
+          let expand_relax r = E.expand_edges r Misc.cons in
+          List.fold_right expand_relax rs []
+
+        let relax_to_sequence relax = match relax with
+        | [] -> Warn.fatal "relax should not be empty."
```
TiberiuBucur: As a user, I would think I did something wrong if I saw this message after I passed no relaxations, whereas this is an internal bug. Do you think we could find a clearer message? Maybe something like "Default value for relax not propagated".. Or anything you might find better.

gen/common/edge.ml
```diff
@@ -661,9 +567,64 @@ let fold_tedges f r =
...
+    | None ->
+        match lookup_edge_prefix input with
+        | None -> Warn.fatal "Bad edge: %s" input
+        | Some (edge, "") -> edge
+        (* If there is a suffix, it must parse as two annotations. *)
+        | Some (edge, suffix) ->
+            begin match parse_edge_annotations suffix with
+            | Some (a1, a2) -> {edge with a1; a2}
+            | _ -> Warn.fatal "Bad edge: %s" input
+            end
```
TiberiuBucur: This current version of the function performs the check in the opposite order compared to its previous version. Before, we prioritised edges, and if none was found, we would check for annotations. This one does the opposite.

Is there a case where an input can be both an annotation (by itself) and an edge / edge with an annotation? Is such a case relevant?

ShaleXIONG: Annotation and edge are distinct. I have a quick thought in the sense that, annotation are much smaller table in the scale of 10-20 elements while edge table contains 100+ elements. Searching the annotation first leads to slightly better performance.

gen/relax.ml
ShaleXIONG: I think `fatal` is often to an unreachable code location in general, while `user_error` is related to some invalid input. In the case here, I expected the first `[]` is never reachable from the code. If this answer your question? Let me write a better message here too.

TiberiuBucur: Precisely because the code is supposed to be unreachable, my opinion was that the error message should convey that intention. Do you agree?

gen/common/edge.ml
TiberiuBucur: Codex seems to suggest a clash for PPC between R as an annotation and the legacy name for the Read "pseudo-edge". Perhaps there is no current regression test covering that. If you think it's not important to treat this as an edge vs an annotation, then I'm fine with this change.

gen/relax.ml
ShaleXIONG: Agree here. I change to "Relax is parsed incorrectly.".

gen/common/edge.ml
ShaleXIONG: Although PPC is de facto deprecated, that is a good comment. I swap back to priority parse edge first. Thanks for the catch.

TiberiuBucur (approved):

PR #1854, opened 2026-06-10
ACL2 ASL updates

solswords:
Miscellaneous improvements to interpreter and reasoning support:
 - Added type normalization function to detect functionally equivalent types
 - Add reasoning support for determining when code can't produce trace output/aborts
 - Divide stdlib proofs into separate partial correctness and termination components


slobodova (approved):
The changes affect only herdtools7/acl2 .

PR #1855, opened 2026-06-11
[herd] Petty optimisation of the Cat interpreter

maranget:
Some optimisations that may help and never harm.

### DIFF
--- a/lib/innerRel.ml
+++ b/lib/innerRel.ml
@@ -60,6 +60,9 @@ module type S =  sig
 (* Transitive closure *)
   val transitive_closure : t -> t
 
+  (* [restrict_domain_transitive_closure s r] returns [s];r+ *)
+  val restrict_domain_transitive_closure : Elts.t -> t -> t
+
 (* Direct cycles *)
   val is_reflexive : t -> bool
   val is_irreflexive : t -> bool
@@ -239,7 +242,13 @@ module Make(O:MySet.OrderedType) : S
           (res,{id=0; visit=M.empty; stack=[];} ) in
       res
 
-    (* Graph is given as set of notes + relation *)
+(*
+ * Graph is given as a set of nodes [nodes] and a relation [m].
+ * Notice that for a node [n] in [nodes] but not in [m],
+ * [dfs ....] is called on [n] and this is important,
+ * for instance for the equation solving as performed
+ * by [scc_kont ...] below.
+ *)
     let scan_nodes_rel kont kont_scc res nodes m =
       let dfs = dfs kont kont_scc m in
       let (res,_) =
@@ -249,6 +258,27 @@ module Make(O:MySet.OrderedType) : S
           nodes
       in
       res
+
+(*
+ * Scan the graph [m] in topological order,
+ * computing scc. Excluding paths that do not start from
+ * the set [nodes]. By contrast with [scan_nodes_rel] above,
+ * given [n] in [nodes] but not in [m],
+ * [dfs] will *not* be called on [n].
+ *)
+    let scan_restrict_domain_rel kont kont_scc res nodes m =
+      let dfs = dfs kont kont_scc m in
+      let (res,_) =
+        M.fold
+          (fun n _ r ->
+            if Elts.mem n nodes then
+              let _,r = dfs n r in r
+            else r)
+          m
+          (res,{id=0; visit=M.empty; stack=[];})
+      in
+      res
+
   end
 
 
@@ -363,22 +393,25 @@ module Make(O:MySet.OrderedType) : S
    *        1995.
    *)
 
-  let transitive_closure (m:t) =
-    let kont n m res =
+  let tr_edge n m res =
       let vr = succs res m |> Elts.add m in
       add_set n vr res
 
-    and kont_scc n scc res =
-      match scc with
-      | [] -> assert false
-      | [_] -> res
-      | ms ->
-          let vn = try M.find n res with Not_found -> assert false in
-          if Elts.is_empty vn then res
-          else List.fold_left (fun res m -> add_set m vn res) res ms
-    in
+  and tr_scc n scc res =
+    match scc with
+    | [] -> assert false
+    | [_] -> res
+    | ms ->
+       let vn = try M.find n res with Not_found -> assert false in
+       if Elts.is_empty vn then res
+       else List.fold_left (fun res m -> add_set m vn res) res ms
 
-    FullSCC.scan_map kont kont_scc M.empty m
+  let transitive_closure (m:t) =
+    FullSCC.scan_map tr_edge tr_scc M.empty m
+
+  let restrict_domain_transitive_closure vs m =
+    FullSCC.scan_restrict_domain_rel tr_edge tr_scc M.empty vs m
+    |>  restrict_domain (fun v -> Elts.mem v vs)
 
   (* Reflexivity check *)
 
@@ -582,7 +615,6 @@ module Make(O:MySet.OrderedType) : S
       let kont _ _ r = r
       and kont_scc _ ns r = kont ns r in
       FullSCC.scan_nodes_rel kont kont_scc res nodes edges
-
 
     (* Check hierarchy *)
 
--- a/lib/interpreter.ml
+++ b/lib/interpreter.ml
@@ -699,6 +699,15 @@ module Make
           eprintf "this is not a relation: '%s'" (pp_val v) ;
           assert false
 
+    let as_notrans ks = function
+      | Rel r -> r
+      | TransRel tr -> tr
+      | Empty -> E.EventRel.empty
+      | Unv -> Lazy.force ks.unv
+      | v ->
+          eprintf "this is not a relation: '%s'" (pp_val v) ;
+          assert false
+
     let as_classrel = function
       | ClassRel r -> r
       | Empty -> ClassRel.empty
@@ -1218,7 +1227,9 @@ module Make
 (* Interpreter *)
 (***************)
     type rel =
-      | Rid of (E.event -> bool)
+      | Rid of E.EventSet.t
+      | Runv
+      | Rtrans of E.EventRel.t
       | Revent of E.EventRel.t
       | Rclass of ClassRel.t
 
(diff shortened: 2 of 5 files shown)


lib/innerRel.ml
```diff
@@ -249,6 +252,20 @@ module Make(O:MySet.OrderedType) : S
...
+
+    let scan_set_rel kont kont_scc res nodes m =
+      let dfs = dfs kont kont_scc m in
+      let (res,_) =
+        M.fold
+          (fun n _ r ->
+            if Elts.mem n nodes then
+              let _,r = dfs n r in r
+            else r)
+          m
```
HadrienRenaud, line 277:
```suggestion
        Elts.fold (fun n r -> dfs n r |> snd) m
```

But is it really different from `scan_nodes_rel`?

lib/innerRel.ml
```diff
@@ -363,23 +380,26 @@ module Make(O:MySet.OrderedType) : S
...
+       let vn = try M.find n res with Not_found -> assert false in
+       if Elts.is_empty vn then res
+       else List.fold_left (fun res m -> add_set m vn res) res ms
 
+  let transitive_closure (m:t) =
     FullSCC.scan_map kont kont_scc M.empty m
 
+  let restrict_domain_transitive_closure vs m =
+    FullSCC.scan_set_rel kont kont_scc M.empty vs m
+    |>  restrict_domain (fun v -> Elts.mem v vs)
```
HadrienRenaud, line 414: It's a shame that we can't recurse on both the set and the map at the same time

HadrienRenaud (commented):
LGTM, thanks Luc

HadrienRenaud (approved):

lib/innerRel.ml
maranget: I guess you mean:
```OCaml
Elts.fold (fun n r -> dfs n r |> snd) nodes
```
This is slightly different from `scan_nodes_rel` as regards the type of `nodes` (set vs. list). 

I choosed to iterate over the map and not over the set because I worried about the cost of `dfs n r` in the case where there is no binding for `n` in the map `m`. However if `nodes` is small it would be better to iterate over the set. I have no good solution here.

lib/innerRel.ml
maranget: I do not undestand what you mean here: perhaps computing some kind of intersection of set and map by merging trees?

maranget: My best approximation of the idea:
```OCaml
  let scan_set_rel kont kont_scc res nodes m =
      let dom = Elts.inter nodes (domain m) in
      let dfs = dfs kont kont_scc m in
      let (res,_) =
        Elts.fold dfs
          dom (res,{id=0; visit=M.empty; stack=[];})
      in
      res
```

lib/innerRel.ml
HadrienRenaud: Oh I see what you mean, I had forgotten the case where `nodes ⊄ domain(m)`.
I think probably your solution is better than mine.

In an ideal world it'd be nice to iterate over both at the same time, for example what [baby](https://cambium.inria.fr/~fpottier/baby/doc/baby/) does to compute the domain of a map.

lib/innerRel.ml
HadrienRenaud: > perhaps computing some kind of intersection of set and map by merging trees?

Yes but it's not something we can do easily, at least without re-implementing sets and maps... I was thinking of something like what [baby](https://github.com/fpottier/baby/blob/main/src/W.cppo.ml#L51) does.

I am not sure this has better performance than what we're doing already.

lib/interpreter.ml
```diff
@@ -1233,6 +1233,10 @@ module Make
       | Revent of E.EventRel.t
       | Rclass of ClassRel.t
 
+    let check_empty_rel is_empty v= if is_empty v then raise Exit else v
+    let check_empty_id s =  check_empty_rel E.EventSet.is_empty s
+    and check_empty_rel s = check_empty_rel E.EventRel.is_empty s
```
HadrienRenaud: Do you think we can name it `exit_if_empty` or something like this? I find the name quite confusing

lib/interpreter.ml
```diff
@@ -1831,6 +1783,86 @@ module Make
...
+            | Runv -> seq_args_id env s es
+            | Rid t ->
+                seq_args_id env
+                  (E.EventSet.inter s t |> check_empty_id) es
+            | Revent r ->
+                seq_args_rel
+                  env
+                  (E.EventRel.restrict_domain
+                     (fun e -> E.EventSet.mem e s) r |> check_empty_rel)
+                  es
```
HadrienRenaud, line 1822: I find this far more readable:

```suggestion
                let r' =
                  E.EventRel.restrict_domain
                    (fun e -> E.EventSet.mem e s)
                    r
                  |> check_empty_rel in
                seq_args_rel env r' es
```

lib/interpreter.ml
```diff
@@ -1831,6 +1783,86 @@ module Make
...
+        | e::es ->
+            match eval_rels env e with
+            | Runv ->  seq_args_rel env r es
+            | Rid s ->
+                begin
+                  match seq_args_id env (check_empty_id s) es with
+                  | Runv -> Revent r
+                  | Rid s ->
+                      Revent
+                        (E.EventRel.restrict_codomain_to_set s r)
```
HadrienRenaud, line 1845: No `check_if_empty` here?

Maybe it'd be interesting to move the `check_empty` tests at the top of the functions so that the checks are always performed. For example for `seq_args_rel`:
```ocaml
     and seq_args_rel env r es =
        if E.EventRel.is_empty r then raise Exit;
        match es with
        | [] -> Revent r
        | e::es ->
```

lib/innerRel.ml
maranget: I keep the initial code, changing the name of the function from `scan_set_rel` to `scan_restrict_domain_rel`. 

There is an important semantical difference between `scan_nodes_rel .... kont_scc nodes r` and  `scan_restrict_domain_rel ... scc_kont set r`. Consier some node `n` that is part of the set (either `nodes` or `set`) but is not part of any edge of `r`. In the first case, `n` is part of the graph and `scc_kont ` must be called called on the trivial strongly connected component `[n]`. In the second case, we do not care, as the intent is tu exclude the paths that start from `n` ,

lib/interpreter.ml
maranget: Good catch, your name is much better.

lib/interpreter.ml
maranget: I had the same idea but noticed that, in one occasion, the second argument `r` does not change:
```OCaml
 and seq_args_rel env r = function
        | [] -> Revent r
        | e::es ->
            match eval_rels env e with
            | Runv ->  seq_args_rel env r es
...
```
However it is likely that this case does not occur very often of at all. So, I tend to adopt your solution,

lib/interpreter.ml
maranget: No more relevant, see below.

maranget:
Hi @HadrienRenaud, I'd like to merge soon. May I have your opinion on this last version?

lib/innerRel.ml
```diff
@@ -239,7 +242,13 @@ module Make(O:MySet.OrderedType) : S
           (res,{id=0; visit=M.empty; stack=[];} ) in
       res
 
-    (* Graph is given as set of notes + relation *)
+(*
+ * Graph is given as a set of nodes [nodes] and a relation [r].
+ * Notice for a node [n] in [nodes] but not in [r],
+ * [dfs ....] is called on [n] and this is important,
+ * for instance for the equation solving as performed
+ * by [scc_kont ...] below.
```
HadrienRenaud:
```suggestion
 * Graph is given as a set of nodes [nodes] and a relation [m].
 * Notice for a node [n] in [nodes] but not in [m],
 * [dfs ....] is called on [n] and this is important,
 * for instance for the equation solving as performed
 * by [scc_kont ...] below.
```

lib/innerRel.ml
```diff
@@ -249,6 +258,26 @@ module Make(O:MySet.OrderedType) : S
           nodes
       in
       res
+
+(*
+ * Scan the graph `r` in topological order,
+ * computing scc. Excluding paths that start from
+ * the set [nodes]. By contrast, for [n] in [nodes] but
+ * not in [r], [dfs] will *not* be called on [n].
```
HadrienRenaud:
```suggestion
 * Scan the graph `m` in topological order,
 * computing scc. Excluding paths that start from
 * the set [nodes]. By contrast, for [n] in [nodes] but
 * not in [m], [dfs] will *not* be called on [n].
```

Also maybe would be good to fill the lines here (`M-q` on emacs, `gq` on vim, [source](https://jaredkrinke.github.io/til/wrap-lines-vim-emacs.html))

lib/innerRel.ml
```diff
@@ -363,23 +392,26 @@ module Make(O:MySet.OrderedType) : S
...
-    and kont_scc n scc res =
-      match scc with
-      | [] -> assert false
-      | [_] -> res
-      | ms ->
-          let vn = try M.find n res with Not_found -> assert false in
-          if Elts.is_empty vn then res
-          else List.fold_left (fun res m -> add_set m vn res) res ms
-    in
+  and kont_scc n scc res =
```
HadrienRenaud: I think those 2 need a better name, maybe prefix with `tr_`?

HadrienRenaud (approved):
Just some documentation comments. Please merge at your convenience.

lib/innerRel.ml
maranget: Fixed, thanks. Automatic line filling does not work properly for OCaml comments.

lib/innerRel.ml
maranget: Changing the names to `tr_edge` and `tr_scc`.

maranget:

lib/innerRel.ml
HadrienRenaud: Sad, on vim it does

akiyks:
Hi @maranget ,

Another report of LKMM issue.  After this merge, I get this fatal error in running herd7 on many LKMM litmus tests:

```
File "./linux-kernel.bell", line 85, characters 12-39: primitive different-values: argument mismatch
```

Said line in linux-kernel.bell looks like this:

```
flag ~empty different-values(srcu-rscs) as srcu-bad-value-match
```

I see a lot of activity in herd these days and this might be a transitory issue, but can you please have a look?

FYI, reverting 1e2307095993 on current master can paper over the error, but ends up in behavior changes of most litmus tests.

maranget:
Thanks for this @akiyks. This is indeed a "petty" bug. I publish a fix soon.

maranget:
Fix is PR #1912.

PR #1856, opened 2026-06-11
[herd] Delay evaluation of transitive closures

maranget:
The value of transitive closures are now `TransRel (t,s,th)`, where `t` and `s` are relations and th is a lazy relation. Such a value represents `t+|s`.

Hence, in some contexts, the  non-evaluated form can be used, so as *not* to compute `t+`. For instance, the interpreter computes `(t+|s)+` as `(t|s)+`. Similarly, an irreflexivity test on `(t+|s)` is performed as an acyclicity test on `(t|s)`.

In contexts where the optimised form cannot be used, the lazy value is forced. As a result the transitive closure operation is performed at most once. The is the case for instance of [S];(t+|s).

The `s` component is instrumental in not forcing transitive closures in unions resulting in more transitive closure operations being avoided. For instance:
```
let r1 = t1+|s1    /* Represented as Transrel (v1,w1,_),
                    * where t1 and s1 evaluates to v1 and w1, respectively. */
                   */
let r2 = (r1|r3)+  /* Represented as Transrel (v1,w4,_),
                    * where w4 is the union of w1 and v3 */
irreflexive r2     /* Performed as acyclic (v1 union w4) */
```
Here, the optimisation consists in no transitive closure operation at all.

Was the last commit of PR #1851.

### DIFF
--- a/lib/interpreter.ml
+++ b/lib/interpreter.ml
@@ -276,7 +276,9 @@ module Make
         | Empty | Unv
         | Pair of (S.event * S.event)
         | Rel of S.event_rel
-        | TransRel of S.event_rel (** Implicitly transitive relation *)
+        (** Implicitly transitive relation:
+            [TransRel (r,s,_)] represents [r+|s].  *)
+        | TransRel of S.event_rel * S.event_rel * S.event_rel Lazy.t
         | ClassRel of ClassRel.t
         | Event of S.event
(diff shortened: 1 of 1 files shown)


maranget:
Tests failing, convert to draft.

maranget:
Tests succeed.

PR #1859, opened 2026-06-13
[ASL Reference] fixes following review

Roman-Manevich:
Fixed an issue raised by Anna Slobodova; thanks Anna.
Fixed issues found by Hrutvik.
Fixed issues found during an AI-assisted review.

Added an option to automatically generate lists of rules by writing
```
\begin{TypingRuleList}{list-name}
... rules ...
\end{TypingRuleL}
```
and similarly for `Semantics` and `AST`.
The `\TypingRuleDef{...}`, `\SemanticsRuleDef{...}`, and `\ASTRulDef{...}` now take optional description arguments that are used to describe them in the generated list of rules.
The LaTeX code for this was generated via an AI.

Populated the description argument for all rule definitions appearing in lists of rules.


asllib/Typing.ml
```diff
@@ -2644,6 +2644,8 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
              let+ () = check_type_satisfies ~loc env t_e t in
              ( LE_SetField (le2, field) |> here,
                ses |: TypingRule.LESetStructuredField )
+         (* End *)
+         (* Begin LESetCollectiondField *)
```
hrutvik:
```suggestion
         (* Begin LESetCollectionField *)
```

asllib/instrumentation.ml
```diff
@@ -378,6 +378,7 @@ module TypingRule = struct
     | LESlice
     | LESetArray
     | LESetStructuredField
+    | LESetCollectiondField
```
hrutvik:
```suggestion
    | LESetCollectionField
```

asllib/tests/ASLTypingReference.t/TypingRule.SFor.bad4.asl
```diff
@@ -1,11 +1,11 @@
 var g : integer = 0;
 func upper_bound() => integer
 begin
-    g = 1; // writing to a global variables implies impurity.
+    g = 1; // writing to a global variable implies impurity.
     return 5;
 end;
 
-func ill_typed_for_loop_3()
+func ill_typed_for_loop_4()
```
hrutvik, line 8: Why the rename here?

asllib/tests/ASLTypingReference.t/TypingRule.DeclareConst.asl
hrutvik, line 1: Why remove this?

asllib/doc/AbstractSyntax.tex
```diff
@@ -588,21 +590,10 @@ \section{Building Parameterised Productions\label{sec:BuildingParameterisedProdu
...
-  \torexpr(\LESetArray(\vle, \ve)) &=& \EGetArray(\torexpr(\vle), \ve)\\
-  \torexpr(\LESetEnumArray(\vle, \ve)) &=& \EGetEnumArray(\torexpr(\vle), \ve)\\
-  \torexpr(\LESetField(\vle, \vx)) &=& \EGetField(\torexpr(\vle), \vx)\\
-  \torexpr(\LESetFields(\vle, [\vx_{1..k}])) &=& \EGetFields(\torexpr(\vle), [\vx_{1..k}])\\
-  \torexpr(\LEDiscard) &=& \EVar(\discardvar)\\
-  \torexpr(\LEDestructuring([\vle_{1..k}])) &=& \ETuple([i=1..k: \torexpr(\vle_i)])\\
-\end{array}
-\]
+
+Notice that $\LESetFields$ has to cases --- one for the \untypedast{} and one for the \typedast{}.
```
hrutvik:
```suggestion
Notice that $\LESetFields$ has two cases --- one for the \untypedast{} and one for the \typedast{}.
```

asllib/doc/Statements.tex
```diff
@@ -453,7 +453,7 @@ \subsection{Typing}
 \listingref{typing-sdecl} shows well-typed declaration statements.
 \ASLListing{Typing declaration statements}{typing-sdecl}{\typingtests/TypingRule.SDecl.asl}
 
-The specifications in \listingref{typing-sdecl-bad1} ill-typed,
+The specifications in \listingref{typing-sdecl-bad1} are ill-typed,
```
hrutvik:
```suggestion
The specification in \listingref{typing-sdecl-bad1} is ill-typed,
```

asllib/doc/TypeChecking.tex
```diff
@@ -126,12 +126,6 @@ \section{Static Environments\label{sec:StaticEnvironments}}
 \]
 \end{definition}
 
-The global environment and local environment consist of various components.
```
hrutvik, line 129: Is this no longer needed?

asllib/doc/TypeDeclarations.tex
```diff
@@ -161,10 +161,9 @@ \chapter{Type Declarations\label{chap:TypeDeclarations}}
 
 We also define the following helper rules:
 \begin{itemize}
-  \item \TypingRuleRef{DeclareType}
-  \item \TypingRuleRef{AddGlobalStorage}
   \item \TypingRuleRef{DeclareType}
```
hrutvik: Perhaps we could generate these in LaTeX (maybe with AI assistance)?

asllib/doc/asl.spec
```diff
@@ -1823,6 +1878,228 @@ constant empty_env : envs
...
+    k < zero;
+    --
+    empty_set;
+  }
+
+  case t_bits_empty {
+    t =: T_Bits(e, _);
+    eval_expr_sef(env, e) -> ResultExprSEF(nvint(zero), _) | ;
+    --
+    make_set(nvbitvector(empty_list));
```
hrutvik: Is there a reason not to fold this into `bitvector_domain`?

asllib/tests/ASLTypingReference.t/TypingRule.SFor.bad4.asl
Roman-Manevich: to match the file name

asllib/tests/ASLTypingReference.t/TypingRule.DeclareConst.asl
Roman-Manevich: Despite `declare_const` being a general function, it is only used for enumeration labels. I considered renaming it to `declare_enumeration_label_const`, but then decided against as its definition is generatic.
I replaced the example with one where the function is actually used to declare a constant for an enumeration label. Since `TypingRule.DeclareConst.asl` became an orphan and it was so trivial, I though it would be best to remove it.

asllib/doc/TypeChecking.tex
Roman-Manevich: no longer needed.

asllib/doc/asl.spec
Roman-Manevich: done

asllib/doc/TypeDeclarations.tex
Roman-Manevich: Definitely. Done, and this actually fixed a few lists. Thanks for the suggestion.

asllib/doc/ASLmacros.tex
```diff
@@ -226,6 +226,9 @@
 \newcommand\TypingRuleDefExample[1]{\emph{TypingRule.#1}}
 \newcommand\SemanticsRuleDef[1]{\subsubsection{SemanticsRule.#1\label{sec:SemanticsRule.#1}}}
 \newcommand\SemanticsRuleDefExample[1]{\emph{SemanticsRule.#1}}
+% Metadata consumed by doclint.py -cm for generated rule lists.
+% Place immediately after a TypingRuleDef or SemanticsRuleDef.
+\newcommand\RuleDescription[1]{}
```
hrutvik: Should this perhaps be integrated into TypingRuleDef or SemanticsRuleDef instead, so that we can't forget one?

asllib/doc/extended_macros.py
```diff
@@ -251,6 +261,142 @@ def transform(cls, block_lines: list[str]) -> list[str]:
         return transformed_lines
 
 
+class RuleListMacro(ABC):
```
hrutvik: Hmm this is a bit surprising to me - could this not be achieved via a LaTeX macro rather than source rewriting with a Python script?

asllib/doc/extended_macros.py
```diff
@@ -251,6 +261,142 @@ def transform(cls, block_lines: list[str]) -> list[str]:
...
+            error_message = (
+                f"there are {len(begin_lines)} occurrences of {cls.BEGIN}"
+                f" and {len(end_lines)} occurrences of {cls.END}! Aborting."
+            )
+            raise ValueError(error_message)
+        blocks: list[Block] = []
+        for begin_line_number, end_line_number in zip(begin_lines, end_lines):
+            if begin_line_number >= end_line_number:
+                raise ValueError(f"{cls.BEGIN} must appear before {cls.END}!")
+            blocks.append(Block(begin_line_number, end_line_number))
```
hrutvik: I'm starting to wonder if this is a frequently-repeated pattern in this file

asllib/doc/TypeSystemUtilities.tex
```diff
@@ -18,19 +20,17 @@ \chapter{Type System Utility Rules\label{chap:TypeSystemUtilityRules}}
...
   \item \TypingRuleRef{LookupImmutableExpr} looks up the initialising expression associated with an immutable storage element.
-  \item \TypingRuleRef{AddGlobalImmutableExpr} updates a \globalstaticenvironmentterm{} by associating a global immutable storage element
-        with its initialising expressions.
-  \item \TypingRuleRef{AddLocalImmutableExpr} updates a \localstaticenvironmentterm{} by associating a local immutable storage element
-        with its initialising expressions.
-  \item \TypingRuleRef{ShouldRememberImmutableExpression} tests whether to associate an immutable storage element
-        with its initialising expressions based on its \sideeffectdescriptorsetsterm.
-  \item \TypingRuleRef{AddImmutableExpr} conditionally updates a \staticenvironmentterm{} by associating
-        a local immutable storage element with its initialising expressions.
+  \item \TypingRuleRef{AddGlobalImmutableExpr} updates a \globalstaticenvironmentterm{} by associating a global immutable storage element with its initialising expressions.
```
hrutvik: "expressions" --> "expression" for all of these entries?

asllib/doc/SemanticsUtilities.tex
```diff
@@ -4,32 +4,33 @@ \chapter{Semantics Utility Rules\label{chap:SemanticsUtilityRules}}
...
-  \item \SemanticsRuleRef{SetIndex} updates a vector \nativevalueterm{} at a given index;
-  \item \SemanticsRuleRef{GetField} reads a given field in a \nativevalueterm{} record;
-  \item \SemanticsRuleRef{SetField} updates a given field in a \nativevalueterm{} record;
-  \item \SemanticsRuleRef{DeclareLocalIdentifier} updates an environment by associating an identifier with a given \nativevalueterm{}
-        and also generating an \executiongraphterm{} with a corresponding Write Effect;
-  \item \SemanticsRuleRef{DeclareLocalIdentifierM} updates an environment by associating an identifier with a given \nativevalueterm{}
-        and also updating a given \executiongraphterm{} with a corresponding Write Effect and an $\asldata$ edge;
-  \item \SemanticsRuleRef{DeclareLocalIdentifierMM} updates an environment by associating an identifier with a given \nativevalueterm{}
-        and also updating a given \executiongraphterm{} with a corresponding Write Effect and an $\aslpo$ edge.
+  \item \SemanticsRuleRef{GetPendingCalls} retrieves the value of the $\pendingcalls$ value from an environment.
```
hrutvik: "value of the ... value" for all of these is a bit unwieldy

asllib/doc/SemanticsUtilities.tex
```diff
@@ -4,32 +4,33 @@ \chapter{Semantics Utility Rules\label{chap:SemanticsUtilityRules}}
...
+  \item \SemanticsRuleRef{SetPendingCalls} updates the value of the $\pendingcalls$ value in an environment.
+  \item \SemanticsRuleRef{IncrPendingCalls} updates the value of the $\pendingcalls$ value in an environment by incrementing it by one.
+  \item \SemanticsRuleRef{DecrPendingCalls} updates the value of the $\pendingcalls$ value in an environment by decrementing it by one.
```
hrutvik: increments/decrements the value of $\pendingcalls$ in an environment ...

asllib/doc/SemanticsUtilities.tex
```diff
@@ -4,32 +4,33 @@ \chapter{Semantics Utility Rules\label{chap:SemanticsUtilityRules}}
...
+  \item \SemanticsRuleRef{WriteToBitvector} updates a slice of a bitvector \nativevalueterm{}.
+  \item \SemanticsRuleRef{GetIndex} reads a vector \nativevalueterm{} at a given index.
+  \item \SemanticsRuleRef{SetIndex} updates a vector \nativevalueterm{} at a given index.
+  \item \SemanticsRuleRef{GetField} reads a given field in a \nativevalueterm{} record.
+  \item \SemanticsRuleRef{SetField} updates a given field in a \nativevalueterm{} record.
+  \item \SemanticsRuleRef{DeclareLocalIdentifier} updates an environment by associating an identifier with a given \nativevalueterm{} and also generating an \executiongraphterm{} with a corresponding Write Effect.
+  \item \SemanticsRuleRef{DeclareLocalIdentifierM} updates an environment by associating an identifier with a given \nativevalueterm{} and also updating a given \executiongraphterm{} with a corresponding Write Effect and an $\asldata$ edge.
+  \item \SemanticsRuleRef{DeclareLocalIdentifierMM} updates an environment by associating an identifier with a given \nativevalueterm{} and also updating a given \executiongraphterm{} with a corresponding Write Effect and an $\aslpo$ edge.
```
hrutvik: "also generating/updating" --> "also generates/updates"

asllib/doc/TypeDomains.tex
```diff
@@ -37,382 +37,53 @@ \subsection{The Dynamic Domain of a Type\label{sec:DynDomain}}
...
+  \tint &
+    \{ \nvintop{n} \;|\; n \in \Z \}\\
+  \treal &
+    \{ \NVLiteral(\LReal(q)) \;|\; q \in \Q \}\\
+  \tstring &
+    \{ \nvstringop{s} \;|\; s \in \Strings \}\\
+  \bitvectordomainop{k} &
+    \begin{cases}
+      \{ \nvbitvectorop{\vb_{1..k}} \;|\; \forall i=1..k: \vb_i \in \Bit \} & \text{if }k > 0\\
+      \emptyset & \text{otherwise}
```
hrutvik: Is this right? There should be one value for an empty bit-vector, `\nvbitvectorop{\emptylist}` or something

asllib/doc/ASLmacros.tex
Roman-Manevich: I decided against this, since I thought something like `\TypingRuleDef[long description]{rule-name}` looks a bit ugly. Also, it would necessitate changing the way we parse rules, which could raise some unexpected.
That said, I'm willing to be convinced otherwise.

asllib/doc/extended_macros.py
Roman-Manevich: Yes, but it would mean we have a generated file, which we won't commit. That means, it would be easy to miss changes to the list, since they won't appear in any diff.

hrutvik: Sorry to clarify - I don't mean generating a LaTeX macro populated with the correct list by scraping the source, I mean defining a LaTeX macro that will automatically generate the correct list dynamically during the LaTeX build (much like a table of contents is generated). Why would we need to diff the list in this case?

asllib/doc/ASLmacros.tex
hrutvik: I'm not sure ugliness should be a concern here over correctness, especially because we can easily split a two-argument macro over multiple lines e.g.
```
\TypingRuleDef{rule-name}
    {long-description}
```
This would also preserve the current parsing of rules wouldn't it?

asllib/doc/extended_macros.py
Roman-Manevich: factored out and deduplicated.

asllib/doc/extended_macros.py
Roman-Manevich: That's a good idea, but one complication is that we need to reference only rules in the remainder of one file. We would need to introduce collection boundaties for lists of rules. Something like:
```
\BeginTypingRuleList
...
\EndTypingRuleList
```
What do you think?

asllib/doc/ASLmacros.tex
Roman-Manevich: done with an optional argument.

asllib/doc/extended_macros.py
Roman-Manevich: I've switched to macros.

asllib/doc/doclint.py
```diff
@@ -1730,6 +1731,91 @@ def check_duplicate_usepackage(latex_files: list[str]) -> int:
     return num_errors
 
 
+def check_rule_list_boundaries(filename: str) -> int:
```
hrutvik: I'm a bit surprised this is necessary - e.g. if `PrintRuleList` is always followed directly by `BeginRuleList`, would one macro not suffice? And then can we not leverage LaTeX to check for unmatched begin/end?

asllib/doc/LexicalStructure.tex
```diff
@@ -62,7 +62,6 @@ \section{Lexical Regular Expressions\label{sec:Lexical Regular Expressions}}
 $A+$                                  & One or more repetitions of A\\
 \verb|"a_string"|                     & The string \verb|a_string| verbatim\\
 \texttt{<}r\texttt{>}                 & The lexical regular expression defined for \texttt{<}r\texttt{>}\\
-\hline
```
hrutvik, line 65: Out of curiosity, why delete these?

asllib/doc/extended_macros.py
hrutvik: > we need to reference only rules in the remainder of one file

For reference, where does this occur?

Roman-Manevich: Are you asking where do we need to reference rules up to the remainder of a file? If so, this is true in all files where we create list-of-rules outlines:
`GlobalStorageDeclarations.tex`, `SubprogramCalls.tex`, `TypeAttributes.tex`, etc.

asllib/doc/doclint.py
Roman-Manevich: we can certainly merge the Print and Begin macros. But merging the Begin and End macros would make the code fragile, since they contain things like `\verb`. I think it's better to keep those apart and move to  environment-style macros.

asllib/doc/LexicalStructure.tex
Roman-Manevich: In some places the combination of table lines and underscores didn't look good so I thought it would be best to move to a uniform table format without the bottom line.

asllib/doc/extended_macros.py
hrutvik: Ah I see sorry - but each file tends to correspond to a chapter boundary in these cases right? So it's the same as "all rules up to the end of the chapter"?

hrutvik (approved):
- I haven't checked that the new lists look as expected - please do generate a PDF diff to confirm
- I'd like to understand whether: rather than environments, could we use chapter boundaries to better delimit lists of rules (i.e. a single `\RuleList` macro that grabs all rules defined in the chapter and  lists them)?

asllib/doc/extended_macros.py
Roman-Manevich: The list of rules for a given file any given kind (typing / semantics) tends to be the same as the list of rules in a chapter, but not always:
* "Chapter 32 Symbolic Reduction and Equivalence Testing": current list has 26 rules; chapter-wide would have 29. It would add `PolynomialOfVar`, `AddPolynomials`, and `AddPolynomialList`, because the corresponding `\TypingRuleDefs` appear before the current list marker. This is intentional, although we can restructure the chapter to move them into the list if we wanted to.
* "Chapter 12: Types": current list has 8 rules; a chapter-wide would have 57. It would add the main type typing rules from `Types.tex`, all relation rules from `RelationsOnTypes.tex`, and base-value rules from `BaseValues.tex`.

PR #1860, opened 2026-06-13
[ASL][ASL Refernce] deprecated some slicing shorthands from ASL

Roman-Manevich:
This PR deprecates the slicing shorthands `[:N]` and `[index *: len]`. These shorhands can be enabled via the new `--version-eac1` flag.
* The shorthand `[:N]` can be replaced by `[0 +: N]` and `[index *: len]`  can be replaced by `[index * len +: len]`.
* The standard library is updated accordingly.
* The shorthands are completely removed from ASL Reference.
* The tests `SemanticsRule.SliceFromZero` and `SemanticsRule.SliceStar` are moved to `tests/regressions.t` and tested with `--version-eac1`.


HadrienRenaud:
Hi @Roman-Manevich, as mentioned in private message, I think the ASL code that we use for the instruction semantics still use the features removed here. Could we please just hide it behind a command line flag instead of removing it from the executable? Thanks.

asllib/doc/Slicing.tex
```diff
@@ -41,18 +41,9 @@ \chapter{Bitvector Slices\label{chap:BitvectorSlicing}}
 \Nslice \derives \ & \Nexpr &\\
             |\  & \Nexpr \parsesep \Tcolon \parsesep \Nexpr &\\
             |\  & \Nexpr \parsesep \Tpluscolon \parsesep \Nexpr &\\
-            |\  & \Nexpr \parsesep \Tstarcolon \parsesep \Nexpr &\\
             |\  & \Tcolon \parsesep \Nexpr &
```
hrutvik: This line also needs removing

asllib/doc/Slicing.tex
hrutvik: This also needs removing

asllib/doc/Syntax.tex
```diff
@@ -641,7 +641,6 @@ \section{ASL Grammar\label{sec:ASLGrammar}}
 \Nslice \derives \ & \Nexpr &\\
               |\  & \Nexpr \parsesep \Tcolon \parsesep \Nexpr &\\
               |\  & \Nexpr \parsesep \Tpluscolon \parsesep \Nexpr &\\
-              |\  & \Nexpr \parsesep \Tstarcolon \parsesep \Nexpr &\\
               |\  & \Tcolon \parsesep \Nexpr &
```
hrutvik: This also needs removing

asllib/tests/ASLDefinition.t/run.t
```diff
@@ -207,7 +207,6 @@ Examples used in ASL High-level Definition:
   $ aslref --no-exec GlobalStorageDeclarations.asl
   $ aslref --no-exec IR.asl
   $ aslref Bit.asl
-  $ aslref PrefixSlice.asl
```
hrutvik, line 210: Is this intentional? The test still exists

Roman-Manevich: removed file

asllib/ParserConfig.mli
```diff
@@ -20,5 +20,8 @@
 (* herdtools7 github repository.                                              *)
 (******************************************************************************)
 
-module type CONFIG = sig end
 (** The CONFIG module signature for the ASL1 Parser *)
+module type CONFIG = sig
+  val version_eac1 : bool
+  (** Errors if a deprecated feature is used. *)
```
HadrienRenaud: I don't think this comment is correct. Maybe this should read


```suggestion
  (** Enables the EAC1 deprecated features. *)
```

asllib/error.ml
```diff
@@ -112,6 +112,7 @@ type error_desc =
   | UnexpectedCollection
   | BadPrimitiveArgument of identifier * string
   | NoEntryPoint
+  | ObsoleteSyntax of string
```
HadrienRenaud: This leads to some not-so-pretty error messages (see breaks at strange places in one of the `run.t`

I would suggest:

```suggestion
  | ObsoleteSyntax of (Format.formatter -> unit)
```

Then in printing you have something like:
```ocaml
    | ObsoleteSyntax f -> pp_err parse "Obsolete syntax:@ @[%t@]" f
```

And when building them:
```ocaml
      let msg fmt =
        Format.fprintf fmt "Deprecated slice syntax, use \"%a*%a +: %a\" instead."
          PP.pp_expr e1 PP.pp_expr e2 PP.pp_expr e2
      in
      Error.fatal_here $startpos $endpos @@
      Error.ObsoleteSyntax msg
```

This makes the error message look like:
```
ASL Grammar error: Obsolete syntax:
  Deprecated slice syntax, use "3*2 +: 2" instead.
```

Full patch here: [0000.not-comitted.use-fun-in-errors.patch](https://github.com/user-attachments/files/29088523/0000.not-comitted.use-fun-in-errors.patch)

HadrienRenaud (approved):
Thanks for this. This looks good to me on the implementation side, just suggesting a better error message formatting.

Disclaimer: I did not look precisely in the diff in the reference.

lib/ASLBase.ml
```diff
@@ -226,7 +226,7 @@ let asl_generic_parser version lexer lexbuf =
 let stmts_from_string s =
   let open Asllib in
   let lexbuf = Lexing.from_string s in
-  let module Parser = Parser.Make(struct end) in
+  let module Parser = Parser.Make(struct let version_eac1 = false end) in
```
hrutvik: Should this not be `true` to match above?

asllib/builder.ml
```diff
@@ -27,10 +27,10 @@ let _debug = false
 type token = Tokens.token
 type ast_type = [ `Opn | `Ast ]
 type version = [ `ASLv0 | `ASLv1 ]
-type parser_config = { v0_use_split_chunks : bool }
+type parser_config = { v0_use_split_chunks : bool; version_eac1 : bool }
 type version_selector = [ `ASLv0 | `ASLv1 | `Any ]
 
-let default_parser_config = { v0_use_split_chunks = false }
+let default_parser_config = { v0_use_split_chunks = false; version_eac1 = true }
```
hrutvik:
```suggestion
let default_parser_config = { v0_use_split_chunks = false; version_eac1 = false }
```

hrutvik (approved):

PR #1864, opened 2026-06-16
Refine AArch64 BBM modelling and add regression coverage

relokin:
## Summary

  This branch refines the AArch64 Break-Before-Make (BBM) modelling in `herd`, fixes related helper logic used by the model checker/interpreter, and adds a dedicated regression catalogue for BBM-oriented tests.

  Compared to `github/master`, the branch includes 7 commits covering:
  - BBM rule updates in the AArch64 cat files
  - a fix for `writable2` in `machModelChecker`
  - support for `PTEtoPA` in `machAction`
  - documentation/macros for the new BBM relations
  - a new `catalogue/aarch64-BBM` regression suite
  - a typo fix in `aarch64fences.cat`

  ## Main changes

  - Move the definition of `ca` out of `aarch64.cat` into `cos.cat` so it can be reused by the BBM-specific logic.
  - Rework BBM detection in `aarch64.cat` / `aarch64bbm.cat` around TTD/TLB-cacheability concepts rather than the older `PTE*` naming, including:
    - `TLBCacheableTTD` / `TLBUncacheableTTD`
    - `TTD-update-needsBBM`
    - updated BBM sequencing check and warning flag
  - Fix `writable2` in `herd/machModelChecker.ml` so writability is evaluated per event using the correct dirty-bit context.
  - Simplify the interpreter helpers by introducing shared relation filtering logic, and use it to implement `same_oa`, `oa_changes`, and `at_least_one_writable`.
  - Extend `tools/miaou.ml` to handle the `domain` procedure.
  - Add TeX macros documenting the new BBM/TTD relations in `catdefinitions.tex`.
  - Add a new `catalogue/aarch64-BBM/tests` suite with expected outputs and a `cata-test-all` target for running it.
  - Update the affected AArch64 KVM expected output to match the revised BBM behaviour.
  - Fix the typo in the deprecated DSB-ST maintenance-scope warning.

  ## Testing

  - `make test.herd.cata-ext.aarch64-BBM`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -285,6 +285,19 @@ cata-test:: test.herd.cata.aarch64-readers-guide
 cata-test:: test.herd.cata.bpf
 cata-test:: test.herd.cata.x86_64
 
+test.herd.cata-extended.%:
+	@ echo
+	$(HERD_REGRESSION_TEST) \
+		-j $(J) \
+		$(NOHASH) \
+		-herd-path $(HERD) \
+		-libdir-path ./herd/libdir \
+		-litmus-dir  catalogue/$*/tests \
+		$(REGRESSION_TEST_MODE)
+	@ echo "herd7 catalogue extended $* tests: OK"
+
+cata-test-all:: test.herd.cata-extended.aarch64-BBM
+
 test.herd-mixed.cata.%:
 	@ echo
 	$(HERD_CATALOGUE_REGRESSION_TEST) \
--- a/catalogue/aarch64-BBM/cfgs/web.cfg
+++ b/catalogue/aarch64-BBM/cfgs/web.cfg
@@ -1,0 +1,32 @@
+graph cluster
+showevents noregs
+withbox true
+labelbox true
+fontsize 12
+xscale 2.0
+yscale 1.5
+arrowsize 0.8
+splines spline
+pad 0.1
+
+doshow ets-ob
+doshow bob,DSB-ob
+unshow dmb.sy,dmb.st,dmb.ld,dsb.sy,isb,dsb.st,dsb.ld,isb
+
+edgeattr iico_data,color,violetred
+edgeattr iico_ctrl,color,lightblue
+edgeattr iico_order,color,teal
+edgeattr rmw,color,indigo
+
+edgeattr po,color,lightgray
+
+edgeattr tc-before,color,aquamarine4
+edgeattr sca-class,color,lightgray
+edgeattr DSB-ob,color,gray
+edgeattr IFB-ob,color,navyblue
+edgeattr dob,color,darkgoldenrod1
+edgeattr bob,color,orange3
+edgeattr aob,color,darkgoldenrod1
+edgeattr ets-ob,color,skyblue4
+edgeattr TLBI-ob,color,purple
+edgeattr TLBI-after,color,forestgreen
--- a/catalogue/aarch64-BBM/tests/CoRR+PteOA.DB0.val.litmus
+++ b/catalogue/aarch64-BBM/tests/CoRR+PteOA.DB0.val.litmus
@@ -1,0 +1,12 @@
+AArch64 CoRR+PteOA.DB0.val
+Variant=vmsa
+{
+ [x]=1; [y]=1;
+ [PTE(x)]=(oa:PA(x),db:0);
+ 0:X4=(oa:PA(y),db:0); 0:X5=PTE(x); 
+ 1:X1=x;
+}
+ P0          | P1          ;
+ STR X4,[X5] | LDR W0,[X1] ;
+             | LDR W2,[X1] ;
+exists(1:X0=1 /\ 1:X2=1)
--- a/catalogue/aarch64-BBM/tests/MP+tlbi-sync.ishsRpteaf0pte-amo.casptepteoa.af1+pos.litmus
+++ b/catalogue/aarch64-BBM/tests/MP+tlbi-sync.ishsRpteaf0pte-amo.casptepteoa.af1+pos.litmus
@@ -1,0 +1,18 @@
+AArch64 MP+tlbi-sync.ishsRpteaf0pte-posptepteoa.af1+pos
+Variant=vmsa
+TTHM=P1:HA
+{
+ [x]=1;
+ [y]=2;
+ 0:X0=PTE(x); 0:X1=(oa:PA(x), af:0); 0:X3=(oa:PA(y));
+ 0:X4=x; 1:X4=x;
+}
+ P0              | P1               ;
+ STR X1,[X0]     | L01: LDR W2,[X4] ;
+ DSB ISH         | L00: LDR W5,[X4] ;
+ LSR X5,X4,#12   |                  ;
+ TLBI VAAE1IS,X5 |                  ;
+ DSB ISH         |                  ;
+ CAS X1,X3,[X0]  |                  ;
+
+exists (1:X2=2 /\ 1:X5=1)
--- a/herd/libdir/aarch64bbm.cat
+++ b/herd/libdir/aarch64bbm.cat
@@ -46,43 +46,46 @@
  *)
 include "aarch64memattrs.cat"
 
-(* Can we move this to stdlib? *)
-(* Coherence-after *)
-let ca = fr | co
-
-let PTE-MT-update =
+let TTD-MT-update =
   [Normal]; ca; [Device]
   | [Device]; ca; [Normal]
 
-let PTE-SH-update =
+let TTD-SH-update =
   [NSH]; ca; [ISH | OSH]
   | [ISH]; ca; [OSH | NSH]
   | [OSH]; ca; [NSH | ISH]
 
-let PTE-ICH-update =
+let TTD-ICH-update =
   [iNC]; ca; [iWT | iWB]
   | [iWT]; ca; [iWB | iNC]
   | [iWB]; ca; [iNC | iWT]
 
-let PTE-OCH-update =
+let TTD-OCH-update =
   [oNC]; ca; [oWT | oWB]
   | [oWT]; ca; [oWB | oNC]
   | [oWB]; ca; [oNC | oWT]
 
-let PTE-DT-update =
+let TTD-DT-update =
   [Device-GRE]; ca; [Device-nGRE | Device-nGnRE | Device-nGnRnE]
   | [Device-nGRE]; ca; [Device-GRE | Device-nGnRE | Device-nGnRnE]
   | [Device-nGnRE]; ca; [Device-GRE | Device-nGRE | Device-nGnRnE]
   | [Device-nGnRnE]; ca; [Device-GRE | Device-nGRE | Device-nGnRE]
 
-let PTE-OA-update = ([PTE]; ca; [PTE & oa-changes(PTE, ca^-1)])
-let PTE-OA-update-writable = PTE-OA-update &
-        ([PTE]; ca; [PTE & at-least-one-writable(PTE, ca^-1)])
+let TTD-OA-update = [TTD & M]; ca \ same-oa((TTD & M) * (TTD & M)); [TTD & W]
+let TTD-at-least-one-writable = at-least-one-writable((TTD & M) * (TTD & M))
+let TTD-memory-contents-mismatch = 
+  ([TTD & M]; PTEtoPA; different-values(W * W); PTEtoPA^-1; [TTD & M]) & loc
 
-let PTE-update-needsBBM = ([PTEV]; ca \ (ca; [PTEV]; ca); [PTEV]) &
-  (PTE-MT-update | PTE-SH-update | PTE-ICH-update | PTE-OCH-update | PTE-DT-update
-  | PTE-OA-update)
+let TLBUncacheableTTD = TTDINV | TTDAF0
+let TLBCacheableTTD = TTD & M \ TLBUncacheableTTD
 
-let TTDV = PTEV
-let TTDINV = PTEINV
-let TTD-update-needsBBM = PTE-update-needsBBM
+let TTD-update-BBM-cand =
+  TTD-MT-update
+  | TTD-SH-update
+  | TTD-ICH-update
+  | TTD-OCH-update
+  | TTD-DT-update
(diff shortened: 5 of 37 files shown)


relokin:
@fsestini @psafont @TiberiuBucur, I would like to merge this soon. If you have any feedback please do let me know.

psafont:
Looks like this PR is a good model for adding support for an architectural feature, I appreciated the commit-by commit structure and explanations in the commit messages.

Other than that, I need to learn more about the domain to understand what certain terms mean and how they interact (OA, PTE, etc)

tools/miaou.ml
```diff
@@ -564,7 +564,14 @@ and cons_seqs (fs:exp list) (es:exp list) =
...
         end
-      | e ->
+      | App (_,Var (_,"domain"), rel) ->
+        let e3 = Next.next () in
+        begin match tr_rel e1 e3 rel with
+        | List ({ items; _ } as l) ->
+           List { l with items = List.rev items; }
+        | _ as i -> i
+        end
+       | e ->
```
fsestini:
```suggestion
      | e ->
```

Makefile
```diff
@@ -284,6 +284,19 @@ cata-test:: test.herd.cata.aarch64-ETS3
 cata-test:: test.herd.cata.bpf
 cata-test:: test.herd.cata.x86_64
 
+test.herd.cata-ext.%:
```
fsestini: Does `ext` stand for "extra"/"extended"? Minor, but perhaps I would write the word in full here, and/or add a one-line comment explaining what differentiates this target from the normal `cata` target.

Makefile
```diff
@@ -284,6 +284,19 @@ cata-test:: test.herd.cata.aarch64-ETS3
...
+	@ echo
+	$(HERD_REGRESSION_TEST) \
+		-j $(J) \
+		$(NOHASH) \
+		-herd-path $(HERD) \
+		-libdir-path ./herd/libdir \
+		-litmus-dir  catalogue/$*/tests \
+		$(REGRESSION_TEST_MODE)
+	@ echo "herd7 catalogue $* instructions tests: OK"
```
fsestini: Should "instructions" be included here?

herd/machAction.ml
```diff
@@ -648,7 +648,30 @@ end = struct
...
+            let open Constant in
+            function
+            | Some (A.V.Val (PteVal v)) -> Some v
+            | Some _ | None -> None in
+          List.filter_map get_pteval [read_of act; written_of act] in
+        let is_physical_location pteval act =
+          match A.V.Cst.PteVal.as_physical pteval, location_of act with
+          | Some s, Some loc ->
+              begin match A.symbol loc with
+              | Some (Constant.Physical (s', _)) -> Misc.string_eq s s'
```
fsestini: Personally I would use the stdlib operators directly when possible, for added clarity. (`Misc.string_eq` is an alias of `String.equal`, so frankly I'm not entirely sure why `string_eq` needs to exist in the first place).
```suggestion
              | Some (Constant.Physical (s', _)) -> String.equal s s'
```

fsestini (commented):
Thanks @relokin . Just a few minor comments.

Makefile
relokin: You are right, it shouldn't be included. I will change this.

catalogue/aarch64-BBM/tests/MP+tlbi-sync.ishsRpteaf0pte-amo.casptepteoa.af1+pos.litmus
```diff
@@ -0,0 +1,18 @@
+AArch64 MP+tlbi-sync.ishsRpteaf0pte-posptepteoa.af1+pos
+Variant=vmsa
+TTHM=P1:HA
+{
+ [x]=1;
+ [y]=2;
+ 0:X0=PTE(x); 0:X1=(oa:PA(x), af:0); pteval_t 0:X2=0; 0:X3=(oa:PA(y)); 0:X4=x;
```
TiberiuBucur: Probably not relevant for the context of this catalogue / this whole PR, but 0:X2 is not used or inspected at all. Was this test generated with diyone7?

Makefile
TiberiuBucur: I'm wondering the same thing as @fsestini. Is the reason why BBM was kept in an "extended" catalogue to do with the CI running time? I know running the whole catalogue takes quite long.

herd/libdir/aarch64bbm.cat
```diff
@@ -71,12 +71,13 @@ let TTD-DT-update =
...
-let TTD-at-least-one-writable = at-least-one-writable(TTD*TTD)
-let TTD-memory-contents-mismatch = (PTEtoPA; different-values(IW * IW); PTEtoPA^-1) & loc
+let TTD-OA-update = [TTD & M]; ca \ same-oa((TTD & M) * (TTD & M)); [TTD & W]
+let TTD-at-least-one-writable = at-least-one-writable((TTD & M) * (TTD & M))
+let TTD-memory-contents-mismatch = 
+  ([TTD & M]; PTEtoPA; different-values(W * W); PTEtoPA^-1; [TTD & M]) & loc
 
 let TLBUncacheableTTD = TTDINV | TTDAF0
-let TLBCacheableTTD = TTD \ TLBUncacheableTTD
+let TLBCacheableTTD = TTD & M \ TLBUncacheableTTD
```
TiberiuBucur: I assume specifying `TTD & M` as opposed to `TTD` is important for `miaou7`?

herd/libdir/catdefinitions.tex
```diff
@@ -98,18 +98,31 @@
...
 %
 
-\newcommand{\TTDupdateneedsBBMemph}[2]{The TTD update from #1 to #2 needs BBM}
+\newcommand{\TTDM}[1]{#1 is a TTD Memory Effect}
+\newcommand{\TTDR}[1]{#1 is a TTD Memory Read Effect}
+\newcommand{\TTDW}[1]{#1 is a TTD Memory Write Effect}
+\newcommand{\TTDAF}[1]{#1 is a TTD Memory Effect with the Access flag set to 0}
+\newcommand{\TTDINV}[1]{#1 is a TTD Memory Effect with the valid bit set to 0}
 \newcommand{\TTDupdateneedsBBM}[2]{The TTD update from #1 to #2 needs BBM}
+\newcommand{\TTDupdateneedsBBMemph}[2]{\TTDupdateneedsBBM{#1}{#2}}
```
TiberiuBucur, line 107: I presume there is no need for `\emph` in this case? Same for `\TTDupdateBBMcandemph`, `\TTDmemorycontentsmismatchemph` and `\TTDatleastonewritableemph`.

herd/libdir/catdefinitions.tex
```diff
@@ -275,6 +275,8 @@
 \newcommand{\introbemph}[2]{#1 is \emph{\introbname{}} #2}
 
 \newcommand{\sameoa}[2]{#1 and #2 have the Same Output Address}
+\newcommand{\atleastonewritable}[2]{Either #1 or #2 grant write permission}
```
TiberiuBucur, line 278: This coincides with `\TTDatleastonewritable`, maybe the latter can use this one.

catalogue/aarch64-BBM/tests/CoRR+PteOA.DB0.val.litmus
TiberiuBucur, line 1: Is this to illustrate that the `Warning-BBM-expected` is only flagged when the values in the physical locations differ?

TiberiuBucur (commented):
A few minor, non-blocking comments, but overall LGTM.

Makefile
relokin: It is more about the level of detail in the testing we perform.

Before this PR, the CI would run each test in the catalogue and check whether the post-condition was satisfied. It would then compare that result against the corresponding entry in the kinds file, which records whether the test is allowed, forbidden, or required. If the observed result did not match the intent recorded in the kinds file, the CI would report a failure.

The extended testing, which is already used for `herd/tests/` and which this PR introduces for the new catalogue, goes further by checking that the complete set of final states matches the expectations recorded in the .expected file.

catalogue/aarch64-BBM/tests/MP+tlbi-sync.ishsRpteaf0pte-amo.casptepteoa.af1+pos.litmus
relokin: No it was not generated with diyone7, but I tried to get a base test with diyone7 and then modified it by hand.

Makefile
HadrienRenaud: Thanks for the explanation. I'm also curious about why this catalogue in particular needs the extended check? Are there some litmus tests where the post-condition does not capture enough information and having the full output reflects better the behaviours?

herd/libdir/aarch64bbm.cat
relokin: Indeed

catalogue/aarch64-BBM/tests/CoRR+PteOA.DB0.val.litmus
relokin: yes

Makefile
relokin: The new catalogue is intended to illustrate the `Warning-BBM-expected` flag. It contains a set of tests covering cases where the Arm architecture recommends a Break-before-make (BBM) sequence. Some of the tests perform the recommended BBM sequence and therefore should not trigger the warning, while others deliberately omit it and are expected to be flagged.

herd/libdir/catdefinitions.tex
relokin: The addition of these cat definitions is a little opportunistic. We are not going to describe these definitions in the Arm ARM for now.

Makefile
HadrienRenaud: Ah indeed warnings are not checked with the non-extended catalogue, that makes sense thank you!

relokin:
Thank @psafont, @TiberiuBucur and @fsestini for the reviews. I am planning to merge this tomorrow. Please let me know if you have any other feedback or if you need more time to review.

TiberiuBucur (approved):
LGTM

Makefile
fsestini: For my understanding: does it mean that for this particular catalogue, it might be sufficient to just check against expected outcome (allowed/forbidden/required) _and_ triggered flags (thus ignoring the rest of herd7's output including final states), if our testing infra allowed a way to do so?

catalogue/aarch64-BBM/cfgs/web.cfg
fsestini, line 1: Probably not something to do with this PR, but when running herd-www locally, the generated graphs look oversized compared to the rest of the UI and won't fit in the designated box. However other catalogue besides the BBM one behave similarly, so this is likely an issue we need to address in herd-www separately at some point.

tools/miaou.ml
```diff
@@ -564,6 +580,13 @@ and cons_seqs (fs:exp list) (es:exp list) =
            List { l with items = List.rev items; }
         | _ as i -> i
         end
+      | App (_,Var (_,"domain"), rel) ->
+        let e3 = Next.next () in
+        begin match tr_rel e1 e3 rel with
+        | List ({ items; _ } as l) ->
+           List { l with items = List.rev items; }
```
fsestini, line 580: Do we need to reverse `items` here?

fsestini (commented):
Thanks @relokin , sorry, I have a couple more comments/questions.

A couple of small comments in the code, and a more general comment/question regarding `miaou`: the PR adds support for some cat builtin functions like `domain`, `same-oa`, etc. However there are no changes to the end-to-end/regression test suite in `tools/tests/miaou.t` that reflect this change. Not a problem per se, but my question is:
- do these changes to miaou affect the transliteration of any relation of interest in aarch64, and if so, what are those relations, and should we add them to the test suite?
- if this change to miaou does not affect any aarch64 relation of interest, then I wonder what motivated the change in the first place.

catalogue/aarch64-BBM/cfgs/web.cfg
relokin: @TiberiuBucur is looking at this. It will be addressed separately for all catalogues.

tools/miaou.ml
relokin: I think it's a matter of how we want to print out a clause that uses domain. In earlier versions of this PR, I had a clause ending with [domain(...)] and I thought it would be better if we described E2 last as we do in other clauses.

However, since then I moved away from using domain for BBM. I kept it just because I thought it might be useful for miaou. We have to make some decisions about domain which are not in the scope for this PR, so I will drop it. @artkhyzha is also using domain in https://github.com/herd/herdtools7/pull/1880 so we can have that discussion in the other PR.

relokin:
> Thanks @relokin , sorry, I have a couple more comments/questions.
> 

Thanks for the review @fsestini.

> A couple of small comments in the code, and a more general comment/question regarding `miaou`: the PR adds support for some cat builtin functions like `domain`, `same-oa`, etc. However there are no changes to the end-to-end/regression test suite in `tools/tests/miaou.t` that reflect this change. Not a problem per se, but my question is:
>

`tools/tests/miaou.t` is includes the definitions we transliterate and include in B2.3. I decided to keep that way and since the BBM clauses will not appear in B2.3 I thought it would be ok to not test them for now.

If you would like to see them tested I will create another .t file.

>     * do these changes to miaou affect the transliteration of any relation of interest in aarch64, and if so, what are those relations, and should we add them to the test suite?

Not really, these new clauses are not going to be included in B2.3. However, I think it *might* be useful to have a way to transliterate them.

> 
>     * if this change to miaou does not affect any aarch64 relation of interest, then I wonder what motivated the change in the first place.

I will drop the commit that adds support for domain.

fsestini:
Thanks for the explanation @relokin . I agree with leaving the treatment of `domain` for a later PR.

> I will drop the commit that adds support for domain.

Happy to approve this PR without any other changes once this commit is dropped.

fsestini (approved):

relokin:
Thanks @fsestini, @psafont and @TiberiuBucur for your reviews!

PR #1865, opened 2026-06-16
[herd] Measure some timings

maranget:
A timer support is addded (cf. lib/timer.ml). The new feature is  used for timing:
  + Cat model execution and
  + code semantics, _i.e._ monad construction by the "Sem" modules.

Timers are activated by the command-line option `-debug timers`. For instance:
```
% herd7 catalogue/aarch64-VMSA/tests/Marc03.litmus -variant vmsa -speedcheck true -debug timers
Test Marc03 Forbidden
...
Timers: model=13.25, run=18.61, semantics=0.02
```

### DIFF
--- a/herd-www/jerd.ml
+++ b/herd-www/jerd.ml
@@ -217,6 +217,7 @@ let run_herd bell cat litmus cfg cat_label =
     let check_filter = !check_filter
     let debug = !debug
     let debuglexer = debug.Debug_herd.lexer
+    module Timer = Timer.No
     let verbose = !verbose
     let hexa = !PP.hexa
     let unroll = !unroll
--- a/herd/AArch64ASLSem.ml
+++ b/herd/AArch64ASLSem.ml
@@ -1663,10 +1663,12 @@ module Make (TopConf : AArch64Sig.Config) (V : Value.AArch64ASL) :
           profile "build AArch64 semantics from ASL" @@ fun () ->
           let test_asl, eqs_test = fake_test ii fname args in
           let model = build_model_from_file "asl.cat" in
+          TopConf.C.Timer.start Timer.sem_key ;
           let { MC.event_structures = rfms; _ }, test_asl =
             profile "run ASL Semantics" @@ fun () ->
             MC.glommed_event_structures ~is_pgm:false test_asl
           in
+          TopConf.C.Timer.stop Timer.sem_key ;
           let () =
             if _dbg then
               Printf.eprintf "Got rfms back: %d of them.\n%!" (List.length rfms)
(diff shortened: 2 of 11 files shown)


fsestini:
Hi @maranget . This PR looks like it would create quite broad conflicts with https://github.com/herd/herdtools7/pull/1833 , which has been waiting in the pipeline for quite some time and I'd be keen to see merged within this week. In light of this, I'd appreciate if you could rebase this PR on https://github.com/herd/herdtools7/pull/1833 and develop this timer feature based on those changes.

maranget:
Hi @festini, fair enough.  I had forgotten PR #1833, which has priority,

fsestini:
> Hi @Festini, fair enough. I had forgotten PR #1833, which has priority,

Thank you!

maranget:
Hi  @fsestini . I have just rebased this "global timers" PR on top of master, _i.e._ on top of the new organisation. I have noticed that `Top_herd.Printer`, in charge of outputing test results now uses the `Format` module, in place of ordinary  output channels, I assume the rationale is to have output to a buffer of string easily. 

In any case would you please check  that timers ouput is at the right place (at the end of `pp_stats` in `Top_Herd.Printer`).

fsestini:
Thanks for rebasing this. I think this is a useful feature, and the printing logic is in the right place. Indeed, `Format` is used for the reasons you state.

Having said that, w.r.t. the PR as a whole, in my opinion the proposed timer infrastructure might be more complex than it needs to be. It would be great if you could take the time to simplify this a bit.

In particular:

- I do not think timers need to be yet another functor. The presence of that functor means the timer module has to be threaded through configs, and it prevents us from using timers as immediate first-class values. Considering this is only to support a single config value, I'm not sure it's really paying off.
- I would avoid representing timers indirectly as string keys, as that adds indirection and consequently extra ceremony when one tries to use timer keys that "don't exist". I'd represent timers as first class values with a flat interface like so:

  ```ocaml
  type t

  val create : string -> t
  val start : t -> unit
  val stop : t -> unit
  ...
  ```

- I would separate the concern of collecting metrics with that of reporting them. Whether timings are printed because `-debug timers` is enabled feels like a presentation/reporting decision of the external client (in this case, `cli.ml`), not something that should be tied to the interface of the timer API. I think it would be cleaner if we let timers always collect data, and then conditionally decide whether to print them separately, though we could also add some simple disabling mechanism later on if needed.

I've implemented these suggestions in a commit [here](https://github.com/herd/herdtools7/compare/master...fsestini:herdtools7:global-profile-2) to make things clearer.

EDIT: I've now realised we could make this even simpler: we don't really need to print profiling data in anything except when running herd7 as a CLI, I'm assuming. Therefore, the timers printing logic can live entirely in cli.ml rather than top_herd.ml. I've updated my sample commit accordingly.

maranget:
Hi @fsestini. You are right the indirection through strings does not seen useful. I'all look at your solution,

maranget:
Superseded by PR #1872.

PR #1870, opened 2026-06-18
[asl] maintain a type annotation for every typechecked expression

Roman-Manevich:
This PR enables maintaining the type inferred for each expression during typechecking as an annotation attached to the expression. The intention is to make this available for tools built on top of `aslref`.

### DIFF
--- a/asllib/AST.mli
+++ b/asllib/AST.mli
@@ -33,13 +33,6 @@
 type version = V0 | V1
 type position = Lexing.position
 
-type 'a annotated = {
-  desc : 'a;
-  pos_start : position;
-  pos_end : position;
-  version : version;
-}
-
 type identifier = string
 (** Type of local identifiers in the AST. *)
 
@@ -146,8 +139,22 @@ type subprogram_type =
       (** A setter is a special procedure called with a syntax similar to slice
           assignment. *)
 
+type 'a annotated = {
+  desc : 'a;
+  pos_start : position;
+  pos_end : position;
+  version : version;
+  ty_opt : ty option;
+      (** An optional type annotation, added only to typed versions of [expr]
+          and [lexpr] types. This is added by the typechecker once the type is
+          inferred, and never used by the typechecker itself. In particular,
+          expressions synthesized by the typechecker are not associated with a
+          type annotation. The intended usage is for tools built on top of
+          aslref. *)
+}
+
 (** Expressions. Parametric on the type of literals. *)
-type expr_desc =
+and expr_desc =
   | E_Literal of literal
   | E_Var of identifier
   | E_ATC of expr * ty  (** Asserted type conversion *)
--- a/asllib/AST.mli
+++ b/asllib/AST.mli
@@ -50,6 +50,27 @@ type precision_loss_flag =
       (** A loss of precision comes with a list of warnings that can explain why
           the loss of precision happened. *)
 
+type ('a, 't) t_annotated = {
+  desc : 'a;
+  pos_start : position;
+  pos_end : position;
+  version : version;
+  ty_opt : 't option;
+      (** An optional type annotation, added only to typed versions of [expr]
+          and [lexpr] types. This is added by the typechecker once the type is
+          inferred, and never used by the typechecker itself. In particular,
+          expressions synthesized without a corresponding typing step, such as
+          during symbolic normalization, may have no type annotation. The
+          intended usage is for tools built on top of aslref. *)
+}
+
+(** An empty type to force the type annotation to [None]. *)
+type no_type_annotation = |
+
+type 'a annotated = ('a, no_type_annotation) t_annotated
+(** A ['a annotated] type is a type wrapped in position annotation but without a
+    type annotation. *)
+
 (* -------------------------------------------------------------------------
 
                                    Operations
@@ -139,22 +160,8 @@ type subprogram_type =
       (** A setter is a special procedure called with a syntax similar to slice
           assignment. *)
 
-type 'a annotated = {
-  desc : 'a;
-  pos_start : position;
-  pos_end : position;
-  version : version;
-  ty_opt : ty option;
-      (** An optional type annotation, added only to typed versions of [expr]
-          and [lexpr] types. This is added by the typechecker once the type is
-          inferred, and never used by the typechecker itself. In particular,
-          expressions synthesized by the typechecker are not associated with a
-          type annotation. The intended usage is for tools built on top of
-          aslref. *)
-}
-
 (** Expressions. Parametric on the type of literals. *)
-and expr_desc =
+type expr_desc =
   | E_Literal of literal
   | E_Var of identifier
   | E_ATC of expr * ty  (** Asserted type conversion *)
@@ -179,7 +186,7 @@ and expr_desc =
   | E_Arbitrary of ty
   | E_Pattern of expr * pattern_matcher
 
-and expr = expr_desc annotated
+and expr = (expr_desc, ty) t_annotated
 
 and pattern_desc =
   | Pattern_All
@@ -301,7 +308,7 @@ type lexpr_desc =
           Third argument is a type annotation. *)
   | LE_Destructuring of lexpr list
 
-and lexpr = lexpr_desc annotated
+and lexpr = (lexpr_desc, ty) t_annotated
 
 type local_decl_keyword = LDK_Var | LDK_Let
 
--- a/asllib/ASTUtils.ml
+++ b/asllib/ASTUtils.ml
@@ -67,7 +67,7 @@ let default_version = V1
 let desc v = v.desc
 
 let annotated desc pos_start pos_end version =
-  { desc; pos_start; pos_end; version }
+  { desc; pos_start; pos_end; version; ty_opt = None }
 
 let add_dummy_annotation ?(version = default_version) desc =
   annotated desc dummy_pos dummy_pos version
@@ -88,8 +88,12 @@ let add_pos_range_from pos_from pos_to desc =
     pos_start = pos_from.pos_start;
     pos_end = pos_to.pos_end;
     version = pos_from.version;
+    ty_opt = None;
   }
 
+let expr_ty_annot_from ~src expr = { expr with ty_opt = src.ty_opt }
+let expr_with_ty_annot ty expr = { expr with ty_opt = Some ty }
+let lexpr_with_ty_annot ty lexpr = { lexpr with ty_opt = Some ty }
 let map_desc f thing = f thing |> add_pos_from thing
 let map_annotated thing f = f thing.desc |> add_pos_from thing
 
@@ -108,6 +112,7 @@ let add_pos_from_pos_of ((fname, lnum, cnum, enum), desc) =
     pos_start = { common with pos_cnum = cnum };
     pos_end = { common with pos_cnum = enum };
     version = default_version (* used only in testing *);
+    ty_opt = None;
   }
 
 let list_fold_lefti f accu l =
@@ -184,6 +189,8 @@ let map2_desc f thing1 thing2 =
     pos_start = thing1.pos_start;
     pos_end = thing2.pos_end;
     version = thing1.version;
+    ty_opt =
+      (match thing1.ty_opt with Some t1 -> Some t1 | None -> thing2.ty_opt);
   }
 
 let s_pass = add_dummy_annotation S_Pass
@@ -437,28 +444,33 @@ and pattern_matcher_equal eq (ps1, pk1) (ps2, pk2) =
   List.equal (pattern_equal eq) ps1 ps2 && pattern_kind_equal pk1 pk2
 
 let qualifier_equal (q1 : func_qualifier option) q2 = Option.equal ( = ) q1 q2
-let var_ x = E_Var x |> add_dummy_annotation
+let expr_of_var x = E_Var x |> add_dummy_annotation
 let binop op = map2_desc (fun e1 e2 -> E_Binop (op, e1, e2))
-let unop op = map_desc (fun e -> E_Unop (op, e))
+let neg e = E_Unop (NEG, e) |> add_pos_from e
 let literal v = E_Literal v |> add_dummy_annotation
 let expr_of_int i = literal (L_Int (Z.of_int i))
 let expr_of_z z = literal (L_Int z)
-let e_true = literal (L_Bool true)
-let e_false = literal (L_Bool false)
+let e_true = literal (L_Bool true) |> expr_with_ty_annot boolean
+let e_false = literal (L_Bool false) |> expr_with_ty_annot boolean
 let expr_of_bool b = if b then e_true else e_false
 let zero_expr = expr_of_z Z.zero
 let one_expr = expr_of_z Z.one
 let minus_one_expr = expr_of_z Z.minus_one
 
 let expr_of_rational q =
+  expr_with_ty_annot real
+  @@
   if Z.equal (Q.den q) Z.one then expr_of_z (Q.num q)
   else binop `DIV (expr_of_z (Q.num q)) (expr_of_z (Q.den q))
 
+(** [mul_expr e1 e2] symbolically multiplies the two expression [e1] and [e2].
+*)
 let mul_expr e1 e2 =
   if expr_equal (fun _ _ -> false) e1 one_expr then e2
   else if expr_equal (fun _ _ -> false) e2 one_expr then e1
   else binop `MUL e1 e2
 
+(** [pow_expr e p] symbolically raises the expression [e] to the power [p]. *)
 let pow_expr e = function
   | 0 -> one_expr
   | 1 -> e
@@ -529,12 +541,12 @@ let slice_as_single = function
   | Slice_Single e -> e
   | _ -> raise @@ Invalid_argument "slice_as_single"
 
-let default_t_bits = T_Bits (E_Var "-" |> add_dummy_annotation, [])
+let discard_expr = E_Var "-" |> add_dummy_annotation
+let default_t_bits = T_Bits (discard_expr, [])
 
 let default_array_ty =
-  let len = E_Var "-" |> add_dummy_annotation in
   let ty = T_Named "-" |> add_dummy_annotation in
-  T_Array (len, ty)
+  T_Array (discard_expr, ty)
 
 let identifier_of_decl d =
   match d.desc with
--- a/asllib/ASTUtils.ml
+++ b/asllib/ASTUtils.ml
@@ -69,15 +69,15 @@ let desc v = v.desc
 let annotated desc pos_start pos_end version =
   { desc; pos_start; pos_end; version; ty_opt = None }
 
-let add_dummy_annotation ?(version = default_version) desc =
+let add_dummy_pos ?(version = default_version) desc =
   annotated desc dummy_pos dummy_pos version
 
-let dummy_annotated = add_dummy_annotation ()
+let dummy_annotated = add_dummy_pos ()
 let to_pos pos = { pos with desc = (); ty_opt = None }
-let is_dummy_annotated x = x.pos_end == dummy_pos || x.pos_start == dummy_pos
+let is_dummy_pos x = x.pos_end == dummy_pos || x.pos_start == dummy_pos
 
 let add_pos_from_st pos desc =
-  if pos.desc == desc then pos else { pos with desc }
+  if pos.desc == desc then pos else { pos with desc; ty_opt = None }
 
 let add_pos_from pos desc = { pos with desc; ty_opt = None }
 
@@ -98,7 +98,7 @@ let map_annotated thing f = f thing.desc |> add_pos_from thing
 
 let add_maybe_loc ?loc thing =
   match loc with
-  | None -> add_dummy_annotation thing
+  | None -> add_dummy_pos thing
   | Some loc -> add_pos_from loc thing
 
 let add_pos_from_pos_of ((fname, lnum, cnum, enum), desc) =
@@ -188,16 +188,16 @@ let map2_desc f v1 v2 =
     pos_start = v1.pos_start;
     pos_end = v2.pos_end;
     version = v1.version;
-    ty_opt = (match v1.ty_opt with Some t -> Some t | None -> v2.ty_opt);
+    ty_opt = None;
   }
 
-let s_pass = add_dummy_annotation S_Pass
+let s_pass = add_dummy_pos S_Pass
 let s_then = map2_desc (fun s1 s2 -> S_Seq (s1, s2))
-let boolean = T_Bool |> add_dummy_annotation
-let string = T_String |> add_dummy_annotation
-let real = T_Real |> add_dummy_annotation
+let boolean = T_Bool |> add_dummy_pos
+let string = T_String |> add_dummy_pos
+let real = T_Real |> add_dummy_pos
 let integer' = T_Int UnConstrained
-let integer = integer' |> add_dummy_annotation
+let integer = integer' |> add_dummy_pos
 
 let well_constrained' ?(precision = Precision_Full) cs =
   T_Int (WellConstrained (cs, precision))
@@ -442,10 +442,10 @@ and pattern_matcher_equal eq (ps1, pk1) (ps2, pk2) =
   List.equal (pattern_equal eq) ps1 ps2 && pattern_kind_equal pk1 pk2
 
 let qualifier_equal (q1 : func_qualifier option) q2 = Option.equal ( = ) q1 q2
-let expr_of_var x = E_Var x |> add_dummy_annotation
+let expr_of_var x = E_Var x |> add_dummy_pos
 let binop op = map2_desc (fun e1 e2 -> E_Binop (op, e1, e2))
 let neg e = E_Unop (NEG, e) |> add_pos_from e
-let literal v = E_Literal v |> add_dummy_annotation
+let literal v = E_Literal v |> add_dummy_pos
 let expr_of_int i = literal (L_Int (Z.of_int i))
 let expr_of_z z = literal (L_Int z)
 let e_true = literal (L_Bool true) |> with_ty_annot boolean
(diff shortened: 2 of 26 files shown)


asllib/AST.mli
```diff
@@ -146,8 +139,16 @@ type subprogram_type =
       (** A setter is a special procedure called with a syntax similar to slice
           assignment. *)
 
+type 'a annotated = {
+  desc : 'a;
+  pos_start : position;
+  pos_end : position;
+  version : version;
+  opt_type : ty option;
```
hrutvik: nit: `ty_opt`?

asllib/ASTUtils.ml
```diff
@@ -193,14 +200,17 @@ let string = T_String |> add_dummy_annotation
...
 
 let well_constrained' ?(precision = Precision_Full) cs =
   T_Int (WellConstrained (cs, precision))
 
 let well_constrained ?loc ?precision cs =
   well_constrained' ?precision cs |> add_maybe_loc ?loc
 
-let integer_exact' e = well_constrained' [ Constraint_Exact e ]
+let integer_exact' e =
+  well_constrained' [ Constraint_Exact (e |> with_integer_ty) ]
```
hrutvik: Is this right? I'm a bit surprised that this is editing the type of `e`

asllib/ASTUtils.ml
```diff
@@ -448,38 +458,49 @@ and pattern_equal eq p1 p2 =
   | _ -> false
 
 let qualifier_equal (q1 : func_qualifier option) q2 = Option.equal ( = ) q1 q2
-let var_ x = E_Var x |> add_dummy_annotation
+let int_expr_of_var x = E_Var x |> add_dummy_annotation |> with_integer_ty
```
hrutvik: In general, the usage of `with_integer_ty` seems suspect to me - how do we know whether it should be (un)constrained, parameterised? I think we should be relying on type-checking to fill this in wherever possible, and avoiding overwriting types as we do several times in this file

asllib/AST.mli
```diff
@@ -146,8 +139,16 @@ type subprogram_type =
...
+}
```
HadrienRenaud: I would keep it separated with the following:

```suggestion
type ('term, 'ty) annotated = {
  desc : 'term;
  pos_start : position;
  pos_end : position;
  version : version;
  opt_type : 'ty option;
}
```

And then declare `expr` as `(expr_desc, ty) annotated`, while for example `stmt` could be `(stmt_desc, unit) annotated`

asllib/ASTUtils.ml
```diff
@@ -927,3 +949,210 @@ let plug_primitives ast =
       raise (Invalid_argument "plug_primitives: badly referenced primitive")
   in
   List.map plug_one
+
+(** Checks that all expressions have [opt_type = Some t] *)
+module CheckTypeAnnotations = struct
```
hrutvik: Why are we performing this check? This is just additional information, it's not mandated?

asllib/ASTUtils.ml
HadrienRenaud: Why this? At the very least, the choice of typing those as `integer` needs to be documented and justified. Also what needs to be justified is that this does not break anything, as now a naive implementation of `type_of` would find those of type `integer` and not `integer {e}`, and maybe this breaks the invariant that all expressions in types are bounded.

IMHO if it is possible, we would leave those without a type.

Naive implementation of `type_of` for reference.
```ocaml
let type_of e =
  match e.type_opt with
  | Some t -> t
  | None -> Typing.annotate_expr e |> snd
```
Note that here `type_of one_expr != Typing.annotate_expr one_expr |> snd`

asllib/Typing.ml
```diff
@@ -2426,6 +2438,12 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
       [loc]. Note however that a bit vector with width [N] can always be
       generated using [0[:N]]. *)
   let rec base_value_v1 ~loc env t : expr =
+    let value = base_value_v1_impl ~loc env t in
+    { value with opt_type = Some t }
```
hrutvik: It looks like `base_value` below already does this? I.e. `base_value_v1` and `base_value` are duplicating the addition of `opt_type`

asllib/Typing.ml
```diff
@@ -3971,11 +4007,18 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
           let new_ty =
             if extra_fields = [] then t1
             else
+              let annotated_extra_fields =
+                List.map
+                  (fun (field, ty) ->
+                    let new_ty, _ses = annotate_type ~loc env ty in
+                    (field, new_ty))
+                  extra_fields
+              in
```
hrutvik: This is new behaviour right? Why is it necessary (i.e. is it a bug that you're also fixing)?

asllib/Native.ml
```diff
@@ -311,14 +311,17 @@ module NativeBackend (C : Config) = struct
     let round_towards_zero = wrap_real_to_int "RoundTowardsZero" truncate
 
     let primitives =
-      let e_var x = E_Var x |> add_dummy_annotation in
+      let e_int_var x = E_Var x |> add_dummy_annotation |> with_integer_ty in
```
HadrienRenaud: This also needs to be justified.

asllib/Typing.ml
```diff
@@ -1927,7 +1930,16 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
...
+    let t, e', ses = annotate_expr_impl env e in
+    let e_with_type = { e' with opt_type = Some t } in
+    (t, e_with_type, ses)
+
+  (** [annotate_expr_impl] annotates the expression [e] in [env] without setting
+      a type annotation. *)
+  and annotate_expr_impl env (e : expr) : ty * expr * SES.t =
     let () = if false then Format.eprintf "@[Annotating %a@]@." PP.pp_expr e in
     let here x = add_pos_from ~loc:e x and loc = to_pos e in
     match e.desc with
```
HadrienRenaud, line 1881: You can avoid putting a new function in this very recursive definition with the following:

```suggestion
  (** [annotate_expr] annotates the expression [e] in [env]. *)
  and annotate_expr env (e : expr) : ty * expr * SES.t =
    let () = if false then Format.eprintf "@[Annotating %a@]@." PP.pp_expr e in
    let here x = add_pos_from ~loc:e x and loc = to_pos e in
    set_type3
    @@
    match e.desc with
```

Maybe in `ASTUtils.ml`
```ocaml
let set_type3: 'ty * ('val, 'ty) annotated * 'data -> 'ty * ('val, 'ty) annotated * 'data =
  fun (t, e, d) -> (t, with_ty_annot t e, d)
```

asllib/ASTUtils.ml
Roman-Manevich: I used it for testing and then turned it off. We can probably remove it, but we may want to preserve it until we have more validation by some actual use cases.

asllib/ASTUtils.mli
```diff
@@ -93,6 +93,12 @@ val add_pos_from_st : 'a annotated -> 'a -> 'a annotated
     If both arguments are physically equal, then the result is also physically
     equal. *)
 
+val with_ty_annot : AST.ty -> 'a annotated -> 'a annotated
```
HadrienRenaud: With my proposal for the definition of annotated, it can have a more flexible definition, without losing the type constraints:

```suggestion
val with_ty_annot : 'ty -> ('a, 'ty) annotated -> ('a, 'ty) annotated
```

(with a slight change in how `with_ty_annot` is written, we could also transform it to `'ty -> ('a, 'b) annotated -> ('a, 'ty) annotated`)

asllib/Typing.ml
Roman-Manevich: removed duplication.

asllib/Typing.ml
Roman-Manevich: This is actually a bug fix. We should have annotated the extra fields.

asllib/ASTUtils.mli
```diff
@@ -166,14 +172,19 @@ val expr_of_int : int -> expr
...
 val binop : binop -> expr -> expr -> expr
 (** Builds a binary operation from to subexpressions. *)
 
-val unop : unop -> expr -> expr
-(** Builds a unary operation from its subexpression. *)
+val integer_binop : binop -> expr -> expr -> expr
+(** Builds a binary operation from to subexpressions, with the assumption that
+    both subexpressions are of type [integer]. *)
+
+val integer_unop : unop -> expr -> expr
```
HadrienRenaud: Is there not just one integer unary operation? In this case shouldn't we just export `neg` instead of this?

asllib/ASTUtils.ml
Roman-Manevich: The is needed for the cases where we are constructing typed AST expressions without later annotating them. In such cases, we need to establish the types ourselves.

asllib/ASTUtils.ml
Roman-Manevich: This replaces `e` with a copy that has the integer type as its type annotation

asllib/Typing.ml
```diff
@@ -4429,7 +4468,20 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
...
       (List.rev ast_rev, env)
 
-  let type_check_ast ast = type_check_ast_in_env empty_global ast
+  let type_check_ast ast =
+    let annotated_ast, genv = type_check_ast_in_env empty_global ast in
+    let should_check_type_annotations =
+      match C.check with
+      | TypeCheck ->
+          false (* set to true if debugging missing annotations is needed *)
+      | _ -> false
```
HadrienRenaud: Can you not add a command line flag for this? I'm wondering if adding the type information should also be a command line flag.

HadrienRenaud (commented):
I think this needs to be more carefully documented. For example one of the important thing that is not said is that none of those type informations are used for type-checking.

Another important choice made here is not to use dependent types for some constructed expressions. This needs to be justified. I can see other options:
1. Do not put any type annotations for those.
2. Put only integer, as is done here 
3. Use [recursive values](https://ocaml.org/manual/5.4/letrecvalues.html) to provide full precision types for literal values (this seems dangerous to me).

Maybe this should be hidden behind a command line flag not to impact the reference interpreter.
In the long term, we should probably write a typed AST which would incorporate those informations, and document what precision is given by those.

asllib/Typing.ml
HadrienRenaud: Can we move the bug fix to a new PR please?

asllib/Native.ml
Roman-Manevich: I'm not sure in what sense this needs to be justified as this is a construction function. Each of the usages indeed requires an integer-typed variable (for bitwidth, for an integer exponentiation operation, and for the integer variable `N`.

asllib/Typing.ml
Roman-Manevich: Thanks. I went with `set_expr_type_annotation` instead of `set_type3`.

asllib/ASTUtils.ml
hrutvik: Would we not expect `e` to be type-annotated already? If not, does the integer type annotation add anything?

asllib/ASTUtils.ml
hrutvik: I understand that we may not have annotated these, but how can we know that `integer` is the right type?

asllib/ASTUtils.ml
hrutvik: I think we should remove for the merge of this PR.
 
> we may want to preserve it until we have more validation by some actual use cases.

I think the opposite actually - this is speculating a need for exhaustivity that we don't yet have supported by usage. I suggest we start with a basic level of annotation and expand as necessary

asllib/ASTUtils.ml
hrutvik: > IMHO if it is possible, we would leave those without a type.

Agreed with this - I think any type annotations we introduce in the first instance should be coming out of the existing type-checking

asllib/Native.ml
hrutvik: However it doesn't justify why this shouldn't be a constrained or parameterised integer. For example I see it used in a `T_Bits`, which usually needs a constrained or parameterised integer right?

hrutvik (changes requested):
I think we need to reduce the scope of this PR. I think it should contain only:

- Mechanism for storing type annotations - whether the existing version or Hadrien's proposal.
- Storing type annotations as calculated during type-checking already (i.e. wrap `annotate_{expr,lexpr}`, `base_value`)

I.e. the additional type annotations not provided by the type system, the exhaustivity checks should be removed. In future we can discuss whether we consider "missing" type annotations as truly missing (and if so whether they should be inserted post-hoc or the type system should be changed to calculate them appropriately), and how this could all be encoded into a truly typed AST.

Please let me know if you disagree.

asllib/ASTUtils.ml
HadrienRenaud: Note that it is possible that we could type-check those expressions, even if it's a bit useless.

asllib/ASTUtils.ml
Roman-Manevich: rolled back

asllib/ASTUtils.ml
Roman-Manevich: rolled back

asllib/ASTUtils.ml
Roman-Manevich: rolled back

asllib/ASTUtils.ml
Roman-Manevich: rolled back

asllib/Native.ml
Roman-Manevich: rolled back

asllib/Typing.ml
Roman-Manevich: rolled back

asllib/AST.mli
Roman-Manevich: This is cleaner in that it utilizes the type system, but moving to `(expr_desc, ty) annotated` but it is quite a pervasive change. Many signatures would have to change. Since the type is optional, I think it already signals that not all nodes are expected to have a type associated with them. What do you think?

asllib/AST.mli
```diff
@@ -147,7 +156,7 @@ type subprogram_type =
           assignment. *)
 
 (** Expressions. Parametric on the type of literals. *)
-type expr_desc =
+and expr_desc =
```
HadrienRenaud: Remains of a previous change?

```suggestion
type expr_desc =
```

asllib/ASTUtils.ml
```diff
@@ -178,12 +182,25 @@ let pair x y = (x, y)
 let pair' y x = (x, y)
 let pair_equal f g (x1, y1) (x2, y2) = f x1 x2 && g y1 y2
 
+(** [map2_exprs f e1 e2] applies [f] to [e1] and [e2] and yields an expression
+    with positions ranging from the start of [e1] to the end of [e2] and the
+    first type annotation of [e1] and [e2] that is not [None]. *)
+let map2_exprs f e1 e2 =
```
HadrienRenaud: I think this can be the implementation of `map2_desc` if we change its signature to have a single type annotation type:

```ocaml
val map2_desc :
  (('a, 't) t_annotated -> ('b, 't) t_annotated -> 'c) ->
  ('a, 't) t_annotated ->
  ('b, 't) t_annotated ->
  ('c, 't) t_annotated
```

Then you can unify `map2_exprs` and `map2_desc`

asllib/AST.mli
```diff
@@ -57,6 +50,22 @@ type precision_loss_flag =
...
+  ty_opt : 't option;
+      (** An optional type annotation, added only to typed versions of [expr]
+          and [lexpr] types. This is added by the typechecker once the type is
+          inferred, and never used by the typechecker itself. In particular,
+          expressions synthesized by the typechecker are not associated with a
+          type annotation. The intended usage is for tools built on top of
+          aslref. *)
+}
+
+type 'a annotated = ('a, unit) t_annotated
```
HadrienRenaud: Thinking about it, `unit` here is too big for what we're trying to do. Let's use an empty type, which will guarantee that no `'a annotated` type has ever a type annotation.

```suggestion
type no_type_annotation = |
(** An empty type to force the type annotation to [None]. *)

type 'a annotated = ('a, no_type_annotation) t_annotated
(** A ['a annotated] type is a type wrapped in position annotation but without any type-annotation. *)
```

(I have tested this this looks like it works without any change in the code base except this)

asllib/desugar.mli
```diff
@@ -105,8 +105,7 @@ val desugar_setter_setfields :
     Case statements
    ------------------------------------------------------------------------- *)
 
-val desugar_case_stmt :
-  expr_desc annotated -> case_alt_desc annotated list -> stmt -> stmt_desc
```
HadrienRenaud, line 109: This was such a strange type declaration.

HadrienRenaud (approved):
This looks good to me, thanks @Roman-Manevich, I've just got a few improvements.

asllib/AST.mli
Roman-Manevich: thanks

asllib/ASTUtils.ml
```diff
@@ -178,12 +182,13 @@ let pair x y = (x, y)
...
   {
-    desc = f thing1 thing2;
-    pos_start = thing1.pos_start;
-    pos_end = thing2.pos_end;
-    version = thing1.version;
+    desc = f v1 v2;
+    pos_start = v1.pos_start;
+    pos_end = v2.pos_end;
+    version = v1.version;
+    ty_opt = (match v1.ty_opt with Some t -> Some t | None -> v2.ty_opt);
```
HadrienRenaud: Actually thinking about it this is probably a bad idea. For example, using it for `binop` is lossy, if you had `(x: integer {10..20}) + (y: integer {30..40})` then the result would be `(x + y : integer {10..20})` (!).

```suggestion
    ty_opt = None;
```

asllib/ASTUtils.mli
```diff
@@ -56,49 +56,69 @@ val dummy_pos : Lexing.position
...
 
-val dummy_annotated : unit annotated
-(** A dummy annotation *)
+val dummy_annotated : (unit, _) t_annotated
+(** A dummy annotation.
 
-val is_dummy_annotated : 'a annotated -> bool
+    Type annotation is always [None]. *)
+
+val is_dummy_annotated : _ t_annotated -> bool
```
hrutvik: nit: it may be more informative to rename this to `add_dummy_pos`, `dummy_pos`, `is_dummy_pos`?

asllib/ASTUtils.ml
```diff
@@ -67,19 +67,19 @@ let default_version = V1
...
 let add_dummy_annotation ?(version = default_version) desc =
   annotated desc dummy_pos dummy_pos version
 
 let dummy_annotated = add_dummy_annotation ()
-let to_pos pos = { pos with desc = () }
+let to_pos pos = { pos with desc = (); ty_opt = None }
 let is_dummy_annotated x = x.pos_end == dummy_pos || x.pos_start == dummy_pos
 
 let add_pos_from_st pos desc =
   if pos.desc == desc then pos else { pos with desc }
```
hrutvik: Clear the type annotation here in the case where the two are not physically equal?

asllib/ASTUtils.mli
```diff
@@ -56,49 +56,69 @@ val dummy_pos : Lexing.position
...
+    Type annotation in the result is always [None]. *)
+
+val add_pos_range_from :
+  ('a, _) t_annotated -> ('a, _) t_annotated -> 'b -> ('b, _) t_annotated
 (** [add_pos_range_from loc_from loc_to v] is [v] with the location data given
     by the range of locations starting at [loc_from] and ending to [loc_to]. *)
 
-val add_pos_from_st : 'a annotated -> 'a -> 'a annotated
+val add_pos_from_st : ('a, 't) t_annotated -> 'a -> ('a, 't) t_annotated
 (** [add_pos_from_st a' a] is [a] with the location from [a'].
```
hrutvik, line 99: Type annotation should be removed if the arguments aren't physically equal?

asllib/ASTUtils.mli
```diff
@@ -56,49 +56,69 @@ val dummy_pos : Lexing.position
...
-  'b annotated ->
-  'c annotated
-(** Folder on two annotated types. *)
+  (('a, 'ty) t_annotated -> ('b, 'ty) t_annotated -> 'c) ->
+  ('a, 'ty) t_annotated ->
+  ('b, 'ty) t_annotated ->
+  ('c, 'ty) t_annotated
+(** [map2_desc f v1 v2] folds two annotated values via [f] and yields a value
+    with positions ranging from the start of [v1] to the end of [v2] and the
+    first type annotation of [v1] and [v2] that is not [None]. *)
```
hrutvik: Echoing @HadrienRenaud's comment, and this comment will also need updating

asllib/error.ml
```diff
@@ -671,7 +671,7 @@ module PPrint = struct
   let pp_warning f e =
     fprintf f "@[<v 0>%a%a@]" pp_pos_begin e pp_warning_desc e
 
-  let error_desc_to_string = asprintf "%a" pp_error_desc
+  let error_desc_to_string e = asprintf "%a" pp_error_desc e
```
hrutvik, line 674: nit: why this change?

asllib/Typing.ml
```diff
@@ -36,11 +36,19 @@ let undefined_identifier ~loc x =
 
 let invalid_expr e = fatal_from ~loc:e (Error.InvalidExpr e)
 let add_pos_from ~loc = add_pos_from loc
+let set_expr_type_annotation (t, e, ses) = (t, with_ty_annot t e, ses)
+
+(** [set_lexpr_type_annotation (lexpr_ty, lexpr, ses)] sets [lexpr_ty] as the
```
hrutvik: nit: could also add a docstring for the new `set_expr_type_annotation` above?

asllib/Typing.ml
```diff
@@ -36,11 +36,19 @@ let undefined_identifier ~loc x =
...
+    type annotation of [lexpr] and returns the updated assignable expression
+    together with [lexpr_ty] and [ses]. *)
+let set_lexpr_type_annotation (lexpr_ty, lexpr, ses) =
```
hrutvik: Hmm both `set_{expr,lexpr}_type_annotation` are used only once - I think it would make sense to define them where they are used

asllib/Typing.ml
```diff
@@ -1868,9 +1877,13 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
       ret_ty_opt,
       ses3 )
 
+  (** [annotate_expr] annotates the expression [e] in [env] and sets a type
+      annotation. *)
   and annotate_expr env (e : expr) : ty * expr * SES.t =
     let () = if false then Format.eprintf "@[Annotating %a@]@." PP.pp_expr e in
     let here x = add_pos_from ~loc:e x and loc = to_pos e in
+    set_expr_type_annotation
```
hrutvik, line 1879:
```suggestion
    let here x = add_pos_from ~loc:e x and loc = to_pos e in
    let set_expr_type_annotation (t, e, ses) = (t, with_ty_annot t e, ses) in
    set_expr_type_annotation
```

asllib/Typing.ml
```diff
@@ -2379,12 +2394,11 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
             if length < 0 then fatal_from ~loc @@ Error.BaseValueEmptyType t
             else L_BitVector (Bitvector.zeros length) |> lit
         | _ ->
-            let zero = L_Int Z.zero |> lit in
-            let slice = Slice_Length (zero, e) in
-            E_Slice (zero, [ slice ]) |> here)
+            let slice = Slice_Length (zero_expr, e) in
```
hrutvik, line 2391: Nice, thanks for tidying these

asllib/Typing.ml
```diff
@@ -2465,17 +2479,22 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
...
+      and sets its actual type. *)
+  let rec annotate_lexpr_ty env le t_e =
     let () =
       if false then
         Format.eprintf "Typing lexpr: @[%a@] to @[%a@]@." PP.pp_lexpr le
           PP.pp_ty t_e
     in
     let loc = to_pos le in
     let here x = add_pos_from ~loc x in
+    set_lexpr_type_annotation
```
hrutvik, line 2490: As above

asllib/Typing.ml
```diff
@@ -2521,38 +2540,38 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
...
-              let width =
-                slices_width env slices_annotated
-                |> StaticModel.try_normalize env
-              in
-              let t = T_Bits (width, []) |> here in
-              check_type_satisfies ~loc env t_e t ()
+            let width =
+              slices_width env slices_annotated |> StaticModel.try_normalize env
             in
+            let t = T_Bits (width, []) |> here in
```
hrutvik, line 2548: This looks much cleaner, thanks

hrutvik (approved):

asllib/error.ml
HadrienRenaud: Oh this is a type polymorphism problem, you need the argument to indicate to the compiler that this is a polymorphic function and that different calls can be instantiated with different types

hrutvik: Ahh OK, thanks

asllib/ASTUtils.ml
Roman-Manevich: agreed.

asllib/ASTUtils.mli
Roman-Manevich: renamed `add_dummy_annotation` to `add_dummy_pos` and `is_dummy_annotated` to `is_dummy_pos`.
I think `dummy_annotated` makes sense as. it is a fully-annotated dummy object and `dummy_pos` could be confused with `Lexing.dummy_pos`. I made an effort to document it.

PR #1871, opened 2026-06-18
[gen] Refactor communication edge representation

ShaleXIONG:
Unify constructors `Rf`, `Fr`, and `Ws`  in `relax.ml`  and the similar constructors `CRf`, `CFr` and `CWs` in `code.ml`. The former has been wrapped behind `Communication (com, ie)` and rename the latter constructors to `Rf`, `Fr`, and `Co`. Update edge folding, parsing, naming, relaxation handling, and generation code to use the unified communication representation while preserving existing printed syntax.

### DIFF
--- a/gen/common/code.ml
+++ b/gen/common/code.ml
@@ -160,7 +160,7 @@ let pp_com = function
   | Fr -> "Fr"
   | Co -> "Co"
 
-let fold_com f r = f Rf (f Fr (f Co r))
+let fold_com f r = f Co (f Fr (f Rf r))
 
 (* Info in tests *)
 type info = (string * string) list
--- a/gen/common/code.ml
+++ b/gen/common/code.ml
@@ -153,14 +153,14 @@ let checks =
 
 
 (* Com relation *)
-type com =  CRf | CFr | CWs
+type com =  Rf | Fr | Co
 
 let pp_com = function
-  | CRf -> "Rf"
-  | CFr -> "Fr"
-  | CWs -> "Co"
+  | Rf -> "Rf"
+  | Fr -> "Fr"
+  | Co -> "Co"
 
-let fold_com f r = f CRf (f CFr (f CWs r))
+let fold_com f r = f Rf (f Fr (f Co r))
 
 (* Info in tests *)
 type info = (string * string) list
--- a/gen/common/code.ml
+++ b/gen/common/code.ml
@@ -172,20 +172,20 @@ let checks =
 
 
 (* Com relation *)
-type com =  CRf | CFr | CWs
+type com =  Rf | Fr | Co
 
 let equal_com c1 c2 = match c1,c2 with
-  | CRf,CRf
-  | CFr,CFr
-  | CWs,CWs -> true
-  | (CRf|CFr|CWs),_ -> false
+  | Rf,Rf
+  | Fr,Fr
+  | Co,Co -> true
+  | (Rf|Fr|Co),_ -> false
 
 let pp_com = function
-  | CRf -> "Rf"
-  | CFr -> "Fr"
-  | CWs -> "Co"
+  | Rf -> "Rf"
+  | Fr -> "Fr"
+  | Co -> "Co"
 
-let fold_com f r = f CRf (f CFr (f CWs r))
+let fold_com f r = f Rf (f Fr (f Co r))
 
 (* Info in tests *)
 type info = (string * string) list
--- a/gen/common/edge.ml
+++ b/gen/common/edge.ml
@@ -50,7 +50,7 @@ module type S = sig
 
 (* edge proper *)
   type tedge =
-    | Rf of ie | Fr of ie | Ws of ie
+    | Communication of com * ie
     | Po of sd*extr*extr | Fenced of fence*sd*extr*extr
     | Dp of dp*sd*extr
     | Leave of com (* Leave thread *)
@@ -228,7 +228,7 @@ and module RMW = A.RMW = struct
 
 (* edge proper *)
   type tedge =
-    | Rf of ie | Fr of ie | Ws of ie
+    | Communication of com * ie
     | Po of sd*extr*extr | Fenced of fence*sd*extr*extr
     | Dp of dp*sd*extr
     | Leave of com
@@ -243,23 +243,23 @@ and module RMW = A.RMW = struct
 
   let is_id = function
     | Id -> true
-    | Store|Insert _|Hat|Rmw _|Rf _|Fr _|Ws _|Po (_, _, _)
+    | Store|Insert _|Hat|Rmw _|Communication _|Po (_, _, _)
     | Fenced (_, _, _, _)|Dp (_, _, _)|Leave _|Back _|Node _ -> false
 
   let is_insert_store = function
     | Store|Insert _ -> true
-    | Id|Hat|Rmw _|Rf _|Fr _|Ws _|Po (_, _, _)
+    | Id|Hat|Rmw _|Communication _|Po (_, _, _)
     | Fenced (_, _, _, _)|Dp (_, _, _)|Leave _|Back _|Node _ -> false
 
   let is_node = function
     | Node _ -> true
-    | Id|Hat|Rmw _|Rf _|Fr _|Ws _|Po (_, _, _)
+    | Id|Hat|Rmw _|Communication _|Po (_, _, _)
     | Fenced (_, _, _, _)|Dp (_, _, _)|Leave _|Back _|Insert _
     | Store -> false
 
   let is_non_pseudo = function
     | Store|Insert _ |Id|Node _-> false
-    | Hat|Rmw _|Rf _|Fr _|Ws _|Po (_, _, _)
+    | Hat|Rmw _|Communication _|Po (_, _, _)
     | Fenced (_, _, _, _)|Dp (_, _, _)|Leave _|Back _ -> true
 
   let is_dp_addr = function
@@ -285,13 +285,16 @@ and module RMW = A.RMW = struct
   | _, None, None -> ""
   | _, _, _ -> sprintf "%s%s" (pp_atom_option a1) (pp_atom_option a2)
 
+  let pp_communication_compat compat com ie =
+    let com = match com with
+    | Co when compat -> "Ws"
+    | _ -> pp_com com in
+    match ie with
+    | UnspecCom -> com
+    | _ -> sprintf "%s%s" com (pp_ie ie)
+
   let pp_tedge_compat compat = function
-    | Rf UnspecCom -> sprintf "Rf"
-    | Fr UnspecCom -> sprintf "Fr"
-    | Ws UnspecCom -> if compat then sprintf "Ws" else sprintf "Co"
-    | Rf ie -> sprintf "Rf%s" (pp_ie ie)
-    | Fr ie -> sprintf "Fr%s" (pp_ie ie)
-    | Ws ie -> if compat then sprintf "Ws%s" (pp_ie ie) else sprintf "Co%s" (pp_ie ie)
+    | Communication (com,ie) -> pp_communication_compat compat com ie
     | Po (UnspecLoc,Irr,Irr) -> "Po"
     | Po (sd,e1,e2) ->
       sprintf "Po%s%s%s" (pp_sd sd) (pp_extr e1) (pp_extr e2)
@@ -344,17 +347,17 @@ and module RMW = A.RMW = struct
 let pp_dp_default tag sd e = sprintf "%s%s%s" tag (pp_sd sd) (pp_extr e)
 
   let do_dir_tgt_com = function
-    | CRf -> Dir R
-    | CWs|CFr -> Dir W
+    | Rf -> Dir R
+    | Co|Fr -> Dir W
 
  and do_dir_src_com = function
-   | CRf|CWs -> Dir W
-   | CFr -> Dir R
+   | Rf|Co -> Dir W
+   | Fr -> Dir R
 
   let do_dir_tgt e = match e with
   | Po(_,_,e)| Fenced(_,_,_,e)|Dp (_,_,e) -> e
-  | Rf _| Hat -> Dir R
-  | Ws _|Fr _|Rmw _  -> Dir W
+  | Communication (Rf,_)| Hat -> Dir R
+  | Communication ((Co|Fr),_)|Rmw _  -> Dir W
   | Leave c|Back c -> do_dir_tgt_com c
   | Id -> NoDir
   | Insert _ -> NoDir
@@ -364,8 +367,8 @@ let pp_dp_default tag sd e = sprintf "%s%s%s" tag (pp_sd sd) (pp_extr e)
 
   and do_dir_src e = match e with
   | Po(_,e,_)| Fenced(_,_,e,_) -> e
-  | Dp _|Fr _|Hat|Rmw _ -> Dir R
-  | Ws _|Rf _ -> Dir W
+  | Dp _|Communication (Fr,_)|Hat|Rmw _ -> Dir R
+  | Communication ((Co|Rf),_) -> Dir W
   | Leave c|Back c -> do_dir_src_com c
   | Id -> NoDir
   | Insert _ -> NoDir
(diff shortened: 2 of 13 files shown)


gen/common/edge.ml
```diff
@@ -344,17 +347,17 @@ and module RMW = A.RMW = struct
...
-   | CFr -> Dir R
+   | Rf|Co -> Dir W
+   | Fr -> Dir R
 
   let do_dir_tgt e = match e with
   | Po(_,_,e)| Fenced(_,_,_,e)|Dp (_,_,e) -> e
-  | Rf _| Hat -> Dir R
-  | Ws _|Fr _|Rmw _  -> Dir W
+  | Communication (Rf,_)| Hat -> Dir R
+  | Communication ((Co|Fr),_)|Rmw _  -> Dir W
```
TiberiuBucur: Minor comment, but we could use the function we just defined above, to avoid duplication:

```
| Communication (com, _) -> do_dir_tgt_com com
```

gen/common/edge.ml
```diff
@@ -364,8 +367,8 @@ let pp_dp_default tag sd e = sprintf "%s%s%s" tag (pp_sd sd) (pp_extr e)
 
   and do_dir_src e = match e with
   | Po(_,e,_)| Fenced(_,_,e,_) -> e
-  | Dp _|Fr _|Hat|Rmw _ -> Dir R
-  | Ws _|Rf _ -> Dir W
+  | Dp _|Communication (Fr,_)|Hat|Rmw _ -> Dir R
+  | Communication ((Co|Rf),_) -> Dir W
```
TiberiuBucur: Ditto.

gen/common/edge.ml
```diff
@@ -374,19 +377,17 @@ let pp_dp_default tag sd e = sprintf "%s%s%s" tag (pp_sd sd) (pp_extr e)
...
-  let r = fold_ie wildcard (fun ie -> f (Ws ie)) r in
+  let r = fold_ie wildcard (fun ie -> f (Communication (Co,ie))) r in
   let r = RMW.fold_rmw_compat (fun rmw -> f (Rmw rmw)) r
   in r
 
 let fold_tedges f r =
-  let r = fold_ie wildcard (fun ie -> f (Rf ie)) r in
-  let r = fold_ie wildcard (fun ie -> f (Fr ie)) r in
-  let r = fold_ie wildcard (fun ie -> f (Ws ie)) r in
+  let r = fold_com (fun com r -> fold_ie wildcard (fun ie -> f (Communication (com,ie))) r) r in
```
TiberiuBucur, line 393: Is it guaranteed that the `f` parameter will return the same result when applied to the same operators no matter the order? Because the old version did `Rf -> Fr -> Ws` whereas `fold_com` does the opposite.

gen/common/edge.ml
```diff
@@ -638,14 +639,14 @@ let fold_tedges f r =
   | Po(sd,src,_) -> Po (sd,src,Dir d)
   | Fenced(f,sd,src,_) -> Fenced(f,sd,src,Dir d)
   | Dp (dp,sd,_) -> Dp (dp,sd,Dir d)
-  | Rf _ | Hat
-  | Insert _|Store|Id|Node _|Ws _|Fr _|Rmw _|Leave _|Back _-> e
+  | Communication (Rf,_) | Hat
+  | Insert _|Store|Id|Node _|Communication ((Co|Fr),_)|Rmw _|Leave _|Back _-> e
```
TiberiuBucur:
```suggestion
  | Communication _ | Hat
  | Insert _|Store|Id|Node _|Rmw _|Leave _|Back _-> e
```

gen/common/edge.ml
```diff
@@ -638,14 +639,14 @@ let fold_tedges f r =
...
 
   and do_set_src d e = match e with
   | Po(sd,_,tgt) -> Po(sd,Dir d,tgt)
   | Fenced(f,sd,_,tgt) -> Fenced(f,sd,Dir d,tgt)
-  | Fr _|Hat|Dp _
-  | Insert _|Store|Id|Node _|Ws _|Rf _|Rmw _|Leave _|Back _ -> e
+  | Communication (Fr,_)|Hat|Dp _
+  | Insert _|Store|Id|Node _|Communication ((Co|Rf),_)|Rmw _|Leave _|Back _ -> e
```
TiberiuBucur:
```suggestion
  | Communication _|Hat|Dp _
  | Insert _|Store|Id|Node _|Rmw _|Leave _|Back _ -> e
```

gen/namer.ml
```diff
@@ -57,12 +59,7 @@ module Make
          | Fenced (f,Diff,_,_) -> Some (Misc.lowercase (pp_fence f))
          | Dp (dp,Same,_) -> Some (Misc.lowercase (pp_dp dp) ^ "s")
          | Dp (dp,Diff,_) -> Some (Misc.lowercase (pp_dp dp))
-         | Rf Int -> Some "rfi"
-         | Ws Int -> Some "coi"
-         | Fr Int -> Some "fri"
-         | Rf Ext -> Some "rfe"
-         | Ws Ext -> Some "coe"
-         | Fr Ext -> Some "fre"
+         | Communication (com,(Int|Ext as ie)) -> Some (pp_communication com ie)
```
TiberiuBucur, line 62: Why omit `UnspecCom` from the `ie` match? Can it never reach this line?

gen/namer.ml
```diff
@@ -116,25 +113,25 @@ module Make
          | [] -> None
 
        let rec count_a = function
-         | {edge=(Rf Ext|Fr Ext|Ws Ext); a2=Some a; _}::
-           ({edge=(Rf Ext|Fr Ext|Ws Ext);a1=Some _; _}::_ as es) ->
+         | {edge=Communication (_,Ext); a2=Some a; _}::
+           ({edge=Communication (_,Ext);a1=Some _; _}::_ as es) ->
```
TiberiuBucur, line 117: I understand this behaviour existed before this patch, but why do this function and the one below only concern themselves with external communication edges?

TiberiuBucur (commented):
A few minor comments / requests for clarifications :)

gen/common/edge.ml

gen/common/edge.ml
ShaleXIONG: Done

gen/common/edge.ml
ShaleXIONG: I change the order back to Rf -> Fr -> Ws(Co). 

There is no guarantee on the same behaviour on different order. However the order does not matter in the the way we use all the `fold_` functions. They are used to build a look-up table.

gen/common/edge.ml
ShaleXIONG: Done

gen/common/edge.ml
ShaleXIONG: Done

gen/namer.ml
ShaleXIONG: Yes you are right, it should never reach the line. Here the module/file heuristically generate a litmus test name based on a concrete cycle.

gen/namer.ml
ShaleXIONG: I do not have a definitive answer but I believe the function `count_a` is used to heuristically to generate file name.

relokin:
Sorry I’m still not convinced we want Com to be a constructor in edge.ml. Can you please explain what is it that you would like to achieve with this PR? Let’s focus on that first and then I can help with the how.

ShaleXIONG:
> Sorry I’m still not convinced we want Com to be a constructor in edge.ml. Can you please explain what is it that you would like to achieve with this PR? Let’s focus on that first and then I can help with the how.

This is purely code improve for now but the main goal here is to implement `ExpObs` edges which will be represented as `Com(UnspecCom,Ext)`. 

Also there were two datatypes that were the same thing but used in two separate places in the code.
- `type com` of `CRf`, `CFr` and `CWs` in the `code.ml`. They are `pretty-print` to `Rf`, `Fr` and `Co` , However they were not used for the common communication edges `Rfe`, `Fre` and `Coe`. They were used for edges `LeaveRf`, `BackRf` etc. For example you will see in the `type tedge` in `edge.ml`, there are constructors `Leave (com)` and `Back (com)`
- The actual communication edges `Rfe`, `Fre`, and `Coe` were defined in `edge.ml` with three explicit constructors `Rf(ie)`, `Fr(ie)`, and `Ws(ie)`. The `pretty-print` for these three constructors were also `Rf` and `Fr` and `Co`.

The two datatypes were the same, so the best way is to fold the second datatype for communication edge into `Com(com_type, ie)`, which is now in similar style as `Leave(com_type)` and `Back(com_type)`.

relokin:
> > Sorry I’m still not convinced we want Com to be a constructor in edge.ml. Can you please explain what is it that you would like to achieve with this PR? Let’s focus on that first and then I can help with the how.
> 
> This is purely code improve for now but the main goal here is to implement `ExpObs` edges which will be represented as `Com(UnspecCom,Ext)`.
> 

Exp-obs is not the only derived relation we need for aarch64.cat. We have many more, for example dob, are you suggesting we do that for data dependencies as well?

I’m not convinced this PR is the right solution to the problem. I think we need to come up with a way to define new symbols which group together multiple relaxations. This would work for Exps-obs as well as other relations in the cat file. This would be more syntactic sugar for the use of diy7 and not a fundamental relaxation. What I mean is that the user would write 

ExpObs and diy7

would expand it to

Rfe
Coe
Fre

but the gen code will know nothing else about ExpObs.

> Also there were two datatypes that were the same thing but used in two separate places in the code.
> 
> * `type com` of `CRf`, `CFr` and `CWs` in the `code.ml`. They are `pretty-print` to `Rf`, `Fr` and `Co` , However they were not used for the common communication edges `Rfe`, `Fre` and `Coe`. They were used for edges `LeaveRf`, `BackRf` etc. For example you will see in the `type tedge` in `edge.ml`, there are constructors `Leave (com)` and `Back (com)`
> * The actual communication edges `Rfe`, `Fre`, and `Coe` were defined in `edge.ml` with three explicit constructors `Rf(ie)`, `Fr(ie)`, and `Ws(ie)`. The `pretty-print` for these three constructors were also `Rf` and `Fr` and `Co`.
> 

This part of the change seems sensible.

ShaleXIONG:
> > > Sorry I’m still not convinced we want Com to be a constructor in edge.ml. Can you please explain what is it that you would like to achieve with this PR? Let’s focus on that first and then I can help with the how.
> > 
> > 
> > This is purely code improve for now but the main goal here is to implement `ExpObs` edges which will be represented as `Com(UnspecCom,Ext)`.
> 
> Exp-obs is not the only derived relation we need for aarch64.cat. We have many more, for example dob, are you suggesting we do that for data dependencies as well?
> 
> I’m not convinced this PR is the right solution to the problem. I think we need to come up with a way to define new symbols which group together multiple relaxations. This would work for Exps-obs as well as other relations in the cat file. This would be more syntactic sugar for the use of diy7 and not a fundamental relaxation. What I mean is that the user would write
> 
> ExpObs and diy7
> 
> would expand it to
> 
> Rfe Coe Fre
> 
> but the gen code will know nothing else about ExpObs.

I totally agree here. This is similar to issue #1784. I think we should treat `ExpObs` and many other wildcard for example `Po` as a macro or in other word, a syntactic sugar. A macro in `diy*` is basically a string-based shorthand for example `AllRW`. `diy*` will directly map the input string to a list of unfolded concrete datatype. While currently syntax like `Po` and `ExpObs` (if we do minimum change to achieve `ExpObs`), are treated as a wildcard. A wildcard has an internal datatype to represent them, for example `Po` as `Po(Unspec,Unspec,Unspec)`. Then a separate step to `expand` an wildcard to the list of concrete datatype. 

There is nothing to stop us to remove the use of wildcard completely, in fact I already have a local PR for that. Yet it breaks some test cases on `cat2config` which I already chatted with @fsestini, and I will try to patch those test cases on `cat2config`. Separately, the PR itself is very large. Although I organised it in theme, for example, (1) move `Po` to macro, (2) move `Dp` to macro, etc., yet I think we probably need to merge them in phases.

> 
> > Also there were two datatypes that were the same thing but used in two separate places in the code.
> > 
> > * `type com` of `CRf`, `CFr` and `CWs` in the `code.ml`. They are `pretty-print` to `Rf`, `Fr` and `Co` , However they were not used for the common communication edges `Rfe`, `Fre` and `Coe`. They were used for edges `LeaveRf`, `BackRf` etc. For example you will see in the `type tedge` in `edge.ml`, there are constructors `Leave (com)` and `Back (com)`
> > * The actual communication edges `Rfe`, `Fre`, and `Coe` were defined in `edge.ml` with three explicit constructors `Rf(ie)`, `Fr(ie)`, and `Ws(ie)`. The `pretty-print` for these three constructors were also `Rf` and `Fr` and `Co`.
> 
> This part of the change seems sensible.

The change here is purely for the second part, hence I decide to put forwards first, which can be independent from how we want to implement `ExpObs`.

gen/common/edge.ml
TiberiuBucur: I see. I don't have an issue either way if you think it doesn't make a difference, but in that case I would personally keep the definition of `fold_com` in the original order (`Rf` as the outermost parameter, `Co` as the innermost) for the sake of consistency.

TiberiuBucur:
> I think we should treat ExpObs and many other wildcard for example Po as a macro or in other word, a syntactic sugar. A macro in diy* is basically a string-based shorthand for example AllRW. diy* will directly map the input string to a list of unfolded concrete datatype. 

If the plan is to do this thing for all kinds of aarch64 dependencies, do you think diy can use the cat file as some sort of ground truth for what "primitive edges" (coe, rfe, fre) a dependency is made of? It would help with future changes to the cat file not requiring updates to diy.

fsestini:
> > I think we should treat ExpObs and many other wildcard for example Po as a macro or in other word, a syntactic sugar. A macro in diy* is basically a string-based shorthand for example AllRW. diy* will directly map the input string to a list of unfolded concrete datatype.
> 
> If the plan is to do this thing for all kinds of aarch64 dependencies, do you think diy can use the cat file as some sort of ground truth for what "primitive edges" (coe, rfe, fre) a dependency is made of? It would help with future changes to the cat file not requiring updates to diy.

FWIW, this is precisely what `cat2config` does (for a subset of the cat file).

ShaleXIONG:
> > > I think we should treat ExpObs and many other wildcard for example Po as a macro or in other word, a syntactic sugar. A macro in diy* is basically a string-based shorthand for example AllRW. diy* will directly map the input string to a list of unfolded concrete datatype.
> > 
> > 
> > If the plan is to do this thing for all kinds of aarch64 dependencies, do you think diy can use the cat file as some sort of ground truth for what "primitive edges" (coe, rfe, fre) a dependency is made of? It would help with future changes to the cat file not requiring updates to diy.
> 
> FWIW, this is precisely what `cat2config` does (for a subset of the cat file).

@TiberiuBucur @fsestini That is a good comment. I think it is certainly possible to gradually bring in some cat language, for example treating them as macros first, as the ground true into `diy7`, esp. those cat axioms that can be easily translated to `diy` relax/edge and ultimately to the instruction sequences. However we have some cat axioms are non-trivial and challenging to map to instruction sequences; one example here is implicit event which may be triggered by many explicit memory events. Here I also want to emphasise instruction sequences because `diy*` are tools that helps us generate interaction sequences and eventual litmus tests. This means the `diy*` input should give some idea of the final litmus tests. 

I think some of this comments are much more related to #1784 , if you @TiberiuBucur  are interested, we can collaborate !

fsestini:
> > > > I think we should treat ExpObs and many other wildcard for example Po as a macro or in other word, a syntactic sugar. A macro in diy* is basically a string-based shorthand for example AllRW. diy* will directly map the input string to a list of unfolded concrete datatype.
> > > 
> > > 
> > > If the plan is to do this thing for all kinds of aarch64 dependencies, do you think diy can use the cat file as some sort of ground truth for what "primitive edges" (coe, rfe, fre) a dependency is made of? It would help with future changes to the cat file not requiring updates to diy.
> > 
> > 
> > FWIW, this is precisely what `cat2config` does (for a subset of the cat file).
> 
> @TiberiuBucur @fsestini That is a good comment. I think it is certainly possible to gradually bring in some cat language, for example treating them as macros first, as the ground true into `diy7`, esp. those cat axioms that can be easily translated to `diy` relax/edge and ultimately to the instruction sequences. However we have some cat axioms are non-trivial and challenging to map to instruction sequences; one example here is implicit event which may be triggered by many explicit memory events. Here I also want to emphasise instruction sequences because `diy*` are tools that helps us generate interaction sequences and eventual litmus tests. This means the `diy*` input should give some idea of the final litmus tests.
> 
> I think some of this comments are much more related to #1784 , if you @TiberiuBucur are interested, we can collaborate !

For the record, I'm not proposing a solution here. Just wanted to point out that _if_ we do decide to involve the cat model for this sort of feature, `cat2config` constitutes a potentially useful precedent to draw inspiration (and maybe implementation) from :)

gen/common/code.ml
```diff
@@ -153,14 +153,14 @@ let checks =
...
 let pp_com = function
-  | CRf -> "Rf"
-  | CFr -> "Fr"
-  | CWs -> "Co"
+  | Rf -> "Rf"
+  | Fr -> "Fr"
+  | Co -> "Co"
 
-let fold_com f r = f CRf (f CFr (f CWs r))
+let fold_com f r = f Co (f Fr (f Rf r))
```
TiberiuBucur: Since you said the order of the edges where this is used does not matter, can we please switch back to the original order for the sake of consistency? Thanks.

ShaleXIONG: Done.

TiberiuBucur (approved):

PR #1872, opened 2026-06-18
[herd] Measure some timings 2

maranget:
Add timer support (cf. lib/timer.ml). The new feature is  used for timing:
  + Cat model execution and
  + code semantics, _i.e._ monad construction by the "Sem" modules.

Timers are activated by the command-line option `-debug timers`. For instance:
```
% herd7 catalogue/aarch64-VMSA/tests/Marc03.litmus -variant vmsa -speedcheck true -debug timers
Test Marc03 Forbidden
...
Timers: model=13.25, run=18.61, semantics=0.02
```

Supersedes PR #1865.

### DIFF
--- a/herd/top_herd.mli
+++ b/herd/top_herd.mli
@@ -26,6 +26,7 @@ module type CommonConfig = sig
   val statelessrc11 : bool
   val skipchecks : StringSet.t
   val dumpallfaults : bool
+  module Timer : Timer.S
 end
 
 module type Config = sig
--- a/herd-www/jerd.ml
+++ b/herd-www/jerd.ml
@@ -217,6 +217,7 @@ let run_herd bell cat litmus cfg cat_label =
     let check_filter = !check_filter
     let debug = !debug
     let debuglexer = debug.Debug_herd.lexer
+    module Timer = Timer.No
     let verbose = !verbose
     let hexa = !PP.hexa
     let unroll = !unroll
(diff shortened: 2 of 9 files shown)


maranget:
Hi @fsestini. I have followed your idea of avoiding a functor. There are three commits:
  1. Your commit a26f14374128535596f08c595e1183f34472f6f7
  2. Add a `Timer.start` call that was missing, ca24635869147e42652de06be2f749a216ceab4c
  3. Include the `Timer` module as part of the configuration module argument of relevant modules (functors), 5062394e0d2c7aa80987661429b8c9f81c18e80f

The commit number 3 above permits passing an inactive Timer module when timers are not activated, as illustrated by the web interface.

herd/top_herd.mli
```diff
@@ -120,6 +121,7 @@ end
 
 module type PrinterConfig = sig
   module PC : PrettyConf.S
+  module Timer : Timer.S
```
fsestini: I think this is no longer needed?

fsestini (commented):
Thanks for re-visiting this @maranget . I just noticed one extra line in `PrinterConfig` that I think isn't needed, but otherwise this LGTM.

herd/top_herd.mli
maranget: Nice catch.

PR #1874, opened 2026-06-19
[asl][ci] Build HTML version of the ASL Reference in CI

hrutvik:
As part of the ASL Reference CI run, also build the HTML version and upload as an artifact.

Also merge `asllib/doc/zyva.sh` into `asllib/doc/Makefile`.


.github/workflows/build-asl-reference.yml
```diff
@@ -53,3 +53,45 @@ jobs:
...
+          apt-get update
+          apt-get install -y build-essential m4 gcc libgmp-dev pkg-config python3 bzip2
+
+      - name: Checkout tree
+        uses: actions/checkout@v4
+
+      - name: Set-up OCaml
+        uses: ocaml/setup-ocaml@v3
+        with:
+          ocaml-compiler: dune,zarith,menhir,logs,conf-which
```
HadrienRenaud: Add hevea here?


```suggestion
          ocaml-compiler: dune,zarith,menhir,logs,conf-which,hevea
```

.github/workflows/build-asl-reference.yml
```diff
@@ -53,3 +53,45 @@ jobs:
...
+          opam-disable-sandboxing: true
```
HadrienRenaud: Why this?

.github/workflows/build-asl-reference.yml
```diff
@@ -53,3 +53,45 @@ jobs:
         with:
          name: ASL Reference Document
          path: asllib/doc/ASLReference.pdf
+
+  make-asldoc-html:
+    runs-on: ubuntu-latest
+    container: texlive/texlive:latest-full
```
HadrienRenaud: Why do we need `texlive`? AFAIK it's only a dependency of hevea if you need to generate images

.github/workflows/build-asl-reference.yml
```diff
@@ -53,3 +53,45 @@ jobs:
...
+
+    env:
+      OPAMROOTISOK: 1 # Suppress warnings about running opam as root
```
HadrienRenaud: We're in root because of the container?

asllib/doc/Makefile
```diff
@@ -21,14 +21,22 @@ ASLReference.pdf: generated_spec *.tex *.bib
 	fi
 	python3 doclint.py --console_macros --aslref $(ASLREF)
 
+html: generated_spec *.tex *.bib *.hva
+	mkdir -p HTML
+	hevea -fix -O -o HTML/ASL.html book.hva macros.hva ASLReference.tex
```
HadrienRenaud: The `zyva.sh` script had the following:
```
  hevea -fix0 -O -o ...
```
Is the difference expected?

HadrienRenaud (commented):
I've got a lot of questions sorry

.github/workflows/build-asl-reference.yml
hrutvik: Ah I had not realised this - will build without a container instead

.github/workflows/build-asl-reference.yml
hrutvik: Yes, unfortunately GitHub Actions seems to require containers to run with `root` user, but then `opam` emits warnings without this flag when not installed by the GitHub action. The same appears on the `static-binaries.yaml` workflow. However, it should go when not using a container instead

.github/workflows/build-asl-reference.yml
hrutvik: Will move from line 86 below

.github/workflows/build-asl-reference.yml
hrutvik: Again container-related - should also disappear

asllib/doc/Makefile
hrutvik: I believe the `-fix0` option is for fast iteration during development (does one pass only, like running `pdflatex` once), whereas `-fix` iterates to a fixpoint (like `pdflatex` multiple times) - I didn't think that we needed the fast iteration option, but I don't feel strongly

.github/workflows/build-asl-reference.yml
hrutvik: It looks like there is a dependency on `bibtex`, which itself requires the `.aux` file. I'll try to add this as `apt-get install texlive-bibtex-extra`, which doesn't take too long overall - but we can keep an eye on CI times and consider using a container if things seem slower than we would like.

asllib/doc/Makefile
Roman-Manevich: I think we need it. I don't think the script would work with `-fix` here. If it does, that's fine.

Roman-Manevich (approved):

asllib/doc/Makefile
HadrienRenaud: Is it because of the bibhva after? You want to first generate a `.aux` file and then run bibhva to produce the good result and then you iterate until fix point?

.github/workflows/build-asl-reference.yml
HadrienRenaud: Ok thanks! CI times look similar?

HadrienRenaud: ahhh indeed for bibhva thanks

.github/workflows/build-asl-reference.yml
```diff
@@ -53,3 +53,36 @@ jobs:
...
          path: asllib/doc/ASLReference.pdf
+
+  make-asldoc-html:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Install dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y build-essential m4 gcc libgmp-dev pkg-config python3 bzip2 texlive-bibtex-extra
```
HadrienRenaud: Just by curiosity (it's already installed so it makes no difference to me) but why do we need `gcc`? Or the other dependencies?

HadrienRenaud (approved):
LGTM

.github/workflows/build-asl-reference.yml
hrutvik: It looks like many of these are dependencies of `opam` (or of the `opam` packages installed) - but perhaps we do not need to specify them. Will try it out

asllib/doc/Makefile
hrutvik: I think various other commands may also need multiple runs after `bibhva` (e.g. for math alignment, for auto-generation of the section lists at the starts of chapters when this gets merged) - my understanding was that it was much like running `pdflatex` twice after `bibtex`

.github/workflows/build-asl-reference.yml
hrutvik: Looks like it works!

asllib/doc/Makefile
hrutvik: Discussed offline - first run of `hevea` can be `fix0`, second run should be `fix`

PR #1876, opened 2026-06-22
[herd] Fix non explicit atomic pairs

maranget:
Compute `rmw` on non-explicit memory accesses from `iico_data`

### DIFF
--- a/herd/mem.ml
+++ b/herd/mem.ml
@@ -140,7 +140,7 @@ module type S = sig
     S.E.event -> S.E.A.location
 
   val make_atomic_load_store :
-    S.E.event_structure -> S.E.EventRel.t
+    S.test -> S.E.event_structure -> S.E.EventRel.t
 
 end
 
@@ -1911,11 +1911,11 @@ let add_eq v1 v2 eqs =
 (*
  * Reconstruct load/store atomic pairs,
  *   By definition such a pair exists when the
- * store precedes the load in generalised program order
+ * load precedes the store in generalised program order
  * (_i.e._ the union of program order and of iico), and
  * that there is no atomic effect to the same location
  * in-between (w.r.t generalised po) the load and the store.
- *   Computation proceeds as follows:
+ *   Computation for explicit accesses proceeds as follows:
  *   First, atomic events are grouped first by thread
  * and then by location. Then, to each atomic load, we
  * associate the closest generalised po successor store,
@@ -1925,15 +1925,25 @@ let add_eq v1 v2 eqs =
  * to use a set of all atomic events (by a given thread and
  * with a given location) ordered by po because some atomic loads
  * may be unrelated.
- *   Finally, such atomic pairs can be spurious, that is not performed
- * by a specific thread. In that case, pairs are given
- * simply by the intra causality data relation.
+ *   Finally, the computation is simpler for non-explicit such atomic
+ * pairs. In that case, pairs are given simply by the intra causality
+ * data relation. Namely, those pairs all stem from the atomic modification
+ * of page table entries.
  *)
 
-    let make_atomic_load_store es =
-      let atms,spurious = U.collect_atomics es in
+    let make_atomic_load_store _test es =
+      let atms,nexps = U.collect_atomics es in
+
       let module StoreSet = EvtSetByPo(struct let es = es end) in
-      let make_atomic_pairs es k =
+
+      let make_atomic_pairs_nexp evts =
+        let rs,ws = List.partition E.is_load evts in
+        let rs = E.EventSet.of_list rs
+        and ws = E.EventSet.of_list ws in
+        E.EventRel.restrict_domains_to_sets
+             rs ws es.E.intra_causality_data
+
+      and make_atomic_pairs es k =
         let rs,ws = List.partition E.is_load es in
         let ws = StoreSet.of_list ws
         and intervening_read er ew =
@@ -1958,23 +1968,10 @@ let add_eq v1 v2 eqs =
         List.map
           (fun (_,m) ->
            U.LocEnv.fold
-             (fun _loc es k ->
-                let exps,nexps = List.partition E.is_explicit es in
-                make_atomic_pairs exps @@ make_atomic_pairs nexps k)
+             (fun _loc exps k -> make_atomic_pairs exps k)
              m E.EventRel.empty)
         atms |> E.EventRel.unions
-      and r2 =
-        let iico = es.E.intra_causality_data in
-        List.map
-          (fun e ->
-             if E.is_load e then
-               match
-                 E.EventRel.succs iico e |> E.EventSet.as_singleton
-               with
-               | None -> assert false (* spurious updates are by pairs *)
-               | Some w -> E.EventRel.singleton (e,w)
-             else E.EventRel.empty)
-          spurious |> E.EventRel.unions in
+      and r2 = make_atomic_pairs_nexp nexps in
       E.EventRel.union r1 r2
 
 (* Retrieve last store from rfmap *)
@@ -2549,7 +2546,7 @@ let add_eq v1 v2 eqs =
                     | OptAce.True -> check_rfmap es rfm
                   then
                     (* Atomic load/store pairs *)
-                    let atomic_load_store = make_atomic_load_store es in
+                    let atomic_load_store = make_atomic_load_store test es in
                     if
                       C.variant Variant.OptRfRMW
                       && some_same_rf_rmw rfm atomic_load_store
--- a/herd/mem.ml
+++ b/herd/mem.ml
@@ -2547,6 +2547,17 @@ let add_eq v1 v2 eqs =
                   then
                     (* Atomic load/store pairs *)
                     let atomic_load_store = make_atomic_load_store test es in
+                    let () =
+                      if dbg then
+                        E.EventRel.iter_succs
+                          (fun _er ews ->
+                             if E.EventSet.cardinal ews <> 1 then begin
+                               let module PP = Pretty.Make(S) in
+                               eprintf "Non-affine rmw relation!\n%!" ;
+                               PP.show_es_rfm test es S.RFMap.empty ;
+                               Warn.fatal "Non-affine rmw relation!"
+                             end)
+                          atomic_load_store in
                     if
                       C.variant Variant.OptRfRMW
                       && some_same_rf_rmw rfm atomic_load_store
--- a/herd/mem.ml
+++ b/herd/mem.ml
@@ -2549,15 +2549,21 @@ let add_eq v1 v2 eqs =
                     let atomic_load_store = make_atomic_load_store test es in
                     let () =
                       if dbg then
-                        E.EventRel.iter_succs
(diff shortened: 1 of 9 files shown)


herd/mem.ml
```diff
@@ -1925,15 +1925,25 @@ let add_eq v1 v2 eqs =
...
- * simply by the intra causality data relation.
+ *   Finally, the computation is simpler for non-explicit such atomic
+ * pairs. In that case, pairs are given simply by the intra causality
+ * data relation. Namely, those pairs all stem from the atomic modification
+ * of page table entries.
  *)
 
-    let make_atomic_load_store es =
-      let atms,spurious = U.collect_atomics es in
+    let make_atomic_load_store _test es =
```
HadrienRenaud, line 1934: Why do we need `_test` here?

lib/myRel.mli
```diff
@@ -43,7 +43,9 @@ module type S = sig
   val cardinal : t -> int
 
   val iter : ((elt1 * elt2) -> unit) -> t -> unit
+  val iter_succs : (elt1 -> Elts2.t -> unit) -> t -> unit
   val fold : ((elt1 * elt2) -> 'a -> 'a) -> t -> 'a -> 'a
+  val fold_succs : (elt1 -> Elts2.t -> 'a -> 'a) -> t -> 'a -> 'a
```
HadrienRenaud, line 48: I think this is unused?

herd/memUtils.mli
HadrienRenaud: I don't think this is true anymore?

herd/mem.ml
```diff
@@ -2549,7 +2546,18 @@ let add_eq v1 v2 eqs =
...
+                      if dbg then
+                        E.EventRel.iter_succs
+                          (fun _er ews ->
+                             if E.EventSet.cardinal ews <> 1 then begin
+                               let module PP = Pretty.Make(S) in
+                               eprintf "Non-affine rmw relation!\n%!" ;
+                               PP.show_es_rfm test es S.RFMap.empty ;
+                               Warn.fatal "Non-affine rmw relation!"
+                             end)
+                          atomic_load_store in
```
HadrienRenaud: Wondering about 2 things:
1. Should we have a `is_affine` function in `EventRel`? Like should we move the cardinality check to `EventRel`?
Thinking of something like:
```ocaml
val is_affine : t -> (elt1, Elt2s.t) option
(** [is_affine rel] is [None] if [rel] is affine, otherwise returns a witness [e1, e2s] where [e1] is related to each element of [e2s] in [rel]. *)
```
2. Should we check that it is a bijection?

herd/mem.ml
maranget: I used the test argument to display some graphs. We can remove it now.

lib/myRel.mli
maranget: `iter_succs` is used in the affinity check (see below). `fold_succs` is a logical companion...

herd/mem.ml
maranget: 1. No real opinion. This is so specfic that I did not see the need for generality.
2. We could do that with the affinity check on the inverse relation.

herd/memUtils.mli
maranget: You think right. Nice catch, who reads comments nowdays?

maranget: Done.

relokin:
Thanks for this Luc, I will try to have a look at the code as soon as possible.

I have a high level question. We discussed this with Hadrien as well, why can't we get this information from the instruction semantics? I understand why we can't do that for exclusives, but for atomic instructions (for example, CAS, SWP) and for atomic hardware updates, we know that the Memory Read Effect and the Memory Write Effect need to be paired together from the instruction semantics. Rather than searching through the list of Effect to find pairs of Effects that satisfy certain conditions, wouldn't it be preferable to retain that information from the instruction semantics?

maranget:
> I have a high level question. We discussed this with Hadrien as well, why can't we get this information from the instruction semantics? I understand why we can't do that for exclusives, but for atomic instructions (for example, CAS, SWP) and for atomic hardware updates, we know that the Memory Read Effect and the Memory Write Effect need to be paired together from the instruction semantics. Rather than searching through the list of Effect to find pairs of Effects that satisfy certain conditions, wouldn't it be preferable to retain that information from the instruction semantics?

I guess the process has been as follows: we first have implemented  exclusives first (for Power) and then generalised without thinking of a different solution. The case of AArch64+ASL is also to be considered.

HadrienRenaud:
> The case of AArch64+ASL is also to be considered.

I think the case of AArch64+ASL is easy enough: we could run this algorithm, or implement it in `asl.cat` (a bit like [Nikos' try](https://github.com/relokin/herdtools7/commit/5fe1484ac282e849437874f815cdc6c3b6d63223#diff-10cfa720a0fd73ed89579543d77baa9d3f62a5d1273cab86effef115e1d997b4)). So I don't think AArch64+ASL would be the biggest problem here.

relokin:
> > The case of AArch64+ASL is also to be considered.
> 
> I think the case of AArch64+ASL is easy enough: we could run this algorithm, or implement it in `asl.cat` (a bit like [Nikos' try](https://github.com/relokin/herdtools7/commit/5fe1484ac282e849437874f815cdc6c3b6d63223#diff-10cfa720a0fd73ed89579543d77baa9d3f62a5d1273cab86effef115e1d997b4)). So I don't think AArch64+ASL would be the biggest problem here.

I agree and longer term ASL will provide the means to identify pairs of Memory Read and Memory Write Effects that are intended to be atomic.

@maranget for Power and other architecture, is exclusives the only concern or is there support for atomics as well. Otherwise, it might be simple to do this hybrid approach, for atomic instructions and hardware updates we use the instruction semantics and for exclusives we use a version of this code.

maranget:
> I think the case of AArch64+ASL is easy enough: we could run this algorithm, or implement it in `asl.cat` (a bit like [Nikos' try](https://github.com/relokin/herdtools7/commit/5fe1484ac282e849437874f815cdc6c3b6d63223#diff-10cfa720a0fd73ed89579543d77baa9d3f62a5d1273cab86effef115e1d997b4)). So I don't think AArch64+ASL would be the biggest problem here.

Nice, we can implement rmw as `amo|lssx` in Cat. Notice that other architectures provide amo's or lxsx. In fact, all architectures provides at least one of them. As a consequence:
1. We need a general set of atomics. As far as I know the name of the set may differ from architecture. we also need to generalise the exclusives annotation (EXCL in your code).
2. stdlib.hva may be a beter place for your definitions.

maranget:
Additonally, I am a bit concerned by the added complexity of an hybrid approach. Don't you think that an "all cat" or an "all mem.ml" approach would be more easy to maintain?

maranget:
And to answer one of your questions : ARM -> exclusives, LISA -> amo, BPF -> amo, C -> amo, Java -> amo, MIPS -> exclusives, PPC -> exclusives, RISCV -> exclusives,amo, X86_64 -> amo, X86 -> amo.

herd/mem.ml
HadrienRenaud: tried in #1879.

relokin:
> > I think the case of AArch64+ASL is easy enough: we could run this algorithm, or implement it in `asl.cat` (a bit like [Nikos' try](https://github.com/relokin/herdtools7/commit/5fe1484ac282e849437874f815cdc6c3b6d63223#diff-10cfa720a0fd73ed89579543d77baa9d3f62a5d1273cab86effef115e1d997b4)). So I don't think AArch64+ASL would be the biggest problem here.
> 
> Nice, we can implement rmw as `amo|lssx` in Cat. Notice that other architectures provide amo's or lxsx. In fact, all architectures provides at least one of them. As a consequence:

`amo` relies on the same code as `rmw` doesn't it? https://github.com/herd/herdtools7/blob/161ed8e9a1826b3aac30ba52621a2ca9668e6d87/herd/machModelChecker.ml#L440 and lxsx is derived in stdlib.cat as `let lxsx = rmw \ amo`. I am not understand the suggestion.

> 
>     1. We need a general set of atomics. As far as I know the name of the set may differ from architecture. we also need to generalise the exclusives annotation (EXCL in your code).
>     2. stdlib.hva may be a beter place for your definitions.

agreed.

maranget:
I meant that amo an lxsx can be computed directly from the event sets EXCL and X as you do [here](https://github.com/relokin/herdtools7/commit/5fe1484ac282e849437874f815cdc6c3b6d63223#diff-10cfa720a0fd73ed89579543d77baa9d3f62a5d1273cab86effef115e1d997b4). That is:
```
let lxsx = [Exp & R & EXCL]; po & same-loc \ (po; [Exp & M & EXCL]; po); [Exp & W & EXCL]
let amo = [Exp & R & X]; same-instance & same-loc; [Exp & W & X]
   | [Imp & TTD & R]; iico_data & same-loc; [Imp & TTD & W]
let rmw = amo | lxsx
```
To that aim we have to define EXCL and X for all architectures.

HadrienRenaud:
> I meant that amo an lxsx can be computed directly from the event sets EXCL and X as you do [here](https://github.com/relokin/herdtools7/commit/5fe1484ac282e849437874f815cdc6c3b6d63223#diff-10cfa720a0fd73ed89579543d77baa9d3f62a5d1273cab86effef115e1d997b4). That is:
> 
> ```
> let lxsx = [Exp & R & EXCL]; po & same-loc \ (po; [Exp & M & EXCL]; po); [Exp & W & EXCL]
> let amo = [Exp & R & X]; same-instance & same-loc; [Exp & W & X]
>    | [Imp & TTD & R]; iico_data & same-loc; [Imp & TTD & W]
> let rmw = amo | lxsx
> ```
> 
> To that aim we have to define EXCL and X for all architectures.

This is one possibility. At least this is how we could do it for AArch64+ASL.

Another possibility (which might be what Nikos was talking about) would be to have the instruction semantics define `amo` and `EXCL`: because `amo` is always in a single instruction, the instruction semantics seem a good place to do this, and have all the information necessary. `EXCL` could be simply defined by a new filter function in `actions.ml` called `is_exclusive` (note that `is_atomic` might not be precise enough here).

maranget:
In any case we need support for the `EXCL` set  or/and `is_exclusive`. As you noticed `is_atomic` is not precise enough. Consider the following test:
```
AArch64 LXSX06
{
 0:X0=x;
 1:X0=x;
}
 P0              |  P1         ;
 LDXR W1,[X0]    | MOV W1,#3   ;
 MOV W5,#2       | STR W1,[X0] ;
 STADD W5,[X0]   |             ;
 MOV W3,#1       |             ;
 STXR W4,W3,[X0] |             ;
exists 0:X1=0 /\ [x]=1
```
At the moment. **herd7** allows the test because it consider that ` STADD` and `STXR` behave the same.

HadrienRenaud:
> In any we need support for `EXCL`.

Should we still go ahead with this PR or do you think this is blocking?

maranget:
Superseded by PR #1889?

maranget:
Superseded by PR #1889.

PR #1877, opened 2026-06-22
[ci] Fix OCaml version to 5.4 when js-of-ocaml is required

HadrienRenaud:


maranget:
Thanks for this @HadrienRenaud. It is likely that `js_of_ocaml` will catch up soon.

HadrienRenaud:
> Thanks for this @HadrienRenaud. It is likely that `js_of_ocaml` will catch up soon.

Indeed it has caught up, see [this succeeding workflow](https://github.com/herd/herdtools7/actions/runs/28015980615).

PR #1878, opened 2026-06-22
[asl] Flag to print compact lisp output

hrutvik:
Add the `--print-lisp-compact` flag to output Lisp like the `--print-lisp` flag, but:
- with minimal whitespace
- without location information (i.e. for `'a annotated`, simply print the `'a` and ignore the `annotated`)
- without comments

Also fix the `www` CI as it seemed to have some OCaml version clash between base compiler and its dependencies.

### DIFF
--- a/asllib/lispobj.ml
+++ b/asllib/lispobj.ml
@@ -51,11 +51,13 @@ let rec equal x y =
 module type PrinterConf = sig
   val defaultpkg : string
   val downcase : bool
+  val compact : bool
 end
 
 module MakePrinter (Conf : PrinterConf) = struct
   let defaultpkg = Conf.defaultpkg
   let downcase = Conf.downcase
+  let compact = Conf.compact
 
   let symbol_part_str x =
     (* to get this actually right we'd need much better analysis of whether the symbol needs escaping.
@@ -95,9 +97,10 @@ module MakePrinter (Conf : PrinterConf) = struct
           | _ -> pp_obj_as_list f x
         else pp_obj_as_list f x
     | Comment (str, obj) ->
-        fprintf f "@[<v>%a@]%a" pp_comment_lines
-          (String.split_on_char '\n' str)
-          pp_obj obj
+        if not Conf.compact then
+          fprintf f "@[<v>%a@]%a" pp_comment_lines
+            (String.split_on_char '\n' str)
+            pp_obj obj
 
   and pp_comment_lines f lines =
     match lines with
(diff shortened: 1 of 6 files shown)


asllib/lispobj.ml
HadrienRenaud, line 81: Is it normal that the parenthesis is not closed here?

solswords (approved):
Looks fine to me

asllib/toLisp.ml
hrutvik, line 1: This file is mostly indentation changes - key changes are:
- wrap in `module Make`
- modify `of_annotated` to account for `Conf.compact`

asllib/lispobj.ml
solswords: > Is it normal that the parenthesis is not closed here?

Seems like a long-standing bug to me. No occasion to use any complex numbers so far though

asllib/lispobj.ml
```diff
@@ -112,17 +114,25 @@ module MakePrinter (Conf : PrinterConf) = struct
...
+
+  let pp fmt =
+    if not compact then pp_obj fmt
+    else
+      let fns = Format.pp_get_formatter_out_functions fmt () in
+      let fns =
+        { fns with out_newline = (fun _ -> ()); out_indent = (fun _ -> ()) }
+      in
+      let fmt = Format.formatter_of_out_functions fns in
+      pp_obj fmt
```
HadrienRenaud, line 127: We can also produce a fast-track output to stdout, something like (yes I was finding this a fun exercise):
```ocaml
module MakeCompactPrinter = struct
  open Printf

  let defaultpkg = "ACL2"
  let downcase = true

  let symbol_part_str chan x =
    (* to get this actually right we'd need much better analysis of whether the symbol needs escaping.
       For now we don't expect to see bad characters in symbols much, but we might need mixed case *)
    if x = String.uppercase_ascii x then
      output_string chan @@ if downcase then String.lowercase_ascii x else x
    else fprintf chan "|%s|" x

  let rec pp_obj chan = function
    | Num (r, i) ->
        if Q.( = ) i Q.zero then Q.output chan r
        else fprintf chan "#C(%a %a)" Q.output r Q.output i
    | String s -> fprintf chan "\"%s\"" (escape_backslashes_and_quotes s)
    | Symbol ("KEYWORD", name) -> fprintf chan ":%a" symbol_part_str name
    | Symbol ("COMMON-LISP", name) -> symbol_part_str chan name
    | Symbol (pkg, name) ->
        (* BOZO neither packages nor escaping are handled correctly *)
        if String.equal pkg defaultpkg then symbol_part_str chan name
        else fprintf chan "%a::%a" symbol_part_str pkg symbol_part_str name
    | Char c -> fprintf chan "#\\%s" (Char.escaped c)
    | Cons (car, Cons (cadr, cddr)) when equal car quote && equal cddr nil ->
        output_char chan '\'';
        pp_obj chan cadr
    | Cons _ as x ->
        output_char chan '(';
        pp_list chan x;
        output_char chan ')'
    | Comment (str, obj) ->
        pp_comment_lines chan (String.split_on_char '\n' str);
        pp_obj chan obj

  and pp_comment_lines f lines =
    List.iter (fun line -> fprintf f ";; %s\n" line) lines

  and pp_list f x =
    match x with
    | Cons (car, cdr) ->
        if equal cdr nil then pp_obj f car
        else (
          pp_obj f car;
          output_char f ' ';
          pp_list f cdr)
    | _ ->
        output_string f ". ";
        pp_obj f x
end
```

asllib/lispobj.ml
HadrienRenaud, line 99: Wondering if we could also skip comments on this version?

HadrienRenaud (approved):
In general this PR looks good to me. If we wanted, we could improve performance of the printer.

asllib/lispobj.ml
hrutvik: Good point! Will add

hrutvik: I'll leave this off this PR unless you'd like me to fix at the same time

asllib/lispobj.ml
hrutvik: Thanks for this! For now, I'd like to reuse the existing implementation if possible - but I agree that if performance becomes a concern we can use this approach

PR #1885, opened 2026-06-24
[aslref][ASL Reference] removing subtyping from code

Roman-Manevich:
This PR removes subtyping from ASL:
- Simplified type declarations so they declare only a type name and type body.
- Replaced named-subtype relationships with same-named type identity checks where named identity is still required.
- Synced the ASL Reference documentation, `asl.spec`, ASLTypingReference tests, and regression tests with the implementation.


asllib/types.ml
```diff
@@ -645,15 +613,9 @@ let rec lowest_common_ancestor ~loc env s t =
...
-    | T_Named name_s, T_Named name_t -> (
-        (* If S and T are both named types: the (unique) common supertype of S
-         and T that is a subtype of all other common supertypes of S and T. *)
-        match find_named_lowest_common_supertype env name_s name_t with
-        | Some name -> Some (T_Named name |> here)
-        | None ->
-            let anon_s = make_anonymous env s
-            and anon_t = make_anonymous env t in
-            lowest_common_ancestor ~loc env anon_s anon_t)
+    | T_Named s_name, T_Named t_name when not (String.equal s_name t_name) ->
```
hrutvik: Reuse `same_named_type` here?

hrutvik: Actually, I also wonder whether we can just delete this case and use the case below?

asllib/Interpreter.ml
```diff
@@ -1427,15 +1415,14 @@ module Make (B : Backend.S) (C : Config) = struct
...
     (* [catcher_matches t c] returns true if the catcher [c] match the raised
        exception type [t]. *)
     (* Begin EvalFindCatcher *)
     let catcher_matches =
-      let static_env = { StaticEnv.empty with global = env.global.static } in
-      fun v_ty (_e_name, e_ty, _stmt) ->
-        subtypes static_env v_ty e_ty |: SemanticsRule.FindCatcher
-      (* End *)
+     fun v_ty (_e_name, e_ty, _stmt) ->
+      Types.same_named_type v_ty e_ty |: SemanticsRule.FindCatcher
```
hrutvik: I wonder about this - it is an invariant that both `v_ty` and `e_ty` are named types here? In which case, could we just use `type_equal` and avoid defining `same_named_type`?

asllib/Interpreter.ml
```diff
@@ -1427,15 +1415,14 @@ module Make (B : Backend.S) (C : Config) = struct
...
+  and eval_catchers catchers otherwise_opt s_m : stmt_eval_type =
     (* [catcher_matches t c] returns true if the catcher [c] match the raised
        exception type [t]. *)
     (* Begin EvalFindCatcher *)
     let catcher_matches =
-      let static_env = { StaticEnv.empty with global = env.global.static } in
-      fun v_ty (_e_name, e_ty, _stmt) ->
-        subtypes static_env v_ty e_ty |: SemanticsRule.FindCatcher
-      (* End *)
+     fun v_ty (_e_name, e_ty, _stmt) ->
```
HadrienRenaud: Why not?
```ocaml
    let catcher_matches v_ty (_e_name, e_ty, _stmt) =
```

asllib/parser_errors.messages
```diff
@@ -317,7 +317,7 @@ spec: CONFIG IDENTIFIER XOR
 ## The known suffix of the stack is as follows:
 ## CONFIG ignored_or_identifier
 ##
-spec: CONFIG IDENTIFIER COLON BIT WITH
+spec: CONFIG IDENTIFIER COLON BIT XOR
```
HadrienRenaud, line 320: Can I double-check that this is generated by `menhir`?

asllib/types.ml
```diff
@@ -533,8 +525,8 @@ and subtype_satisfies env t s =
 and type_satisfies env t s =
   ((* Type T type-satisfies type S if and only if at least one of the following
       conditions holds: *)
-   (* T is a subtype of S *)
-   subtypes env t s
+   (* T is a same-named type as S *)
+   same_named_type t s
```
HadrienRenaud: Wondering if it wouldn't be simpler to say that they are equal?

asllib/types.ml
```diff
@@ -566,9 +558,9 @@ let rec type_clashes env t s =
         type-clash
       • they both have the structure of tuples of the same length whose
         corresponding element types type-clash
-      • S is either a subtype or a supertype of T *)
+      • S and T are same-named types *)
   (* We will add a rule for boolean and boolean. *)
-  ((subtypes env s t || subtypes env t s)
+  (same_named_type s t
```
HadrienRenaud: Same question: wouldn't it be simpler to simply say `equal s t` ?

asllib/types.ml
HadrienRenaud: I think the `when ...` is always true here, am I correct?
Why not convert it into an assert?
```suggestion
    | T_Named s_name, T_Named t_name ->
        assert (not (String.equal s_name t_name)); (* per case above *)

File: asllib/doc/TopLevel.tex [hunk 7]
```diff
@@ -210,15 +210,15 @@ \section{Renaming Local Storage Elements in the Standard Library\label{sec:Renam
 \begin{itemize}
   \item view $\vf$ as a subprogram description consisting of a subprogram named $\name$,
         list of parameters $\params$, list of arguments $\args$, body $\vbody$,
-        optional return type $\rettyopt$, optional subtype $\vsubtype$,
+        optional return type $\rettyopt$, optional subprogram type $\vsubtype$,
```
hrutvik: Should this also be `\vsubprogramtype` or similar?

asllib/tests/ASLTypingReference.t/TypingRule.AddNewFunc.bad1.asl
```diff
@@ -1,10 +1,10 @@
 type shape of integer;
```
hrutvik: nit: we may want to rename `shape` --> `circle` in all these examples, as conceptually a square is still a shape so it seems a bit confusing like this

asllib/tests/regressions.t/run.t
```diff
@@ -5,12 +5,7 @@ Hello world should work:
 
 Type-checking errors:
 
-  $ aslref subtype-satisfaction-arrray-illegal.asl
-  File subtype-satisfaction-arrray-illegal.asl, line 4, characters 0 to 38:
-  type o of array[[10]] of n subtypes m;
-  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-  ASL Type error: a subtype of m was expected, provided array [[10]] of n.
-  [1]
+  $ aslref subtype-satisfaction-arrray.asl
```
hrutvik:
```suggestion
  $ aslref subtype-satisfaction-array.asl
```
(extra "r")

asllib/tests/regressions.t/subtype-satisfaction-example.asl
hrutvik, line 1: This is a bit bizarre as a regression test, in that most of the illegal things are commented out. Perhaps we should factor them out into separate tests?

asllib/Interpreter.ml
```diff
@@ -1427,15 +1415,14 @@ module Make (B : Backend.S) (C : Config) = struct
 
   (* Evaluation of Catchers *)
   (* ---------------------- *)
-  and eval_catchers env catchers otherwise_opt s_m : stmt_eval_type =
+  and eval_catchers catchers otherwise_opt s_m : stmt_eval_type =
     (* [catcher_matches t c] returns true if the catcher [c] match the raised
        exception type [t]. *)
     (* Begin EvalFindCatcher *)
     let catcher_matches =
```
hrutvik: I wonder if we should inline this below?

asllib/parser_errors.messages
Roman-Manevich: Running `aslref` on `config x : bit xor` results in `A `config` declaration must introduce a single name, and have both a type annotation and initialising expression: config name : type = initial_expression;`

asllib/Interpreter.ml
Roman-Manevich: `same_named_type` is syntactic and doesn't need the static environment, which is a big plus.

asllib/Interpreter.ml
Roman-Manevich: It's a bit more readable like this. Also easier to model in `asl.spec`.

asllib/tests/regressions.t/subtype-satisfaction-example.asl
Roman-Manevich: I split the illegal bits into separate tests and renamed the identifiers that suggest subtyping into more neutral ones.

solswords (approved):

asllib/types.ml
Roman-Manevich: We need it to handle something like
```
type T1 of integer;
type T2 of T1;
type T3 of T1;
func main () => integer
 begin
   let x = if ARBITRARY: boolean then 3 as T3 else 2 as T2;
   let a: real = x;
end;
```
to ensure `x` gets typed as `integer` rather than  `T3`, which is what the next case results in.

hrutvik: Thanks, makes sense on deleting - but we can ruse `same_named_type` right?

asllib/Interpreter.ml
hrutvik: There's something a bit unusual here - am I right in thinking that a `catcher` must always have a named type? In which case, it seems a bit odd that we don't:

- enforce named types in the parser/AST (i.e. just have an `identifier`)
- even after removing subtyping, we're carrying around a type in `Throwing` and comparing types in the interpreter

Perhaps we should:
- restrict to named types in the parser/AST
- remove the type from `Throwing`
- compare the record label in the `v` of `Throwing (v, ...)` against the named type (`identifier`) in the `catcher`

?

asllib/types.ml
Roman-Manevich: This doesn't work, since `type_clashes` is used during renaming of stdlib parameters, which requires already having the parameters substituted for signatures.

asllib/types.ml
Roman-Manevich: The names don't have to be equal as for `T3` and `T2` below:
```
type T1 of integer;
type T2 of T1;
type T3 of T1;
func main () => integer
 begin
   let x = if ARBITRARY: boolean then 3 as T3 else 2 as T2;
   let a: real = x;
end;
```

asllib/types.ml
HadrienRenaud: After experimentation, the following change does not work:

```suggestion
  (equal env s t
```

This is because the parameters for `s` or `t` live in different environments, making the type-checking fail to resolve function calls with parameters.
It is also unsound.

asllib/types.ml
HadrienRenaud: Avoid having to have an extra rule `same_named_type` by inlining the code to `get_structure`

```suggestion
match (s.desc, t.desc) with
| T_Int _, T_Int _
| T_Real, T_Real
| T_String, T_String
| T_Bits _, T_Bits _
| T_Bool, T_Bool ->
true
| T_Named s_name, T_Named t_named when String.equal s_name t_name -> true
| T_Named _, _ | _, T_Named _ ->
  type_clashes (make_anonymous env s) (make_anonymous env t)
```

asllib/parser_errors.messages
HadrienRenaud: Thanks but I don't understand how this answer my question

asllib/types.ml
HadrienRenaud: Yes, I'm saying that for this to match, they have to be different otherwise they would fall into the `type_equal env s t` just above

Roman-Manevich: oh, okay.

asllib/parser_errors.messages
Roman-Manevich: Based on https://cambium.inria.fr/~fpottier/menhir/manual.html#sec77
1. ```
   menhir --external-tokens Tokens asllib/Tokens.mly asllib/Parser.mly --base asllib/Parser --update-errors asllib/parser_errors.messages > /tmp/parser_errors.messages
   ```
   Regenerates Menhir's auto-generated comments.

2. ```
   diff -u asllib/parser_errors.messages /tmp/parser_errors.messages
   ```
   The diff only contains differences to states and lookahead tokens.
   Then use the new messages file.

3. ```
   dune build asllib/parser_errors.ml
   ```
   Verifies the file, checking that the sample sentences are correct and irredundant, then generates
   `parser_errors.ml`.

asllib/Interpreter.ml
Roman-Manevich: Let's postpone this to a subsequent PR.

asllib/parser_errors.messages
HadrienRenaud: Thanks for this Roman. However, if I remove the whole sentence and re-run those commands, I still don't get any errors.

For example:
```
❯ diff -u asllib/parser_errors.messages.roman asllib/parser_errors.messages
--- asllib/parser_errors.messages.roman 2026-06-29 16:39:46
+++ asllib/parser_errors.messages       2026-06-29 16:43:15
@@ -316,17 +316,7 @@
 ##
 ## The known suffix of the stack is as follows:
 ## CONFIG ignored_or_identifier
-##
-spec: CONFIG IDENTIFIER COLON BIT XOR
-##
-## Ends in an error in state: 886.
-##
-## decl -> CONFIG ignored_or_identifier as_ty . EQ expr SEMI_COLON [ VAR TYPE READONLY PURE PRAGMA NORETURN LET IMPLEMENTATION IMPDEF FUNC EOF CONSTANT CONFIG ACCESSOR ]
-## decl -> CONFIG ignored_or_identifier as_ty . EQ elided_param_call SEMI_COLON [ VAR TYPE READONLY PURE PRAGMA NORETURN LET IMPLEMENTATION IMPDEF FUNC EOF CONSTANT CONFIG ACCESSOR ]
 ##
-## The known suffix of the stack is as follows:
-## CONFIG ignored_or_identifier as_ty
-##

 A `config` declaration must introduce a single name, and have both a type annotation and initialising expression:
   config name : type = initial_expression;

❯ menhir --external-tokens Tokens asllib/Tokens.mly asllib/Parser.mly --base asllib/Parser --compare-errors asllib/parser_errors.messages --compare-errors asllib/parser_errors.messages.roman
Read 41 sample input sentences and 6 error messages.
Read 42 sample input sentences and 6 error messages.
```

Whole patch available [here](https://github.com/user-attachments/files/29472305/0000.not-comitted.remove-useless-phrase-asl-922.patch).

HadrienRenaud: Ahh but I do get some differences on `asllib/tests/parser-errors.t/run.t` for something like: `config x: integer;`, that sounds good to me, thanks

HadrienRenaud: For reference, the check that I should have run is the following (the difference is not symmetric):
```
❯ menhir --external-tokens Tokens asllib/Tokens.mly asllib/Parser.mly --base asllib/Parser --compare-errors asllib/parser_errors.messages.roman --compare-errors asllib/parser_errors.messages
Read 42 sample input sentences and 6 error messages.
Read 41 sample input sentences and 6 error messages.
File "asllib/parser_errors.messages.roman", line 320, characters 0-37:
Error: this sentence leads to an error in state 879.
No sentence that leads to this state exists in "asllib/parser_errors.messages".
```

asllib/types.ml
Roman-Manevich: That won't work, since once we get the structure of `s` and `t` the names disappear. We need to say that named types of different names do not clash.

asllib/types.ml
Roman-Manevich: Sure, I'm using this now:
```
| T_Named _, T_Named _ ->
        assert (not (same_named_type s t));
        let anon_s = make_anonymous env s and anon_t = make_anonymous env t in
        lowest_common_ancestor ~loc env anon_s anon_t
  ```

File: asllib/types.ml
Comment (HadrienRenaud) [reply asllib/types.ml]: Apologies, I was suggesting to remove the call to `get_structure` at the top of the function

File: asllib/types.ml (hunk 1 above)
Comment (HadrienRenaud) [reply asllib/types.ml]: I'm not sure it's that helpful here, because you've already unpacked `s` and `t`, and probably assertions are not transliterated to the documentation no?

IMHO the following is more readable
```ocaml
| T_Named s_name, T_Named t_name ->
        assert (not (String.equal s_name t_name));
        let anon_s = make_anonymous env s and anon_t = make_anonymous env t in
        lowest_common_ancestor ~loc env anon_s anon_t
```

asllib/types.ml
```diff
@@ -533,8 +525,8 @@ and subtype_satisfies env t s =
 and type_satisfies env t s =
   ((* Type T type-satisfies type S if and only if at least one of the following
       conditions holds: *)
-   (* T is a subtype of S *)
-   subtypes env t s
+   (* T is a same-named type as S *)
```
HadrienRenaud: Comment needs to be updated

asllib/types.ml
Roman-Manevich: assertions do not appear in the ASL Reference.

asllib/doc/asl.spec
```diff
@@ -6423,14 +6409,14 @@ typing function type_satisfies(tenv: static_envs, t: ty, s: ty) -> (b: Bool) | t
...
+  case same_named {
+    type_equal(tenv, t, s) -> True;
     --
     True;
   }
 
-  case not_subtypes {
-    is_subtype(tenv, t, s) -> False;
+  case not_same_named {
+    type_equal(tenv, t, s) -> False;
```
hrutvik: Should these two rules be renamed `case equal` and `case not_equal`?

asllib/doc/asl.spec
```diff
@@ -6493,48 +6479,39 @@ typing relation lowest_common_ancestor(tenv: static_envs, t: ty, s: ty) -> (ty:
   case type_equal {
     type_equal(tenv, t, s) -> True;
     --
-    s;
+    t;
```
hrutvik: Why this change?

asllib/types.ml
hrutvik: What's the outcome of this discussion?

asllib/doc/asl.spec
Roman-Manevich: I'm not sure. I'm reverting it.

asllib/types.ml
HadrienRenaud: Because I don't think GitHub rendered my diff well, I've made this PR: https://github.com/herd/herdtools7/pull/1897

Roman-Manevich: adopted suggestion and removed `samed_named_type` from the code, ASL Reference, and tests.

HadrienRenaud (approved):

asllib/doc/asl.spec
```diff
@@ -4810,15 +4807,17 @@ semantics function find_catcher(tenv: static_envs, v_ty: ty, catchers: list0(cat
...
+    v_name = e_name;
     --
     some(c);
   }
 
   case no_match {
     catchers =: match_cons(c, catchers1);
     c =: (name_opt, e_ty, s);
-    is_subtype(tenv, v_ty, e_ty) -> False;
+    not(v_ty = T_Named(_) && e_ty = T_Named(_) && v_ty = e_ty);
```
hrutvik, line 4719: Is the De Morgan of this a bit more understandable? I.e. a disjunction of inequalities?

hrutvik (approved):

asllib/doc/asl.spec
Roman-Manevich: the advantage of this is that it is syntactically a negation of the case above so it makes it easy to validate mutual exclusion of these two cases.

hrutvik: There is no validation in pace though right? I think we should be prioritising readability

Roman-Manevich: Sorry, by "validation" I meant manual auditing. For me, it's easy to check for mutual exclusion when I see that one case has `expr` and the other has `not(expr)`.

PR #1886, opened 2026-06-25
[gen] Add pair-cell offset annotation for AArch64.

ShaleXIONG:
Expose a non-mixed AArch64 offset annotation to generate ordinary accesses to the second cell of a pair-backed location. The surface spelling follows the current natural type size. 
- with the default 32-bit `int`, the added annotation is `w4`
- with `-type int64_t`, it is `q8`.

We also remove the synthetic `ADD` after `AArch64` pair loads but check both register values in the final condition. 
Given the new annotation and change, we can generate test, e.g. `diyone7 -metadata false -arch AArch64 -oneloc Rfe PaIQ Fre w4 PosWW`
```
AArch64 WW+R+posw4p+PaIQ
{
 int x[2]={0,0};
 uint32_t 0:X0=0; 0:X1=x;
 1:X1=x;
}
 P0                                                      | P1                ;
 MOV W0,#2                                               | LDIAPP W0,W2,[X1] ;
 STR W0,[X1,#4] (* w4 annotation to access second cell *)|                   ; (* Add W0,W0,W2 is removed *)
 MOV W2,#1                                               |                   ;
 STR W2,[X1]                                             |                   ;

exists (x={1,2} /\ 1:X0=1 /\ 1:X2=0) (* Value check for both 1:X0 and 1:X2 *)
```

### DIFF
--- a/gen/AArch64Compile_gen.ml
+++ b/gen/AArch64Compile_gen.ml
@@ -1619,7 +1619,8 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
 
     let emit_ldp_reg opt st init rA =
       let r1,r2,st = next_reg2 st in
-      r1,init,pseudo [do_ldp opt r1 r2 rA;add vloc r1 r1 r2;],st
+      let st = A.set_friends r1 [r2] st in
+      r1,init,pseudo [do_ldp opt r1 r2 rA;],st
 
     let emit_ldp_reg opt idx st _p init rA =
       match opt,idx with
--- a/gen/AArch64Compile_gen.ml
+++ b/gen/AArch64Compile_gen.ml
@@ -1673,7 +1673,11 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
 
 
     let emit_obs t = match t with
-    | Code.Ord | Code.Instr-> emit_load_mixed naturalsize 0
+    | Code.Ord | Code.Instr->
+        fun st p init loc ->
+        let r,init,cs,st = emit_load_mixed naturalsize 0 st p init loc in
+        let st = A.add_type (A.of_reg p r) Cfg.typ st in
+        r,init,cs,st
     | Code.Pte->
         fun st p init loc ->
         let r,init,cs,st = LDR.emit_load_var A64.V64 st p init (Misc.add_pte loc) in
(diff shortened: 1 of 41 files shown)


relokin (commented):
This is great, I haven't had a very thorough look at the code but I am happy with the tests we can now generate. One issue that would be good to address is the type mismatch is tests like this:
```
$> diyone7 -arch AArch64 -metadata false w4 PosWW Rfe PaIQ Fre -oneloc
AArch64 WW+R+posw4p+PaIQ
{
 int x[2]={0,0};
 uint32_t 0:X0=0; 0:X1=x;
 1:X1=x;
}
 P0             | P1                ;
 MOV W0,#2      | LDIAPP W0,W2,[X1] ;
 STR W0,[X1,#4] |                   ;
 MOV W2,#1      |                   ;
 STR W2,[X1]    |                   ;

exists (x={1,2} /\ 1:X0=1 /\ 1:X2=0)
```
Notice that x is an array of int but 0:X0 and 0:X1 are uint32_t.

ShaleXIONG:
> This is great, I haven't had a very thorough look at the code but I am happy with the tests we can now generate. One issue that would be good to address is the type mismatch is tests like this:
> 
> ```
> $> diyone7 -arch AArch64 -metadata false w4 PosWW Rfe PaIQ Fre -oneloc
> AArch64 WW+R+posw4p+PaIQ
> {
>  int x[2]={0,0};
>  uint32_t 0:X0=0; 0:X1=x;
>  1:X1=x;
> }
>  P0             | P1                ;
>  MOV W0,#2      | LDIAPP W0,W2,[X1] ;
>  STR W0,[X1,#4] |                   ;
>  MOV W2,#1      |                   ;
>  STR W2,[X1]    |                   ;
> 
> exists (x={1,2} /\ 1:X0=1 /\ 1:X2=0)
> ```
> 
> Notice that x is an array of int but 0:X0 and 0:X1 are uint32_t.

Should we remove `unit32_t 0:X0=0` as there is a `MOV W0 #2`? So we keep it similar to the test that only change the first cell, which we can generate before this PR: `diyone7 -arch AArch64 -metadata false PosWW Rfe PaIQ Fre -oneloc`
```
AArch64 WW+R+PaIQ
{
 int x[2]={0,0};
 0:X1=x;
 1:X1=x;
}
 MOV W0,#2   | LDIAPP W0,W2,[X1] ;
 STR W0,[X1] | ADD W0,W0,W2      ;
 MOV W2,#1   |                   ;
 STR W2,[X1] |                   ;

exists (x={2,0} /\ 1:X0=1)
```
If so, I think I fixed it in 4d72160cacdfd1a344cc675d37fe3a6a85b30069.

relokin (approved):
Thanks Shale that fixes the problem.

PR #1889, opened 2026-06-29
[herd] Split exclusive accesses from the X event set

relokin:
## Summary

  This PR makes exclusive accesses explicit in `herd` by introducing a distinct `EX` event set instead of treating them as
  part of `X`.

  The main goal is to let CAT models distinguish reservation-based exclusive operations from other atomic/RMW events more
  precisely, while also exposing reusable helpers for reasoning about RMWs.

  ## What changed

  - add dedicated exclusive annotations for AArch64 (`EX`, `EXA`, `EXL`) and thread them through the AArch64/ASL semantics
  - use `EX` for exclusive/LR-SC style instructions in ARM, RISC-V, MIPS, and PPC
  - keep `X` separate from `EX`, rather than conflating all of these events in one set
  - extend `stdlib.cat` with reusable `EX`, `lxsx`, `amo`, and `rmw` helpers
  - update `ppc-checks.cat` so `scXX` is defined over exclusive events
  - remove the AArch64 `A162` test which duplicates `A44`

  ## Why

  Previously, exclusive accesses were folded into the generic `X` set. That made it harder to express model rules that
  specifically target reservation-based exclusives rather than all atomic accesses.

  By separating `EX` from `X`, the model becomes clearer and CAT code can talk about exclusives and RMWs directly.


relokin:
One question I had is whether we want to remove some of the code that's used to compute `atomic_load_store`?

herd/ASLSem.ml
HadrienRenaud: I think that this should now be just atomicity:

```suggestion
      let is_ax x n = if is_atomic then x else n in
```

Maybe it would be worth adding an assertion that no instruction can be exclusive and atomic at the same time:
```ocaml
      assert (not (is_exclusive && is_atomic));
```

herd/ASLSem.ml
```diff
@@ -314,10 +314,13 @@ module Make (Conf : Config) = struct
...
       let an =
-        if (not is_read) && is_release then is_ax XL L
+        if (not is_read) && is_release && is_exclusive then EXL
+        else if (not is_read) && is_release then is_ax XL L
         else if is_noret then NoRet
+        else if is_read && is_acquiresc && is_exclusive then EXA
         else if is_read && is_acquiresc then is_ax XA A
         else if is_read && is_acquirepc then is_ax XQ Q
+        else if is_exclusive then EX
         else is_ax X N
```
HadrienRenaud:
```suggestion
        if (not is_read) && is_release then is_eax EXL XL L
        else if is_noret then NoRet
        else if is_read && is_acquiresc then is_eax EXA XA A
        else if is_read && is_acquirepc then is_ax XQ Q
        else is_eax EX X N
```

with `is_eax` defined with `is_ax` as:
```ocaml
let is_eax ex x n = if is_exclusive then ex else is_ax x n
```

herd/libdir/stdlib.cat
```diff
@@ -79,3 +79,12 @@ let map f =
   do_map
 
 let LKW = try LKW with emptyset
+
+let PTE = try PTE with emptyset
+let EX = try EX with emptyset
+let X = try X with emptyset
+
+let lxsx = [R & EX]; po \ (po; [M & EX]; po); [W & EX]
```
HadrienRenaud, line 90: This is different semantics to what it was before (where it needed to be the same memory location). IIUC this is wanted for AArch64, but is this what is wanted for the other arch?

HadrienRenaud (commented):
> One question I had is whether we want to remove some of the code that's used to compute `atomic_load_store`?

At the current state of the PR, I don't see why we would want to keep it.

maranget:
> * remove the obsolete AArch64 `A162` test, which relied on the previous treatment of exclusives

Perhaps we can keep this test?

relokin:
> > * remove the obsolete AArch64 `A162` test, which relied on the previous treatment of exclusives
> 
> Perhaps we can keep this test?

Sorry Luc, the message is incorrect - I didn't remember correctly why I removed this test. `herd/tests/instructions/AArch64/A162.litmus` is a duplicate of `herd/tests/instructions/AArch64/A44.litmus`.

maranget:
There is a difficulty with RISCV, the following test yields no outcome at all:
```
RISCV TST
{
 0:x5=x;
 1:x5=x;
}
 P0                   | P1          ;
 ori x7,x0,1          | ori x6,x0,2 ;
 amoswap.w x6,x7,(x5) | sw x6,0(x5) ;

exists ([x]=1 /\ 0:x6=0)
```
The reason may be that RISCV AMO instructions are represented by a specific "read and write" event.

relokin:
> There is a difficulty with RISCV, the following test yields no outcome at all:
> 
> ```
> RISCV TST
> {
>  0:x5=x;
>  1:x5=x;
> }
>  P0                   | P1          ;
>  ori x7,x0,1          | ori x6,x0,2 ;
>  amoswap.w x6,x7,(x5) | sw x6,0(x5) ;
> 
> exists ([x]=1 /\ 0:x6=0)
> ```
> 
> The reason may be that RISCV AMO instructions are represented by a specific "read and write" event.

Thanks Luc, I wasn't really familiar with the Amo actions. I've excluded `id` from the definition of `amo` here: https://github.com/herd/herdtools7/pull/1889/commits/491e75917741f7583cf5ebecd739f6e6b88b5d09

herd/libdir/stdlib.cat
relokin: this is true but as we've discussed in the past, this test is allowed by the architecture:
```
AArch64 CoRW1+podxx
{
 0:X0=x; 0:X2=y;
}
 P0              ;
 LDXR W1,[X0]    ;
 MOV W3,#1       ;
 STXR W4,W3,[X2] ;
exists([y]=1)
```
Which I took it to mean that the requirement for same location is not valid.

HadrienRenaud: Sorry my comment was confusing. I understand that the change you made for all the architectures is wanted for AArch64. What I'm not sure is if this applies for other architectures?

In other words, I agree that this AArch64 test should behave like what your implementation does, but I'm wondering how a similar test should behave for risc-v or ppc.

herd/ASLSem.ml
relokin: Thanks @HadrienRenaud, fixes in https://github.com/herd/herdtools7/pull/1889/commits/1bd06f1ac4b9c931e01290ac8eb360f64f88ecc6

herd/libdir/stdlib.cat
relokin: That's a good point, and I am happy to start to split out the definitions for AArch64/AArch32/arm and all other architectures. Also the case of hardware updates is very specific to AArch64 too.

relokin: @maranget, I am not very familiar with the ordering requirements for ll/sc for other architectures. Do you think the definition of lxsx should be restricted to same location for MIPS, RISCV and PPC?

maranget: No idea! The RISCV description clearly states that the "_reserved bytes_" _contains_ (and not is equal to) the bytes written, Additionally the size of these "reserved bytes" is not specified. So I guess that the new definitiobn is acceptable for RISCV.

relokin:
> > One question I had is whether we want to remove some of the code that's used to compute `atomic_load_store`?
> 
> At the current state of the PR, I don't see why we would want to keep it.

Commits https://github.com/herd/herdtools7/pull/1889/commits/222e5094556e31b0f29b5867364dbaa7b9a9e2cf and https://github.com/herd/herdtools7/pull/1889/commits/5b6af157952fae9ebc34d44f75573ddad021606b are the first steps in this process. Very quickly, I am stepping into the territory of other architectures and rc11.ml and I am not sure what the right thing would be.

@maranget do you have a opinion about this?

herd/libdir/stdlib.cat
relokin: Should we retain the old behaviour for architectures other than AArch64/AArch32/arm then?

maranget:
I guess we can get rid of dead code, BTW, there is a shadowed definition of `lxsx` in stdlib.cat, let us get rid of it too!
```
diff --git a/herd/libdir/stdlib.cat b/herd/libdir/stdlib.cat
index e48ddda7f..74e51a8e8 100644
--- a/herd/libdir/stdlib.cat
+++ b/herd/libdir/stdlib.cat
@@ -33,9 +33,6 @@ let fencerel(B) = (po & (_ * B)) ;  po
 (* Control+cfence *)
 let ctrlcfence(ctrl,CFENCE) = (ctrl & (_ * CFENCE)) ; po
 
-(* Make the difference between load-reserve/store conditional and amo insructions *)
-let lxsx = rmw \ amo
-
 (* Backward compatibility *)
 let inv-field = try inv-domain with 0
```

herd/libdir/stdlib.cat
HadrienRenaud: Maybe the version in the standard library needs to have the old behaviour, and `aarch64utils` could redefine it with the Arm-specific definition?

But at this point, I'm wondering why not keep the old computed-in-mem.ml behaviour for non-Arm architectures. And maybe at this point #1876 should be amended for simply removing the handling of implicit atomic accesses because only Arm architectures generate them.

relokin: Happy either way. One of the problems with the old computed-in-mem.ml code is that it is also used for reasons other than computing `rmw`. For example, the optace optimisation relies on having `rmw`and then it does some checks (one of them is wrong though and started the whole discussion).

herd/libdir/stdlib.cat
```diff
@@ -79,3 +76,23 @@ let map f =
...
+let PTE = try PTE with emptyset
+
+(* For architectures that do not have exclusives or LL/SC or LR/SC *)
+let EX = try EX with emptyset
+
+(* For architectures that do not have atomic instructions *)
+let X = try X with emptyset
+
+(* Define pairs of exclusives *)
+let lxsx = [R & EX]; po \ (po; [R & EX]; po); [W & EX]
```
relokin: BTW, I've relaxed this definition even further. AFAIU, for AArch64 at least, we need to pair a LDXR with the next in-po STXR (regardless of location) and any other following in-po STXR, unless we have a new LDXR.

herd/libdir/stdlib.cat
HadrienRenaud: > One of the problems with the old computed-in-mem.ml code is that it is also used for reasons other than computing `rmw`

Oh I had not realised. Maybe this is a good reason to keep the old computed-in-mem.ml code then?

I'm wondering if we could do a "best of both" and check that the two computations are equal in cat? This way it is explicitly defined in cat, and still enables the optimisations or the other uses of `rmw` before the start of the cat interpretation

herd/libdir/stdlib.cat
maranget: Is it possible that in an instruction sequence LDXR; STXR; STXR; the second STXR may succeed? I thought that STXR was cancelling any reservation, whether it succeeds or not.

maranget: And, as far as I know, this is what our semantics does.

relokin: Sorry I think I got confused and I will revert this.

If the STXR is successful then the global monitor has to to go to open access: https://developer.arm.com/documentation/ddi0487/mb/-Part-B-The-AArch64-Application-Level-Architecture/-Chapter-B2-The-AArch64-Application-Level-Memory-Model/-B2-12-Synchronization-and-semaphores/-B2-12-1-Exclusive-access-instructions-and-Non-shareable-memory-locations . In our case the presence of an Explicit Memory Write Effect is the consequence of a successful STXR.

maranget:
FYI I have made a draft PR #1894 that computes the EX and X sets in OCaml differently. The main feature is a (hopfully) fixed computation of rmw in OCaml.  Namely, the computation of PR #1876 is incorrect by lack of making the difference between exclusives and AMO's. The PR also adds a predicate `is_exclusive` on events.

herd/libdir/x86tso-mixed.cat
```diff
@@ -22,7 +22,7 @@ let rec ob = obs; si
         | ob; ob
 
 (* Internal visibility requirement *)
-acyclic po-loc | ca | rf as internal
+acyclic po-loc | amo | ca | rf as internal
```
maranget, line 25: You may consider applying the same fix  to x86tso.cat, which is the default model of the architecture X86. We may also add at lease the corresponding test for this (obsolete I admit) architecture:
```
X86 CoRW2+rmw
{
}
 P0           | P1         ;
 MOV EAX,$1   | MOV [x],$2 ;
 XCHG [x],EAX |            ;

exists ([x]=2 /\ 0:EAX=2)
```

maranget (approved):
LGTM. Some suggestions for a future PR
1. Add LR/SC and AMO tests for all architectures.
2.  As long as there is is_atomic on events, we can define teh set "X" in machModelChecker.ml and avoid arch-specific defintions.

herd/libdir/x86tso-mixed.cat
relokin: Done

relokin:
> As long as there is is_atomic on events, we can define teh set "X" in machModelChecker.ml and avoid arch-specific defintions.

Wouldn't it be preferable to move towards a solution where the `amo` is computed when we build semantics. For example, in `herd/AArch64Sem.ml` when we build the semantics of a SWP we know that the Explicit Memory Read Effect and the Explicit Memory Write Effect should be in `amo`. Similarly for a hardware update of af and db. If we introduce a new field in `event_structure` then we would only need to modify the monadic operators to populate it correctly.

relokin:
@maranget @HadrienRenaud are you happy with this PR? If you are, I'll rebase squash and merge.

maranget:
> > As long as there is is_atomic on events, we can define teh set "X" in machModelChecker.ml and avoid arch-specific defintions.
> 
> Wouldn't it be preferable to move towards a solution where the `amo` is computed when we build semantics. For example, in `herd/AArch64Sem.ml` when we build the semantics of a SWP we know that the Explicit Memory Read Effect and the Explicit Memory Write Effect should be in `amo`. Similarly for a hardware update of af and db. If we introduce a new field in `event_structure` then we would only need to modify the monadic operators to populate it correctly.

Both solutions are fine (although mutually exclusive...). The first one (in MachModelChecker.ml) is compatible with the idea that such "atomic" accesses are a common to all architectures and avoids some duplication of code. The second (in semantics)n highlights the connexion with semantics.

maranget:
> @maranget @HadrienRenaud are you happy with this PR? If you are, I'll rebase squash and merge.

I'd like to have a last look at the last commit.

maranget:
Hi @relokin. LGTM, merge at your earliest convenience.

HadrienRenaud (approved):
This looks good to me, thanks.

relokin:
Merged, thanks @maranget and @HadrienRenaud !

PR #1890, opened 2026-06-30
[herd] Bump asl pseudocode 2026-06

HadrienRenaud:
This PR bumps the version of the pseudocode from 2025-12 to 2026-06.

Main changes on the herd+ASL side:
1. For barriers, `DMB` non `SY` are deprecated, and unsupported by the ASL code in the Arm ARM. I have chosen to edit the test that was testing this feature, because I wasn't able to have the same behaviour as what herd currently supports.
2. For VMSA, the function `SetAccessFlag` now takes the `fault_perm` argument, similarly as what is done in `SetDirtyState`. They also now check for the `fault_perm` permissions faults and have the correct behaviour wrt the access flag. I've thus removed our patched functions that were implementing this behaviour. @maranget could you have a look and tell me if I've missed anything with this?

Note: the tests probably don't pass between each commit, so this PR will need to be squashed.


Makefile
```diff
@@ -171,6 +171,8 @@ test:: test.herd-asl.inst.AArch64.kvm
 
 test-local:: test.herd-asl.inst.AArch64.sve
 
+test-all-asl:: test.herd.inst.ASL
```
hrutvik, line 174: Were these completely missing before, or tested in some other target?

hrutvik (approved):
I'm afraid I can't say much about the updates in `patches-vmsa.asl`, litmus tests, or `herd/{ASLSem,AArch64ASLSem}.ml`

Makefile
HadrienRenaud: They were just in another target but they are quite relevant for `make test-all-asl`

herd/libdir/asl-pseudocode/implementations.asl
```diff
@@ -338,3 +344,9 @@ begin
     return FALSE;
   end;
 end;
+
+// =============================================================================
+
+// Not defined in the arm arm
+constant NUM_VMIDBITS = 16;
+constant NUM_ASIDBITS = 16;
```
hrutvik: Repeated in `herd/libdir/asl-pseudocode/type-checking-implementations.asl` - is this intentional?

hrutvik: Discussed offline that it's intentional, but suggested to copy comments about the provenance of these over to this file

hrutvik (approved):

maranget (approved):
LGTM.

The new implementation of the check for setting the access flag is much more direct.

Changing the test FENCES.litmus is also a good idea, as not changing it would mean not testing the ASL implementation of DMB.

solswords (approved):

PR #1895, opened 2026-07-02
[gen] Change CSEL dependency instruction sequences

ShaleXIONG:
To fix a problem that memory tag annotation combined with Csel such as `T DpDataCseldW` generates tests that triggered error `Illegal operation readbit63 on x:green (User error)` in `herd7`, we change the generated instruction sequences. In the new sequence, `AArch64` CSEL dependency uses register self-compare instead of hardcode comparison against zero. This makes the selected CSEL path deterministic. It also removes the extra register instruction, e.g., `diyone7 -arch AArch64 -variant memtag DpAddrCselsW Rfi -oneloc` now generates:
```
`AArch64 CoRW1+addrcsels-rfi
Variant=memtag
Com=Rf
Orig=DpAddrCselsW Rfi
"DpAddrCselsW Rfi"
{
 0:X0=x:green;
}
 P0                  ;
 LDR W1,[X0]         ;
 CMP W1,W1           ; (* previously MOV W3,#1  CMP W1, #0 *)
 CSEL W2,W3,W4,EQ    ; (* previously an extra clear-up AND W2,W2,#2  to set W2=0 *)
 MOV W5,#1           ;
 STR W5,[X0,W2,SXTW] ;
```
The similar sequence is also applied to `DpDataCsel*`. While for `DpCtrlCsel*`, for example `diyone7 -arch AArch64 -variant memtag DpCtrlCselsW Rfi -oneloc`, now generates:
```
AArch64 CoRW1+ctrlcsels-rfi
Variant=memtag
Com=Rf
Orig=DpCtrlCselsW Rfi
"DpCtrlCselsW Rfi"
{
 0:X0=x:green;
}
 P0                ;
 LDR W1,[X0]       ;
 CMP W1,W1         ; (* previously CMP W1,#0 *)
 CSINC W2,W3,W4,EQ ; (* previous the false branch is W2 but now is a fresh register W4 *)
 CBNZ W2,LC00      ;
 LC00:             ;
 MOV W5,#1         ;
 STR W5,[X0]       ;

exists (0:X1=1)
```

### DIFF
--- a/gen/AArch64Compile_gen.ml
+++ b/gen/AArch64Compile_gen.ml
@@ -225,9 +225,9 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
     let cmpi r i = do_cmpi vloc r i
 
     let do_csel v r1 r2 r3 = I_CSEL (v,r1,r2,r3,EQ,Cpy)
-    let do_cinc v r1 r2 r3 = I_CSEL (v,r1,r2,r3,EQ,Inc)
 
-    let cmp r1 r2 = op3r vloc SUBS ZR r1 r2
+    let do_cmp v r1 r2 = op3r v SUBS ZR r1 r2
+    let cmp r1 r2 = do_cmp vloc r1 r2
 
     let b lbl = I_B (BranchTarget.Lbl lbl)
     let bne lbl = I_BC (NE,BranchTarget.Lbl lbl)
@@ -2275,10 +2275,7 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
       | NoCsel -> fun src dst -> [calc0 vdep src dst],st
       | OkCsel ->
          fun dst src ->
-           let r3,st = next_reg st in
-           let r4,st = next_reg st in
-           [do_movi vdep r3 1; do_cmpi vdep src 0;
-            do_csel vdep dst r3 r4; andi vdep dst dst 2;],st
+           [do_cmp vdep src src; do_csel vdep dst ZR ZR;],st
 
     let emit_access_dep_addr csel vdep st p init e rd =
       let r2,st = next_reg st in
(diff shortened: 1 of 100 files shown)


gen/AArch64Compile_gen.ml
```diff
@@ -2741,8 +2741,9 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
       | OkCsel ->
          let r2,st = next_reg st in
          let r3,st = next_reg st in
```
relokin: Not the core of this change but in practice r3 and r4 are initialized to 0 and never set of used. We could also use ZR.

gen/AArch64Compile_gen.ml
```diff
@@ -2741,8 +2741,9 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
...
+         let r4,st = next_reg st in
          pseudo
-           [do_cmpi vdep r 0; do_cinc vdep r2 r3 r2;]@
+           [do_cmp vdep r r; do_cinc vdep r2 r3 r4;]@
```
relokin: Should also change this to CSEL? I think it would rather easy to add for all variants of CSEL and introduce new relaxations if we wanted to (not that we should do it).

gen/AArch64Compile_gen.ml
```diff
@@ -2266,8 +2267,7 @@ module Make(Cfg:Config) : XXXCompile_gen.S =
          fun dst src ->
            let r3,st = next_reg st in
            let r4,st = next_reg st in
-           [do_movi vdep r3 1; do_cmpi vdep src 0;
-            do_csel vdep dst r3 r4; andi vdep dst dst 2;],st
+           [do_cmp vdep src src; do_csel vdep dst r3 r4;],st
```
relokin: Same here r3 and r4 could have been the ZR.

gen/AArch64Compile_gen.ml
ShaleXIONG: I will change to ZR consistently to all three cases

gen/AArch64Compile_gen.ml
ShaleXIONG: Oh sure. That will be easy change. I will double check on herd7 results before and after.

ShaleXIONG: Done.

gen/AArch64Compile_gen.ml
ShaleXIONG: Done

relokin (approved):
LGTM Shale, please feel free to rebase and merge.

PR #1896, opened 2026-07-03
[herd] Load-Reserve Store-Conditional may pair on different addresses

maranget:
The following test may succeed on hardware and is legal.
```
AArch64 L020
(* Observed on many systems with `-mode presi` or `-s 4` *)
{
 0:X0=x; 0:X2=y;
}
 P0              ;
 LDXR W1,[X0]    ;
 MOV W3,#1       ;
 STXR W4,W3,[X2] ;
exists([y]=1)
```
Have **herd7** to perform store-exclusive even when the reserved address is different. In practive we encode the absence of a reservation with the ResAddr (or `RES`) register holding zero. Hence the store exclusive may succeeed whenever ResAddr does not hold zero. Further notice that executing a store exclusive instruction voids the reservation, regardless of  the write being performed or not.

### DIFF
--- a/herd/AArch64ASLSem.ml
+++ b/herd/AArch64ASLSem.ml
@@ -1051,18 +1051,20 @@ module Make (TopConf : AArch64Sig.Config) (V : Value.AArch64ASL) :
     let tr_cst tr =
       Constant.map tr Misc.identity Misc.identity Misc.identity
 
-    let aarch64_to_asl_bv_cst sz = function
+    let aarch64_to_asl_bv var_ val_ sz = function
       | V.Var _ as v ->
           let nv = V.fresh_var () in
-          let nid =
-            match nv with
-            | V.Var n -> n
-            | _ -> assert false in
+          let nid = match nv with V.Var n -> n | _ -> assert false in
           let eq =
             let op = AArch64Op.Extra1 (ASLOp.ToBV sz) in
             M.VC.Assign (nv,M.VC.Unop (Op.ArchOp1 op,v))  in
-          (ASLS.A.V.freeze nid, [eq])
-      | V.Val cst -> (tr_cst (ASLScalar.convert_to_bv sz) cst, [])
+          (var_ nid, [eq])
+      | V.Val cst -> (val_ (tr_cst (ASLScalar.convert_to_bv sz) cst), [])
+
+    let aarch64_to_asl_bv_cst = aarch64_to_asl_bv ASLS.A.V.freeze Fun.id
+
+    let aarch64_to_asl_bv_v =
+      aarch64_to_asl_bv (fun n -> ASLS.A.V.Var n) (fun cst -> ASLS.A.V.Val cst)
 
     let aarch64_to_asl = function
       | V.Var v -> ASLS.A.V.Var v
@@ -1149,12 +1151,21 @@ module Make (TopConf : AArch64Sig.Config) (V : Value.AArch64ASL) :
       let pc_val =
         ASLS.A.V.scalarToV (ASLScalar.bv_of_int_sized 64 ii.A.addr)
       in
+      let resaddr_val, eqs =
+        match A.look_reg AArch64Base.ResAddr ii.A.env.A.regs with
+        | Some v ->
+          let v, eqs2  = aarch64_to_asl_bv_v (if is_vmsa then 56 else 64) v in
+          Some v, eqs2 @ eqs
+       | None -> None, eqs
+      in
       let st =
         ASLS.A.state_empty
         |> state_add (global_loc "PSTATE") pstate_val
         |> state_add (global_loc "_PC") pc_val
         |> List.fold_right add_arch_reg_if_present ASLBase.gregs
-        |> add_reg_if_present AArch64Base.ResAddr (global_loc "RESADDR")
+        |> (match resaddr_val with
+            | Some v -> state_add (global_loc "RESADDR") v
+            | None -> Fun.id)
         |> add_reg_if_present AArch64Base.SP (global_loc "_SP_EL0")
         |> (if is_vmsa then
               state_add (global_loc "D128")
--- a/herd/AArch64ASLSem.ml
+++ b/herd/AArch64ASLSem.ml
@@ -1135,10 +1135,13 @@ module Make (TopConf : AArch64Sig.Config) (V : Value.AArch64ASL) :
       let state_add loc v st =
         ASLS.A.state_add st (ASLS.A.Location_reg (ii.A.proc, loc)) v
       in
-      let add_reg_if_present reg loc st =
+      let add_reg_if_present ?(setzero=false) reg loc st =
         match A.look_reg reg ii.A.env.A.regs with
         | Some v -> state_add loc (aarch64_to_asl v) st
-        | _ -> st
+        | _ ->
+           if setzero then
+             state_add loc ASLS.V.zero st
+           else st
       in
       let add_arch_reg_if_present reg =
         add_reg_if_present reg (ASLBase.ArchReg reg)
--- a/herd/libdir/asl-pseudocode/physmem-std.asl
+++ b/herd/libdir/asl-pseudocode/physmem-std.asl
@@ -123,7 +123,7 @@ begin
   SuccessVA = SomeBoolean();
 
   if SuccessVA then
-    CheckEq(address, reserved);
+    CheckProp(reserved != Zeros{64});
   end;
 
   return SuccessVA;
--- a/herd/libdir/asl-pseudocode/physmem-vmsa.asl
+++ b/herd/libdir/asl-pseudocode/physmem-vmsa.asl
@@ -249,7 +249,7 @@ begin
   SuccessPA = SomeBoolean();
 
   if SuccessPA then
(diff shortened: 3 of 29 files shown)


herd/tests/instructions/AArch64/L106.litmus
```diff
@@ -0,0 +1,10 @@
+AArch64 L106
```
relokin, line 1: Is this practically the same as L020?

herd/tests/instructions/AArch64/L020.litmus
```diff
@@ -1,6 +1,7 @@
 AArch64 L020
 (* LX/SX with different address, success is Constrained Upredictable.
    Normaly excluded, appears with  -variant CU *)
+Variant=CU
```
relokin: Does CU make a difference for this test?

I think, we might want to demonstrate that with CU a STXR will succeed regardless. Perhaps, a test with a STRX only - without a LDXR?

relokin (commented):
Overall looks good to me but I have a couple of comments about the tests.

herd/tests/instructions/AArch64/L106.litmus
maranget: Nice catch, I delete L106 and remove the added line (Variant=CU) to L020. Notice that L107 is also a duplicate (of L021). Hence I also delete L107.

herd/tests/instructions/AArch64/L020.litmus
maranget: You mean this test:
```
AArch64 L106
(* Without -variant CU, this test fails *)
Variant=CU
{
0:X0=x; 0:X1=1;
}
  P0             ;
 STXR W2,W1,[X0] ;
exists [x]=1
```

maranget:
Hi @relokin. What this PR basically does is changing the test that controls performing the store conditional.

First observe that the reservation (a.k.a exclusive monitor) is implemented by using a unique register RES (for AArch64, VMSA and ASL. VMSA+ASL uses an additional flag). Schematic behaviour is a follows
1. Initially this register holds the null pointer (_i.e._ 0).
2. LDXR sets RES to its effective address.
3. STRX perform a check on RES : (a) old check compare STXR effective address with RES; or (b) new check ,  checks that the contents of RES is different from null. Perform store when the check passes. Always reset RES to null.

See also the tests L020 (LDXR and STXR to different addresses, store was always failing now it may succeed) and L021 (LDXR followed by two STXR, all to the same address, the second STXR never succeeds).

Finally an new test L106 consists in a simple STXR, without LDXR, the store normally always fails but may succeed when `-variant ConstrainedUnredictable` is specified.

maranget:
I have also added L107 (L021 with `-variant CU`). The current implementation of this PR can perform the second STXR. I am not sure at all that the result is correct, as the description of the local monitor does suggest that this variant have no impact on the local monitor being cleared ([here](https://developer.arm.com/documentation/ddi0487/mb/-Part-B-The-AArch64-Application-Level-Architecture/-Chapter-B2-The-AArch64-Application-Level-Memory-Model/-B2-12-Synchronization-and-semaphores/-B2-12-1-Exclusive-access-instructions-and-Non-shareable-memory-locations)).
```
AArch64 L107
(* Can the second STXR succeed? *)
Variant=CU
{
int x=1;
0:X0=x;
}
 P0               ;
 MOV W3,#2        ;
 MOV W5,#3        ;
 LDXR W1,[X0]     ;
 STXR W4,W3,[X0]  ;
 STXR W6,W5,[X0]  ;
exists 0:X4=1 /\ 0:X6=0 /\ x=3
```

herd/tests/instructions/AArch64/L020.litmus
relokin: Yes!

relokin:
Looks good to me. Perhaps one concern I have is that for AArch64 `-variant CU` is not modelling any architecturally valid behaviours. In other words, it shouldn't be used by most users.

maranget:
> Looks good to me. Perhaps one concern I have is that for AArch64 `-variant CU` is not modelling any architecturally valid behaviours. In other words, it shouldn't be used by most users.

Thanks @relokin. I'll delay considering merging until PR #1889 is merged.

herd/tests/instructions/AArch64/L020.litmus
maranget: Added

maranget:
Hi @relokin and @HadrienRenaud. Rebased on master since prerequisite PR #1889 is merged, all commits squashed into two commits, ready for review.

herd/AArch64ASLSem.ml
```diff
@@ -1133,10 +1133,13 @@ module Make (TopConf : AArch64Sig.Config) (V : Value.AArch64ASL) :
...
       in
-      let add_reg_if_present reg loc st =
+      let add_reg_if_present ?(setzero=false) reg loc st =
         match A.look_reg reg ii.A.env.A.regs with
         | Some v -> state_add loc (aarch64_to_asl v) st
-        | _ -> st
+        | _ ->
+           if setzero then
+             state_add loc ASLS.V.zero st
+           else st
```
HadrienRenaud, line 1144: I don't really understand why this is needed. In both `physmem-*.asl`, the address is initialised to `Zeros{_}`

Also the initial environment would not be correctly typed checked: this is an integer and we expect a 56 or 64 bits bitvector (depending on whether VMSA is activated, i.e. whether `RESADDR` stores a physical or virtual address)

herd/libdir/asl-pseudocode/physmem-std.asl
```diff
@@ -123,7 +123,7 @@ begin
   SuccessVA = SomeBoolean();
 
   if SuccessVA then
-    CheckEq(address, reserved);
+    CheckProp(reserved != 0);
```
HadrienRenaud: `reserved` has type `bits(64)` here

```suggestion
    CheckProp(!IsZeros(reserved));
```

or


```suggestion
    CheckProp(reserved != Zeros{64});
```

herd/libdir/asl-pseudocode/physmem-vmsa.asl
```diff
@@ -249,7 +249,7 @@ begin
   SuccessPA = SomeBoolean();
 
   if SuccessPA then
-    CheckEq(paddress.address, reserved);
+    CheckProp(reserved != 0);
```
HadrienRenaud: `reserved` has type `bits(56)` here

```suggestion
    CheckProp(!IsZeros(reserved));
```

Or

```suggestion
    CheckProp(reserved != Zeros{56});
```

HadrienRenaud (approved):
This looks ok to me but I think the handling of reserved addresses in ASL could be cleaner.

relokin:
Not really the point of this PR so you can go ahead and merge as soon as @HadrienRenaud is happy but I don't understand why we support `-variant cu` in AArch64. -variant cu models behaviours which are architecturally forbidden. If the point of `-variant cu` is to enable forbidden behaviours then I think we need to change the name of this variant. CONSTRAINED UNPREDICTABLE has a specific meaning for the AArch64 architecture and the behavious of `-variant cu` are not CONSTRAINED UNPREDICTABLE. Perhaps these should have been behaviours one would expect from `-through all`?

herd/libdir/asl-pseudocode/physmem-vmsa.asl
maranget: Consider two STXR in a row. The first STXR sets the `RES` register to zero (integer zero) and this value is seen as zero (integer) by the next STXR. In the case we have no LDXR nor STXR before an exclusive instruction, `RES` is not initialised, hence the `setzero` flag above...

maranget:
I do not understand the intent of `-variant CU`. I'd gladly keep it out of LR/SC if you think this is the thing to do. As far as I remember, some usage of this variant in mixed-size mode was less problematic.

maranget:
> some usage of this variant in mixed-size mode was less problematic.

My bad, this usage is for generators... Hence we can suppress `-variant CU` in **herd7**, having STXR to succeed whatever the context is being the only occurrence of this variant.

herd/libdir/asl-pseudocode/physmem-vmsa.asl
HadrienRenaud: Hi @maranget, I think there was a misunderstanding for what I was suggesting, so I implemented it in #1914. Please consider merging it in this branch before merging this PR.

maranget: I  no longer have time to do this in a safe manner, I leave it to you.

herd/AArch64ASLSem.ml
```diff
@@ -1135,10 +1135,13 @@ module Make (TopConf : AArch64Sig.Config) (V : Value.AArch64ASL) :
       let state_add loc v st =
         ASLS.A.state_add st (ASLS.A.Location_reg (ii.A.proc, loc)) v
       in
-      let add_reg_if_present reg loc st =
+      let add_reg_if_present ?(setzero=false) reg loc st =
```
HadrienRenaud, line 1138: I believe this is not necessary anymore? I don't think `add_reg_if_present` is called with `setzero`?

PR #1903, opened 2026-07-08
[tools] Add litmus2desc

fsestini:
This PR introduces a new tool `litmus2desc`. The tool can be used to produce English prose describing an AArch64 litmus test. It explains candidate executions that satisfy the test's final condition, covering a selection of the test's instructions and the Effects they generate.

For example, we can run the following:

```sh
$ cat ./tools/litmus2desc/test/rg_litmus_tests/MP.litmus
AArch64 MP
{
 0:X1=x; 0:X3=y;
 1:X1=x; 1:X3=y;
}
 P0          | P1          ;
 MOV W0,#1   | LDR W4,[X3] ;
 STR W0,[X1] | LDR W6,[X1] ;
 MOV W2,#1   |             ;
 STR W2,[X3] |             ;

exists (1:X4=1 /\ 1:X6=0)

$ litmus2desc -set-libdir ./herd/libdir ./tools/litmus2desc/test/rg_litmus_tests/MP.litmus
This test asks whether the following execution is architecturally Allowed:

* The first Store instruction in program order on P0, i.e. STR W0,[X1], generates an Explicit Memory Write Effect of Location x with value 1;
* The second Store instruction in program order on P0, i.e. STR W2,[X3], generates an Explicit Memory Write Effect of Location y with value 1;
* The first Load instruction in program order on P1, i.e. LDR W4,[X3], generates an Explicit Memory Read Effect of Location y;
* The second Load instruction in program order on P1, i.e. LDR W6,[X1], generates an Explicit Memory Read Effect of Location x;
* The Register X4 on P1 holds the value 1 in the end, because the Explicit Memory Read Effect of Location y on P1 generated by LDR W4,[X3] Reads-from the Explicit Memory Write Effect of Location y on P0 generated by STR W2,[X3];
* The Register X6 on P1 holds the value 0 in the end, because the Explicit Memory Read Effect of Location x on P1 generated by LDR W6,[X1] Reads-from the initial state, and therefore is Coherence-before the Explicit Memory Write Effect of Location x on P0 generated by STR W0,[X1].
```

The `README.md` and included test suite should be good starting places to
better understand what the tool can accomplish at the moment.


psafont:
This tools looks very interesting, I've tried this on some basic examples to try and understand the tests better
```
$ cat herd/tests/instructions/AArch64/A01.litmus
AArch64 A01
(* Tests ldrb (register), UXTW *)
{ uint64_t 0:X0; 0:X1=x; uint8_t x; int64_t 0:X2=0; }

P0;
  LDRB W0, [X1, X2, UXTW];

exists (0:X0=0)
```

```
This test asks whether the following execution is architecturally Allowed:

* The LDRB instruction on P0, i.e. LDRB W0,[X1,X2,UXTW], generates an Explicit Memory Read Effect of Location x;
* The Register X0 on P0 holds the value 0 in the end, because the Explicit Memory Read Effect Reads-from the initial state.
This test asks whether the following execution is architecturally Allowed:
```
From the description, what causes the relation between Location x and P0 is not made explicit. Do you have plans to make it so? Maybe it's worth printing again the litmus test to have all the information together, or maybe to add some prose explaining that X1 holds the value x initially, similar to how the value at the end appears in the output.

fsestini:
> From the description, what causes the relation between Location x and P0 is not made explicit. Do you have plans to make it so? Maybe it's worth printing again the litmus test to have all the information together, or maybe to add some prose explaining that X1 holds the value x initially, similar to how the value at the end appears in the output.

Good idea, I could add a flag `--print-litmus` or something, that re-prints the input litmus test to stdout alongside the prose explanation.

lib/misc.ml
```diff
@@ -898,3 +898,11 @@ module Option = struct
     let* x = x_opt in
     Some (f x)
 end
+
+module Result = struct
+  include Result
```
psafont: While I understand this new addition is following the current style in the module, I've found that shadowing the standard library module with a module that includes it tends to bring problems:
- It can cause typing headaches under certain complex scenarios where the base modules are also overriden to inject code with ppxs, this happens because the ocaml typechecker decides that two types are not equivalent.
- Maintenance-wise, it's easy to forget removing a function that is available on all supported compilers.
- When combined with `open`, it can mislead the reader with regards to the provenance and the behaviour of a function. But I consider this to be one of problems with using global `open`s, which are problematic because of the implicit context of the code, which can make reviews much harder than they should be. Note that there are valid alternatives to this, including local opens: https://github.com/lindig/ocaml-style#avoid-opening-modules-globally

fsestini: > While I understand this new addition is following the current style in the module, I've found that shadowing the standard library module with a module that includes it tends to bring problems

I was the one who introduced this style. The intent was to allow one to use `Misc.XXX` as an inclusive extension of `Stdlib.XXX` (in the same way one would use modules from a stdlib extension library like `containers`). The `include` was to minimise refactoring, so that call sites referring to `Misc.XXX.foo` would not need to be changed even if at some point `Stdlib.XXX.foo` gets introduced. That said, I'm not married to this approach, and we can change it if needed. Such changes would be mechanic and guided by the compiler anyway.

> * It can cause typing headaches under certain complex scenarios where the base modules are also overriden to inject code with ppxs, this happens because the ocaml typechecker decides that two types are not equivalent.

That sounds plausible.

> * Maintenance-wise, it's easy to forget removing a function that is available on all supported compilers.

Wouldn't that be the case regardless of `include`?

> * When combined with `open`, it can mislead the reader with regards to the provenance and the behaviour of a function. But I consider this to be one of problems with using global `open`s, which are problematic because of the implicit context of the code, which can make reviews much harder than they should be.

I'm completely with you in considering global opens problematic, but I see it as an orthogonal issue. In other words, given that the intent of `Misc.XXX` is to be an extension of `Stdlib.XXX`, `open Misc.XXX` shouldn't be more of a problem than `open Stdlib.XXX` in my view.

psafont: > Wouldn't that be the case regardless of include?

Well, it's much easier to spot since all the uses of the function will not look like they are using the standard library. This is because there's a local open nearby, or it's fully qualified

fsestini: Patched this in [ee10fc4](https://github.com/herd/herdtools7/pull/1903/commits/ee10fc4c97e3747a69302aa797a2a671f0da692a)

fsestini:
> > From the description, what causes the relation between Location x and P0 is not made explicit. Do you have plans to make it so? Maybe it's worth printing again the litmus test to have all the information together, or maybe to add some prose explaining that X1 holds the value x initially, similar to how the value at the end appears in the output.
> 
> Good idea, I could add a flag `--print-litmus` or something, that re-prints the input litmus test to stdout alongside the prose explanation.

I've added a flag `--print-litmus` that does this.

tools/litmus2desc/lib/top.ml
```diff
@@ -0,0 +1,745 @@
...
+  let render_instruction ctx ins =
+    LitmusTest.render_instruction ~latex:ctx.latex_compat ins
+
+  let opcode_name (instr : LitmusTest.instr) =
+    match instr.instruction with
+    | I_FENCE b -> AArch64Base.pp_barrier b
+    | _ ->
+        let str = LitmusTest.render_instruction ~latex:false instr in
+        let splitted = String.split_on_char ' ' str in
+        List.hd splitted
```
psafont, line 155: Is it guaranteed that a space is present in the string? In other words, can the List.hd throw an exception here? I did see you put quite a bit of effort into non-empty lists :)

tools/litmus2desc/lib/top.ml
```diff
@@ -0,0 +1,745 @@
...
+    | None ->
+        let spoi = ins.static_poi in
+        let similar_instructions =
+          ctx.code
+          |> List.filter (fun (other_ins : LitmusTest.instr) ->
+              Int.equal other_ins.proc proc
+              && (not (Int.equal spoi other_ins.static_poi))
+              && String.equal (opcode_name other_ins) ins_name)
+        in
+        if List.length similar_instructions = 0 then
```
psafont: Using `List.length` can needlessly go through long lists unnecessarily, and so I tend to prefer it as a matter of style, even if it doesn't look consequential here:
```suggestion
        if similar_instructions = [] then
```

psafont (approved):
Thanks! I only have a couple of nits to contribute.

tools/litmus2desc/lib/top.ml
fsestini: Ah yes, completely agree. I think this used to be a `... > 1` check or similar, but then got refactored.
I'll probably just use `Misc.List.is_empty` here.

tools/litmus2desc/lib/top.ml
fsestini: As long as `str` is non-empty, the result of that split will always be a non-empty list regardless of the presence of spaces, so `List.hd` won't throw an exception here:

```ocaml
utop # String.split_on_char ' ' "foo";;
- : string list = ["foo"]
```

Having said that, it doesn't hurt to raise a more informative error message in the unlikely case that the input instruction pretty-prints to an empty string. I'll add that.

fsestini: Actually the documentation of `split_on_char` specifically mention that the output list is always non-empty, so I might leave this as is for now.

fsestini:
Merged, thanks @psafont

PR #1905, opened 2026-07-08
[herd] Expose herd as a dune library

fsestini:
This PR implements step 4 of the plan documented in issue #1782: it makes herd available as a dune library, `herd_core`, so that other OCaml tools in the repository can depend on its internals directly.

Note that the broader plan to make herd's core code properly library-like and client-agnostic is still work in progress. In other words, this PR does not claim to provide a polished or stable API, nor does it address all remaining issues and TODOs described in the aforementioned issue #1782.

That said, the API that herd _does_ currently offer (mainly through its modules `ParseTest` and `Top_herd`) is already useful enough for other internal tools to take advantage of it. See, for example, my upcoming tool litmus2desc (#1903), which runs herd directly in OCaml via the `herd_core` library added here.

Since the proposed `herd_core` library is intended for internal use only (at least of now), I think it is reasonable to expose it now as an incremental step, while continuing the remaining refactoring work tracked in #1782.

Let me know your thoughts.

PS: This PR also updates `herd/rc11.ml` to use explicit namespacing for record field accessors. Adding the module to the wrapped `herd_core` library made the compiler report warnings 40 and 42 for various record accesses. The PR only adds explicit module names to those field accesses, it does not change any of the logic. As an aside, `rc11.ml` does not currently appear to be used anywhere in herd.

### DIFF
--- a/herd/dune
+++ b/herd/dune
@@ -22,10 +22,12 @@
 
 (ocamllex lexConf_herd)
 
-(executable
- (name herd)
- (public_name herd7)
+(library
+ (name herd_core)
  (libraries unix herdtools)
+ (wrapped true)
+ (modules
+  (:standard \ herd cli itimer))
  (modules_without_implementation
   AArch64Sig
   action
@@ -34,3 +36,10 @@
   sem
   XXXMem
   port))
+
+(executable
+ (name herd)
+ (public_name herd7)
+ (package herdtools7)
+ (modules herd cli itimer)
+ (libraries unix herdtools herd_core))
--- a/herd/cli.ml
+++ b/herd/cli.ml
@@ -14,6 +14,7 @@
 (* "http://www.cecill.info". We also give a copy in LICENSE.txt.            *)
 (****************************************************************************)
 
+open Herd_core
 module TR = Top_herd.TestResult
 
 let iter_count (i : ('a -> unit) -> 'b) : ('a -> unit) -> int * 'b =
--- a/herd/herd.ml
+++ b/herd/herd.ml
@@ -16,6 +16,7 @@
 
 (** Entry point to Herd  *)
 
+open Herd_core
 open Printf
 open Archs
 open Opts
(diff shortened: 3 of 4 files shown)


HadrienRenaud (approved):
Hi @fsestini, this looks good to me, I don't see any negative to this.

By curiosity, did you notice a difference in compilation time (from clean or just with small edits)?

fsestini:
> Hi @fsestini, this looks good to me, I don't see any negative to this.
> 
> By curiosity, did you notice a difference in compilation time (from clean or just with small edits)?

Thanks @HadrienRenaud . I did not felt a noticeable difference as I was developing, but did not measure it. I've now run a simple `time` measurement, both from clean. The difference seems mostly noise:

```
This PR:
dune build herd/herd.exe  136.25s user 26.95s system 509% cpu 32.016 total

master:
dune build herd/herd.exe  134.48s user 26.46s system 490% cpu 32.804 total
```

herd/dune
```diff
@@ -34,3 +36,9 @@
   sem
   XXXMem
   port))
+
+(executable
+ (name herd)
+ (public_name herd7)
```
psafont, line 42: Are there any plans on having more opam metadata files on the top level? it might be worth adding `(package herdtools7)` if that's the case

fsestini: > Are there any plans on having more opam metadata files on the top level? it might be worth adding `(package herdtools7)` if that's the case

Added: https://github.com/herd/herdtools7/pull/1905/commits/92904d80bda132de9c71aeea87286f42a4331e1f

herd/dune
```diff
@@ -22,10 +22,12 @@
...
 
-(executable
- (name herd)
- (public_name herd7)
+(library
+ (name herd_core)
  (libraries unix herdtools)
+ (wrapped true)
+ (modules
+  (:standard \ herd cli lexConf_herd itimer))
```
relokin: This is more of a question than a concern. Shouldn't lexConf_herd be part of the library? For example www also needs it. Is there any herd interface that wouldn't need it?

fsestini: Yes, you're right.

relokin (approved):
LGTM, thanks Filippo!

PR #1910, opened 2026-07-09
[ACL2][ASL] ACL2 fixes for pattern simplification change

solswords:


acl2/asl/ast.lisp
```diff
@@ -241,7 +241,7 @@ the given type.")
      (kind pattern_kind-p))
     :short "Pattern matcher: a list of patterns and whether they are matched positively or negatively."
     :layout :fulltree
-    :measure (acl2::two-nats-measure (acl2-count x) 10))
+    :measure (acl2::two-nats-measure (acl2-count x) 20))
```
Roman-Manevich, line 244: out of curiosity: what is this number and where does it come from?

Roman-Manevich (approved):

acl2/asl/ast.lisp
solswords: It's a termination argument for a mutual recursion, in particular the fixing functions that coerce an arbitrary object to one of one of the types described. For a recursive function (or set of mutually recursive functions) to be logically admissible, they have to be proved to terminate -- this rules out inconsistent definitions like f(x) = not(f(x)) or f(x) = 1+f(x). To show that, we provide a measure, which is an ordinal that provably decreases on every recursive call. Since there's no infinite decreasing sequence of ordinals, that means any call of the function terminates. In this case if you're fixing some bad input object into a pattern_matcher, it needs to call the fixing function for pattern_list on something that might not be smaller according to `(acl2-count x)`, so the second argument needs to decrease between those two functions.

Roman-Manevich: right, but why specifically 10 or 20? was it chosen heuristically / experimentally?
Also, thanks for this PR.

PR #1911, opened 2026-07-10
[libdir] Backport new version of stdlib.cat

fsestini:
herd7 is complaining about a missing `rmw` symbol when run on the snapshot versions of the concurrency model:

```sh
$ dune exec herd7 -- ./catalogue/aarch64/tests/MP.litmus -set-libdir ./herd/libdir
Test MP Allowed
States 4
1:X0=0; 1:X2=0;
...

$ dune exec herd7 -- ./catalogue/aarch64/tests/MP.litmus -set-libdir ./herd/libdir/aarch64/ArmARM-L.b
File "./herd/libdir/aarch64/ArmARM-L.b/stdlib.cat", line 37, characters 11-14: unbound var: rmw
$ dune exec herd7 -- ./catalogue/aarch64/tests/MP.litmus -set-libdir ./herd/libdir/aarch64/ArmARM-M.a
File "./herd/libdir/aarch64/ArmARM-M.a/stdlib.cat", line 37, characters 11-14: unbound var: rmw
$ dune exec herd7 -- ./catalogue/aarch64/tests/MP.litmus -set-libdir ./herd/libdir/aarch64/ArmARM-M.b
File "./herd/libdir/aarch64/ArmARM-M.b/stdlib.cat", line 37, characters 11-14: unbound var: rmw
$ dune exec herd7 -- ./catalogue/aarch64/tests/MP.litmus -set-libdir ./herd/libdir/aarch64/ArmARM-M.c
File "./herd/libdir/aarch64/ArmARM-M.c/stdlib.cat", line 37, characters 11-14: unbound var: rmw
```

I believe this is due to recent changes in https://github.com/herd/herdtools7/pull/1889 , so a quick and easy fix to this is to copy the current `./herd/libdir/stdlib.cat`, introduced by that PR, into the snapshot directories. However, perhaps there's some other way.


relokin:
Good catch @fsestini! Have you checked whether the regressions run correctly with M.a, M.b and M.c? We need something not automatic in the CI to test these older releases.

fsestini:
> Good catch @fsestini! Have you checked whether the regressions run correctly with M.a, M.b and M.c?

I ran these targets:

```
test.herd.inst.AArch64
test.herd.inst.AArch64.mixed
test.herd.inst.AArch64.PAC
test.herd.inst.AArch64.kvm
test.herd.inst.AArch64.self
test.herd.inst.AArch64.MTE
test.herd.inst.AArch64.vmsa+mte
test.herd.inst.AArch64.vmsa+ifetch
test.herd.inst.AArch64.neon
test.herd.inst.AArch64.sve
test.herd.inst.AArch64.sme
test.herd.inst.AArch64.gcs
test.herd.cata.aarch64
test.herd.cata.aarch64-mixed
test.herd.cata.aarch64-pick
test.herd.cata.aarch64-faults
test.herd.cata.aarch64-MTE
test.herd.cata.aarch64-PAC
test.herd.cata.aarch64-ifetch
test.herd.cata.aarch64-cas
test.herd.cata.aarch64-VMSA
test.herd.cata.aarch64-ETS2
test.herd.cata.aarch64-ETS3
test.herd-mixed.cata.aarch64
test.herd-mixed.cata.aarch64-pick
```

using two versions of each snapshot libdir `L.b`, `M.a`, `M.b`, `M.c`, i.e. from this PR and from a previous `master` commit predating both this PR and #1889 (68edc64a1b67cc84fde7792093ada64adb63811d). This is how they compare:

- `L.b`: same failures before and after:
  - `test.herd.inst.AArch64.kvm`
  - `test.herd.inst.AArch64.self`
  - `test.herd.inst.AArch64.neon`
  - `test.herd.inst.AArch64.gcs`
- `M.a`: same failures before and after:
  - `test.herd.inst.AArch64.kvm`
  - `test.herd.inst.AArch64.self`
  - `test.herd.inst.AArch64.neon`
  - `test.herd.inst.AArch64.gcs`
- `M.b`: same failure before and after:
  - `test.herd.inst.AArch64.gcs`
- `M.c`: no failures before or after.

The results seem stable before vs after: targets that fail with this PR would fail at 68edc64a1b67cc84fde7792093ada64adb63811d as well, and for reasons that do not appear to be related to `stdlib.cat`. For example, many tests fail under `L.b` and `M.a` due to unbound `same-translation-context` (which is
understandable if `same-translation-context` is a newly-introduced symbol).

I did not run asl regressions, as the snapshots do not include `asl-pseudocode`.

> We need something not automatic in the CI to test these older releases.

Agreed.

relokin (approved):
Amazing thank you very much for this!

relokin:
Something that just occurred to me. Is perhaps the wrong approach to save stdlib.cat in the snapshots?

psafont:
> We need something not automatic in the CI to test these older releases.

I don't think I understand the meaning of "not automatic" here. Do you mean set up a regression test that doesn't use the latest version of everything?

relokin:
> > We need something not automatic in the CI to test these older releases.
> 
> I don't think I understand the meaning of "not automatic" here. Do you mean set up a regression test that doesn't use the latest version of everything?

Sorry, what I meant is that it shouldn't be part of the standard regressions we run on every PR (make test`). It should be a (Makefile?) target that we can run manually.

fsestini:
> Something that just occurred to me. Is perhaps the wrong approach to save stdlib.cat in the snapshots?

If we don't need these snapshots to be self-contained libdirs, removing `stdlib.cat` is also an option, yes.

On a practical level, this means to run tests on the snapshots, one needs to specify the location of `stdlib.cat` with an extra argument:

```sh
$ herd7 -I ./herd/libdir -set-libdir ./herd/libdir/aarch64/ArmARM-M.c ./catalogue/aarch64/tests/MP.litmus
```

however this is not that much of an issue if it's mostly just us who would be running the herd7 CLI on these snapshot anyway (for example, as part of dune tests). The snapshots also get deployed on the herd7 web interface, however in that case in appears the webapp already provides the default `stdlib.cat` as a fallback, so no issues on that front AFAICT.

relokin:
> > Something that just occurred to me. Is perhaps the wrong approach to save stdlib.cat in the snapshots?
> 
> If we don't need these snapshots to be self-contained libdirs, removing `stdlib.cat` is also an option, yes.
> 
> On a practical level, this means to run tests on the snapshots, one needs to specify the location of `stdlib.cat` with an extra argument:
> 
> ```shell
> $ herd7 -I ./herd/libdir -set-libdir ./herd/libdir/aarch64/ArmARM-M.c ./catalogue/aarch64/tests/MP.litmus
> ```
> 
> however this is not that much of an issue if it's mostly just us who would be running the herd7 CLI on these snapshot anyway (for example, as part of dune tests). The snapshots also get deployed on the herd7 web interface, however in that case in appears the webapp already provides the default `stdlib.cat` as a fallback, so no issues on that front AFAICT.

True, I guess then the choice is between updating `stdlib.cat` or making it a symlink to the one in `herd/libdir`.

fsestini:
Thanks for the review @relokin . As per offline discussion, I've decided to symlink `herd/libdir/stdlib.cat` as I don't see any downsides with it and it actually makes future maintenance easier.

I'm planning to merge this at the end of today but do shout out if you have any last minute concerns.

relokin:
Thanks Filippo, the change looks good to me. Just a question have you checked whether herd-www still works?

fsestini:
> Thanks Filippo, the change looks good to me. Just a question have you checked whether herd-www still works?

Indeed I checked the build and run a few tests on the snapshotted libdirs through the web interface. It all looked fine.

fsestini:
> We need something not automatic in the CI to test these older releases.

Tracking this here https://github.com/herd/herdtools7/issues/1916

PR #1912, opened 2026-07-10
[herd,cat] Fix some problems with the C architecture

maranget:
This PR fixes two problems:

1. Since PR #1855, the polymorphic value `V.Empty`  circulates more. Some primitives were performing set and relation  normalisation by themselves, overlooking the cases of empty and universe. We fix this by using the already present and complete normalisation functions `as_set` and `as_rel` slightly generalised to handle different error reporting.
2. PR #1889 changed how the relation  `rmw` is computed  by defining it in stdlb.cat. The Cat definition of the relation `rmw` assumes that atomic effects are gathered in the set `X`, which was not the case of the C semantics. As a result the `rmw` relation was empty when the read-modify-write instructions are implemented by two (atomic) effects.

### DIFF
--- a/herd/CAction.ml
+++ b/herd/CAction.ml
@@ -321,9 +321,11 @@ end = struct
   | _ -> false
 
 (* Architecture-specific sets *)
+  let is_rmw_or_atomic e = is_rmw e || is_atomic e
 
   let arch_sets = [
-    "RMW",(fun e -> is_rmw e || is_atomic e);
+(* Both names below are needed for stdlib.cat and LKMM model respectively *)
+    "X",is_rmw_or_atomic; "RMW",is_rmw_or_atomic;
     "LK", is_lock; "LKR", is_lock_read; "LKW",is_lock_write;
     "LS", is_successful_lock;"LF", is_failed_lock;
     "UL", is_unlock;
--- a/herd/tests/instructions/C/C14.litmus
+++ b/herd/tests/instructions/C/C14.litmus
@@ -1,0 +1,12 @@
+C C14
+{}
+
+P0 (atomic_int* y) {
+  atomic_fetch_add_explicit(y,1,memory_order_relaxed);
+}
+
+P1 (atomic_int* y) {
+  atomic_fetch_add_explicit(y,1,memory_order_relaxed);
+}
+
+forall y=2
--- a/herd/tests/instructions/C/C14.litmus.expected
+++ b/herd/tests/instructions/C/C14.litmus.expected
@@ -1,0 +1,10 @@
+Test C14 Required
+States 1
+[y]=0x2;
+Ok
+Witnesses
+Positive: 2 Negative: 0
+Condition forall ([y]=0x2)
+Observation C14 Always 2 0
+Hash=4ba3dd96c70ac3dd5dac4c7337261858
+
--- a/herd/tests/instructions/C/C15.litmus
+++ b/herd/tests/instructions/C/C15.litmus
@@ -1,0 +1,12 @@
+C C15
+{}
+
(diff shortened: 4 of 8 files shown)


akiyks:
The error is fixed, but I see behavior changes in LKMM litmus tests.

Litmus test:

```
C add_unless_mb

(*
 * Result: Never
 *
 * This litmus test demonstrates that a successful atomic_add_unless
 * acts as a full memory barrier, ensuring that *x=1 propagates to P1
 * before P1 executes *x=2.
 *)

{
	atomic_t z;
}

P0(int *x, int *y, atomic_t *z)
{
	int r0;

	WRITE_ONCE(*x, 1);
	r0 = atomic_add_unless(z,1,5);
	WRITE_ONCE(*y, 1);
}

P1(int *x, int *y)
{
	int r0;

	r0 = READ_ONCE(*y);
	if (r0 == 1)
		WRITE_ONCE(*x, 2);
}

locations [z;]
exists (1:r0=1 /\ x=1)
```

Expected output

```
Test add_unless_mb Allowed
States 2
1:r0=0; [x]=1; [z]=1;
1:r0=1; [x]=2; [z]=1;
No
Witnesses
Positive: 0 Negative: 2
Condition exists (1:r0=1 /\ [x]=1)
Observation add_unless_mb Never 0 2
Time add_unless_mb 0.00
Hash=740ecaca893f828d9079c0c827815245
```

Output as of fix-some-cat-primitives

```
Test add_unless_mb Allowed
States 3
1:r0=0; [x]=1; [z]=1;
1:r0=1; [x]=1; [z]=1;
1:r0=1; [x]=2; [z]=1;
Ok
Witnesses
Positive: 1 Negative: 2
Condition exists (1:r0=1 /\ [x]=1)
Observation add_unless_mb Sometimes 1 2
Time add_unless_mb 0.00
Hash=740ecaca893f828d9079c0c827815245
```

There seems to be lost orderings implied by atomic accesses.

maranget:
Hi @akiyks. This probably is a consequence of PR #1889. I have a look.

fsestini:
Is `delift({})` expected to work? With this PR, I get the following:

```sh
$ dune exec herd7 -- -set-libdir herd/libdir -model repro.cat herd/tests/instructions/AArch64/A01.litmus
File "./repro.cat", line 1, characters 6-16: primitive delift: argument mismatch
```

where `repro.cat` is:

```
empty delift({})
```

maranget:
> Is `delift({})` expected to work? With this PR, I get the following:

Nice catch! Fix on its way. It is likely that V.Empty will not occur un ClassRel context, but robustness commands fixing.

maranget:
Just pushed a small fix for the ClassRel arguments of primitives. In ambiguous cases, I favored event set and relation.

maranget:
Commit ba788a27e93d3a75f73273a5f21667f5d0f77f2c should fix the `add_unless_mb` problem.

maranget:
Unfortunately, the added tests would not have caught the `add_unless_mb` problem because **herd7** generates RMW effects for all the read-modify-write instructions of standard C (_i.e._ non-lkmm), while `atomic_add_unless` is implemented by emitting an atomic read and an atomic write.

akiyks:
Just confirmed this restores expected behavior.
Thanks!

maranget:
Hi @relokin and @HadrienRenaud. I am leaving next Monday for two weeks. Could we merge this PR that fixes two bugs soon?

HadrienRenaud (approved):
I can't claim to understand all the details nor consequences of those changes, but overall I don't see why we shouldn't merge it

maranget:
Thanks @HadrienRenaud. Merging.

PR #1913, opened 2026-07-10
[gen] Use explicit equality helpers for edge comparison

ShaleXIONG:
Add equality helpers for code, edge components, dependencies, and RMWs. This avoids polymorphic equality for `tedge` internals. Use existing `compare_atom` for `atom` equality check.

This pull request separates from #1835 .

### DIFF
--- a/gen/common/code.ml
+++ b/gen/common/code.ml
@@ -75,6 +75,25 @@ type sd = Same|Diff|UnspecLoc
 (* Direction of related events *)
 type extr = Dir of dir | Irr | NoDir
 
+let equal_ie ie1 ie2 = match ie1,ie2 with
+  | Int,Int
+  | Ext,Ext
+  | UnspecCom,UnspecCom -> true
+  | (Int|Ext|UnspecCom),_ -> false
+
+let equal_sd sd1 sd2 = match sd1,sd2 with
+  | Same,Same
+  | Diff,Diff
+  | UnspecLoc,UnspecLoc -> true
+  | (Same|Diff|UnspecLoc),_ -> false
+
+let equal_extr e1 e2 = match e1,e2 with
+  | Dir W,Dir W
+  | Dir R,Dir R
+  | Irr,Irr
+  | NoDir,NoDir -> true
+  | (Dir _|Irr|NoDir),_ -> false
+
 (* Associated pretty print & generators *)
 let pp_dir = function
   | W -> "W"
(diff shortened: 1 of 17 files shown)


gen/common/edge.ml
```diff
@@ -423,6 +425,35 @@ let fold_tedges f r =
 
   let get_access_atom = A.get_access_atom
 
+  let equal_atomo = Option.equal (fun a1 a2 -> A.compare_atom a1 a2 = 0)
```
psafont, line 428: This is a lot of code, which doesn't seem trivial to maintain. I think it might be worthwhile to use a ppx processor for the data types to generate this code: https://github.com/ocaml-ppx/ppx_deriving#plugins-eq-and-ord

ShaleXIONG: In short, I agree. However, I believe the general consensus is to not introduce extra dependency. I will be in favour of use `ppx` to deriving many `pp` and `equal` function in `diy*` but not sure how it will impact the compilation and dependencies?  @psafont  Do you have insight on this?

gen/common/code.ml
```diff
@@ -75,6 +75,25 @@ type sd = Same|Diff|UnspecLoc
 (* Direction of related events *)
 type extr = Dir of dir | Irr | NoDir
 
+let equal_ie ie1 ie2 = match ie1,ie2 with
+| Int,Int
+| Ext,Ext
+| UnspecCom,UnspecCom -> true
+| (Int|Ext|UnspecCom),_ -> false
```
fsestini: Could you please indent the pattern matching bars (here and elsewhere), for easier readability and consistency with the rest of the code?

```suggestion
  | Int,Int
  | Ext,Ext
  | UnspecCom,UnspecCom -> true
  | (Int|Ext|UnspecCom),_ -> false
```

ShaleXIONG: I systematically added two spaces for similar code.

gen/common/edge.ml
fsestini: In principle I agree that writing a lot of repetitive boilerplate code by hand can be a maintenance burden. And while I think LLMs have changed the landscape a bit in that respect by making the _writing_ process of such boilerplate more “automatic” and less error-prone, maintenance is also about reading code, and in some cases it does help to hide boring repetitive code behind macros, especially when the macro’s behaviour is straightforward and understood.
IMO the process of generating “canonical” implementations of `equal`, `compare`, and `pp` is indeed an example where macros/ppx would make sense, so I’d personally be in favour of us gradually transitioning to using ppx derivers for auto-generating those.

> not sure how it will impact the compilation and dependencies?

That's a good question. In terms of number of dependencies being transitively pulled in, it doesn't seem too bad from a quick look at [the opam page](https://opam.ocaml.org/packages/ppx_deriving/). Assessing the impact on compilation speed with require benchmarking.

fsestini (approved):
LGTM, thanks @ShaleXIONG

PR #1917, opened 2026-07-15
[herd] Fix atomic pairs matching algorithm

HadrienRenaud:
The situation, reported by @akiyks, arrises when there is an atomic read that should stay unmatched, for example in the case of a failed compare-and-swap. However, if there is a second atomic instruction at the same location in the same thread, then the algorithm would try and match the read of the first atomic instruction with the second write, which hit an assertion error.
This commit transforms this assertion error in a condition where the pairs between atomic elements on different instructions are not matched.

---

On the test `C-atomic-04` given by @akiyks in [the bug report](https://github.com/herd/herdtools7/pull/1908#issuecomment-4977821169), results are identical now with on the following command line:
```bash
herd7 C-atomic-04.litmus -conf catalogue/linux/cfgs/linux-kernel.cfg -I catalogue/linux/cats -I herd-www/linux/
```

### DIFF
--- a/herd/tests/instructions/C/C17.litmus
+++ b/herd/tests/instructions/C/C17.litmus
@@ -1,0 +1,15 @@
+C C17
+
+{
+  atomic_t x = ATOMIC_INIT(0);
+}
+
+P0 (atomic_int* x) {
+  int r0;
+  r0 = __atomic_add_unless{mb}(x, 10, 0);
+
+  int r1;
+  r1 = __atomic_add_unless{mb}(x, 10, 10);
+}
+
+forall (0: r0 = 0 /\ 0: r1 = 1 /\ x = 10)
--- a/herd/mem.ml
+++ b/herd/mem.ml
@@ -1978,11 +1978,8 @@ let add_eq v1 v2 eqs =
                 List.fold_left
                   (fun k er ->
                     match EvtSet.find_first_after er ws with
-                    | Some ew ->
-                       assert (E.same_instance er ew) ;
-                       (er,ew)::k
-                    | None -> (* Cann occur for failing CAS *)
-                       k)
+                    | Some ew when E.same_instance er ew -> (er, ew) :: k
+                    | _ -> k (* Can occur for failing CAS *))
                   k rs)
               atms k)
         [] atms |> E.EventRel.of_list
--- a/herd/tests/instructions/C/C17.litmus.expected
+++ b/herd/tests/instructions/C/C17.litmus.expected
@@ -1,0 +1,10 @@
+Test C17 Required
+States 1
+0:r0=0x0; 0:r1=0x1; [x]=0xa;
+Ok
+Witnesses
+Positive: 1 Negative: 0
+Condition forall (0:r0=0x0 /\ 0:r1=0x1 /\ [x]=0xa)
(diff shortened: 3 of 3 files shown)


fsestini:
Could we add a regression test case to guard against future occurrences of this?

akiyks:
Just tested and confirmed this restores previous behavior.
Thanks.

HadrienRenaud:
> Could we add a regression test case to guard against future occurrences of this?

Done! Note that this cannot happen in AArch64 because in a failed CAS, the read memory effect does not have the atomic annotation, and so it is not collected for the atomic pair matching algorithm

herd/tests/instructions/C/C17.litmus
```diff
@@ -0,0 +1,8 @@
+C C17
+
+{}
```
relokin: Do we need to be explicit with the fact that this is an atomic_t?
```suggestion
{
  atomic_t x = ATOMIC_INIT(0);
}
```

herd/tests/instructions/C/C17.litmus
```diff
@@ -0,0 +1,8 @@
...
+
+P0 (atomic_int* x) {
+  __atomic_add_unless{mb}(x, 10, 0);
+  __atomic_add_unless{mb}(x, 10, 10);
```
relokin: The point of this test is to illustrate a failing __atomic_add_unless isn't it? Should we only have:

```suggestion
  int r0;
  r0 = __atomic_add_unless{mb}(x, 10, 0);
```

herd/tests/instructions/C/C17.litmus
```diff
@@ -0,0 +1,8 @@
...
+}
```
relokin, line 13:
```suggestion
}
forall (0:r0=0 /\ x=0)
```

relokin (commented):
Thanks for this Hadrien, this change looks good to me, I have some minor comments regarding the test. Should we make it more specific to this particular scenario?

herd/tests/instructions/C/C17.litmus
HadrienRenaud: We do not do this with the other C regression tests. That being said, I'm not against it. Note that it is already mentioned in the argument of thread P0:

```C
P0 (atomic_int* x) {
```

herd/tests/instructions/C/C17.litmus
HadrienRenaud: We do need the second atomic_add_unless, otherwise it does not fail earlier: there is no write to pair it with.

relokin: I see. Just for my understanding, we handle a failed amo correctly, but if there is a failed amo followed by a successful amo then it was the assertion you removed that fired.

relokin (approved):
LGTM, thanks Hadrien!

herd/tests/instructions/C/C17.litmus
HadrienRenaud: Exactly!

herd/tests/instructions/C/C17.litmus
HadrienRenaud: I used it in the end.

PR #1918, opened 2026-07-15
[asl][doc] Minor fix of auto-generated content

hrutvik:


hrutvik:
@Roman-Manevich would it be worth adding a flag that would make ASL Reference builds fail if the auto-generated content hasn't been checked in (which I think is the case here)? We could use this flag in CI for example

Roman-Manevich:
> @Roman-Manevich would it be worth adding a flag that would make ASL Reference builds fail if the auto-generated content hasn't been checked in (which I think is the case here)? We could use this flag in CI for example

Perhaps instead of splicing in the console output, we should switch to a macro generator approach like `aslspec` and store the console content as macros under something like `generated_console_macros.tex`. It shouldn't be hard to get the python code to do that.

hrutvik:
Closed in favour of #1919

PR #1919, opened 2026-07-15
[asl reference] replace console splicing by generated console macros

Roman-Manevich:
Previously, console output was spliced directly into `.tex` files by manually running `doclint.py -cm`. Changes to referenced ASL tests could therefore leave the documented output stale. This change generates console output automatically as part of the document build.

Replace inline generated console output with `\RenderConsoleFor`:
* The makefile now regenerates console output automatically during PDF and HTML builds.
* Added HeVeA-compatible output files.
* Migrated existing console examples.
* Fixed bug with multiline `\ASLListing` spellchecking.
* Removed obsolete extended-macro code.
* Fixed HTML ASL Reference builds by using the correct dune build-tree path for `aslref`. Previous attempts to build aslref defaulted to a path that might not exist. `asllib/doc/Makefile` now builds `aslref` automatically.

### DIFF
--- a/asllib/doc/ASLmacros.tex
+++ b/asllib/doc/ASLmacros.tex
@@ -127,6 +127,37 @@
 \usepackage{mathpartir}  % For deduction rules and equations paragraphs
 \newcommand\hva[0]{} % Disable the macro, just for the current PR.
 \usepackage{fancyvrb}
+
+% HeVeA defines \ifhevea itself. Define the false branch for LaTeX so the same
+% rendering macro can select the appropriate console-output representation.
+\makeatletter
+\@ifundefined{ifhevea}{\newif\ifhevea\heveafalse}{}
+\makeatother
+% Render the output produced by running aslref on a test file:
+%   \RenderConsoleFor[<options>]{<ID>}{<test-file>}
+%
+% <ID> must be unique and consist of letters, digits, and hyphens, beginning
+% with a letter. <test-file> must use one of \definitiontests, \syntaxtests,
+% \typingtests, or \semanticstests. Both identify the generated output; TeX
+% itself uses only <ID>.
+%
+% <options> contains options passed to aslref. It may additionally contain the
+% generator directives `expect-error`, to capture stdout and stderr from an
+% intentionally failing test, and `show-command`, to include the command in the
+% rendered output. The complete invocation must occupy one source line so that
+% doclint.py can find it. The Makefile regenerates generated_console_macros.tex
+% and the HTML console-output files before building the document.
+\newcommand\RenderConsoleFor[3][]{%
+  % HeVeA does not implement fancyvrb's SaveVerbatim/UseVerbatim mechanism, so
+  % it reads the separately generated text file instead. LaTeX uses the named
+  % SaveVerbatim block from generated_console_macros.tex.
+  \ifhevea
+    \VerbatimInput[fontsize=\footnotesize,frame=single]{generated_console_output/#2.txt}%
+  \else
+    \UseVerbatim[fontsize=\footnotesize,frame=single]{console-#2}%
+  \fi
+}
+
 \usepackage[
   % Notes on the right, should be less than outer
   marginparwidth=100pt,
--- a/asllib/doc/ASLmacros.tex
+++ b/asllib/doc/ASLmacros.tex
@@ -128,11 +128,6 @@
 \newcommand\hva[0]{} % Disable the macro, just for the current PR.
 \usepackage{fancyvrb}
 
-% HeVeA defines \ifhevea itself. Define the false branch for LaTeX so the same
-% rendering macro can select the appropriate console-output representation.
-\makeatletter
-\@ifundefined{ifhevea}{\newif\ifhevea\heveafalse}{}
-\makeatother
 % Render the output produced by running aslref on a test file:
 %   \RenderConsoleFor[<options>]{<ID>}{<test-file>}
 %
@@ -141,21 +136,14 @@
 % \typingtests, or \semanticstests. Both identify the generated output; TeX
 % itself uses only <ID>.
 %
-% <options> contains options passed to aslref. It may additionally contain the
-% generator directives `expect-error`, to capture stdout and stderr from an
-% intentionally failing test, and `show-command`, to include the command in the
-% rendered output. The complete invocation must occupy one source line so that
-% doclint.py can find it. The Makefile regenerates generated_console_macros.tex
-% and the HTML console-output files before building the document.
+% <options> contains generator keys translated to aslref options by doclint.py.
+% Supported keys are `type-check-no-warn`; `expect-error`, to capture stdout and
+% stderr from an intentionally failing test; and `show-command`, to include the
+% command in the rendered output. The complete invocation must occupy one source
+% line so that doclint.py can find it. The Makefile regenerates the console-output
+% text files before building the document.
 \newcommand\RenderConsoleFor[3][]{%
-  % HeVeA does not implement fancyvrb's SaveVerbatim/UseVerbatim mechanism, so
-  % it reads the separately generated text file instead. LaTeX uses the named
-  % SaveVerbatim block from generated_console_macros.tex.
-  \ifhevea
-    \VerbatimInput[fontsize=\footnotesize,frame=single]{generated_console_output/#2.txt}%
-  \else
-    \UseVerbatim[fontsize=\footnotesize,frame=single]{console-#2}%
-  \fi
+  \VerbatimInput[fontsize=\footnotesize,frame=single]{generated_console_output/#2.txt}%
 }
 
 \usepackage[
--- a/asllib/doc/Makefile
+++ b/asllib/doc/Makefile
@@ -1,4 +1,4 @@
-ASLREF=../../_build/install/default/bin/aslref
+ASLREF=../../_build/default/asllib/aslref.exe
 LATEX=pdflatex
 BIBTEX=bibtex
 
@@ -46,6 +46,10 @@ ASLReference.pdf: generated_spec $(GENERATED_CONSOLE_MACROS_TEX) *.tex *.bib
 
 $(GENERATED_CONSOLE_MACROS_TEX): extended_macros.py doclint.py $(ASLREF) $(ASL_TESTS) $(filter-out $(GENERATED_CONSOLE_MACROS_TEX),$(wildcard *.tex))
 	python3 doclint.py --console-macros-only --aslref $(ASLREF)
+
+$(ASLREF):
+	@# Build the aslref target in the parent asllib directory.
+	dune build ../aslref.exe
 
 html: generated_spec $(GENERATED_CONSOLE_MACROS_TEX) *.tex *.bib *.hva
 	mkdir -p HTML
--- a/asllib/doc/Makefile
+++ b/asllib/doc/Makefile
@@ -3,6 +3,8 @@ LATEX=pdflatex
 BIBTEX=bibtex
 
 GENERATED_MACROS_TEX=generated_macros.tex
+GENERATED_CONSOLE_MACROS_TEX=generated_console_macros.tex
+ASL_TESTS=$(shell find ../tests -name '*.asl')
 
 # HeVeA and Hacha can emit warnings on stderr while still exiting with success.
 # Capture stderr from the command, preserve real command failures, and make
@@ -29,7 +31,7 @@ all: ASLReference.pdf
 
 LATEXMK=latexmk
 
-ASLReference.pdf: generated_spec *.tex *.bib
+ASLReference.pdf: generated_spec $(GENERATED_CONSOLE_MACROS_TEX) *.tex *.bib
 	if command -v $(LATEXMK) >/dev/null 2>&1; \
 	then \
 		$(LATEXMK) -pdf ASLReference.tex; \
@@ -40,9 +42,12 @@ ASLReference.pdf: generated_spec *.tex *.bib
 		$(BIBTEX) ASLReference; \
 		$(LATEX) ASLReference.tex; \
 	fi
-	python3 doclint.py --console_macros --aslref $(ASLREF)
+	python3 doclint.py
 
-html: generated_spec *.tex *.bib *.hva
+$(GENERATED_CONSOLE_MACROS_TEX): extended_macros.py doclint.py $(ASLREF) $(ASL_TESTS) $(filter-out $(GENERATED_CONSOLE_MACROS_TEX),$(wildcard *.tex))
+	python3 doclint.py --console-macros-only --aslref $(ASLREF)
+
+html: generated_spec $(GENERATED_CONSOLE_MACROS_TEX) *.tex *.bib *.hva
 	mkdir -p HTML
 	@# The first run is expected to generate warnings because bibhva
 	@# and the second hevea pass have not run, so we silence them via -s.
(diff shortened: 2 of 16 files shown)


asllib/doc/doclint.py
```diff
@@ -929,7 +934,7 @@ def spellcheck(reference_dictionary_path: str, latex_files: list[str]) -> int:
     ]
     extract_patterns = [
         # Patterns for extracting words from specific macros:
-        r"\\ASLListing\{(.*?)\}\{.*?\}\{.*?\}",
+        r"\\ASLListing\{(.*?)\}\s*\{.*?\}\s*\{.*?\}",
```
hrutvik, line 937: Please could this and related changes (e.g. `dictionary.txt`) be moved to a separate commit (or ideally a separate PR)?

asllib/doc/Makefile
```diff
@@ -1,8 +1,10 @@
-ASLREF=../../_build/install/default/bin/aslref
+ASLREF=../../_build/default/asllib/aslref.exe
```
hrutvik, line 1: Could something like `dune show location aslref` work here instead?

asllib/doc/Makefile
```diff
@@ -40,9 +42,16 @@ ASLReference.pdf: generated_spec *.tex *.bib
...
-	python3 doclint.py --console_macros --aslref $(ASLREF)
+	python3 doclint.py
 
-html: generated_spec *.tex *.bib *.hva
+$(GENERATED_CONSOLE_MACROS_TEX): extended_macros.py doclint.py $(ASLREF) $(ASL_TESTS) $(filter-out $(GENERATED_CONSOLE_MACROS_TEX),$(wildcard *.tex))
+	python3 doclint.py --console-macros-only --aslref $(ASLREF)
+
+$(ASLREF):
+	@# Build the aslref target in the parent asllib directory.
+	dune build ../aslref.exe
```
hrutvik: I'm wondering if there is a better way to achieve this

asllib/doc/ASLmacros.tex
```diff
@@ -127,6 +127,37 @@
...
+\makeatletter
+\@ifundefined{ifhevea}{\newif\ifhevea\heveafalse}{}
+\makeatother
+% Render the output produced by running aslref on a test file:
+%   \RenderConsoleFor[<options>]{<ID>}{<test-file>}
+%
+% <ID> must be unique and consist of letters, digits, and hyphens, beginning
+% with a letter. <test-file> must use one of \definitiontests, \syntaxtests,
+% \typingtests, or \semanticstests. Both identify the generated output; TeX
+% itself uses only <ID>.
```
hrutvik, line 137: What does the ID do, and is it useful? I wonder if we should just remove them, one less name to write

asllib/doc/ASLmacros.tex
```diff
@@ -127,6 +127,37 @@
...
+% intentionally failing test, and `show-command`, to include the command in the
+% rendered output. The complete invocation must occupy one source line so that
+% doclint.py can find it. The Makefile regenerates generated_console_macros.tex
+% and the HTML console-output files before building the document.
+\newcommand\RenderConsoleFor[3][]{%
+  % HeVeA does not implement fancyvrb's SaveVerbatim/UseVerbatim mechanism, so
+  % it reads the separately generated text file instead. LaTeX uses the named
+  % SaveVerbatim block from generated_console_macros.tex.
+  \ifhevea
+    \VerbatimInput[fontsize=\footnotesize,frame=single]{generated_console_output/#2.txt}%
```
hrutvik: Why not use this approach for both LaTeX and HeVeA?

asllib/doc/extended_macros.py
```diff
@@ -54,238 +52,149 @@ def execute_and_capture_output(command: str, error_expected: bool) -> str:
...
-    if num_changes > 0:
-        print(f"Performed {num_changes} line transformations")
+def write_if_changed(filename: str, content: str):
+    try:
+        with open(filename, encoding="utf-8") as file:
+            if file.read() == content:
+                return
+    except FileNotFoundError:
+        pass
+    temporary_filename = filename + ".tmp"
```
hrutvik: Do we need a temporary file here?

asllib/doc/doclint.py
Roman-Manevich: moved to a new commit

asllib/doc/Makefile
Roman-Manevich: `dune show location aslref` points to public install path (`./../_build/install/default/bin/aslref`), but we need the path created by the makefile rule for `$(ASLREF)`. I don't know how to make it point to the right path.

asllib/doc/Makefile
Roman-Manevich: I don't know of one, sorry. Then again, I'm not a dune expert.

asllib/doc/ASLmacros.tex
Roman-Manevich: I uniquely namesone `\RenderConsoleFor` instance in order to match the correct generated macro. Generating it from the test file and command line options is tricky - both latex/python/hevea have to agree on the same key.

asllib/doc/extended_macros.py
Roman-Manevich: It makes updates atomic, preventing LaTeX from reading a partially written output file. I'll add a comment explaining this.

hrutvik: Why would a partial file be written and then read again? This Python script runs to completion ahead of LaTeX right?

Roman-Manevich: The script indeed runns ahead of LaTeX on the CI. This is defensive, but nice to have. (Locally I sometimes run LaTeX concurrently with make.)

hrutvik: Does this not run the risk of seeing the old file, before `replace` has been called?

asllib/doc/Makefile
```diff
@@ -1,8 +1,11 @@
-ASLREF=../../_build/install/default/bin/aslref
+ASLREF=../../_build/default/asllib/aslref.exe
 LATEX=pdflatex
 BIBTEX=bibtex
 
 GENERATED_MACROS_TEX=generated_macros.tex
+# Empty file marking successful generation of every file in generated_console_output/.
```
hrutvik, line 6: I'm a bit surprised by this - is a `.PHONY` target in the `Makefile` not sufficient to model this?

asllib/doc/extended_macros.py
```diff
@@ -54,238 +51,125 @@ def execute_and_capture_output(command: str, error_expected: bool) -> str:
...
-                    first_changed_line_number = len(output_lines) + 1
-                    last_changed_line_number = first_changed_line_number + len(
-                        block_output_lines
+GENERATED_CONSOLE_OUTPUT_STAMP = "generated_console_output.stamp"
+GENERATED_CONSOLE_OUTPUT_DIR = "generated_console_output"
+CONSOLE_INVOCATION_RE = re.compile(
+    r"^\s*\\RenderConsoleFor(?:\[([^]]*)\])?"
+    r"\{([A-Za-z][A-Za-z0-9-]*)\}\{([^{}]+)\}\s*$"
+)
+TEST_PATH_REPLACEMENTS = {
```
hrutvik: Seems to be used only once in `resolve_test_path` - why not define it there instead? Similarly with other ones

asllib/doc/Makefile
Roman-Manevich: The command generates one file per `\RenderConsoleFor` macro invocation. There are currently 27 of those files (`HeVeA` needs those files seprately). A `.PHONY` target would regenerate them and rebuild the document every time. The stamp gives you incremental builds.

asllib/doc/extended_macros.py
Roman-Manevich: done

asllib/doc/Makefile
```diff
@@ -40,9 +43,17 @@ ASLReference.pdf: generated_spec *.tex *.bib
...
+	python3 doclint.py
 
-html: generated_spec *.tex *.bib *.hva
+$(GENERATED_CONSOLE_OUTPUT_STAMP): extended_macros.py doclint.py $(ASLREF) $(ASL_TESTS) $(wildcard *.tex)
+	python3 doclint.py --console-macros-only --aslref $(ASLREF)
+	touch $@
+
+$(ASLREF):
+	@# Build the aslref target in the parent asllib directory.
+	dune build ../aslref.exe
```
HadrienRenaud: Force recompilation of `$(ASLREF)`, otherwise this will never get updated by make:
```Makefile
.PHONY: FORCE_REMAKE
$(ASLREF): FORCE_REMAKE
  dune build $@
```

This shouldn't trigger useless call to remake `$(GENERATED_CONSOLE_OUTPUT_STAMP)` because it looks like `dune` doesn't rewrite targets that don't change (or this is what `stat` tells me on my machine)

asllib/doc/extended_macros.py
```diff
@@ -54,272 +49,132 @@ def execute_and_capture_output(command: str, error_expected: bool) -> str:
...
+                    raise ValueError(
+                        f"{source}:{line_number}: malformed RenderConsoleFor invocation; "
+                        "the invocation must occupy one line"
                     )
-                    print(
-                        f"{source} lines {first_changed_line_number}-{last_changed_line_number} are new."
+                option_text, identifier, test_path = match.groups()
+                option_tokens = shlex.split(option_text or "")
+                expect_error = "expect-error" in option_tokens
+                show_command = "show-command" in option_tokens
```
HadrienRenaud: Why putting the command line options directly in the macro? why not translate it? for example by using similar keys as the metadata for the test suite?

This way when `aslref` cli change you only need to update this file and not all of the `.tex`?
Also you've already changed the cli syntax by adding the `--expect-error` and `--show-command` options

asllib/doc/Makefile
```diff
@@ -1,8 +1,11 @@
-ASLREF=../../_build/install/default/bin/aslref
+ASLREF=../../_build/default/asllib/aslref.exe
 LATEX=pdflatex
 BIBTEX=bibtex
 
 GENERATED_MACROS_TEX=generated_macros.tex
+# Empty file marking successful generation of every file in generated_console_output/.
+GENERATED_CONSOLE_OUTPUT_STAMP=generated_console_output.stamp
+ASL_TESTS=$(shell find ../tests -name '*.asl')
```
HadrienRenaud: If you only have 4 parent directories you could just enumerate them here and use Make's `wildcard`

HadrienRenaud (approved):
This looks ok to me.

Can we document what was the problem with the previous build system in the PR please?

I wanted to try and put as much things into `make`, but it doesn't seem possible because we don't know the command line arguments. Makes you want to add some metadata to your ASL test files...

I did not check the `.tex` files.

asllib/doc/Makefile
hrutvik: How about `dune show location ../aslref.exe`?

asllib/doc/Makefile
hrutvik: Was this comment supposed to be on the other thread?

asllib/doc/Makefile
Roman-Manevich: That returns `_build/default/asllib/aslref.exe`

asllib/doc/extended_macros.py
Roman-Manevich: Okay. I removed the temporary file and write directly.

asllib/doc/Makefile
Roman-Manevich: sure

asllib/doc/extended_macros.py
Roman-Manevich: thanks, done.

asllib/doc/Makefile
hrutvik: I'm not sure I understand - that can be prefixed by `../../` to get the right result can't it?

Roman-Manevich: `ASLREF=../../$(shell dune show location --no-print-directory ../aslref.exe 2>&1)` seems to work.

asllib/doc/Makefile
Roman-Manevich: switched to a `.PHONY` target.

asllib/doc/Makefile
```diff
@@ -1,8 +1,14 @@
...
+ASLREF=../../$(shell dune show location --no-print-directory ../aslref.exe 2>&1)
 LATEX=pdflatex
 BIBTEX=bibtex
 
 GENERATED_MACROS_TEX=generated_macros.tex
+ASL_TESTS=$(wildcard \
+	../compliance/tests/ASLDefinition.t/*.asl \
+	../compliance/tests/ASLSyntaxReference.t/*.asl \
+	../compliance/tests/ASLTypingReference.t/*.asl \
+	../compliance/tests/ASLSemanticsReference.t/*.asl)
```
hrutvik: Are these still accurate?

asllib/doc/extended_macros.py
```diff
@@ -54,272 +49,144 @@ def execute_and_capture_output(command: str, error_expected: bool) -> str:
...
-        )
-        # `command`` is potentially included in the LaTeX output (if `CONSOLE_CMD``
-        # appears in the macro).
-        # Since we don't want to include any user-specific paths `command` is used
-        # for inclusion whereas the command with the actual path to aslref is
-        # used in `executable_command`.
-        executable_command = command.replace("aslref", ASLREF_EXE)
+                expect_error = "expect-error" in option_keys
+                show_command = "show-command" in option_keys
+                aslref_options = tuple(
```
hrutvik: nit: could we not pass in ASLRef options directly in the macro rather than having a translation?

asllib/doc/extended_macros.py
```diff
@@ -54,272 +49,144 @@ def execute_and_capture_output(command: str, error_expected: bool) -> str:
...
+        r"\semanticstests": "../compliance/tests/semantics",
+    }
+    for latex_macro, directory in test_path_replacements.items():
+        if test_path.startswith(latex_macro):
+            return directory + test_path[len(latex_macro) :]
+    raise ValueError(f"Unsupported test path {test_path!r}")
+
+
+def generate_console_outputs(invocations: list[ConsoleInvocation], aslref_exe: str):
+    output_dir = "generated_console_output"
```
hrutvik: Do we need to add the `generated_console_output` directory to `.gitignore`?

asllib/doc/extended_macros.py
```diff
@@ -54,272 +49,144 @@ def execute_and_capture_output(command: str, error_expected: bool) -> str:
...
+def write_if_changed(filename: str, content: str):
+    try:
+        with open(filename, encoding="utf-8") as file:
+            if file.read() == content:
+                return
+    except FileNotFoundError:
+        pass
+    with open(filename, "w", encoding="utf-8") as file:
+        file.write(content)
+    print(f"Generated {filename}")
```
hrutvik: This seems a bit odd, and perhaps more complicated than it needs to be - could we not instead:
- delete any files in the output dir
- regenerate all of them (this should be very quick still I think?)

?

asllib/doc/Makefile
Roman-Manevich: Replaced by `ASL_TESTS=$(wildcard ../compliance/tests/*/*.asl)`

asllib/doc/extended_macros.py
Roman-Manevich: It's already there.

asllib/doc/extended_macros.py
Roman-Manevich: I'm not entirely sure I understand. The translation is needed for options that don't correspond to ASLRef options (expect-error and show-command). ASLRef options are rejected. Are you suggesting letting the macro list ASLRef options? It's possible, but not needed as of now.

asllib/doc/extended_macros.py
Roman-Manevich: I was worried it could needlessly trigger LaTeX builds, but that doesn't seem to be the case so I'm simplifying as you suggested.

PR #1920, opened 2026-07-17
AARCH-25948: Replace Hardware Update Effects with Implicit TTD Memory Write Effects

fsestini:


herd/libdir/aarch64deps.cat
```diff
@@ -67,7 +67,7 @@ let rec dtrm =
 (** Data, Address and Control dependencies *)
 
 let data = [Exp & R]; dtrm & po; [Rreg]; iico_data & ii_data; [Exp & W]
-let addr = [Exp & R]; dtrm & po; [Rreg]; iico_data & ii_addr; [Exp & M | Imp & Tag & R | Imp & TTD & R | HU | TLBI | DC.CVAU | IC.IVAU]
+let addr = [Exp & R]; dtrm & po; [Rreg]; iico_data & ii_addr; [Exp & M | Imp & Tag & R | Imp & TTD & M | TLBI | DC.CVAU | IC.IVAU]
```
TiberiuBucur, line 70: I know `TTD & M` has been used elsewhere, but cat2table currently assumes `TTD` is a subset of `M`, which is also confirmed by the `aarch64assumptions.cat` file, ran across the test suite. I'm wondering if we can leave this occurrence as `Imp & TTD`. Perhaps `Imp & TTD & M` might still be better in order to keep `miaou7` output consistent with Arm ARM proses. What do you think?

fsestini: > Perhaps `Imp & TTD & M` might still be better in order to keep `miaou7` output consistent with Arm ARM proses.

Yes.

TiberiuBucur:
LGTM, but I'm going to hold off on this until tickets are made official and communicated.

PR #1923, opened 2026-07-21
[INSTALL.md] Provide more instructions on how to build and install herdtools

psafont:
The instructions on creating the switch and activating it were missing. There's no need for users to explicitly install the dependency packages (and maintain that list) when opam can determine the list using existing metadata.

The explanation on changing PREFIX has been moved to the Install section and now it contains a full command on doing an installation while changing the variable.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -13,10 +13,6 @@ Then, to get the newest version:
 Source build
 ============
 
-Tools will be installed in PREFIX/bin, and various files in PREFIX/share/herdtools7.
-By default (see Makefile) PREFIX is $HOME.
-You can change PREFIX by editing the Makefile, or by running ``make ...`` as ``make PREFIX=yourprefix ...``.
-
 Requirements
 ------------
 
@@ -27,11 +23,12 @@ Requirements
 - logs
 
 We strongly recommend to have this base software installed through the opam
-package manager.
+package manager. This means an opam switch needs to be prepared before
+installing the dependencies of the project. For example:
 
-    % opam install dune menhir zarith logs
-
-Make sure to run `eval $(opam config env)` to make tools available in your PATH.
+    % opam switch create herdtools7 --empty
+    % eval $(opam config env --switch=herdtool7 --set-switch)
+    % opam install . --deps-only
 
 Notice: Compilation with ocamlbuild is not longer possible
 
@@ -43,7 +40,7 @@ Build
 Testing
 -------
 
-The optionnal dependency `qcheck` can be installed with `opam` as follows:
+The optional dependency `qcheck` can be installed with `opam` as follows:
 
     % opam install qcheck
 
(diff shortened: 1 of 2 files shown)


INSTALL.md
```diff
@@ -27,11 +23,12 @@ Requirements
...
 
 We strongly recommend to have this base software installed through the opam
-package manager.
-
-    % opam install dune menhir zarith logs
+package manager. This means an opam switch needs to be prepared to install the
+software:
 
-Make sure to run `eval $(opam config env)` to make tools available in your PATH.
+    % opam switch create herdtools7 ocaml.4.14.4
```
relokin: Why are we advising users to install OCaml 4.14?

INSTALL.md
```diff
@@ -27,11 +23,12 @@ Requirements
 - logs
 
 We strongly recommend to have this base software installed through the opam
-package manager.
-
-    % opam install dune menhir zarith logs
+package manager. This means an opam switch needs to be prepared to install the
```
relokin: Should we turn this into a recommendation? Creating a switch could be recommended but strictly speaking is not required.

relokin (commented):
Thanks for this Pau. I’ve got two comments.

INSTALL.md
psafont: No reason in particular, other than it's one of the supported versions

INSTALL.md
psafont: > Creating a switch could be recommended but strictly speaking is not required.

I can introduce the sequence as an example of steps to do when installing dependencies with opam. It's good for newcomers to enumerate the steps. from scratch, this also allows experts users to sidestep the instructions and apply the ones that make sense for them.

README-tests.md
```diff
@@ -16,6 +16,8 @@
 
 The basic litmus test consists in compiling series of tests. Some series are also executed but not all of them. Tests that are not run by default would take too much time or raise errors.
 
+The litmus tests require the package `qemu-system` in order to run the tests that use kvm mode.
```
psafont, line 19: Found this information was missing as well, even if the information can be gathered from the github workfows

INSTALL.md
```diff
@@ -27,11 +23,12 @@ Requirements
...
 
 We strongly recommend to have this base software installed through the opam
-package manager.
-
-    % opam install dune menhir zarith logs
+package manager. This means an opam switch needs to be prepared before
+installing the dependecies of the project. For example:
 
-Make sure to run `eval $(opam config env)` to make tools available in your PATH.
+    % opam switch create herdtools7 ocaml.5.5.0
```
relokin: I would omit the ocaml version and let it just use whatever is the default otherwise we would need to update this document periodically.

Also, I tried this locally and opam install herdtools7 as well. If that happens and the path of opam is before $PREFIX in PATH then the user will end up using the opam build binary. Or am I doing something wrong?

psafont: Using `opam install herdtools7` will normally install the version from the official repository.

The local version of herdtools7 can be installed by running either `opam install .`, `opam install ./herdtools7.opam`, or running `opam pin add herdtool7 . --no-action` followed by `opam install herdtools7`.

In any case opam will fail to install the package, because the dune library aslref is not available. We need to fix it somehow, since opam does not not about the version of aslref present in the repository.

This is the reason why I didn't add an `opam install` and left the make install as the only option for installing the tools

INSTALL.md
psafont: The new opam switch creation command does not depend on a compiler version, and it lets opam pick a compiler version later on

fsestini (approved):
LGTM, thanks @psafont

PR #1926, opened 2026-07-22
[litmus] add new tests that record the programs' output

psafont:
These tests capture the output of the resulting programs to detect
regressions and are meant to run in native architecture that the litmus
tests test. They are also run along the rest of unit tests driven by
dune.

Currently a few tests have been added to demonstrate how these can be
developed into a full suite.

The pre-commit hooks had to be changed to avoid modifying the output
present in the cram test, in both the endline and whitespace checks

Further enhancements for the tests include, adding more parameters to
test, more target architectures.

There's also the addition of cross-architecture tests, using cross-compilers and qemu-user package. These allow running litmus tests on arbitrary architectures, but the dependency setup is more complex. This is why they are not run by default. I've prepared an opam metapackage to submit upstream if we decide to proceed with the approach: https://github.com/ocaml/opam-repository/compare/master...psafont:opam-repository:dev/pau/qemu

The first two commits are code changes I made while inspecting the codebase, and may be dropped if they are too disruptive, I thought there were worthwhile since they show a style I like: they remove code by increasing shared code, and adds unit-tests for the newly shared code to ensure it behaves as expected and won't regress.

### DIFF
--- a/internal/herd_regression_test.ml
+++ b/internal/herd_regression_test.ml
@@ -16,9 +16,6 @@
 
 (** A tool that runs regression tests of herd7, against .expected files. *)
 
-module Fun = Base.Fun
-
-
 (* Flags. *)
 
 type path = string
--- a/internal/lib/base.mli
+++ b/internal/lib/base.mli
@@ -18,21 +18,6 @@
  *  earlier versions of OCaml, or to add extra functionality. *)
 
 module Fun : sig
-  (** [Finally_raised e] is raised by [protect ~finally f] if [~finally] raises
-   *  an exception [e], to disambiguate it from exceptions raised by [f].
-   *  If [Finally_raised] is raised, it is either an unexpected exception (e.g.
-   *  [Out_of_memory]), or programmer error. *)
-  exception Finally_raised of exn
-
-  (** [negate f] negates the predicate function [f]. *)
-  val negate : ('a -> bool) -> ('a -> bool)
-
-  (** [protect ~finally f] calls [f], then calls [~finally]. If [f] raises an
-   *  exception [e], it calls [~finally] before re-raising [e]. If [~finally]
-   *  raises an exception [e], [e] is re-raised as [Finally_raised e].
-   *  It is equivalent to [Fun.protect] from OCaml >= 4.08. *)
-  val protect : finally:(unit -> unit) -> (unit -> 'a) -> 'a
-
   (** [open_out_protect f name] applies f to a channel
    *  to file whose name is [name]. Close the file under
       all circumstances. *)
--- a/internal/lib/filesystem.ml
+++ b/internal/lib/filesystem.ml
@@ -47,7 +47,17 @@ let rec remove_recursive path =
       Sys.remove path
   end
 
+(* Seed the PRNG when loading the module. *)
+let () = Random.self_init ()
+
+let temp_name root =
+  let name = Random.bits64 () in
+  Printf.sprintf "%s/%Ld" root name
+
 let new_temp_dir () =
-  let path = ref "" in
-  Command.run ~stdout:(fun c -> path := input_line c) "mktemp" ["-d"] ;
-  !path
+  let rec try_name counter =
+    let name = temp_name (Filename.get_temp_dir_name ()) in
+    try Sys.mkdir name 0o700 ; name
+    with Sys_error _ as e ->
+      if counter >= 10 then raise e else try_name (counter + 1)
+  in try_name 0
--- a/litmus/tests/cross/AArch64/A04.t
+++ b/litmus/tests/cross/AArch64/A04.t
@@ -1,0 +1,27 @@
+Use litmus7 to generate code from a litmus test
+
+  $ TEST="A04"
+  $ RUNNER='qemu-aarch64'
+  $ GCC='aarch64-linux-gnu-gcc'
+  $ mkdir "$TEST"
+  $ litmus7 -set-libdir ../../../libdir -gcc="$GCC" -o "$TEST" \
+  > "../../../../herd/tests/instructions/AArch64/$TEST.litmus" \
+  > -mode std -a 4 -s 1k -r 100 \
+  > -mach aarch64 -ccopts='-march=armv8-a+lse'
+
+Compile using a cross-compiler and run the litmus test using qemu user mode, natively, avoid printing the timing, it's not
+stable
+
+  $ cd "$TEST"
+  $ make > /dev/null
+  $ $RUNNER -L /usr/aarch64-linux-gnu -- "./$TEST.exe" | sed '$d'
+  Test A4 Allowed
+  Histogram (1 states)
+  4000000*>0:X0=0;
+  Ok
+  
+  Witnesses
+  Positive: 4000000, Negative: 0
+  Condition exists (0:X0=0) is validated
+  Hash=d64e911ddfaa8df8f0c4d17557b7562d
+  Observation A4 Always 4000000 0
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,7 +24,10 @@ repos:
             )|
 
             # Expected files should be left verbatim.
-            \.expected(-failure)?$
+            \.expected(-failure)?$|
+
+            # cram tests should be left verbatim
+            \.t$
           )
 
     -   id: trailing-whitespace
@@ -45,7 +48,7 @@ repos:
             \.expected(-failure)?$|
 
             # Cram tests may include trailing whitespace.
-            (^|/)run\.t$
+            \.t$
           )
 - repo: https://github.com/arenadotio/pre-commit-ocamlformat
   rev: 0439858f79b3bcb49e757278eb1312e212d7dd4f
--- a/asllib/aslspec/dune-project
+++ b/asllib/aslspec/dune-project
@@ -1,4 +1,4 @@
-(lang dune 3.11)
+(lang dune 3.12)
 (name aslspec)
 (using menhir 2.0)
 (cram enable)
--- a/asllib/dune-project
+++ b/asllib/dune-project
@@ -1,4 +1,4 @@
-(lang dune 3.11)
+(lang dune 3.12)
 (using menhir 2.0)
 (cram enable)
 
--- a/dune-project
+++ b/dune-project
@@ -1,4 +1,4 @@
-(lang dune 3.11)
+(lang dune 3.12)
 (name herdtools7)
 (using menhir 2.0)
 (cram enable)
--- a/herd/dune
+++ b/herd/dune
@@ -1,6 +1,8 @@
-(dirs :standard \ tests/instructions libdir)
-
 (data_only_dirs libdir)
+
+(subdir
+ tests
+ (data_only_dirs instructions))
 
 (rule
  (copy ../Version.ml Version.ml))
--- a/herdtools7.opam
+++ b/herdtools7.opam
@@ -22,6 +22,7 @@ depends: [
   "conf-which"
   "logs"
   "conf-python-3" {= "9.0.0" & with-test}
+  "conf-gcc" {with-test}
 ]
 conflicts: ["ocaml-option-bytecode-only"]
 url {
--- a/internal/herd_catalogue_regression_test.ml
+++ b/internal/herd_catalogue_regression_test.ml
@@ -14,7 +14,6 @@
 (* "http://www.cecill.info". We also give a copy in LICENSE.txt.            *)
 (****************************************************************************)
 
-module Fun = Base.Fun
 module Option = Base.Option
 
 exception Error of string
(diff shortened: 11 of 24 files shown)


psafont:
- [x] I need to control better the output of the programs:

```diff
--- a/_build/.sandbox/1e71485e728b3185aca35345f52c5b7e/default/litmus/tests/X86_64/A009.t
+++ b/_build/.sandbox/1e71485e728b3185aca35345f52c5b7e/default/litmus/tests/X86_64/A009.t.corrected
@@ -12,12 +12,12 @@ stable
   $ "./$TEST.exe" | sed '$d'
   Test A009 Required
   Histogram (1 states)
-  12000000:>0:rcx=-1;
+  4000000:>0:rcx=-1;
   Ok
   
   Witnesses
-  Positive: 12000000, Negative: 0
+  Positive: 4000000, Negative: 0
   Condition forall (0:rcx=-1) is validated
   Hash=7ca3c35015d75a877ccf509d75062e79
-  Observation A009 Always 12000000 0
+  Observation A009 Always 4000000 0
```

- [x] The lower bounds for alcotest also needs to be set correclty, it can now pick an incompatible version

internal/lib/filesystem_test.ml
```diff
@@ -0,0 +1,35 @@
...
+   with Sys_error _ as e ->
+  if counter >= 10 then raise e else try_name (counter + 1)
+ in try_name 0
+
+let with_dir f =
+  let root = temp_dir () in
+  let destroy () = Filesystem.remove_recursive root in
+  Fun.protect ~finally:destroy (fun () -> f root)
+
+let ( let@ ) f x = f x
```
fsestini: Nice trick, I've used it quite a lot as well, in the past.

internal/lib/base.mli
```diff
@@ -40,6 +40,20 @@ module Fun : sig
...
+module Seq : sig
+  val split_at : int -> 'a Seq.t -> ('a Seq.t * 'a Seq.t)
+  (** [split_at k seq] returns a pair of sequences, the first one containing
+      the first [k] elements of [seq], and the second one the rest. When [k]
+      less than 0, it's processed as it if were 0. When [seq] has less elements
+      than [k], the first element contains all the elements and the second one
+      is empty. Calling this function evaluates the first [k] elements of
+      [seq]. *)
+
+  val for_every_element : ('a -> bool) -> 'a Seq.t -> bool
```
fsestini: Why not `Stdlib.Seq.for_all`?

internal/herd_regression_test.ml
```diff
@@ -105,69 +69,64 @@ let show_tests_seq flags =
...
     TestHerd.herd_args
       ~bell:None ~cat:None
       ~conf:flags.conf
       ~variants:flags.variants
       ~libdir:flags.libdir
       ~timeout:None ~checkfilter:None ~speedcheck:None in
+  let mapply = Filename.concat (Filename.dirname herd) "mapply7" in
+  mapply, args
+
+let parallelize ?j ~s ~p args = match j with None -> s args | Some j -> p j args
```
fsestini: Nit: could we use slightly more informative names for `~s` and `~p`? E.g. `~para` for `~p`.

internal/lib/dune
```diff
@@ -1,7 +1,13 @@
 (library
  (name internal_lib)
+ (modules :standard \ base_test filesystem_test)
  (libraries herdtools str unix)
  (modes native)
  (wrapped false))
 
 (ocamllex lexState)
+
+(tests
```
fsestini: Note that I think this means `dune build @default` becomes dependent on `alcotest` being installed, which is not what we advertise in `INSTALL.md` and probably also not desirable. Could we make this test stanza optional based on whether `alcotest` is installed?

fsestini (commented):
I'm still looking through the PR, but here's some early comments.

internal/lib/base.mli
psafont: That's because `for_all` doesn't guarantee that the `p` is evaluated for every element, i.e it stops evaluating elements as soon as `p` does not hold.

internal/lib/dune
psafont: The commit changes both INSTALL.md and the opam file.

I can make the tests optional on the presence of alcotest, but I think that removes quite a bit of value from having tests in the first place.

fsestini: My point was that in `INSTALL.md` we don't require users to install test dependencies in order to run `make all`. However, I was wrong about `alcotest` being needed for `make all`. I though the chain of calls would be `make all -> dune build -> dune build @default`, and indeed `dune build @default` does require `alcotest` now.
But actually, `dune build` only resolves to `@@default`, not `@default`, so no issues there...

psafont: Ah, I see. The names of those rules are not confusing at all...

internal/lib/base.mli
fsestini: Got it, makes sense. Maybe just a personal preference, but I'd probably call it something like `fold_with_predicate` or similar. I feel like seeing the term "fold" would make it unambiguously clear in my head that this function is fully traversing the structure. Not a big deal though.

internal/lib/base.mli
```diff
@@ -40,6 +40,20 @@ module Fun : sig
...
+  (** [for_all p seq] returns whether all elements in the finite sequence [seq]
```
fsestini:
```suggestion
  (** [for_every_element p seq] returns whether all elements in the finite sequence [seq]
```

internal/lib/filesystem.ml
```diff
@@ -47,6 +47,19 @@ let rec remove_recursive path =
       Sys.remove path
   end
 
+let list_dir dir () =
```
fsestini: Wouldn't this leak the open `handler` if the sequence is only partially consumed?

INSTALL.md
```diff
@@ -40,6 +40,10 @@ Build
 Testing
 -------
 
+For testing, additional dependencies are needed, this can be installed using opam:
+
+    % opam install --deps-only -with-test .
```
fsestini:
```suggestion
    % opam install --deps-only --with-test .
```

internal/lib/filesystem.ml
psafont: Yes, it would, good catch. Need to think more about it on how to force closure of the handle.

fsestini: Perhaps I would use an iterator-like approach (as in `type 'a iter = ('a -> unit) -> unit` instead of `Seq.t`. With `Seq.t` the traversal is controlled by the caller, which is not ideal in this case.

```ocaml
let list_dir dir : dir iter = fun k ->
  let handle = ... in
  Fun.protect ~finally:(fun () -> (* release handle *)) (fun () ->
    try
       (* for all entries, call k entry *)
     with End_of_file -> ()
    )
```

psafont: This ends up with a bit more code on the caller, but it stills ends up being a net gain, IMO

internal/lib/base.mli
```diff
@@ -48,6 +37,10 @@ module List : sig
    *  [to_ocaml_string String.to_ocaml_string ["a"; "b"]] returns
    *  "[\"a\"; \"b\"]". *)
   val to_ocaml_string : ('a -> string) -> 'a list -> string
+
+  val for_every_element : ('a -> bool) -> 'a list -> bool
+  (** [for_every_element p lst] returns whether all elements in [lst] hold the
+      property [p]. Every element in the list is evaluated. *)
```
fsestini: Maybe we should reword this now that the container is a list rather than a `Seq.t`:
```suggestion
      property [p]. [p x] is evaluated for every element [x] in the list. *)
```

internal/herd_assumptions_test.ml
```diff
@@ -14,6 +14,8 @@
 (* "http://www.cecill.info". We also give a copy in LICENSE.txt.            *)
 (****************************************************************************)
 
+open! Base.Fun.Syntax
```
fsestini: Can we move this to be a `let open Base.Fun.Syntax` local to `get_each_litmus_in_dir`, since it's only used there anyway?

internal/lib/filesystem.ml
```diff
@@ -47,7 +47,26 @@ let rec remove_recursive path =
...
 
+let list_dir dir k =
+  let handle = Unix.opendir dir in
+  let close () = Unix.closedir handle in
+  let rec read () =
+    try
+      let entry = Unix.readdir handle in
+      k entry ;
+      read ()
+    with End_of_file -> ()
```
fsestini: On a second look, I think this `try/with` should only be around the call to `Unix.readdir`. Right now it wraps `k entry` and `read ()` too, which makes `read` non-tail-recursive, and silently swallows EOF exceptions raised by `k`.

I recall the idiom `match ... with | exception Foo -> ...` being useful in these cases:

```
  let rec read () =
    match Unix.readdir handle with
    | entry -> k entry; read ()
    | exception End_of_file -> ()
```

internal/herd_regression_test.ml
```diff
@@ -39,61 +38,30 @@ type flags = {
...
-    )
-
-let read_litmus_dir litmus_dir =
-  let litmuses = ref [] in
-  let () =
-    for_each_litmus_in_dir litmus_dir
-      (fun litmus -> litmuses :=  litmus :: !litmuses) in
-  List.rev !litmuses
-
+  (* append in not tail-recursive in 4.14, rev_append forces the reverse sort*)
```
fsestini: But then again, `List.map` is also not tail recursive (modulo cons) in 4.14. 

I general it looks to me like this function might potentially allocate quite a bit more compared to the previous version (e.g. it materialises the whole output in a list and does a `List.map` on it). I would consider implementing `litmuses_in` as an iterator-returning function to avoid list allocations and preserve the memory behaviour of the previous implementation.

INSTALL.md
fsestini: I think this is still unresolved.

litmus/tests/cross/AArch64/A04.t
```diff
@@ -0,0 +1,27 @@
+Use litmus7 to generate code from a litmus test
+
+  $ TEST="A04"
+  $ RUNNER='qemu-aarch64'
+  $ GCC='aarch64-linux-gnu-gcc'
+  $ mkdir "$TEST"
+  $ litmus7 -set-libdir ../../../libdir -gcc="$GCC" -o "$TEST" \
+  > "../../../../herd/tests/instructions/AArch64/$TEST.litmus" \
+  > -mode std -a 4 -s 1k -r 100
```
fsestini: Should there be a trailing `\` here?

internal/lib/filesystem.ml
```diff
@@ -47,7 +47,26 @@ let rec remove_recursive path =
...
+  in
+  Fun.protect ~finally:close read
+
+let temp_name root =
+  let name = Random.bits64 () in
```
fsestini, line 54: How does this generator get seeded? Perhaps we need a `Random.self_init ()` somewhere?

fsestini:
Added a few more comments. I think the cram tests look fine, but the changes to `internal/` need more work/discussion I think. In light of that, I would consider pulling either the cram tests or the changes to `internal/` into a separate PR to be reviewed and merged independently.

internal/lib/filesystem.ml
psafont: The Random module has a default value for the seed. This means that the sequence is deterministic. In theory this could be used to generate a Denial of Service why blocking the creation of a file, but other than being very annoying, I don't see it having any other implication on security.

The default is defined in:
https://github.com/ocaml/ocaml/blob/4.14/stdlib/random.ml#L199
https://github.com/ocaml/ocaml/blob/5.5/stdlib/random.ml#L334

psafont:
I've removed the most problematic changes in `internal` from the series

internal/lib/filesystem.ml
fsestini: The problem is that every process will get the same initial seed unless you call `self_init` or similar:

```ocaml
$ cat random_bits.ml
let temp_name root =
  let name = Random.bits64 () in
  Printf.sprintf "%s/%Ld" root name

let () = print_endline (temp_name "root")
$ ocamlopt -o random_bits random_bits.ml
$ ./random_bits
root/4985260161842195760
$ ./random_bits
root/4985260161842195760
$ ./random_bits
root/4985260161842195760
```

This can be a problem if this function is invoked from different processes running in parallel, as all of them will get the same name of temporary directory and overwrite each other's data.

psafont: Now seeding happens when fIlesystem.ml is loaded. This ensures the seeding will happen at least once when any executable uses this module.

This could be done at the topmost level for each binary, but it's more error prone, and none of them do it at the moment, producing more churn.

internal/herd_assumptions_test.ml
psafont: Done

internal/herd_regression_test.ml
psafont: Done

fsestini (approved):
Thanks @psafont !

PR #1927, opened 2026-07-22
[aslspec] type parameter refactoring and type inference

Roman-Manevich:
This PR is the first in a sequence of changes to support grammar definitions in `aslspec` (and also to model generic helper relations). Grammar definitions need parameterized relations to define reusable rules that build the ASL AST. A follow-up PR will unify operators and relations; this PR prepares for that by making type-parameter inference reusable and completing support for parameterized types.

Parameterized types are currently not instantiated during type inference. For example:

```
typedef Pair[[T]] =
  | PairValue(left: T, right: T)
;
```

When type checking `PairValue(x, y)`, we need to infer a concrete type for `T` from the types of `x` and `y`, then instantiate the variant `PairValue` before comparing it with the expression.

For example, if both `x` and `y` have type `Int`, `PairValue(x, y)` is inferred as `Pair[[Int]]`, so accessing `left` from a `Pair[[Int]]` produces an `Int`. If `x` has type `Int` but `y` has type `Bool`, the two inferred types for `T` cannot be unified and `aslspec` reports an error.

This PR:
- Represents operator type parameters explicitly in the AST.
- Factors out type-parameter inference and substitution so parameterized types, and later parameterized relations, can use them.
- Instantiates parameterized variants during type checking.
- Detects duplicate operator type parameters.
- Adds documentation and end-to-end tests.


asllib/aslspec/tests.t/parameterized_type_instantiation.spec
```diff
@@ -0,0 +1,109 @@
+typedef Int;
+
+typedef Tree[[T]] =
+  | Leaf(value: T)
+;
+
+typing relation make_leaf(x: Int) -> (Tree[[Int]])
+{
+  prose_transition = ""
```
HadrienRenaud, line 9: Can we make `prose_transition` optional?

asllib/aslspec/spec.ml
```diff
@@ -804,6 +788,25 @@ let filter_rule_for_path { Relation.name; rule_opt; loc } path_str =
...
+      let _ =
+        List.fold_left
+          (fun seen parameter ->
+            if StringSet.mem parameter seen then
+              Error.duplicate_type_parameter loc ~owner:("operator " ^ name)
+                parameter
+            else StringSet.add parameter seen)
+          StringSet.empty parameters
+      in
+      ()
```
HadrienRenaud, line 804: Consider using `|> ignore` instead of `let _ = ... in ()` to decrease indentation and simplify code.

asllib/aslspec/spec.ml
```diff
@@ -804,6 +788,25 @@ let filter_rule_for_path { Relation.name; rule_opt; loc } path_str =
...
+      declares each type parameter at most once. *)
+  let check_unique_parameters { ast; _ } =
+    let check_relation { Relation.loc; name; parameters; _ } =
+      let _ =
+        List.fold_left
+          (fun seen parameter ->
+            if StringSet.mem parameter seen then
+              Error.duplicate_type_parameter loc ~owner:("operator " ^ name)
+                parameter
+            else StringSet.add parameter seen)
```
HadrienRenaud, line 801: optionally: avoid doing 2 accesses by checking physical equality:

```ocaml
            let seen' = StringSet.add parameter seen in
            if seen' == seen then
              Error.duplicate_type_parameter loc ~owner:("operator " ^ name)
                parameter;
            seen'
```

asllib/aslspec/spec.ml
```diff
@@ -1750,7 +1749,7 @@ module Check = struct
       *)
       check_well_instantiated spec term
 
-    let check ({ ast; id_to_defining_node } as spec) =
+    let check ({ ast; _ } as spec) =
```
HadrienRenaud, line 1752: At this point, do not deconstruct it here and use `spec.ast` at line 1772:

```suggestion
    let check spec =

File: asllib/aslspec/spec.ml [hunk 5]
```diff
@@ -353,6 +357,58 @@ let record_variant_for_expr spec expr =
 let is_defined_id self id = StringMap.mem id self.id_to_defining_node
 let elements self = self.ast
 
+(** Low-level, error-free transformations of explicit type parameters in type
+    terms. Parameter binding inference, which depends on type unification, is
+    implemented by [TypeInference.TypeParameterInference]. *)
```
HadrienRenaud, line 362: If you could put a small example here that would help readability a lot

asllib/aslspec/spec.ml
```diff
@@ -353,6 +357,58 @@ let record_variant_for_expr spec expr =
...
+          }
+    | Function
+        ({ from_type = from_name, from_term; to_type = to_name, to_term } as
+         node) ->
+        Function
+          {
+            node with
+            from_type = (from_name, substitute_in_term from_term);
+            to_type = (to_name, substitute_in_term to_term);
+          }
```
HadrienRenaud, line 408: nit:
```suggestion
    | Function node ->
        let from_name, from_term = node.from_type
        and to_name, to_term = node.to_type in
        let from_type = (from_name, substitute_in_term from_term)
        and to_type = (to_name, substitute_in_term to_term) in
        Function { node with from_type; to_type }

File: asllib/aslspec/spec.ml [hunk 7]
```diff
@@ -2219,43 +2276,52 @@ module Check = struct
...
       Utils.list_get_all_option unified_args_opt
 
-    module InstantiateOperator : sig
-      val instantiate_operator_types_from_inferred_types :
+    (** Infers bindings for explicit type parameters using the enclosing type
+        inference module's unification operations. Operations that can report
+        errors take an [owner] argument describing the calling definition,
+        allowing this machinery to be used for both operators and parameterized
+        types. Low-level substitution and typedef expansion are provided by
+        [TypeParameterOps]. *)
```
HadrienRenaud, line 2284: Once again, having examples would be good, this is very hard to understand otherwise

asllib/aslspec/spec.ml
```diff
@@ -2312,98 +2376,62 @@ module Check = struct
...
+          the bindings in [parameter_env]. *)
+      let substitute parameter_env term =
+        TypeParameterOps.substitute parameter_env term
 
+      (** [infer spec ~owner ~parameters ~formal_terms ~actual_terms ~loc]
+          infers bindings for [parameters] from corresponding formal and actual
+          terms. [owner] and [loc] are used for error reporting. *)
+      let infer spec ~owner ~parameters ~formal_terms ~actual_terms ~loc =
+        let parameter_env =
+          List.fold_left2
```
HadrienRenaud, line 2415: Adding a comment on why the two list must be of same length would be good

asllib/aslspec/spec.ml
```diff
@@ -2312,98 +2376,62 @@ module Check = struct
...
+            (fun curr_env formal_type arg_type ->
+              infer_parameter_type spec ~owner formal_type arg_type curr_env)
+            StringMap.empty formal_terms actual_terms
+        in
+        let () =
+          List.iter
+            (fun param ->
+              if not (StringMap.mem param parameter_env) then
+                Error.uninstantiated_parameter loc param ~owner)
```
HadrienRenaud, line 2424: In what context can this happen?

asllib/aslspec/spec.ml
```diff
@@ -244,20 +244,31 @@ let make_variant_id_to_containing_type ast =
         acc_map variants)
     StringMap.empty typedefs
 
-let make_field_to_containing_variant ast =
+type containing_record = {
+  record_variant : TypeVariant.t;
+  containing_type : Type.t;
+}
+(** A record variant together with the type definition containing it. *)
```
HadrienRenaud, line 251: Can we have an example please?

asllib/aslspec/spec.ml
```diff
@@ -407,6 +425,21 @@ module TypeParameterOps = struct
...
+    match defining_node_opt_for_id spec typename with
+    | Some (Node_Type { Type.param_opt = Some parameter; variants; _ }) ->
+        List.map
+          (fun ({ TypeVariant.term; _ } as variant) ->
+            {
+              variant with
+              TypeVariant.term =
+                substitute (StringMap.singleton parameter argument) term;
+            })
+          variants
```
HadrienRenaud, line 441: Only build the map once, and write it a bit more succinctly by opening `TypeVariant`:
```suggestion
        let env = StringMap.singleton parameter argument in
        let open TypeVariant in
        List.map
          (fun variant ->
            { variant with term = substitute env variant.term })
          variants

File: asllib/aslspec/spec.ml [hunk 12]
```diff
@@ -1563,6 +1610,31 @@ module Check = struct
             String.equal sub_typename super_typename
             && subsumed_rec spec expansion_limit expanded_types sub_term
                  super_term
+        | _, ParamType { typename; term = _, argument } ->
+            if expansion_count typename expanded_types >= expansion_limit then
+              false
+            else
+              let expanded_types =
```
HadrienRenaud, line 1617: Maybe?

```suggestion
            expansion_count typename expanded_types >= expansion_limit
            &&
            let expanded_types =

File: asllib/aslspec/spec.ml
Comment (HadrienRenaud) [inline asllib/aslspec/spec.ml:1526]: this is outdated btw

File: asllib/aslspec/spec.ml [hunk 13]
```diff
@@ -1563,6 +1610,31 @@ module Check = struct
...
+                    super_term)
+        | ParamType { typename; term = _, argument }, _ ->
+            if expansion_count typename expanded_types >= expansion_limit then
+              false
+            else
+              let expanded_types =
+                increment_expansion_count typename expanded_types
+              in
+              let variants = TypeParameterOps.variants spec typename argument in
+              (not (Utils.list_is_empty variants))
```
HadrienRenaud, line 1632: why this line? can there be an empty list of variants? Why wouldn't they be subsumed then? the empty type could be understood as a subtype of anything (because it can't get any value anyway)

asllib/aslspec/spec.ml
```diff
@@ -1563,6 +1610,31 @@ module Check = struct
             String.equal sub_typename super_typename
             && subsumed_rec spec expansion_limit expanded_types sub_term
                  super_term
+        | _, ParamType { typename; term = _, argument } ->
```
HadrienRenaud, line 1613: Can we add a negative test that is recursive and uses parameters?

asllib/aslspec/spec.ml
```diff
@@ -2427,6 +2494,77 @@ module Check = struct
         parameter_env
     end
 
+    (** [containing_type_for_variant spec variant] returns the type definition
+        containing [variant]. Labelled variants are found through their label;
+        record variants are found through one of their globally unique fields.
+    *)
+    let containing_type_for_variant spec { TypeVariant.term; _ } =
```
HadrienRenaud, line 2501: If I'm reading this correctly, the only usage of `containing_type_for_variant` has an `assert false` if `containing_type_for_variant` return `None`.
Consider moving this `assert false` up the chain, making `containing_type_for_variant` return a `'a` instead of a `'a option`.
Benefits are simplifying the code , make it more succinct, and simplify debugging (multiple `assert false` give you more information when one of them is hit).

asllib/aslspec/spec.ml
```diff
@@ -2427,6 +2494,77 @@ module Check = struct
...
+          | Some type_name -> (
+              match StringMap.find_opt type_name spec.id_to_defining_node with
+              | Some (Node_Type type_def) -> Some type_def
+              | _ -> None)
+          | None -> None)
+      | Term.Parameter _ | Term.TypeOperator _ | Term.ParamType _
+      | Term.Tuple { label_opt = None; _ }
+      | Term.Record { fields = []; _ }
+      | Term.ConstantsSet _ | Term.Function _ ->
+          None
```
HadrienRenaud, line 2517: Suggestion above:

```suggestion
      match term with
      | Term.Record { fields = { Term.name; _ } :: _; _ } ->
          let { containing_type; _ } =
            StringMap.find name spec.field_to_containing
          in
          containing_type
      | Term.Label { label } | Term.Tuple { label_opt = Some label; _ } -> (
          let type_name =
            StringMap.find label spec.variant_id_to_containing_type
          in
          match StringMap.find type_name spec.id_to_defining_node with
          | Node_Type type_def -> type_def
          | _ -> assert false)
      | Term.Parameter _ | Term.TypeOperator _ | Term.ParamType _
      | Term.Tuple { label_opt = None; _ }
      | Term.Record { fields = []; _ }
      | Term.ConstantsSet _ | Term.Function _ ->
          assert false

HadrienRenaud (commented):
In general, this looks good to me.

I struggle a bit to see the big picture, so consider maintaining a big picture explanation of what is happening.

> Detects duplicate operator type parameters.

I think having a small example in the PR description would make it clearer what this is about

I think it is worth considering implementing a `map_terms` function on your types so that you can implement term rewriting more easily.

asllib/aslspec/spec.ml
hrutvik: +1, could make similar changes to other cases in this function too

PR #1929, opened 2026-07-27
[herd] Mimic Cat code for computing RMW atomic load pairs in OCaml

maranget:
Looking at PR #1917, I realised that the OCaml code did not follow the Cat definition. Namely,  the OCaml code assumes the existence of internal dependencies or ordering from the read to the write of "internal" atomic pairs, while the reference Cat definition does not:
```
[Exp & R & X]; (same-instance \ id) & loc; [Exp & W & X]
```
The new OCaml code of this PR follows the Cat definition more closely.


herd/memUtils.mli
```diff
@@ -106,6 +106,9 @@ module Make : functor (S: SemExtra.S) -> sig
 (* Partition by location *)
   val partition_events : S.event_set -> S.event_set list
 
+(* Collect by program order (i.e. by instruction instance). *)
+  val collect_by_po : S.event list -> S.event list list
```
fsestini: Maybe personal preference, but I would suggest:
```suggestion
(* Group events by program order index (i.e. by instruction instance). *)
  val group_by_po : S.event list -> S.event list list
```

lib/myMap.ml
```diff
@@ -50,6 +50,7 @@ module type S = sig
 
 (* Bind keys to list of values *)
   val accumulate : key -> 'a -> 'a list t -> 'a list t
+  val get_accumulated : 'a list t -> 'a list list
```
HadrienRenaud: This could also be made more generic, with the same implementation (just a different name):

```suggestion
  (** [values m] is the list of values bound in [m].
  
      Equivalent to [bindings m |> List.map snd]. *)
  val values : 'a t -> 'a list
```

herd/mem.ml
```diff
@@ -1959,30 +1956,33 @@ let add_eq v1 v2 eqs =
...
-                let rs,ws = List.partition E.is_load evts in
-                let ws = EvtSet.of_list ws in
-                List.fold_left
-                  (fun k er ->
-                    match EvtSet.find_first_after er ws with
-                    | Some ew when E.same_instance er ew -> (er, ew) :: k
-                    | _ -> k (* Can occur for failing CAS *))
-                  k rs)
+                U.collect_by_po evts
+                |> List.fold_left
```
HadrienRenaud: Couldn't we fold directly on the `IntMap`? Why extract it to a list first?

herd/memUtils.mli
HadrienRenaud: By analogy to the other `collect_` functions, shoudln't this return the `IntMap.t`?

```suggestion
  val collect_by_po : S.event list -> S.event list IntMap.t
```

Or maybe it should be renamed `partition_by_progorder`?

herd/memUtils.ml
```diff
@@ -436,6 +436,19 @@ let lift_proc_info i evts =
...
     LocEnv.fold (fun _ evts k -> E.EventSet.of_list evts::k) env []
 
+(****************************)
+(* Collect by program order *)
+(****************************)
+
+let collect_by_po es =
+  List.fold_left
+    (fun m e ->
+      match E.progorder_of e with
```
HadrienRenaud, line 438: Can 2 events in different threads have the same `progorder`? If not, we should document that this function expect effects to the same proc (and maybe even check for it).

HadrienRenaud (approved):
Thanks @maranget, I like this PR. I have a few comments on the interface between `mem` and `memUtils`

herd/memUtils.mli
fsestini: I think 'partition' might be confusing (to me at least), because 'partition' has a well-defined meaning in set theory that `collect_by_po` does not fully fit: https://en.wikipedia.org/wiki/Partition_of_a_set
In particular it drops events from the input if they don't have a po index.

I've suggested `group_by_po` in another comment, as a more general name.

maranget: Gladly change, thanks @fsestini.

lib/myMap.ml
maranget: Nice catch. Thanks @HadrienRenaud. Given your next suggestion I delete this function though.

herd/memUtils.mli
maranget: I follow the suggestion of `group_by_po` returning a map.

herd/memUtils.ml
maranget: I add a comment in `memUtils.mli`.

maranget:
Hi @fsestini and @HadrienRenaud, I have followed your suggestions. Notice that I am leaving tomorrow,. I'll be back on 10 August, This PR can wait until then or can be merged now if you agree.

HadrienRenaud:
For me you can merge this today @maranget

fsestini:
Some further comments:
- The `EvtSet` module binding at `mem.ml:1939` looks unused now and could be removed. Happy to see this merged once that's cleaned up.
- Not a blocker for this PR, but I think ideally it would be good if at some point we had some focused OCaml unit tests to check this kind of utility functions, as they seem subtle. I think property-based testing of the likes of `qcheck` would be a good fit here to catch edge cases/hidden assumptions.

herd/mem.ml
```diff
@@ -1959,30 +1956,32 @@ let add_eq v1 v2 eqs =
...
       (*
        * Compute `[Exp & R & X]; (same-instance \ id) & loc; [Exp & W & X]`,
-       * by matching each explicit atomic read with its (generalised)
-       * po-successor explicit atomic write to the same address.
-       * It is a fatal error for this write not to be emited
-       * by the same instruction instance as the read.
+       * Notice that is assumed that an instruction execution generates:
+       *   + one atomic effect exactly, or
+       *   + one atomic read and one atomic write.
+       * Other situations are flagged by a fatal error.
```
relokin, line 1954: Sorry I've just started looking at this so I haven't looked at the code yet. Isn't this going to create problems for an atomic instruction (for example a SWP) that also performs a hardware update of the AF?

HadrienRenaud: Hi @relokin 

Some beginning of an answer, I'm not sure this is what you meant but maybe this is enough for you to draw a conclusion about it.

This code is only for explicit atomic effects at a given location in a given processor. 
More precisely, for an effect `e` to be considered here, it needs to satisfy: `E.is_atomic e && Option.is_some (E.proc_of e) && E.is_explicit` (see the first result of [`MemUtils.collect_atomics_and_exclusives`](https://github.com/herd/herdtools7/blob/02cc3f14e6d9eb2de8c670c54eed95460aeb2ca9/herd/memUtils.ml#L396-L397)). I don't think hardware updates satisfy those, unless I'm missing something?

Note those effects are also partitioned by processor, instruction instance and location.

maranget: Hi I have updated  comments in order to clarifiy the point made by @relokin.

maranget:
> * The `EvtSet` module binding at `mem.ml:1939` looks unused now and could be removed. Happy to see this merged once that's cleaned up.

Done, thanks for the sharp eye @fsestini.

maranget:
Hi @relokin, @HadrienRenaud and @fsestini. My vacations have been postponed, so that I could answer your remarks, Do you agree with merging as the PR is now?

herd/mem.ml
relokin: Thank you both that makes sense!

herd/mem.ml
```diff
@@ -1959,50 +1947,53 @@ let add_eq v1 v2 eqs =
...
-              (restrict_domains E.is_load E.is_store po)
-              (sequence po po) (* local  po is already from EX to EX *)
-           ::k)
-        [] excls |> E.EventRel.unions in
+             let open E.EventRel in
+             diff
+               (restrict_domains E.is_load E.is_store po)
+               (sequence po po) (* local  po is already from EX to EX *)
+             ::k)
+          [] excls |> E.EventRel.unions in
```
relokin: Since we don't have a particular style we follow, could we restore the old spacing to avoid polluting the diff?

relokin (approved):
Thanks Luc - sorry to hear that your holiday's have been postponed! I've got only a minor suggestion but I would like Filippo to approve this as well.

herd/memUtils.mli
```diff
@@ -106,6 +106,13 @@ module Make : functor (S: SemExtra.S) -> sig
 (* Partition by location *)
   val partition_events : S.event_set -> S.event_set list
 
+(*
```
fsestini, line 109: Extremely minor, but for comments that serve as documentation for a function, I would use the `(** ... *)` format, so that they show up in documentation files generated via `odoc` as well as editor tooling (e.g. `shift+k` in vim+LSP).

fsestini (approved):
LGTM

herd/memUtils.mli
maranget: Done.

herd/mem.ml
maranget: Done, thanks.

maranget:
Merged, thanks @HadrienRenaud, @fsestini, @relokin.

PR #1931, opened 2026-07-28
Dot2desc multiple reg alternatives

TiberiuBucur:
This patch adds the ability for dot2desc to accept multiple registers as operands of an instruction passed through the `-instr` parameter.

Take as an example the execution graph produced by herd on a simple CSEL instruction:

```
AArch64 CSEL-success
{
 0:X1=1; 0:X2=2;
}
 P0               ;
 CSEL W0,W1,W2,NE ;
```

and let's run dot2desc with `-instr "CSEL Xd,Xn,Xm,cond"`. The final markdown produced by dot2desc (and pandoc) will contain phrases describing the architectural registers passed through `-instr`:

```
-   D~2~: There is an Intrinsic Data Dependency from the Register Read Effect of X~n~ to the Register Write Effect of X~d~.
```

After this patch, dot2desc can accept multiple registers for the same operand. This is useful when one instruction has the same dependencies for multiple register sizes. In the case of CSEL, the execution graph is the same. However, now executing dot2desc with `-instr "CSEL Xd|Wd,Xn|Wn,Xm|Wm,cond"` produces phrases describing all alternative registers in place of their respective operands:

```
-   D~2~: There is an Intrinsic Data Dependency from the Register Read Effect of X~n~ or W~n~ to the Register Write Effect of X~d~ or W~d~.
```

Alternatives for a single operand are separated by the `|` character. This applies to memory address registers as well, in which case the `|` can sit either inside or outside the memory addressing syntax (`[]`). For example, a load instruction can be passed as both `-instr "LDR Xt,[Xn|Wn]"` or `-instr "LDR Xt,[Xn]|[Wn]"`, although using a word (32bit) sized register as an address register does not make sense in practice. This also highlights that one operand can have only 1 possible option, while another can have multiple.

Malformed inputs include using the `|` operator without providing an alternative (eg. `-instr "BLR Xn|"`). In such cases, dot2desc will fail with "Invalid empty operand in instr parameter".

It is worth mentioning that if the `-wbfname` parameter is used, the written-back dot graph will contain effects describing only the first alternative of an operand (in the case described above, the Xn and Xd registers).

### DIFF
--- a/tools/dot2desc.ml
+++ b/tools/dot2desc.ml
@@ -325,6 +325,50 @@ module DotGraph = struct
       true
     with Not_found -> false
 
+  let access_annots = [
+    "";
+    "Acq*"; "Acq"; "AcqPc*"; "AcqPc";
+    "Rel*"; "Rel";
+    "*"; "NoRet"; "^s"; "NT"; "EX"; "AcqEx"; "RelEx";
+  ]
+
+  let explicit_annots = [
+    "";
+    "Exp"; "NExp"; "IFetch"; "NExpAF"; "NExpDB"; "NExpAFDB"; "GCS";
+  ]
+
+  (* Not every combination is possible, but this is good enough for our purposes *)
+  let access_suffixes =
+    List.concat (List.map (fun annot ->
+      List.map (fun explicit -> annot ^ explicit) explicit_annots
+    ) access_annots)
+
+  let is_access_size = function
+    | 'b' | 'h' | 'w' | 'q' | 's' -> true
+    | _ -> false
+
+  let is_access_suffix suffix =
+    List.exists ((=) suffix) access_suffixes
+
+  let strip_bracket_access_size value =
+    let len = String.length value in
+    match String.index_from_opt value 0 ']' with
+    | None -> value
+    | Some bracket_pos ->
+        match String.index_from_opt value bracket_pos '=' with
+        | None -> value
+        | Some eq_pos when eq_pos > bracket_pos + 1 ->
+            let size_pos = eq_pos - 1 in
+            let suffix_start = bracket_pos + 1 in
+            let suffix_len = size_pos - suffix_start in
+            let suffix = String.sub value suffix_start suffix_len in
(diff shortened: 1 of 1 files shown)


tools/dot2desc.ml
```diff
@@ -192,6 +192,73 @@ module DescDict = struct
...
+    Str.regexp {|\(ZA\|[CXWVBHSDQZP]\)\([a-z0-9]+\)$|}
+
+  let valid_param = Str.regexp {|[a-zA-Z0-9\._]+$|}
+
+  let invalid_instr () =
+    Warn.fatal "Invalid format for command. Command must have arguments \
+      separated by commas. The mnemonic and the first argument are separated \
+      by exactly one space (eg. LDR Xn,[Xm])"
+
+  let strip_brackets s =
```
ShaleXIONG: The `strip_brackets` function will erase uneven bracket as well such as input `[Xn`. Not a big deal for me but as long as it will not affect other part.

tools/dot2desc.ml
```diff
@@ -234,6 +301,50 @@ module DotGraph = struct
...
+    "*"; "NoRet"; "^s"; "NT"; "EX"; "AcqEx"; "RelEx";
+  ]
+
+  let explicit_annots = [
+    "";
+    "Exp"; "NExp"; "IFetch"; "NExpAF"; "NExpDB"; "NExpAFDB"; "GCS";
+  ]
+
+  (* Not every combination is possible, but this is good enough for our purposes *)
+  let access_suffixes =
```
ShaleXIONG, line 341: empty string "" from `access_annots` followed by empty string "" from `explicit_annots` is allowed?

TiberiuBucur: Yes

tools/dot2desc.ml
TiberiuBucur: Thanks for pointing this out. It should be fixed now.

ShaleXIONG (approved):

PR #1933, opened 2026-07-28
[herd-www] Simplify build process

fsestini:
This PR simplifies `herd-www`'s build process: instead of copying and patching Herd sources, Jerd now depends directly on the newly-added `herd_core` Dune library. We also remove the `notwww.awk` preprocessing pass and the corresponding `NOTWWW` markers. This also fixes some jerd-related errors when running `dune build @herd-www/check`.

I don't have the full historical context of the reasons for introducing `notwww.awk` (please chime in if you do), but with Codex's help I've done some archeology and run some experiments to make sure the removal of `notwww.awk` wouldn't introduce any regressions.

Git history suggests that the script began as a workaround for an old `Stdint`/js_of_ocaml build problem (cb980e41f; see 365a55f04). `Stdint` was replaced shortly afterwards (8cd04096a), but the filtering remained and was later extended to hide the specialised Morello, Neon, SVE, and SME pattern matching paths in `herd/AArch64ParseTest.ml`. ASL support was separately excluded from `herd/parseTest.ml` in abc5b4fee; that commit does not record a technical reason.

To better understand the consequences of removing `notwww.awk`, I've tried building `jerd.js` without the `notwww.awk` preprocessing step and running it on all litmus tests in `herd/tests/instructions`, as well as all catalogue tests in scope of `make test`. These are the results:

1. Various tests in `herd/tests/instructions/AArch64.sme` and `herd/tests/instructions/AArch64.sve` fail due to missing js stubs for the `zarith` function `ml_z_to_int64_unsigned`. This problem can be addressed by requiring that herd-www be built with `zarith_stubs_js` v0.17, which does define stubs for that function. In turn, this means requiring OCaml >= 5.1 for herd-www builds (as required by `zarith_stubs_js`).

2. Two ASL tests in `herd/tests/instructions/AArch64.ASL` (`MOVI01.litmus` and `MOVI02.litmus`) fail during lexing: some of the numeric literals used in the tests are too big to fit `js_of_ocaml`'s [ 32-bit representation of `int` ](https://ocsigen.org/js_of_ocaml/latest/js_of_ocaml/runtime-representation.html). This issue is unrelated to anything ASL-specific and applies to any litmus test containing big integer literals. It is thus orthogonal to `notwww.awk`.

3. No other test raises failures/exceptions

In light of these results, this PR proposes the following:

- simplify the herd-www build process and remove the `notwww.awk` script;
- set herd-www's minimum version of `zarith_stubs_js` to v0.17 and of OCaml to 5.1, so that SME/SVE tests don't fail due to missing js stubs;
- Add a dedicated `.opam` file for herd-www (cherry-picked from [this PR](https://github.com/herd/herdtools7/pull/1728)), to make herd-www's dependency requirements explicit.

Let me know if this all makes sense.

Note: none of the changes proposed here should affect the functionality of herd-www's deployment at developer.arm.com/herd7, since they do not affect the interface or input/output protocol of `jerd.js`.

### DIFF
--- a/.github/workflows/check-build-www.yml
+++ b/.github/workflows/check-build-www.yml
@@ -40,6 +40,9 @@ jobs:
         uses: ocaml/setup-ocaml@v3
         with:
           ocaml-compiler: dune,menhir,zarith,js_of_ocaml,js_of_ocaml-ppx,zarith_stubs_js,logs
+      - name: Install herd-www dependencies
+        working-directory: herd-www
+        run: opam install . --deps-only --yes
       - run: |
           opam exec -- make Version.ml
           opam exec -- dune build tools/cat2html.exe
--- a/herd-www/dune
+++ b/herd-www/dune
@@ -1,30 +1,13 @@
-(copy_files
- (files ../herd/*.{ml,mli}))
-
 (executable
  (name jerd)
  (modes js)
  (optional)
- (libraries herdtools js_of_ocaml zarith_stubs_js)
- (preprocessor_deps
-  (file notwww.awk))
+ (libraries herdtools herd_core js_of_ocaml zarith_stubs_js)
+ (modules jerd webInput webRun)
  (preprocess
   (per_module
    ((pps js_of_ocaml-ppx)
     jerd
-    webInput)
-   ((action
-     (system "awk -f herd-www/notwww.awk %{input-file}"))
-    AArch64ParseTest
-    ParseTest
-    Top_herd)))
+    webInput)))
  (js_of_ocaml
-  (flags --disable genprim))
- (modules_without_implementation
-  AArch64Sig
-  action
-  arch_herd
-  monad
-  sem
-  XXXMem
-  port))
+  (flags --disable genprim)))
--- a/herd-www/herd-www.opam
+++ b/herd-www/herd-www.opam
@@ -1,0 +1,19 @@
+opam-version: "2.0"
+synopsis: "Dependencies for herd-www (web interface for herdtools7)"
+maintainer: "Luc Maranget <Luc.Maranget@inria.fr>"
+authors: [
+  "Jade Alglave <j.alglave@ucl.ac.uk>"
+  "Luc Maranget <Luc.Maranget@inria.fr>"
+]
+homepage: "https://developer.arm.com/herd7"
+bug-reports: "http://github.com/herd/herdtools7/issues/"
+license: "CECILL-B"
+depends: [
+  "ocaml" {>= "5.1.0"}
+  "dune"  {>= "3.11.0" }
+  "menhir" {>= "20231231"}
+  "js_of_ocaml" {>= "6.4.0"}
+  "js_of_ocaml-ppx" {>= "6.4.0"}
+  "zarith" {>= "1.13"}
+  "zarith_stubs_js" {>= "v0.17.0"}
+]
--- a/herd-www/README.md
+++ b/herd-www/README.md
@@ -12,12 +12,10 @@ Contact: diy-devel@inria.fr
 Compilation and installation
 ============================
 
-Dependencies.
-- ocaml, dune, menhir, js_of_ocaml, js_of_ocaml-ppx, zarith_stubs_js
-
-Install all those with opam!
+Dependencies are declared in `herd-www.opam`. From this directory, use opam to
+solve and install them:
 ```
-% opam install dune menhir js_of_ocaml js_of_ocaml-ppx  zarith_stubs_js
+% opam install . --deps-only
 ```
 
 Build
--- a/herd-www/jerd.ml
+++ b/herd-www/jerd.ml
@@ -345,7 +345,35 @@ let run_herd bell cat litmus cfg cat_label =
         let bell_model_info = bi
         include Config end) in
 
+  let module SP =
+    Splitter.Make
+      (struct
+        let debug = Config.debug.Debug_herd.lexer
+        let check_rename = Config.check_rename
+      end) in
+
+  (* Standalone (non-AArch64) ASL tests are outside Jerd's supported architectures.
+     AArch64+ASL additionally needs pseudocode assets that are not deployed.
+     See also: https://github.com/herd/herdtools7/pull/1933 *)
+  let check_supported_litmus f =
+    let splitted = SP.split_string f litmus in
+    match splitted.Splitter.arch with
+    | `ASL -> Warn.user_error "ASL architecture is not supported"
+    | _ ->
+        let module TestConf =
+          TestVariant.Make
+            (struct
+              include Config
+              module Opt = Variant
+              let info = splitted.Splitter.info
+            end)
+        in
+        if TestConf.variant Variant.ASL then
+          Warn.user_error "ASL variant is not supported"
+  in
+
   let from_file f =
+    check_supported_litmus f ;
     SymbValue.reset_gensym () ;
     T.from_file f in
 
--- a/herd-www/jerd.ml
+++ b/herd-www/jerd.ml
@@ -17,6 +17,7 @@
 (** Entry point to Herd  *)
 open Js_of_ocaml
 open Printf
+open Herd_core
 open Opts
 open OptNames
 
--- a/herd-www/notwww.awk
+++ b/herd-www/notwww.awk
@@ -1,4 +1,0 @@
-BEGIN { out=1 ; }
-/\(\* END NOTWWW/   { out=1; }
-{ if (out)  { print $0; } else { printf("(* %s *)\n",$0); } }
-/\(\* START NOTWWW/ { out=0; }
--- a/herd-www/webRun.ml
+++ b/herd-www/webRun.ml
@@ -14,6 +14,7 @@
 (* "http://www.cecill.info". We also give a copy in LICENSE.txt.            *)
 (****************************************************************************)
 
+open Herd_core
 module TR = Top_herd.TestResult
 
 module Make (O : sig
--- a/herd/AArch64ParseTest.ml
+++ b/herd/AArch64ParseTest.ml
@@ -63,13 +63,7 @@ module Make(Conf:RunTest.Config)(ModelConfig:MemCat.Config) = struct
             end
           module X = RunTest.Make (AArch64S) (P) (AArch64M) (Conf)
         end
-(*
- * Markers START/END below are for excluding source
- * when compiling the web interface
- *)
-
         let run =
-(* START NOTWWW *)
           if is_morello then
             let module  AArch64Value = CapabilityValue.Make(ConfMorello) in
             let module X = AArch64Make(AArch64Value) in
@@ -83,7 +77,6 @@ module Make(Conf:RunTest.Config)(ModelConfig:MemCat.Config) = struct
             let module X = AArch64Make(AArch64Value) in
             X.X.run
           else
-(* END NOTWWW *)
             let module AArch64Value = AArch64Value.Make(ConfMorello) in
             let module X = AArch64Make(AArch64Value) in
             X.X.run
--- a/herd/parseTest.ml
+++ b/herd/parseTest.ml
@@ -206,11 +206,9 @@ end) = struct
       | `LISA ->
          let module X = LISAParseTest.Make(Conf)(ModelConfig) in
          X.run dirty ~filename ~contents env splitted
-(* START NOTWWW *)
       | `ASL ->
          let module X = ASLParseTest.Make(Conf)(ModelConfig) in
          X.run dirty ~filename ~contents env splitted
-(* END NOTWWW *)
       | arch -> Warn.fatal "no support for arch '%s'" (Archs.pp arch)
     end else env, None
 


.github/workflows/check-build-www.yml
```diff
@@ -40,6 +40,9 @@ jobs:
         uses: ocaml/setup-ocaml@v3
         with:
           ocaml-compiler: dune,menhir,zarith,js_of_ocaml,js_of_ocaml-ppx,zarith_stubs_js,logs
+      - name: Install herd-www dependencies
+        working-directory: herd-www
+        run: opam install . --deps-only --yes
```
psafont, line 45: Doesn't this install the normal herd dependencies? to install the herd-www ones, the directory should change:
```suggestion
        run: opam install ./herd-www --deps-only --yes
```

fsestini: Won't `working-directory: herd-www` do the trick here?

psafont: yes, indeed

herd-www/dune
```diff
@@ -1,30 +1,13 @@
-(copy_files
- (files ../herd/*.{ml,mli}))
```
psafont, line 2: This rule copied `Version.ml` to `herd-www` since `herd/dune` has a rule to copy it to `/herd/`. If it's still working, does this mean that it wasn't needed at all?

fsestini: `Version.ml` gets built as part of the `herd_core` library and the `Version` module is therefore imported in `jerd.ml` via `open Herd_core`.

That said, I don't think the current code of `jerd.ml` is using `Version.libdir` in a correct or meaningful way, as `Version.libdir` should refer to a path in the build environment, which doesn't exist in the browser's sandbox.

psafont:
> set herd-www's minimum version of zarith_stubs_js to v0.17 and of OCaml to 5.1, so that SME/SVE tests don't fail due to missing js stubs;

The commit that introduces the change is unhelpful[^1] regarding whether this restriction is needed or whether it was made as a matter of policy. In any case, it's unfortunate.

[^1]: https://github.com/janestreet/zarith_stubs_js/commit/8e3faba4c23c147c21e1f779902567d3077a7bc3

herd-www/herd-www.opam
```diff
@@ -9,10 +9,11 @@ homepage: "https://developer.arm.com/herd7"
...
 depends: [
-  "ocaml" {>= "4.14.0"}
+  "ocaml" {>= "5.1.0"}
   "dune"
   "menhir"
-  "js_of_ocaml" {= "5.4.0"}
-  "js_of_ocaml-ppx" {= "5.4.0"}
-  "zarith_stubs_js" {>= "v0.16.1"}
+  "js_of_ocaml" {= "6.4.0"}
+  "js_of_ocaml-ppx" {= "6.4.0"}
```
psafont: Is there any reason in particular for requiring a specific version? I'm not familiar with js_of_ocaml to understand the implication of using some versions or others

maranget:
Thanks for this @fsestini.  I totally agree with the objective of a more standard installation and of getting rid of the awk script. That said, it does not suffice to run AArch64+ASL tests, as the ASL code itself is not installed. (To see this add `variant asl` to the configuration file, visible by clicking on the grear symbol on top the window that displays the Cat model).

maranget:
I am wondering @fsestini, what is your procedure to run jerd on all the tests?

herd-www/herd-www.opam
maranget: Perhaps specifying minimal versions would do the job?
```suggestion
  "js_of_ocaml" {>= "6.4.0"}
  "js_of_ocaml-ppx" {>= "6.4.0"}
```

herd-www/dune
maranget: You are right `Version.libdir` is of no use to the web interface. However `Version.version` and `Version.rev` could be of some use, although they are not exploited at the time.

maranget:
BTW I have compiled and installed in https://diy.inria.fr/TST/www/. It works perfectly, as far as I have tested.

fsestini:
> Thanks for this @fsestini. I totally agree with the objective of a more standard installation and of getting rid of the awk script. That said, it does not suffice to run AArch64+ASL tests, as the ASL code itself is not installed. (To see this add `variant asl` to the configuration file, visible by clicking on the grear symbol on top the window that displays the Cat model).

Thanks @maranget , you're indeed right. In that case then, perhaps we should just make it clear that the ASL variant is not supported by jerd/herd-www, and raise an error right from the start if the user attempts to run a litmus test under `variant asl`.


> I am wondering @fsestini, what is your procedure to run jerd on all the tests?

I had Codex write a script that loads and runs `jerd.js` in a javascript VM via node.js. The script only runs the `run_herd` function of `jerd.js` on various litmus tests and just reports if any of them throws exceptions. It does not check their output against expectations.
I attach the code below. Note that it's quite hacky and not something I would commit as it, but as a quick and dirty test it seemed to work fine.

<details><summary>Details</summary>
jerd-smoke.js:

```js
"use strict";

const fs = require("node:fs");
const path = require("node:path");
const vm = require("node:vm");

const herdWww = __dirname;
const repository = path.resolve(herdWww, "..");
const www = path.join(herdWww, "www");
const bundle = readFile(path.join(www, "jerd.js"));

function readFile(file) {
  return fs.readFileSync(file, "utf8");
}

class XMLHttpRequest {
  open(_method, url, _async) {
    this.url = url;
  }

  send() {
    const relative = String(this.url)
      .replace(/^\/+/, "")
      .split(/[?#]/, 1)[0];
    const file = path.resolve(www, relative);

    if (!file.startsWith(`${www}${path.sep}`)) {
      this.status = 403;
      this.statusText = "Forbidden";
      this.responseText = null;
      return;
    }

    try {
      this.responseText = readFile(file);
      this.status = 200;
      this.statusText = "OK";
    } catch (error) {
      this.responseText = null;
      this.status = 404;
      this.statusText = error.message;
    }
  }
}

function loadJerd() {
  const context = {
    XMLHttpRequest,
    herd_output() {},
    herd_stderr() {},
  };
  vm.createContext(context);
  vm.runInContext(bundle, context, { filename: "jerd.js" });

  if (typeof context.runHerd !== "function") {
    throw new Error("jerd.js did not export runHerd");
  }
  return context;
}

module.exports = {
  loadJerd,
  readFile,
  repository,
};
```

test-jerd-instructions.js:

```js
"use strict";

const fs = require("node:fs");
const path = require("node:path");
const { loadJerd, readFile, repository } = require("./jerd-smoke");

const instructions = path.join(repository, "herd", "tests", "instructions");

function litmusFiles(directory) {
  return fs.readdirSync(directory, { withFileTypes: true }).flatMap((entry) => {
    const file = path.join(directory, entry.name);
    if (entry.isDirectory()) return litmusFiles(file);
    const isLitmus = entry.name.endsWith(".litmus");
    return isLitmus && (entry.isFile() || entry.isSymbolicLink()) ? [file] : [];
  });
}

function suiteConfig(suite) {
  const config = path.join(suite, "ci.cfg");
  if (!fs.existsSync(config)) return "";

  // The web library flattens herd/libdir, whereas the native C test
  // configuration names rc11.cat as herd/libdir/rc11.cat.
  return readFile(config).replaceAll("herd/libdir/", "");
}

for (const entry of fs.readdirSync(instructions, { withFileTypes: true })) {
  if (!entry.isDirectory() || !entry.name.startsWith("AArch64")) continue;

  const suite = path.join(instructions, entry.name);
  const config = `${suiteConfig(suite)}\nshow none\n`;
  let jerd = loadJerd();

  for (const litmus of litmusFiles(suite)) {
    try {
      jerd.runHerd("", "", readFile(litmus), config, "");
    } catch (error) {
      console.error(path.relative(repository, litmus), error);
      process.exitCode = 1;
      // A failed run may have left mutable OCaml state behind. Start the
      // following test from a clean browser runtime.
      jerd = loadJerd();
    }
  }
}
```

test-jerd-catalogue.js:

```js
"use strict";

const { spawnSync } = require("node:child_process");
const path = require("node:path");
const { loadJerd, readFile, repository } = require("./jerd-smoke");

const catalogue = path.join(repository, "catalogue");
const python = "python3";

// Keep this list in step with the cata-test and test.herd-mixed.cata targets
// in the top-level Makefile. In particular, do not add cata-test-all's VMSA
// shelf or the ASL catalogue targets, which are outside `make test`.
const makeTestCatalogueTargets = [
  { book: "aarch64" },
  { book: "aarch64-mixed" },
  { book: "aarch64-pick" },
  { book: "aarch64-faults" },
  { book: "aarch64-MTE" },
  { book: "aarch64-PAC" },
  { book: "aarch64-ifetch" },
  { book: "aarch64-cas" },
  { book: "aarch64-ETS2" },
  { book: "aarch64-ETS3" },
  { book: "aarch64", variant: "mixed" },
  { book: "aarch64-pick", variant: "mixed" },
];

// Shelf.of_file, used by the native catalogue regression test, evaluates the
// shelf as Python. Do the same here rather than attempting to parse a Python
// source file as text.
const readShelfProgram = String.raw`
import json
import os
import sys

shelf_path = os.path.abspath(sys.argv[1])
os.chdir(os.path.dirname(shelf_path))
scope = {}
with open(os.path.basename(shelf_path), encoding="utf-8") as shelf_file:
    exec(compile(shelf_file.read(), shelf_path, "exec"), scope)

keys = ("cats", "cfgs", "bells", "illustrative_tests", "tests")
print(json.dumps({key: scope.get(key) for key in keys}))
`;

function catalogueFile(book, relative) {
  return path.join(catalogue, book, relative);
}

function loadShelf(book) {
  const shelfPath = path.join(catalogue, book, "ci-shelf.py");
  const result = spawnSync(python, ["-c", readShelfProgram, shelfPath], {
    encoding: "utf8",
  });
  if (result.error) throw result.error;
  if (result.status !== 0) {
    throw new Error(result.stderr);
  }

  const {
    cats,
    cfgs,
    bells = [],
    illustrative_tests,
    tests = [],
  } = JSON.parse(result.stdout);

  return {
    bell: bells?.[0] ?? "",
    cat: cats[0],
    config: cfgs[0],
    tests: illustrative_tests?.length ? illustrative_tests : tests,
  };
}

for (const target of makeTestCatalogueTargets) {
  const shelf = loadShelf(target.book);
  const bell =
    shelf.bell === "" ? "" : readFile(catalogueFile(target.book, shelf.bell));
  const cat = readFile(catalogueFile(target.book, shelf.cat));
  const config = `${[
    readFile(catalogueFile(target.book, shelf.config)),
    target.variant && `variant ${target.variant}`,
    "show none",
  ]
    .filter(Boolean)
    .join("\n")}\n`;
  let jerd = loadJerd();

  for (const test of shelf.tests) {
    const litmus = catalogueFile(target.book, test);
    try {
      jerd.runHerd(bell, cat, readFile(litmus), config, "");
    } catch (error) {
      const variant = target.variant ? ` (${target.variant})` : "";
      console.error(`${path.relative(repository, litmus)}${variant}`, error);
      process.exitCode = 1;
      // A failed run may have left mutable OCaml state behind. Start the
      // following test from a clean browser runtime.
      jerd = loadJerd();
    }
  }
}
```
</details>

herd-www/herd-www.opam
fsestini: No reason, good catch, thanks. We used to require a particular version (`5.4.0`) due to [a bug in jsoo](https://github.com/ocsigen/js_of_ocaml/issues/2122) that would cause stack overflows on Firefox. That is now fixed since `6.4.0` so we only need it to be a lower bound.

maranget (approved):

relokin:
> > Thanks for this @fsestini. I totally agree with the objective of a more standard installation and of getting rid of the awk script. That said, it does not suffice to run AArch64+ASL tests, as the ASL code itself is not installed. (To see this add `variant asl` to the configuration file, visible by clicking on the grear symbol on top the window that displays the Cat model).
> 
> Thanks @maranget , you're indeed right. In that case then, perhaps we should just make it clear that the ASL variant is not supported by jerd/herd-www, and raise an error right from the start if the user attempts to run a litmus test under `variant asl`.
> 
> > I am wondering @fsestini, what is your procedure to run jerd on all the tests?
> 
> I had Codex write a script that loads and runs `jerd.js` in a javascript VM via node.js. The script only runs the `run_herd` function of `jerd.js` on various litmus tests and just reports if any of them throws exceptions. It does not check their output against expectations. I attach the code below. Note that it's quite hacky and not something I would commit as it, but as a quick and dirty test it seemed to work fine.
> Details
> jerd-smoke.js:
> 
> ```js
> "use strict";
> 
> const fs = require("node:fs");
> const path = require("node:path");
> const vm = require("node:vm");
> 
> const herdWww = __dirname;
> const repository = path.resolve(herdWww, "..");
> const www = path.join(herdWww, "www");
> const bundle = readFile(path.join(www, "jerd.js"));
> 
> function readFile(file) {
>   return fs.readFileSync(file, "utf8");
> }
> 
> class XMLHttpRequest {
>   open(_method, url, _async) {
>     this.url = url;
>   }
> 
>   send() {
>     const relative = String(this.url)
>       .replace(/^\/+/, "")
>       .split(/[?#]/, 1)[0];
>     const file = path.resolve(www, relative);
> 
>     if (!file.startsWith(`${www}${path.sep}`)) {
>       this.status = 403;
>       this.statusText = "Forbidden";
>       this.responseText = null;
>       return;
>     }
> 
>     try {
>       this.responseText = readFile(file);
>       this.status = 200;
>       this.statusText = "OK";
>     } catch (error) {
>       this.responseText = null;
>       this.status = 404;
>       this.statusText = error.message;
>     }
>   }
> }
> 
> function loadJerd() {
>   const context = {
>     XMLHttpRequest,
>     herd_output() {},
>     herd_stderr() {},
>   };
>   vm.createContext(context);
>   vm.runInContext(bundle, context, { filename: "jerd.js" });
> 
>   if (typeof context.runHerd !== "function") {
>     throw new Error("jerd.js did not export runHerd");
>   }
>   return context;
> }
> 
> module.exports = {
>   loadJerd,
>   readFile,
>   repository,
> };
> ```
> 
> test-jerd-instructions.js:
> 
> ```js
> "use strict";
> 
> const fs = require("node:fs");
> const path = require("node:path");
> const { loadJerd, readFile, repository } = require("./jerd-smoke");
> 
> const instructions = path.join(repository, "herd", "tests", "instructions");
> 
> function litmusFiles(directory) {
>   return fs.readdirSync(directory, { withFileTypes: true }).flatMap((entry) => {
>     const file = path.join(directory, entry.name);
>     if (entry.isDirectory()) return litmusFiles(file);
>     const isLitmus = entry.name.endsWith(".litmus");
>     return isLitmus && (entry.isFile() || entry.isSymbolicLink()) ? [file] : [];
>   });
> }
> 
> function suiteConfig(suite) {
>   const config = path.join(suite, "ci.cfg");
>   if (!fs.existsSync(config)) return "";
> 
>   // The web library flattens herd/libdir, whereas the native C test
>   // configuration names rc11.cat as herd/libdir/rc11.cat.
>   return readFile(config).replaceAll("herd/libdir/", "");
> }
> 
> for (const entry of fs.readdirSync(instructions, { withFileTypes: true })) {
>   if (!entry.isDirectory() || !entry.name.startsWith("AArch64")) continue;
> 
>   const suite = path.join(instructions, entry.name);
>   const config = `${suiteConfig(suite)}\nshow none\n`;
>   let jerd = loadJerd();
> 
>   for (const litmus of litmusFiles(suite)) {
>     try {
>       jerd.runHerd("", "", readFile(litmus), config, "");
>     } catch (error) {
>       console.error(path.relative(repository, litmus), error);
>       process.exitCode = 1;
>       // A failed run may have left mutable OCaml state behind. Start the
>       // following test from a clean browser runtime.
>       jerd = loadJerd();
>     }
>   }
> }
> ```
> 
> test-jerd-catalogue.js:
> 
> ```js
> "use strict";
> 
> const { spawnSync } = require("node:child_process");
> const path = require("node:path");
> const { loadJerd, readFile, repository } = require("./jerd-smoke");
> 
> const catalogue = path.join(repository, "catalogue");
> const python = "python3";
> 
> // Keep this list in step with the cata-test and test.herd-mixed.cata targets
> // in the top-level Makefile. In particular, do not add cata-test-all's VMSA
> // shelf or the ASL catalogue targets, which are outside `make test`.
> const makeTestCatalogueTargets = [
>   { book: "aarch64" },
>   { book: "aarch64-mixed" },
>   { book: "aarch64-pick" },
>   { book: "aarch64-faults" },
>   { book: "aarch64-MTE" },
>   { book: "aarch64-PAC" },
>   { book: "aarch64-ifetch" },
>   { book: "aarch64-cas" },
>   { book: "aarch64-ETS2" },
>   { book: "aarch64-ETS3" },
>   { book: "aarch64", variant: "mixed" },
>   { book: "aarch64-pick", variant: "mixed" },
> ];
> 
> // Shelf.of_file, used by the native catalogue regression test, evaluates the
> // shelf as Python. Do the same here rather than attempting to parse a Python
> // source file as text.
> const readShelfProgram = String.raw`
> import json
> import os
> import sys
> 
> shelf_path = os.path.abspath(sys.argv[1])
> os.chdir(os.path.dirname(shelf_path))
> scope = {}
> with open(os.path.basename(shelf_path), encoding="utf-8") as shelf_file:
>     exec(compile(shelf_file.read(), shelf_path, "exec"), scope)
> 
> keys = ("cats", "cfgs", "bells", "illustrative_tests", "tests")
> print(json.dumps({key: scope.get(key) for key in keys}))
> `;
> 
> function catalogueFile(book, relative) {
>   return path.join(catalogue, book, relative);
> }
> 
> function loadShelf(book) {
>   const shelfPath = path.join(catalogue, book, "ci-shelf.py");
>   const result = spawnSync(python, ["-c", readShelfProgram, shelfPath], {
>     encoding: "utf8",
>   });
>   if (result.error) throw result.error;
>   if (result.status !== 0) {
>     throw new Error(result.stderr);
>   }
> 
>   const {
>     cats,
>     cfgs,
>     bells = [],
>     illustrative_tests,
>     tests = [],
>   } = JSON.parse(result.stdout);
> 
>   return {
>     bell: bells?.[0] ?? "",
>     cat: cats[0],
>     config: cfgs[0],
>     tests: illustrative_tests?.length ? illustrative_tests : tests,
>   };
> }
> 
> for (const target of makeTestCatalogueTargets) {
>   const shelf = loadShelf(target.book);
>   const bell =
>     shelf.bell === "" ? "" : readFile(catalogueFile(target.book, shelf.bell));
>   const cat = readFile(catalogueFile(target.book, shelf.cat));
>   const config = `${[
>     readFile(catalogueFile(target.book, shelf.config)),
>     target.variant && `variant ${target.variant}`,
>     "show none",
>   ]
>     .filter(Boolean)
>     .join("\n")}\n`;
>   let jerd = loadJerd();
> 
>   for (const test of shelf.tests) {
>     const litmus = catalogueFile(target.book, test);
>     try {
>       jerd.runHerd(bell, cat, readFile(litmus), config, "");
>     } catch (error) {
>       const variant = target.variant ? ` (${target.variant})` : "";
>       console.error(`${path.relative(repository, litmus)}${variant}`, error);
>       process.exitCode = 1;
>       // A failed run may have left mutable OCaml state behind. Start the
>       // following test from a clean browser runtime.
>       jerd = loadJerd();
>     }
>   }
> }
> ```

@psafont can we turn this into some form of basic unit testing for the web please? @fsestini do you agree?

fsestini:
> @psafont can we turn this into some form of basic unit testing for the web please? @fsestini do you agree?

Completely agree. It would also be useful to make sure that future changes to `jerd.js` maintain the external input/output interface that other clients of jerd (like the developer.arm.com web app) expect.

herd-www/herd-www.opam
```diff
@@ -0,0 +1,19 @@
...
+  "Jade Alglave <j.alglave@ucl.ac.uk>"
+  "Luc Maranget <Luc.Maranget@inria.fr>"
+]
+homepage: "https://developer.arm.com/herd7"
+bug-reports: "http://github.com/herd/herdtools7/issues/"
+license: "CECILL-B"
+depends: [
+  "ocaml" {>= "5.1.0"}
+  "dune"
+  "menhir"
```
relokin: Should we mirror the minimums from https://github.com/herd/herdtools7/blob/master/herdtools7.opam ?

fsestini: Yep (done in [7016e74](https://github.com/herd/herdtools7/pull/1933/commits/7016e74828650021fc99ef620d7267a9e726ff1e))

relokin (approved):
LGTM

fsestini:
> > Thanks for this @fsestini. I totally agree with the objective of a more standard installation and of getting rid of the awk script. That said, it does not suffice to run AArch64+ASL tests, as the ASL code itself is not installed. (To see this add `variant asl` to the configuration file, visible by clicking on the grear symbol on top the window that displays the Cat model).
> 
> Thanks @maranget , you're indeed right. In that case then, perhaps we should just make it clear that the ASL variant is not supported by jerd/herd-www, and raise an error right from the start if the user attempts to run a litmus test under `variant asl`.

I've taken a second look at this. It should be noted that while `notwww.awk` preprocessing step did exclude litmus tests with `ASL` architecture, it did _not_ (AFAICT) exclude AArch64 tests with ASL variant, aka AArch64+ASL. So if we wanted this PR to merely preserve the behaviour induced by `notwww.awk`, it would be sufficient to instrument `jerd.ml` to reject tests with `ASL` architecture. That said, as Luc pointed out, jerd is currently not set up to properly run AArch64+ASL tests, so I think it's sensible to also reject those with this PR. This is now done in [5961319](https://github.com/herd/herdtools7/pull/1933/commits/5961319fd92374f279f4a03b3b03d31ebf60f8bc).

fsestini:
Merged, thanks all for the reviews!

PR #1934, opened 2026-07-29
[github] Update action versions

psafont:
There are 3 actions in use by the github workflow, the setup-ocaml ones, and two from github ones. While the ocaml one keept updating the versions with the latest node versions, this is not the case for the gihub ones, which consider it a breaking change. This means that in order to use the latest node version, the versions need to be bumped.

Github also decided to deprecate node 20 late last year and will remove the runtime in the coming months. Since the action versions we use only support node 20, we're forced to update the github actions before the runtime is removed.


fsestini (approved):
This looks sensible to me.

PR #1935, opened 2026-07-29
[aslref] Rework ASL error codes and error reporting

Roman-Manevich:
This PR organizes and standardizes the API for creating, classifying, accessing, and printing ASL errors.

Key changes:
- Organizes specified errors into build, typing, and dynamic categories.
- Maps previously missing or incorrect error codes to the existing ASL error-code taxonomy; no new codes are introduced.
- Replaces over-general errors with structured constructors that retain useful details such as expected and actual arities, operators, expressions, indexes, and bounds.
- Aligns `aslref` and `asl.spec` where they model the same behavior, including slice validation, type satisfaction, overriding, collection access, and global-initialization failures.
- Separates errors that intentionally have no ASL code: implementation limitations, internal invariant failures, ASLv0-only behavior, and failures caused by skipping type checking.
- Preserves source locations through backend operations where available.
- Reorganizes error-code mappings by category and alphabetically within each category, making missing or duplicate mappings easier to identify.
- Standardizes diagnostic wording, capitalization, punctuation, and expected/provided terminology.
- Updates generated documentation and test expectations accordingly.

Specified errors are printed consistently as:
`ASL <error-kind> error (<error-code>): <diagnostic-message>`

Errors without an ASL code use:
`ASL <error-kind> error: <diagnostic-message>`

Cleanup:
- Removed the nonexistent `SemanticsRule.SThrowSTry.asl` entry from `tests/ASLSemanticsReference.t/run.t`. The file does not appear in Git history.
- Removed `SemanticsRule.LEUndefIdentV1.asl` from the semantics-reference tests. It duplicated typing-rule coverage and tests `aslref` behavior when type checking is skipped rather than a rule in the

Co-authored-by: Hrutvik Kanabar <hrutvik.kanabar2@arm.com>
Co-authored-by: Roman-Manevich <Roman.Manevich@arm.com>

This PR was AI-assisted.


asllib/error.ml
```diff
@@ -307,8 +499,14 @@ module PPrint = struct
...
-  let fprintf_err f kind =
-    kdprintf (fun msg -> fprintf f "@[<hov 2>ASL %s error:@ %t@]" kind msg)
+  let fprintf_err f kind code_opt =
+    let code_string =
+      match code_opt with
+      | None -> ""
+      | Some code -> " (" ^ ErrorCode.to_string code ^ ")"
+    in
+    kdprintf (fun msg ->
+        fprintf f "@[<hov 2>ASL %s error%s:@ %t@]" kind code_string msg)
```
HadrienRenaud:
```suggestion
    let pp_code fmt code = fprintf fmt " (%s)" (ErrorCode.to_string code) in
    kdprintf (fun msg ->
        fprintf f "@[<hov 2>ASL %s error%a:@ %t@]" kind (pp_print_option pp_code) code_opt msg)

File: asllib/error.ml [hunk 2]
```diff
@@ -498,23 +498,30 @@ module PPrint = struct
...
       | None -> ""
       | Some code -> " (" ^ ErrorCode.to_string code ^ ")"
     in
+    let () =
+      match code_opt with
+      | Some (Typing _) -> assert (kind = typing || kind = static)
+      | Some (Build _) ->
+          assert (kind = lexical || kind = parse || kind = static)
+      | Some (Dynamic _) -> assert (kind = dynamic)
+      | None -> ()
```
HadrienRenaud, line 603: Could we do this check on the abstract types (i.e. `error_handling_time`) and not on their string representation? Seems britle

asllib/tests/ASLSemanticsReference.t/run.t
```diff
@@ -160,11 +160,11 @@ ASL Semantics Tests:
   $ aslref SemanticsRule.PMask.asl
   $ aslref SemanticsRule.ATCValue.asl
   $ aslref -0 SemanticsRule.LEUndefIdentV0.asl
-  $ aslref SemanticsRule.LEUndefIdentV1.asl
+  $ aslref --no-type-check SemanticsRule.LEUndefIdentV1.asl
```
HadrienRenaud, line 170: Can we also run it with type-check just for completeness?

asllib/error.ml
```diff
@@ -292,7 +292,7 @@ module ErrorCode = struct
     | MismatchType _
     (* dynamic ATC but also mismatched integers for loop limits *) ->
         None
-    | CannotParse _ (* used in lexing too *) -> None
+    | CannotParse _ (* used in lexing too *) -> Some (Build PE)
```
HadrienRenaud: Move from tidy up section to normal then?

HadrienRenaud (commented):
This seems like a good PR.

I'm wondering if the `error_handling_time`/`kind` duality in `error.ml` could be made a little bit cleaner, although I could be satisfied with it being left for another PR.

asllib/backend.mli
```diff
@@ -124,11 +124,11 @@ module type S = sig
   val create_exception : (AST.identifier * value) list -> value m
   (** Creates an exception, with the indicated names. *)
 
-  val get_index : int -> value -> value m
-  (** [get_i i vec] returns value at index [i] inside [vec].*)
+  val get_index : loc:_ AST.t_annotated -> int -> value -> value m
```
HadrienRenaud, line 127: No need for type annotations here no?

```suggestion
val get_index : loc:_ AST.annotated -> int -> value -> value m
```

(similarly in the rest of the file)

herd/ASLSem.ml
```diff
@@ -515,8 +515,11 @@ module Make (Conf : Config) = struct
       | V.Val (Constant.Frozen i) -> return (V.Var i)
       | v -> return v
 
-    let get_index i v = M.op1 (Op.ArchOp1 (ASLOp.GetIndex i)) v >>= unfreeze
-    let set_index i v vec = M.op (Op.ArchOp (ASLOp.SetIndex i)) vec (freeze v)
+    let get_index ~loc:_ i v =
+      M.op1 (Op.ArchOp1 (ASLOp.GetIndex i)) v >>= unfreeze
```
HadrienRenaud, line 519: A todo for me later: add a way to use those position annotations in errors given here

asllib/Native.ml
```diff
@@ -172,20 +172,20 @@ module NativeBackend (C : Config) = struct
...
       List.map
         (fun (start, length) ->
           Slice_Length (expr_of_int start, expr_of_int length))
         positions
     in
-    Error.(fatal_unknown_pos (BadSlices (C.error_handling_time, slices, 0)))
+    Error.fatal_from loc (Error.BadSlices (C.error_handling_time, slices, 0))
 
-  let slices_to_positions positions =
+  let slices_to_positions ~loc positions =
```
HadrienRenaud, line 199: Hum looks like the change to `Backend.mli` went into the wrong commit? Ok only if we squash the commits later

asllib/doc/asl.spec
```diff
@@ -9148,9 +9148,17 @@ semantics relation build_genv(tenv: static_envs, typed_spec: spec) -> (new_env:
...
-  --
-  (new_env, new_g);
+  case normal {
+    eval_globals(typed_spec, (env, empty_graph)) -> (new_env, new_g) | DynErrorConfig(), DivergingConfig();
+    --
+    (new_env, new_g);
+  }
+
+  case throwing {
+    eval_globals(typed_spec, (env, empty_graph)) -> Throwing(_, _, _, _) | DynErrorConfig(), DivergingConfig();
```
HadrienRenaud, line 9183: Shouldn't this error (missing possible config handling) be caught by the type-checker of `spec`?

asllib/Typing.ml
```diff
@@ -1972,7 +1972,7 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
           best_effort t_true (fun _ ->
               match Types.lowest_common_ancestor ~loc:e env t_true t_false with
               | None ->
-                  fatal_from ~loc (Error.UnreconcilableTypes (t_true, t_false))
```
HadrienRenaud, line 1980: Is `UnreconcilableTypes` used anywhere else? What is the difference between this and `NoCommonAncestor`?

Also question: why not throw this in `Types.lowest_common_ancestor`? This would probably simplify quite a bit the formalism. Maybe for another PR unless it's simple?

asllib/Typing.ml
```diff
@@ -3378,8 +3378,7 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
       | E_Cond (e, e1, e2) ->
           parameters_of_expr ~env e @ parameters_of_expr ~env e1
           @ parameters_of_expr ~env e2
-      | E_Tuple _ | _ ->
-          Error.fatal_from (to_pos e) (Error.UnsupportedExpr (Static, e))
+      | E_Tuple _ | _ -> Error.fatal_from (to_pos e) (Error.BadParameterExpr e)
```
HadrienRenaud: This case is so strange: why `E_Tuple _ | _`? why not simply `_`?

asllib/Typing.ml
HadrienRenaud: Overridden in a future commit

asllib/error.ml
HadrienRenaud, line 1: nit: commit b22f149 should have been done while doing the other commits. Ok for merging if the commits are squashed.

asllib/Typing.ml
```diff
@@ -2510,11 +2508,8 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
           | T_Tuple tys ->
               if List.compare_lengths tys les != 0 then
                 Error.fatal_from le
-                  (Error.BadArity
-                     ( Static,
-                       "LEDestructuring",
-                       List.length tys,
-                       List.length les ))
+                  (Error.BadTupleArity
+                     ("LEDestructuring", List.length tys, List.length les))
```
HadrienRenaud: nit: `LEDestructuring` is not understandable. Why not use `Arity error while unpacking a tuple` or something similar?

asllib/Typing.ml
```diff
@@ -2783,11 +2778,8 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
     | T_Tuple lhs_tys, T_Tuple rhs_tys ->
         if List.compare_lengths lhs_tys rhs_tys != 0 then
           fatal_from ~loc
-            (Error.BadArity
-               ( Static,
-                 "tuple initialization",
-                 List.length rhs_tys,
-                 List.length lhs_tys ))
+            (Error.BadTupleArity
+               ("tuple initialization", List.length rhs_tys, List.length lhs_tys))
```
HadrienRenaud: idem: give a complete sentence instead of a keyword

asllib/Typing.ml
```diff
@@ -3386,7 +3386,7 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
       | T_Int UnConstrained
       | T_Real | T_String | T_Bool | T_Array _ | T_Named _ ->
           []
-      | _ -> Error.fatal_from (to_pos ty) (Error.UnsupportedTy (Static, ty))
+      | _ -> Error.fatal_from (to_pos ty) (Error.BadParameterType ty)
```
HadrienRenaud, line 3417: Can we even trigger this error? Is it just something for warning mode? Can we add a test for it?

asllib/StaticInterpreter.mli
```diff
@@ -30,8 +30,8 @@ val static_eval : StaticEnv.env -> AST.expr -> AST.literal
       if the a type error is detected or the expression is not one of the
       following: [E_Literal], [E_Var], [E_Binop], [E_Unop], [E_Slice], or
       [E_Cond].
-    @raise UnsupportedExpr if the given expression cannot evaluate to a literal.
-*)
+    @raise StaticEvaluationFailure
+      if the given expression cannot evaluate to a literal. *)
```
HadrienRenaud:
```suggestion
      if the given expression cannot evaluate to a literal.
    @raise other static exceptions if they are triggered by the static interpreter, such as [RecursionLimitReached (Static, _)]
*)
```

asllib/tests/ASLSemanticsReference.t/run.t
```diff
@@ -160,11 +160,11 @@ ASL Semantics Tests:
   $ aslref SemanticsRule.PMask.asl
   $ aslref SemanticsRule.ATCValue.asl
   $ aslref -0 SemanticsRule.LEUndefIdentV0.asl
```
hrutvik: I'm a bit surprised a V0 test appears in the reference test directory - should this be moved elsewhere?

asllib/error.ml
hrutvik: Is this right? As per the comment, the lexer also produces some `CannotParse _` errors - are those genuine parse errors, or should they be converted to `Build LE`? In any case, should this line be moved out of the "TODO" section?

asllib/error.ml
```diff
@@ -26,6 +26,20 @@ open AST
...
+
+type bad_declaration =
+  | LocalDeclarationWithoutName
+  | LocalTupleDeclarationWithoutName
+  | GlobalDeclarationWithoutName
+  | LocalConstantDeclaration
+  | EmptyRecordTypeDeclaration
+  | EmptyExceptionTypeDeclaration
+  | EmptyCollectionTypeDeclaration
+  | ElidedParameterWithoutBitvectorType
```
hrutvik, line 117: I'm not convinced by this type - it seems to introduce unnecessary indirection/overhead. I this we should keep the strings for this one

asllib/error.ml
```diff
@@ -56,11 +60,14 @@ type error_desc =
       found_call_type : subprogram_type;
     }
   | BadArity of error_handling_time * identifier * int * int
+  | BadCallArity of identifier * int * int
+  | BadTupleArity of { expected : int; actual : int }
```
hrutvik, line 187: Why name the fields of one but not the other?

asllib/error.ml
```diff
@@ -340,6 +347,25 @@ module ErrorCode = struct
...
+        | _ -> None
+      in
+      fatal_from e (StaticEvaluationFailure (e, reason))
+  | Some
+      (ErrorCode.Dynamic
+         ( ErrorCode.UNR | ErrorCode.DAF | ErrorCode.AET | ErrorCode.BO
+         | ErrorCode.LE | ErrorCode.UE | ErrorCode.OSA | ErrorCode.NAL
+         | ErrorCode.NEP )) ->
+      fatal_from e (StaticEvaluationFailure (e, None))
+  | Some (ErrorCode.Build _ | ErrorCode.Typing _) | None -> fatal cause
```
hrutvik: Can all these happen?

asllib/error.ml
hrutvik: nit: would something like
```
 ErrorCode.(Dynamic
         ( UNR | DAF | AET | BO | LE | UE | OSA | NAL | NEP )) ->
```
work?
Similarly throughout this function.
Also, this is a bit mysterious at first - a function-level docstring would be useful

asllib/error.ml
```diff
@@ -26,6 +26,37 @@ open AST
 
 type error_handling_time = Static | Dynamic
 
+(** Failures that evaluation may encounter when it runs without full type
+    checking. They have no error codes. *)
```
hrutvik, line 30: Do we need all this support for these? We don't really support this mode in general

asllib/error.ml
```diff
@@ -52,24 +92,22 @@ type error_desc =
...
-  | BadArity of error_handling_time * identifier * int * int
   | BadCallArity of identifier * int * int
   | BadTupleArity of { expected : int; actual : int }
-  | BadParameterArity of error_handling_time * version * identifier * int * int
+  | BadParameterArity of version * identifier * int * int
   | UnsupportedBinop of error_handling_time * binop * literal * literal
   | UnsupportedUnop of error_handling_time * unop * literal
-  | UnsupportedExpr of error_handling_time * expr
   | StaticEvaluationFailure of expr * static_evaluation_failure_reason option
+  | ImplementationIntegerOverflow of Z.t
```
hrutvik, line 197: It seems odd that this isn't included with other "internal" style errors

asllib/error.ml
```diff
@@ -322,28 +358,30 @@ module ErrorCode = struct
...
+       information or encounter mismatches that would normally be reported
+       during typing. The specification therefore assigns these failures no
+       dynamic error code. *)
+    | UncheckedExecutionError _ -> None
     | CannotParse _ (* also used for targeted lexer diagnostics *) ->
         Some (Build PE)
-    | EmptyConstraints (* does this need to be reflected in reference? *) ->
-        None
+    (* [WellConstrained] contains a non-empty list in the specification, and
+       the parser prevents source programs from violating that invariant. *)
```
hrutvik: Does this make `EmptyConstraints` an `InternalInvariantError` then?

asllib/error.ml
```diff
@@ -516,14 +550,20 @@ module PPrint = struct
...
+          Z.pp_print z
     | BadParameterType ty -> pp_err typing "Unsupported type %a." pp_ty ty
     | InvalidExpr e -> pp_err typing "invalid expression %a." pp_expr e
-    | MismatchType (v, [ ty ]) ->
-        pp_err dynamic "Mismatch type:@ value %s does not belong to type %a." v
-          pp_type_desc ty
-    | MismatchType (v, li) ->
-        pp_err dynamic
-          "Mismatch type:@ value %s@ does not subtype any of those types:@ %a" v
+    | UncheckedExecutionError (t, TypeMismatch (v, [ ty ])) ->
```
hrutvik: If separating unchecked errors, suggest factoring out their stringifying into a separate function. Ideally, separate all their functionality into a dedicated module.

asllib/error.ml
```diff
@@ -668,7 +739,18 @@ module PPrint = struct
         pp_err typing "multiple recursive declarations:@ @[%a@]."
           (pp_comma_list (fun f -> fprintf f "%S"))
           decls
-    | UnrespectedParserInvariant -> pp_err typing "Parser invariant broke."
+    | InternalInvariantError TypedArrayExpressionInAnnotation ->
```
hrutvik: Similar

asllib/error.ml

asllib/doc/asl.spec
Roman-Manevich: it's not very thorough about what appears after the`|`. I removed the extra `DynErrorConfig()`.

asllib/error.ml
```diff
@@ -321,19 +326,23 @@ module ErrorCode = struct
...
+    | ConflictingTypes (TypeSatisfaction, _, _) -> Some (Typing TSF)
     | MismatchedCallType _ | BadCallArity _ | BadParameterArity _
     | NoCallCandidate _ ->
         Some (Typing BC)
     | UnsupportedUnop (Dynamic, _, _) | UnsupportedBinop (Dynamic, _, _, _) ->
         Some (Dynamic BO)
     | AssertionFailed _ | BadPrimitiveArgument _ -> Some (Dynamic DAF)
-    | ImpureExpression _ | MismatchedPurity _ | ConflictingSideEffects _ ->
+    | ImpureExpression _ | MismatchedPurity _ | ConflictingSideEffects _
+    | ConstantTimeBroken _ ->
```
hrutvik: `ConstantTimeBroken` is an internal-only error right?

asllib/Typing.ml
```diff
@@ -33,8 +33,8 @@ let fatal_from ~loc = Error.fatal_from loc
 let undefined_identifier ~loc x = fatal_from ~loc (Error.UndefinedIdentifier x)
 let add_pos_from ~loc = add_pos_from loc
 
-let conflict ~loc expected provided =
-  fatal_from ~loc (Error.ConflictingTypes (expected, provided))
+let conflict ?(reason = Error.UnexpectedType) ~loc expected provided =
```
hrutvik: Why have this as optional, rather than explicit everywhere?

asllib/error.ml
```diff
@@ -111,7 +115,7 @@ type error_desc =
   | StaticEvaluationFailure of expr * static_evaluation_failure_reason option
   | ImplementationIntegerOverflow of Z.t
   | BadParameterType of ty
-  | InvalidExpr of expr
+  | V0InvalidExpr of expr
```
hrutvik: Why prefix with `V0` here when for other errors you have factored out a separate type?

asllib/error.ml
```diff
@@ -580,24 +586,26 @@ module PPrint = struct
...
         pp_err static "Cannot extract from bitvector of length %d slice %a."
           length pp_slice_list slices
     | BadIndex (index, length) ->
         pp_err dynamic "%a" pp_bad_index (index, length)
     | BadTupleIndex (index, length) ->
         pp_err typing "tuple index %d is outside the valid range 0..%d." index
           (length - 1)
-    | BadSlice slice -> pp_err static "invalid slice %a." pp_slice slice
+    | BadSlices (NonPositiveLength slice) ->
+        pp_err static "invalid slice %a." pp_slice slice
```
hrutvik: Could add more information here?

asllib/Typing.ml
Roman-Manevich: Fixed by changing to `Error.NoCommonAncestor` and removing `UnreconcilableTypes`.

asllib/doc/ErrorCodes.tex
```diff
@@ -37,8 +37,8 @@ \section{How Implementations Should Handle Errors\label{sec:How Implementations
   characters 11 to 22:
     return 5 + "hello";
            ^^^^^^^^^^^
-ASL Type error: Illegal application of operator + on types integer {5}
-  and string.
+ASL Type error (TE_BO): Illegal application of operator + on types
+  integer {5} and string.
```
hrutvik: Not sure I understand why this commit would affect these

asllib/Typing.ml
Roman-Manevich: I moved the exception throwing to `Types.lowest_common_ancestor`. 
This does change the error message, since it can now refer to a failure in a nested element (tuple component, record field). 

Also, since `lowest_common_ancestor` unfolds named types, the error may refer to an anoymous type, which is the case for `TypingRule.LowestCommonAncestor.bad.asl`.
Instead of ` Types array [[8]] of Word1 and Exc have no common ancestor.`, we now get `Types array [[8]] of Word1 and exception { i: integer } have no common ancestor.`.
I think this is fine as the error is more localised now. What do you think?

asllib/Typing.ml
Roman-Manevich: agree. removed the tuple case.

asllib/Typing.ml
Roman-Manevich: Yes, `TypingRule.ParametersOfExpr.bad.asl` exercises it.

asllib/Typing.ml
Roman-Manevich: This has been refactored now and the string is entirely in `Error.ml`.

asllib/Typing.ml
Roman-Manevich: This has been refactored, which removed the string altogether.

asllib/Native.ml
Roman-Manevich: right. I'm okay with squashing.

asllib/error.ml
Roman-Manevich: done. replaced with a variant type where each phase corresponds to a constructor.

asllib/error.ml
Roman-Manevich: The lexer (both V0 and V1) now use `UnknownSymbol` and generate an `LE` error.
There are no more todos.

asllib/error.ml
Roman-Manevich: changed to a record.

asllib/error.ml
Roman-Manevich: Some are definitely reachable (`UNR`, `DAF`, and `LE`). The others probably aren't (that's an interesting question about our current semantics).
I wrote it like this to make sure the matching is exhaustive.

asllib/error.ml
Roman-Manevich: We can decide and change in another PR.

asllib/error.ml
Roman-Manevich: The advantage of this approach is that all of text generated for error messages lives in exactly one place.
This is consistent with all of the other type categories.
Besides, the overhead is very small, it's just a constructor per error; there are no helper functions.

asllib/error.ml
Roman-Manevich: All of the other internal errors correspond to invariants, which we expec to always hold. This one has to do with our implementation limitations. We could move it together with other implementation-specific errors, but right now this is the only one we (partially) detect.

asllib/error.ml
Roman-Manevich: Yes. Changed to `InternalInvariantError EmptyWellConstrainedInteger`

asllib/tests/ASLSemanticsReference.t/run.t
Roman-Manevich: moved to `regressions.t`

asllib/error.ml
Roman-Manevich: okay, done.

asllib/error.ml
Roman-Manevich: okay, done.

asllib/error.ml
Roman-Manevich: No. It can be triggered with fine-grained side-effects, for example, `TypingRule.DeclareGlobalStorage.bad1.asl`.

asllib/Typing.ml
Roman-Manevich: `conflict` now doesn't take the optional parameter and we use `type_satisfaction_conflict` in cases where it's currently needed.

asllib/error.ml
Roman-Manevich: I moved most V0 errors into a new type
```
type v0_error =
  | EmptySlice
  | InvalidExpr of expr
  | ParameterWithoutDecl of identifier
  | SetterWithoutCorrespondingGetter of func
 ```
 The exception is `V0SetterWithoutValueArgument`, which is kept under `InternalInvariantError`.

File: asllib/error.ml (hunk 27 above)
Comment (Roman-Manevich) [reply asllib/error.ml]: changed to `NonPositiveLength of { slice : slice; length : int }` to provide more specific information.

File: asllib/doc/ErrorCodes.tex (hunk 28 above)
Comment (Roman-Manevich) [reply asllib/doc/ErrorCodes.tex]: We're still splicing console outputs rather using macros as per https://github.com/herd/herdtools7/pull/1919. Regenerating and splicing the console output reflected the changes to the error printing.

File: asllib/backend.mli (hunk 5 above)
Comment (Roman-Manevich) [reply asllib/backend.mli:127]: That breaks compilation: `AST.annotated` fixes `ty_opt` to `no_type_annotation`, but these functions also receive typed expressions. `_ AST.t_annotated` accepts both typed and untyped nodes.

File: asllib/Operations.ml
Comment (HadrienRenaud) [inline asllib/Operations.ml:1]: Maybe good to put a `Operations.mli` to avoid such unused functions being kept?

File: asllib/error.ml (hunk 16 above)
Comment (HadrienRenaud) [reply asllib/error.ml:117]: Hi @Roman-Manevich ,  it seems to me that you still have this type here, can you comment on why you're keeping it please?

File: asllib/error.ml [hunk 29]
```diff
@@ -26,70 +26,225 @@ open AST
...
+    | UninitialisedImmutableLocal ->
+        Format.fprintf f "An immutable local declaration has no initialiser."
+    | V0SetterWithoutValueArgument ->
+        Format.fprintf f "A setter has no value argument."
+    | GlobalWithoutTypeOrInitialiser ->
+        Format.fprintf f
+          "A global declaration has neither a type nor an initialiser."
+    | ParameterizedIntegerAtRuntime ->
+        Format.fprintf f
+          "A parameterized integer type reached dynamic evaluation."
```
HadrienRenaud: Refacto to `to_string` and then use `pp_print_text`

asllib/error.ml
```diff
@@ -26,70 +26,225 @@ open AST
...
+end
+
+(** Whether conflicting types arose from a general structural expectation or the
+    specification's type-satisfaction check. *)
+type conflicting_types_reason = UnexpectedType | TypeSatisfaction
```
HadrienRenaud: Is there a good reason why we couldn't have 2 variants in the first place?

asllib/error.ml
```diff
@@ -26,70 +26,225 @@ open AST
...
   | AssignToTupleElement of lexpr
   | AlreadyDeclaredIdentifier of string
   | BadReturnStmt of ty option
-  | UnexpectedSideEffect of string
   | UncaughtException of string
   | OverlappingSlices of slice list * error_handling_time
   | BadLDI of AST.local_decl_item
   | BadRecursiveDecls of identifier list
-  | UnrespectedParserInvariant
+  | InternalInvariantError of internal_invariant_error
```
HadrienRenaud, line 222: Why do we need this? I really don't understand why we're making this PR bigger than what it already is

asllib/error.ml
```diff
@@ -26,70 +26,225 @@ open AST
...
+  let label = function
+    | EmptySlice -> "V0EmptySlice"
+    | InvalidExpr _ -> "V0InvalidExpr"
+    | ParameterWithoutDecl _ -> "V0ParameterWithoutDecl"
+    | SetterWithoutCorrespondingGetter _ -> "V0SetterWithoutCorrespondingGetter"
+end
+
+(** Violations of invariants between internal processing stages. They are not
+    ASL errors and therefore have no error codes. *)
+type internal_invariant_error =
```
HadrienRenaud, line 138: Same comment as @hrutvik and on the `bad declaration` type: why was this needed here? Especially, you're adding quite a bit of more details to error codes that the normal user will never see

asllib/error.ml
```diff
@@ -26,70 +26,225 @@ open AST
...
+  | GlobalDeclarationWithoutName
+  | LocalConstantDeclaration
+  | EmptyRecordTypeDeclaration
+  | EmptyExceptionTypeDeclaration
+  | EmptyCollectionTypeDeclaration
+  | ElidedParameterWithoutBitvectorType
+
+type static_evaluation_failure_reason =
+  | IndexOutOfBounds of int * int
+  | ValueOutsideAssertedType of string * type_desc
```
HadrienRenaud: Are there not runtime errors that could be used to represent this with a correct `error_handling_time`?

asllib/error.ml
```diff
@@ -150,92 +311,218 @@ type warning_desc =
...
+        | BadIndex (index, length) -> Some (IndexOutOfBounds (index, length))
+        | DynamicATCFailure (value, ty) ->
+            Some (ValueOutsideAssertedType (value, ty))
+        | _ -> None
+      in
+      fatal_from e (StaticEvaluationFailure (e, reason))
+  | Some ErrorCode.(Dynamic (UNR | DAF | LE | AET | BO | UE | OSA | NAL | NEP))
+    ->
+      fatal_from e (StaticEvaluationFailure (e, None))
+  | Some ErrorCode.(Build _ | Typing _) | None -> fatal cause
```
HadrienRenaud: Can we just not match on those in the first place? This breaks the stack trace

This actually breaks my debugging flow, so I'm going to request changes for this

asllib/StaticInterpreter.ml
```diff
@@ -59,11 +59,13 @@ let static_eval (senv : SEnv.env) (e : expr) : literal =
...
-      l |: Instrumentation.TypingRule.StaticEval
-  | SI.Normal _ | SI.Throwing _ | SI.Cutoff ->
-      Error.fatal_from e (UnsupportedExpr (Static, e))
+  try
+    match eval_from ~loc:e env e with
+    | SI.Normal (Native.NV_Literal l, _env) ->
+        l |: Instrumentation.TypingRule.StaticEval
+    | SI.Normal _ | SI.Throwing _ | SI.Cutoff ->
+        Error.fatal_from e (StaticEvaluationFailure (e, None))
+  with Error.ASLException cause -> Error.fatal_from_static_evaluation e cause
```
HadrienRenaud: As said in another comment, I think this breaks my debugging flow, so I'm going to request change here.

If possible I would like to see here, or at least justified why it's not possible:
1. Inlining (or at least relocate) `fatal_from_static_evaluation` in this file
2. [IMPORTANT] only match on the exception that you want to effectively edit, why do you need to catch all `ASLException`?
3. Why do you even need to edit those exceptions? Isn't the `error_handling_time` enough to discriminate them from the other times they can be raised?
4. Add a way to disable this behaviour when `OCAMLRUNPARAM=b` is set, i.e. when `Printexc.backtrace_status()` is `true`

asllib/Typing.ml
```diff
@@ -1375,9 +1378,10 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
         | T_Collection _ ->
             assert (not decl);
             let+ () =
-              check_true
-                (List.for_all (fun (_, t) -> has_structure_bits env t) fields)
-              @@ fun () -> fatal_from ~loc Error.(UnsupportedTy (Static, ty))
+             fun () ->
+              List.iter
+                (fun (_, ty) -> check_structure_bits ~loc:ty env ty ())
+                fields'
```
HadrienRenaud: Introduce `check_all` as:
```ocaml
  let check_all li f () = List.iter (fun x1 -> f x1 ()) li
```

and then use it like so:
```ocaml
            let+ () =
              check_all fields' @@ fun (_, ty) ->
              check_structure_bits ~loc:ty env ty
            in
```

HadrienRenaud (changes requested):
Thanks for all the work on this @Roman-Manevich. I fear that this PR has turned into a beast.

If possible I would suggest opening smaller PRs for small parts that can be extracted. For example, I think the change to `lowest_common_ancestor` could be a separate PR.

I also don't really understand where the need for refacto the whole structure of errors came from, and if possible in the future I would like us to spend a bit of time to try and isolate them as much as possible from the rest of the code changes.
In particular, the item _"Standardizes diagnostic wording, capitalization, punctuation, and expected/provided terminology"_ was a non-goal for me at the start of this PR. In particular, some of the standardised errors are very specific to non-standard uses of the tool, so where very low priority to me.

I'm requesting changes because the change to `StaticInterpreter` is breaking a behaviour that I rely on to debug: the presence of backtraces. It shouldn't be too much of a change to revert/adapt.

I have not reviewed the changes to the reference nor the association between error codes in the implementation and in the reference.

asllib/Operations.ml
Roman-Manevich: Let's do this in a future PR.

asllib/error.ml
Roman-Manevich: We're matching those because we want to emit the right error codes. See my reply to the next comment.

asllib/StaticInterpreter.ml
Roman-Manevich: The conversion is still required because the specification maps any dynamic error during static evaluation to `TE_SEF`, and not every error carries `error_handling_time`. Specifically, `BadIndex`, `DynamicATCFailure`, `LoopLimitReacged`, `NegativeArrayLength`, `and BadPrimitiveArgument`.

 I’ll move `fatal_from_static_evaluation` here, restrict it to dynamically classified errors, and bypass both rewrites when backtraces are enabled.

asllib/error.ml
Roman-Manevich: I kept this type because these form a closed group of related declaration errors. 
The variants make the complete set and its organization visible in one place.
I think that clarity justifies the small indirection.

asllib/error.ml
Roman-Manevich: I've split them into two separate errors: `UnexpectedType` and `TypeSatisfactionFailure`.

asllib/error.ml
Roman-Manevich: `UnrespectedParserInvariant` was a vague error used for several unrelated internal failures, many unrelated to parsing. This groups them and gives useful diagnostics.
This PR intentionally attempts a one-time comprehensive reorganization of the errors, including classifying existing errors correctly.

asllib/error.ml
Roman-Manevich: This isn't necessarily about what users will see. The idea is to group and classify errors so we can get a handle on them.

asllib/error.ml
Roman-Manevich: Yes, we can probably refactor to achieve this.

asllib/Typing.ml
Roman-Manevich: done

asllib/error.ml
HadrienRenaud: Un resolving because this does not seem to have been addressed

Roman-Manevich: I forgot to push the changes. It's done.

asllib/backend.mli
HadrienRenaud: Good enough thanks

asllib/doc/asl.spec
HadrienRenaud: Maybe worth keeping in a list of improvements for aslspec? It's always interesting to have a list of previous bugs when you want to justify strengthening your type system.

asllib/error.ml
HadrienRenaud: FYI I still can't see the changes

asllib/error.ml
HadrienRenaud: Ok, but no need for refactoring their error messages then no?

asllib/Operations.ml
HadrienRenaud: Fair enough, can you keep track of it please?

asllib/error.ml
Roman-Manevich: It doesn't hurt to have nice error messages, until we remove (or not).

asllib/error.ml
HadrienRenaud: > This PR intentionally attempts a one-time comprehensive reorganization of the errors, including classifying existing errors correctly.

Please next time split this work into 2: the work on error codes on one side and the refacto of the errors on another side.

asllib/error.ml
HadrienRenaud: Please include this in this PR

HadrienRenaud (commented):
Changes requested have been changed. I don't have time to review this PR before Tuesday but I wanted to remove the Changes requested flag.

HadrienRenaud (approved):
Changes requested have been changed. However, I did not re-review the PR.

asllib/Operations.ml
Roman-Manevich: opened a ticket.

asllib/doc/asl.spec
Roman-Manevich: done.

asllib/error.ml
Roman-Manevich: It should be there now:
`BadCallArity { name; expected; provided }` and `BadTupleArity { expected; actual }`.

asllib/error.ml
Roman-Manevich: I made `BadIndex` and `ATCFailure` carry `error_handling_time` and used them both for static evaluation and dynamic evaluation. Static cases yield `TE_SEF`; dynamic yield `DE_BI` and `DE_TAF`. I removed the static failure-reason constructors.

asllib/error.ml
hrutvik: I don't agree that this is clearer, and being "visible in one place" is a non-goal. These types seem to be placeholders for the strings they really represent - they seem to be constructed in a single place, and used in a single place to produce the string. We should just use the strings directly.

asllib/error.ml
hrutvik: The question is really whether this should be part of this PR.

asllib/error.ml
hrutvik: I find this rather unclear - you've spent a lot of time separating language errors from implementation-specific errors, but this is an implementation-specific error

asllib/error.ml
hrutvik: That is a non-goal

asllib/error.ml
hrutvik: I'm sorry, I don't follow. Fine-grained side-effects are not part of the language definition, and `ConstantTimeBroken` is only raised when fine-grained side-effects are enabled. This makes `ConstantTimeBroken` an implementation-specific error, so it should not be causing a language error code like `TE_SEV`. I'm therefore not sure why this comment was resolved.

asllib/error.ml
```diff
@@ -271,7 +289,8 @@ module ErrorCode = struct
     | UnsupportedUnop (Dynamic, _, _) | UnsupportedBinop (Dynamic, _, _, _) ->
         Some (Dynamic BO)
     | AssertionFailed _ | BadPrimitiveArgument _ -> Some (Dynamic DAF)
-    | ImpureExpression _ | MismatchedPurity _ -> Some (Typing SEV)
+    | ImpureExpression _ | MismatchedPurity _ | ConflictingSideEffects _ ->
```
hrutvik, line 292: Similarly to `ConstantTimeBroken`, `ConflictingSideEffects` is not an ASL1 error.

asllib/error.ml
```diff
@@ -305,10 +324,7 @@ module ErrorCode = struct
     | CannotParse _ (* used in lexing too *) -> Some (Build PE)
     | EmptyConstraints (* does this need to be reflected in reference? *) ->
         None
-    | MultipleWrites _
-    (* For desugaring, but uses `check_no_duplicates` which is always TE_IAD? *)
```
hrutvik, line 309: I can't see that this comment has been addressed - in particular:
- documentation: `desugar_lhs_fields_tuple` uses `check_no_duplicates`, which always produces `TE_IAD`
- implementation: `desugar_lhs_fields_tuple` produces `BE_PE` via `MultipleWrites`

asllib/error.ml
```diff
@@ -319,12 +323,14 @@ module ErrorCode = struct
     | UnsupportedUnop (Static, _, _)
     | UnsupportedBinop (Static, _, _, _) ->
         Some (Typing BO)
-    | BadSlices (Static, _, _)
-    | BadSlice _ | EmptySlice
+    | BadSlices (NonPositiveLength _)
+    | BadSlices (OutOfBitvectorBounds _)
+    | BadSlices (NegativeStartOrLength (Static, _))
```
hrutvik, line 328: Combine these patterns?

asllib/error.ml
```diff
@@ -553,28 +553,28 @@ module PPrint = struct
...
     | UnsupportedUnop (t, op, v) ->
         pp_err
           (error_handling_time_to_string t)
-          "Illegal application of operator %s for value@ %a."
-          (unop_to_string op) pp_literal v
+          "Operator %s is not defined for value@ %a." (unop_to_string op)
+          pp_literal v
     | StaticEvaluationFailure (e, None) ->
-        pp_err typing "static evaluation failed for expression %a." pp_expr e
+        pp_err typing "Static evaluation of expression %a failed." pp_expr e
```
hrutvik, line 564: If you've grouped these under the same constructor, that constructor could be used to remove duplication of "Static evaluation of ..." in the code?

asllib/error.ml
```diff
@@ -604,62 +602,59 @@ module PPrint = struct
...
+          "Call to %S has incorrect argument arity:@ expected %d argument(s); \
+           provided %d."
+          name expected provided
     | UncheckedExecutionError
         (t, ParameterArityMismatch { name; expected; provided }) ->
         pp_err
           (error_handling_time_to_string t)
-          "Arity error while calling '%s':@ %d %s expected and %d provided."
-          name expected
-          (if expected = 1 then "parameter" else "parameters")
```
hrutvik, line 640: Why remove this if improving error messages?

asllib/error.ml
```diff
@@ -189,6 +189,11 @@ exception ASLException of error
 
 type 'a result = ('a, error) Result.t
 
+let unknown_symbol id = UnknownSymbol (id, None)
+
+let unknown_symbol_with_alternative ~lexeme ~alternative_lexeme =
+  UnknownSymbol (lexeme, Some alternative_lexeme)
```
hrutvik, line 195: `let unknown_symbol ?alternative ~lexeme = ...`?

asllib/error.ml
```diff
@@ -703,15 +708,24 @@ module PPrint = struct
...
+        in
         let codes = List.map Char.code (List.of_seq (String.to_seq s)) in
         let not_printable code = code < 33 || code > 126 in
-        if String.length s = 0 then pp_err lexical "Unknown symbol."
+        if String.length s = 0 then
+          pp_err lexical "Unknown symbol.%a" pp_alternative_symbol
+            alternative_symbol_opt
         else if List.exists not_printable codes then
-          pp_err lexical "Unknown symbol with byte value(s): %a."
+          pp_err lexical "Unknown symbol with byte value(s): %a.%a"
```
hrutvik, line 723: I've realised this was converted to `byte` rather than `ASCII` - this is a regression, as `Char.code` does in fact return `ASCII` codes

asllib/types.mli
```diff
@@ -80,8 +80,9 @@ val type_clashes : env -> ty -> ty -> bool
 val subprogram_clashes : env -> func -> func -> bool
 (** Subprogram clashing relation. *)
 
-val lowest_common_ancestor : loc:_ t_annotated -> env -> ty -> ty -> ty option
-(** Lowest common ancestor. *)
+val lowest_common_ancestor : loc:_ t_annotated -> env -> ty -> ty -> ty
+(** Lowest common ancestor. Raises [Error.ASLException] carrying
```
hrutvik, line 84: Use `@raise` syntax?

asllib/Operations.ml
HadrienRenaud: Actually done in #1958

asllib/StaticInterpreter.ml
```diff
@@ -59,11 +67,17 @@ let static_eval (senv : SEnv.env) (e : expr) : literal =
...
+    match eval_from ~loc:e env e with
+    | SI.Normal (Native.NV_Literal l, _env) ->
+        l |: Instrumentation.TypingRule.StaticEval
+    | SI.Normal _ | SI.Throwing _ | SI.Cutoff ->
+        Error.fatal_from e (StaticEvaluationFailure e)
+  with
+  | Error.ASLException cause
+  when (not (Printexc.backtrace_status ())) && is_dynamic_error cause
+  ->
+    Error.fatal_from e (StaticEvaluationFailure e)
```
HadrienRenaud, line 80: Use `exception _` matching pattern to simplify `try match`:

```suggestion
  match eval_from ~loc:e env e with
  | SI.Normal (Native.NV_Literal l, _env) ->
      l |: Instrumentation.TypingRule.StaticEval
  | SI.Normal _ | SI.Throwing _ | SI.Cutoff ->
      Error.fatal_from e (StaticEvaluationFailure e)
  | exception (Error.ASLException cause)
  when (not (Printexc.backtrace_status ())) && is_dynamic_error cause
  ->
    Error.fatal_from e (StaticEvaluationFailure e)
```

asllib/StaticInterpreter.ml
```diff
@@ -44,9 +44,17 @@ module SI = Interpreter.Make (Native.StaticBackend) (InterpConf)
...
+  with
+  | Error.(ASLException exn)
+  when (not (Printexc.backtrace_status ())) && is_dummy_pos exn
+  ->
     Error.fatal_from loc exn.desc
 
+let is_dynamic_error cause =
+  match Error.ErrorCode.of_error cause with
+  | Some (Error.ErrorCode.Dynamic _) -> true
+  | Some Error.ErrorCode.(Build _ | Typing _) | None -> false
```
HadrienRenaud, line 56:
```suggestion
  let open Error.ErrorCode in
  match of_error cause with
  | Some (Dynamic _) -> true
  | Some (Build _ | Typing _) | None -> false
```

asllib/Native.ml
```diff
@@ -411,7 +415,9 @@ let rec unknown_of_aggregate_type unknown_of_singular_type ~eval_expr_sef ty =
       |> fun record -> NV_Record record
   | T_Enum li -> NV_Literal (L_Label (List.hd li))
   | T_Tuple types -> NV_Vector (List.map (fun t -> unknown_of_type t) types)
-  | T_Collection _ | T_Named _ -> Error.(fatal_from ty TypeInferenceNeeded)
+  | T_Collection _ | T_Named _ ->
+      Error.(
+        fatal_from ty (UncheckedExecutionError (Dynamic, TypeInferenceNeeded)))
```
HadrienRenaud, line 420: Can we add a comment justifying why we don't use `C.error_handling_time` here?

asllib/StaticInterpreter.ml
HadrienRenaud: Re-opening to say that the following should be addressed by a future PR, for the moment this is good enough I think

> The conversion is still required because the specification maps any dynamic error during static evaluation to `TE_SEF`, and not every error carries `error_handling_time`. Specifically, `BadIndex`, `DynamicATCFailure`, `LoopLimitReacged`, `NegativeArrayLength`, `and BadPrimitiveArgument`.

PR #1937, opened 2026-07-30
[catalogue] Fix aarch64-readers-guide test `MP+rel+acq.litmus`

fsestini:
This PR fixes a test in the `aarch64-readers-guide` catalogue that previously failed to run:

  ```sh
  $ herd7 catalogue/aarch64-readers-guide/tests/MP+rel+acq.litmus
  Size mismatch (eeid=f action=W[x]=0) vs. (eeid=d action=R[x]q=0), ie word vs. quad
  Warning: File "catalogue/aarch64-readers-guide/tests/MP+rel+acq.litmus": Illegal mixed-size test (User error)
  ```

  Relatedly, it also:

  - Adds the `aarch64-readers-guide` catalogue to the regression suite invoked by `cata-test`. The standalone
  target, `make test.herd.cata.aarch64-readers-guide`, takes about half a second on my machine, so the CI
  overhead should be negligible.
  - Fixes `internal/herd_catalogue_regression_test.ml` to return a non-zero exit status when `herd7` reports a
  fatal error. Previously, the regression runner printed the error but exited successfully, so `make
  test.herd.cata.aarch64-readers-guide` also exited with status 0 and CI could not detect the failure.


catalogue/aarch64-readers-guide/tests/MP+rel+acq.litmus
```diff
@@ -1,5 +1,6 @@
 AArch64 MP+rel+acq
```
relokin, line 1: The challenge with this then becomes that we have two tests with the same name in the catalogue and different hashes.

relokin (commented):
One comment otherwise looks good to me.

I think we also have to ensure that the catalogue doesn't replicate the tests that are in https://github.com/herd/herdtools7/tree/master/readers-guide but we can leave this for another PR.

catalogue/aarch64-readers-guide/tests/MP+rel+acq.litmus
fsestini: I agree this could be a problem if at some point we want to reorganise the catalogue tests so that all `.litmus` files are in the same location. Several other tests in `catalogue/aarch64-readers-guide`, like `LB.litmus`, `MP.litmus`, etc., also have slightly different source code (hence different hashes) compared to tests in `catalogue/aarch64-*` with the same name. When that moment comes, we'll need to pick one version of a test and discard the others.

However, as things are now, I don't see the hash difference causing immediate problems, at least when it comes to `make test` and CI checks. herd7 should only compare the hashes of two files if they are both passed as argument to the same herd7 invocation, but as far as I can tell, our current regression test infrastructure runs each litmus test in its own separate herd7 process, so effectively there is no hash comparison taking place. This name/hash difference is brittle and potentially problematic on different workflows, but perhaps not an urgent cause of concern. What do you think?

fsestini:
> I think we also have to ensure that the catalogue doesn't replicate the tests that are in https://github.com/herd/herdtools7/tree/master/readers-guide but we can leave this for another PR.

Agreed, in fact those files should just be symlinks to the catalogue tests.

catalogue/aarch64-readers-guide/tests/MP+rel+acq.litmus
relokin: It's not urgent but if one does
```
$> herd7 catalogue/aarch64/tests/@all catalogue/aarch64-readers-guide/tests/*.litmus
```
herd7 (and other tools) will be unhappy and print warnings.

relokin (approved):
Thanks Filippo let's fix the immediate problem with this PR and plan for another PR to clean up.

fsestini:
Merged, thanks @relokin !

PR #1566, opened 2025-11-13
[ci][asl] Use latest `texlive` image for ASL reference CI

hrutvik:
As suggested by @HadrienRenaud in #1565, an open PR to revert commit 330a11b714403c54e7e02b30a4933a344ef54a2d and go back to using the latest `texlive` image for CI, once it starts working again.

PR #1938, opened 2026-07-30
[internal] Fix parallel promotion for herd tests 

psafont:
When running or promoting litmus tests in parallel, the parallelizer
runs a wrapper, which runs herd7, each of the stages has its own
parameters. Early 2025 some parameters were added to the wrapper that
runs herd7, but these were not added to the promoter, and since then
parameters have accumulated. Unfortunately the handling of these
parameters was not added to the promoter, causing failures to the
promotion when these parameters are present. We therefore have to find a
way to allow this parameters to be present for the runner while the
promoter can ignore these easily.

This commit solves this by using the common practice of separating the
arguments for the launcher and the launched programs using "--" with a
shared function.

Further work may include changing the wrapper's handling of its own
parameters use Arg, or make the promoter be aware of the arguments.

Opening as a draft to allow @Roman-Manevich to test the promotion of ASL tests


internal/lib/args.mli
```diff
@@ -70,3 +70,15 @@ val is_file : spec -> spec
  *  Arg.spec that raises Arg.Bad if the argument is not a valid path to a
  *  directory. *)
 val is_dir : spec -> spec
+
```
Roman-Manevich, line 73: Wow. kudos for the detailed comment.

hrutvik:
Adding @HadrienRenaud to review ASL+herd tests - @psafont said he will provide details on how to run

psafont:
The fix can be verified by running `make test-all-asl REGRESSION_TEST_MODE=promote`

HadrienRenaud:
Hi @psafont, this looks very nice to me, but I'd like to understand a little bit more before approving. Could you please share a before/after of a command that you changed? Thanks

psafont:
>  I'd like to understand a little bit more before approving. 

Of course, let's run the command for promoting on a branch without the fix:
```
% make test-all-asl REGRESSION_TEST_MODE=promote
Fatal error: exception Unix.Unix_error(Unix.ENOENT, "create_process", "-nohash")
Fatal error: exception Unix.Unix_error(Unix.ENOENT, "create_process", "-nohash")
[...]
Fatal error: exception Unix.Unix_error(Unix.ENOENT, "create_process", "-nohash")
Fatal error: exception Unix.Unix_error(Unix.ENOENT, "create_process", "-nohash")
Some tests had errors
make: *** [test.herd.inst.ASL] Error 1
```
With `make -n`, the command that produces these errors can be seen:
```
_build/default/internal/herd_regression_test.exe \
                -j 15 \
                -nohash \
                -herd-path _build/install/default/bin/herd7 \
                -libdir-path ./herd/libdir \
                -litmus-dir ./herd/tests/instructions/ASL \
                -conf ./herd/tests/instructions/ASL/ci.cfg \
                promote
```
Then we modify the `internal/herd_promote.ml` to print the cli arguments it receives, there's already a guarded expression to print to do this.

The arguments it got called with it is:
```
_build/default/internal/herd_promote.exe -nohash _build/install/default/bin/herd7 -exit true -set-libdir ./herd/libdir -conf ./herd/tests/instructions/ASL/ci.cfg ./herd/tests/instructions/ASL/write_mem.litmus
```
We can see that the `-no-hash` argument snuck up before the location of the herd executable. The code is quite clear here, the command (`com`) to launch is supposed to be the first argument. This means there has been a breakdown in the interface between the caller and the callee. Similarly, this also happens when adding a `-verbose` argument.

To fix this, we change the interface that `herd_promote.exe` and `herd_test.exe` expose so they can be used in the same way by `herd_regression_test.exe`.

After the fix the arguments are:
```
_build/default/internal/herd_promote.exe -nohash -- _build/install/default/bin/herd7 -exit true -set-libdir ./herd/libdir -conf ./herd/tests/instructions/ASL/ci.cfg ./herd/tests/instructions/ASL/globals.litmus
```
The new code, however uses a different output format to show the semantic meaning that it gives to the argument:
```
_build/default/internal/herd_promote.exe called with com: _build/install/default/bin/herd7 and args: [-exit; true; -set-libdir; ./herd/libdir; -conf; ./herd/tests/instructions/ASL/ci.cfg; ./herd/tests/instructions/ASL/records.litmus]
```


Similarly, when test is invoked we can see these arguments with the previous code:

```
_build/default/internal/herd_test.exe -nohash _build/install/default/bin/herd7 -exit true -set-libdir ./herd/libdir -conf ./herd/tests/instructions/ASL/ci.cfg ./herd/tests/instructions/ASL/records.litmus
```

and with the new code:
```
_build/default/internal/herd_test.exe -nohash -- _build/install/default/bin/herd7 -exit true -set-libdir ./herd/libdir -conf ./herd/tests/instructions/ASL/ci.cfg ./herd/tests/instructions/ASL/unknown.litmus
```

HadrienRenaud (approved):
Thanks a lot for the explanation, this looks very good to me.

psafont:
I've added unit-tests for `List.split_at`. I've noticed that I'm not too fond of the existing pretty printers, they allocate for each stage of the conversion, which can become quite costly for serializing datastructures. In general I would prefer something based on `Format.asprintf`, which allows to defer the conversion to be done all at once using `Format.asprintf "%a" pretty_printer_for_a a`. A good library that allows this is https://ocaml.org/p/fmt/latest.

Alcotest combines this with equality tests to have quite succint code, for exmaple, to define a pair of lists of characters, like in the code I added:

```
let split_at_result_pp = Fmt.Dump.(pair (list char) (list char))
let split_at_result_equal = Pair.equal (List.equal Char.equal) (List.equal Char.equal) 
let split_at_result = Alcotest.testable split_at_result_pp split_at_result_equal

[...]

Alcotest.check split_at_result "split_at must split correctly" expected actual
```

internal/lib/tests/base_test.ml
```diff
@@ -97,6 +97,35 @@ let tests = [
       tests
   );
 
+  "Base.List.split_at", (fun () ->
```
fsestini: In general I'm very much in favour of improving our testing infrastructure, but I also think different code paths warrant different degrees of testing. `split_at`'s implementation seems straightforward, and being a general operation on lists it's very unlikely to ever change. Therefore I wonder whether it adds value to have a test such as this one checked in and run at every `make test`.

psafont: I've found that for long-running project, encoding function behaviours in unit tests is worth it. These are generally very cheap to run, in the order of milliseconds, and they end up being read, or excercised because someone wants to change them because there are curious or there are better that don't need that function. This also helps catching silly mistakes.

One such example is when I replaced several `chop` functions with `split_at`: https://github.com/xapi-project/xen-api/pull/6733

Maybe I should rename this function to `split_when`, as the `at` seems to indicate a position.

fsestini: > Maybe I should rename this function to `split_when`, as the `at` seems to indicate a position.

I agree. I was thinking `split_with` but `split_when` also works.

fsestini:
I won't approve this yet as I believe @psafont is still working on some parts of this.

psafont:
I believe that the commit [[internal] Recognise the different wrappers for herd (promote and test)](https://github.com/herd/herdtools7/pull/1938/commits/942cc152c80ea8d2e3f36964266d8b9d43cf39eb) Should fix the issue, but I'm still seeing warnings regardings some promotions not having any output.

The last two commits change interfaces for cli tools, and might not be worth merging:
- The first has the issue that the wrapper (herd_test / herd_promote) doesn't drop the last parameter (the litmus test) when launching herd
- The second has the issue that it changes mapply, which is not internal only.

For this reasons I'm leaving this as draft for the time being.

psafont:
The issue about the litmus parameter being passed should be fixed now, while the mapply change needs some input, reopening for reviews again.

We could even elect to drop the change in the interface for herd_test and herd_promote, as I found a way to fix it, although it is somewhat less robust against new arguments.

internal/lib/testHerd.ml
```diff
@@ -189,20 +189,39 @@ let herd_args ~bell ~cat ~conf ~variants ~libdir ~timeout ~speedcheck
...
+  let wrapper_only_args = verbose @ nohash @ check in
+  wrapper_only_args @ "--" :: herd :: args
+
+let redirect_with_args ?(verbose=false) ~com args =
+  let redirect = Filename.(concat (dirname Sys.argv.(0)) "herd_redirect.exe") in
+  let redirect_only_args = if verbose then ["-verbose"] else [] in
+  (redirect, redirect_only_args @ com :: args)
+
+let mapply_args ?(litmuses=[]) ?(exits=false) ~j ~com args =
+  let exit = match exits with true -> ["-exit"; "true"] | false -> [] in
```
HadrienRenaud:
```suggestion
  let exit = if exits then ["-exit"; "true"] else [] in
```

internal/lib/testHerd.ml
```diff
@@ -189,20 +189,39 @@ let herd_args ~bell ~cat ~conf ~variants ~libdir ~timeout ~speedcheck
...
+  exit @ "-j" :: Printf.sprintf "%i" j :: litmuses @ "--" :: com :: args
```
HadrienRenaud:
```suggestion
  exit @ "-j" :: string_of_int j :: litmuses @ "--" :: com :: args
```

internal/lib/testHerd.ml
```diff
@@ -189,20 +189,39 @@ let herd_args ~bell ~cat ~conf ~variants ~libdir ~timeout ~speedcheck
...
+=
+  let args = herd_wrapper_args ~nohash ~verbose ~check ~herd args in
+  mapply_args ~exits:true ~litmuses ~j ~com:wrapper args
+
+let mapply_herd_test_args =
+  let wrapper = Filename.(concat (dirname Sys.argv.(0))) "herd_test.exe" in
+  mapply_herd_wrapper_args ~wrapper
+
+let mapply_herd_promote_args =
+  let wrapper = Filename.(concat (dirname Sys.argv.(0))) "herd_promote.exe" in
```
HadrienRenaud: Worth extracting `Filename.(concat (dirname Sys.argv.(0)))` to a util function? It's used 3 times here

Roman-Manevich (approved):
I don't understand all details very deeply, but the code looks very readable, and I can see how the interfaces became more robust.

internal/lib/args.mli
psafont: It had a mention to a non-existing field, I've removed it. oops

psafont:
The failure is due to https://github.com/ocaml/opam/issues/5968

I'll check why is apparmor being fussy in the CI environment

psafont:
I've rebased it on top of https://github.com/herd/herdtools7/pull/2005 to get rid of the failure

psafont:
Great, it failed on the workflow I didn't change:

[make-asldoc](https://github.com/herd/herdtools7/actions/runs/35243070846/job/105276177510#logs)
```
opam install . --deps-only --yes
[ERROR] Command "/usr/bin/git -C /__w/herdtools7/herdtools7 ls-files" failed:
"/usr/bin/git -C /__w/herdtools7/herdtools7 ls-files" exited with code 128
Error: Process completed with exit code 99.
```

@hrutvik did you see similar errors while developing the workflow?

hrutvik:
Hmm I'm not sure - one thought I had just after we spoke: does the marking of the directory as safe need to happen in the `setup-opam` action before `opam install . --deps-only`?
```
      - name: Mark working directory as "safe"
        run: git config --global --add safe.directory $GITHUB_WORKSPACE
```

psafont:
> Hmm I'm not sure - one thought I had just after we spoke: does the marking of the directory as safe need to happen in the `setup-opam` action before `opam install . --deps-only`?
> 
> ```
>       - name: Mark working directory as "safe"
>         run: git config --global --add safe.directory $GITHUB_WORKSPACE
> ```

That job does not use the new action, so I don't think it's related to the introduction of `setup-opam`

hrutvik:
Ah I see - we have had some flakiness with using the latest `texlive` image, see https://github.com/herd/herdtools7/pull/1566 for example - fixing a version could potentially help?

psafont:
> Ah I see - we have had some flakiness with using the latest `texlive` image, see #1566 for example - fixing a version could potentially help?

I'm trying to avoid using docker images, and using a "normal", javascript-based action, with tectonic: https://github.com/tectonic-typesetting/tectonic/#readme

tools/tests/dune
```diff
@@ -2,7 +2,8 @@
  (applies_to miaou)
  (deps
   (source_tree libdir)
-  %{bin:miaou7})
+  %{bin:miaou7}
+  %{bin:mapply7})
```
fsestini: Note that this stanza contains `(applies_to miaou)`, so in particular it doesn't apply to `mapply.t`. Maybe `mapply.t` should have its own stanza?

BASIC-img/index.tex
fsestini, line 1: What it the purpose of this empty file?

internal/lib/args.mli
```diff
@@ -70,3 +70,16 @@ val is_file : spec -> spec
...
+  args : string list; (** Arguments for wrapper *)
+  com : string; (** Path of wrapped command *)
+  wrapped : string list; (** Arguments for [com] *)
+}
+
+val split_wrapper_args : String.t array -> wrapper_args
+(** [split_wrapper_args inp] returns [{args; com; wrapped; litmus}], where
+    [args] is the shortest prefix of elements in [inp] that are not "--", [com]
+    is the argument immediately after "--", and [wrapped] is the rest of the
+    arguments except the last one, without the argument "--"
```
fsestini: > [wrapped] is the rest of the arguments except the last one

From the implementation, it looks like `wrapped` includes the last argument as well

internal/lib/args.mli
```diff
@@ -70,3 +70,16 @@ val is_file : spec -> spec
...
 val is_dir : spec -> spec
+
+type wrapper_args = {
+  args : string list; (** Arguments for wrapper *)
+  com : string; (** Path of wrapped command *)
+  wrapped : string list; (** Arguments for [com] *)
+}
+
+val split_wrapper_args : String.t array -> wrapper_args
+(** [split_wrapper_args inp] returns [{args; com; wrapped; litmus}], where
```
fsestini:
```suggestion
(** [split_wrapper_args inp] returns [{args; com; wrapped}], where
```

internal/herd_redirect.ml
```diff
@@ -16,20 +16,27 @@
...
+         Printf.eprintf "%s: Could not find litmus among arguments: [%s]\n%!"
+           Sys.argv.(0) (String.concat "; " args) ;
+         exit 1 in
+  let rec gather_args (flags, litmus) args =
+    match args with
+    | [] -> flags, get_litmus litmus
+    | "-verbose" :: args ->
+        gather_args ({verbose=true}, litmus) args
+    | _ :: args ->
+        gather_args (flags, litmus) args
```
fsestini, line 39: Should one of these branches set `litmus` to a non-`None` value? Otherwise I think the top level `litmus` will always be `None`

tools/tests/dune
psafont: yes, it should. Fixing it

BASIC-img/index.tex
psafont: It's a product of running make, it should not be checked-out

internal/herd_redirect.ml
psafont: yep, just like the other tools. I've added the branch

fsestini:
Thanks for the fixes @psafont . I think this is ready to be rebased and merged, once https://github.com/herd/herdtools7/pull/2005 lands.

As an aside, I'm wondering whether you had the chance to run the full test suite on both this PR and current `master`, to compare the generated tool invocations side by side. I'm not sure there is an easy way to do this kind of comparison with our current tools, but thought I'd ask, since this PR changes the input/output protocol across several tools.

psafont:
> As an aside, I'm wondering whether you had the chance to run the full test suite on both this PR and current `master`, to compare the generated tool invocations side by side. I'm not sure there is an easy way to do this kind of comparison with our current tools, but thought I'd ask, since this PR changes the input/output protocol across several tools.

Once I understood what the problem was, and a holistic solution was found was to textually search all the users, in all files, including the Makefile and change the affected usages, as well as tools. I did individually check the make targets directly affected, even showing executions in the chain of executions involving mapply and herd_xxx, which is mostly make -> regression_tools -> mapply -> herd_xxx -> herd

This is on top of the unit-tests for mapply to ensure that the tools using it are as expected; and running as many tests as I could, with
```bash
ALL_TESTS=true make test-all test-all-asl diy-test diymicro-test litmus-aarch64-test
```

I can spend more time adding unit-tests for the herd_xxx tools, if you are not sure about the correct behaviour of the changes

fsestini:
> I can spend more time adding unit-tests for the herd_xxx tools, if you are not sure about the correct behaviour of the changes

No need to hold this PR for that. You run your tests, and the code LGTM. I'm approving.

fsestini (approved):

PR #1939, opened 2026-07-30
[catalogue] Fix catalogue

maranget:
Fix various problems in the catalogue.


lib/interpreter.ml
```diff
@@ -1372,6 +1372,7 @@ module Make
             begin match eval_ord env e with
             | V.Empty -> V.Empty
             | Unv -> Rel (Lazy.force env.EV.ks.id)
+            | Event e -> Rel (E.EventRel.singleton (e,e))
```
fsestini, line 1375: Out of curiosity, what problem is this line addressing? What is an example of a `ToId` operator with an `Event` operand?

maranget: Well I could have made a separate PR the code is [`let succs (e,r) = range ([e];r)`](https://github.com/herd/herdtools7/blob/efbec191d444827bf29e61aed5347b572151f6fa/catalogue/aarch64-down-one-leg/cats/always-executed-writes.cat#L6), where `e` is a single event. This code was working before,because the Cat interpreter was systematically  using the same function (`eval_events`) when an event set is expected. The function `eval_events`  performs the silent promotion of one event into a singleton set. Since for instance PR #1855, the evaluation of sets has become more complex and this case of `[e]` where `e` is an event had been overlooked.

PR #1940, opened 2026-07-30
[herd] Improvements for vmsa+ifetch

artkhyzha:


herd/mem.ml
```diff
@@ -429,6 +429,35 @@ module Make(C:Config) (S:Sem.Semantics) : S with module S = S	=
...
+        List.fold_left (fun acc (a,_) -> 
+          acc + (if pa <= a && a < pa + Pseudo.page_size then 1 else 0)
+          ) 0 code
+      in
+      let check =
+        match page_addrs with
+        | [] -> () (* no checks to perform *)
+        | pa::page_addrs -> begin
+          let eta = page_size pa in
+          List.iter (fun a ->
```
psafont, line 453: Would you be interested in taking some functions in the list and map libraries that make the intention clearer in the code?

In particular:
- extracting values from the map: https://github.com/herd/herdtools7/commit/a15b8c53899bf5d96b92340c8edba8dee30f90f5
- counting elements that satisfy a function: https://github.com/herd/herdtools7/commit/4052e676577a8c9637ddc9703f1f4f64b803b387

These can be adapted into this code like so:
```suggestion
        Label.Map.values prog
        |> Seq.filter (fun addr ->
          (addr mod Pseudo.page_size = 0)
          && (is_on_exported_page (addr2va addr) || is_on_relevant_page (addr2va addr))
          )
      in
      let page_size pa =
        let (_,code) = IntMap.find pa code_segment in
        Misc.List.count (fun (a, _) -> pa <= a && a < pa + Pseudo.page_size) code
      in
      let check =
        match Seq.uncons page_addrs with
        | None -> () (* no checks to perform *)
        | Some (pa, page_addrs) -> begin
          let eta = page_size pa in
          Seq.iter (fun a ->
```

psafont:
Is the first commit only fixing a bug, or are there also enhancements? It would be good to explain in the commit message which bug is being fixed, and if there are unrelated enhancements, separate them from the bugfix.

PR #1942, opened 2026-07-31
[herd] Include some linux tests in the regression testing procedure

maranget:
This PR add a few linux kernel memory model tests to our regression suite. The model, configuration files and most tests are taken from [Linux sources](https://github.com/torvalds/linux/tree/master/tools/memory-model). A few tests are authored by @akiyks.


relokin:
Thanks for this Luc, this is very useful!

Just one question. Should we move the tests to the catalogue? I see two arguments for doing so:
* `herd/test` should focus on unit tests that are focused on the semantics we implement in herd. Arguably, this is not universally true but I think this would be a good principle to follow.
* if we had a catalogue C-lkmm or LKMM we could expose these tests through the web interface.

If you want to maintain the expected files, see here for what I did with a similar requirement https://github.com/herd/herdtools7/pull/1864/changes#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52

maranget:
Exposing some LKMM tests in the web interface looks like a great idea. Nevertheless, I'd realy like to also have some LKMM testing to be part of systematic testing, given the often unforseen impact of some changes on LKMM we experienced in the past. Your solution from PR #1864 can apparently give us both. Should we wait for PR #1864 to be merged?

maranget:
Hi @relokin. I have moved all tests to the now serious "linux" book of the web interface and used your Makefile entry "test.herd.cata-extended.%" to include the tests of the book into the standard test suite (`make test`).

Web interface: https://diy.inria.fr/TST/www/index.html?record=linux&bell=linux-kernel&cat=linux-kernel&litmus=MP%2Brelacq&cfg=linux-kernel

herd/lexConf_herd.mll
```diff
@@ -119,7 +119,13 @@ let handle_key dir main key arg = match key with
...
 | "suffix" ->  suffix := arg
 | "include" ->
-   includes := !includes @ [arg]
+     let arg =
+       let pat = "$dirname" in
+       if String.starts_with ~prefix:pat arg then
+         let lpat = String.length pat in
+         dir ^ String.sub arg lpat (String.length arg - lpat)
+       else arg in
+     includes := arg :: !includes
```
HadrienRenaud: Hi @maranget, just checking I understand well this new syntax.

Why not `./`? is this because relative paths are relative to the current working directory and not relative to the configuration file?

Also is it on purpose that you've switch the order of addition of arguments to this? (`!includes @ [arg]` becomes `arg :: !includes`)

maranget: > Hi @maranget, just checking I understand well this new syntax.
>

Hi @HadrienRenaud 
 
> Why not `./`? is this because relative paths are relative to the current working directory and not relative to the configuration file?

I guess so, `./` refers to the current working directory and I guess we want that meaning not to change.
> 

> Also is it on purpose that you've switch the order of addition of arguments to this? (`!includes @ [arg]` becomes `arg :: !includes`)

A mistake of mine. I'll fix it.

fsestini:
Could you help me understand the motivation for the new `include $dirname`? How come other catalogues do not seem to need it?

herd/lexConf_herd.mll
maranget: Fixed, thanks.

HadrienRenaud: Thinking about it I'm not sure it really makes sense that `./` in a conf file refers to the current working directory, and not the directory of the conf file.
That would mean that any conf file that contains a relative directory can only be run from a single directory? That seems brittle to me

maranget:
> Could you help me understand the motivation for the new `include $dirname`? How come other catalogues do not seem to need it?

For the linux arch, **herd7** loads three files in addition to the configuration file (.cfg): model (.cat), bell file (.bell) and macros (.def). For running **herd7** as `make  test.herd.cata-extended.linux` does, I found it convenient to group all these files in the same directory (catalogue/linux/cats) and to specify them in the configuration file. The $dirname trick avoids ising explicit paths. Additionnaly one can invoke the installed **herd7** directly on the tests as simply as:
```
% herd7 -conf catalogue/linux/cfgs/ci.cfg catalogue/linux/tests/*.litmus
```
Hence the new feature has more to do with **herd7** than with the web interface. However it may be necessary to load the `macro file. I am not so sure about the bell file...

Another option would have been to add the `-I` option to the test driver for it to pass the search path to **herd7**. This would have been slightl;y  more complex.

herd/lexConf_herd.mll
fsestini: The meaning of `.` when used to specify inclusion paths in a config file is tool-dependent, and there are many examples where it is used to mean "this config file's directory" rather than "the cwd at the moment of tool invocation". `dune` and [`git`](https://git-scm.com/docs/git-config#_includes) are two examples that follow the "this config file's directory" semantics. Thus it wouldn't be unreasonable to treat `.` this way in `herd7`, it just needs to be documented clearly.

catalogue/linux/tests/Z6.0+pooncelock+pooncelock+pombonce.litmus
```diff
@@ -0,0 +1,40 @@
+C Z6.0+pooncelock+pooncelock+pombonce
```
relokin, line 1: This catalogue has two tests with almost identical names:
- Z6.0+pooncelock+pooncelock+pombonce
- Z6.0+pooncelock+poonceLock+pombonce

In case-insensitive filesystems (for example, by default APFS on MacOS), this creates problems. Can we change the name of one of the tests?

catalogue/linux/cfgs/ci.cfg
```diff
@@ -0,0 +1,3 @@
+include $dirname/../cats
```
relokin: isn't this very fragile? Is it hard to create a full libdir inside this catalogue?

relokin (commented):
Thanks for this Luc, I've got two comments.

herd/lexConf_herd.mll
fsestini: That said, I realise that treating `./` as "the config's directory" would be incompatible with the way herd7 resolves relative paths currently. In other words, I think it would be surprising if `include foo` and `include ./foo` had different semantics.

catalogue/linux/cfgs/ci.cfg
maranget: I do not see what you mean by  "full libdir inside this catalogue".

catalogue/linux/tests/Z6.0+pooncelock+pooncelock+pombonce.litmus
maranget: Nice catch, I am fixing this as you suggest.

catalogue/linux/cfgs/ci.cfg
relokin: AFAICT, the issue is that the Makefile sets `libdir` to `herd/libdir`, while you want `herd7` to be able to find `linux-kernel.bell`, `linux-kernel.cat`, `linux-kernel.def`, and `lock.cat`, which currently live in `catalogue/linux/cats`.

I can see three possible ways to fix this:

- Your approach: extend the list of paths in which `herd7` looks for files.
- Copy these files into `herd/libdir`.
- Create a complete `libdir` under `catalogue/linux/libdir` containing all the `.cat`, `.bell`, and `.def` files needed to run this catalogue.

I think the second and third options are less fragile. More generally, independently of this particular catalogue, I would expect us to want all the files needed to run `herd7` with the LKMM model to be available in one place, as that would also make things easier for users.

catalogue/linux/tests/Z6.0+pooncelock+pooncelock+pombonce.litmus
relokin: Thank you!

herd/lexConf_herd.mll
HadrienRenaud: That makes sense thanks.

Reading a bit further into the code, there are 2 different behaviours: one for the includes, for which relative paths are always resolved at the current working directory, and for other files, which are resolved in the first path from `includes` that work. Maybe another way would be to allow resolving included paths from the `includes`, as is done today for the conf files.

> In other words, I think it would be surprising if include foo and include ./foo had different semantics.

Sorry yes, I was using `./` as a stand-in for all relative paths.

fsestini:
> > Could you help me understand the motivation for the new `include $dirname`? How come other catalogues do not seem to need it?
> 
> For the linux arch, **herd7** loads three files in addition to the configuration file (.cfg): model (.cat), bell file (.bell) and macros (.def). For running **herd7** as `make test.herd.cata-extended.linux` does, I found it convenient to group all these files in the same directory (catalogue/linux/cats) and to specify them in the configuration file. The $dirname trick avoids ising explicit paths. Additionnaly one can invoke the installed **herd7** directly on the tests as simply as:
> 
> ```
> % herd7 -conf catalogue/linux/cfgs/ci.cfg catalogue/linux/tests/*.litmus
> ```

Thanks for the additional context. As a data point, for the mere purpose of `make test` and/or CI, I believe it would be enough to write this in `ci.cfg`:

```
include catalogue/linux/cats
local-conf linux-kernel.cfg
```

I realise this only works if `herd7` is executed in a working directory that corresponds to the repository root. Such assumption holds for CI, but not necessarily so if a user is running these tests independently from a non-root cwd. @relokin 's suggestions seem to offer more flexibility in that regard.

catalogue/linux/cfgs/ci.cfg
fsestini: I would also be in favour of solving this problem with (2) or (3), or other "internal" solution, before considering working around it by changing the public configuration language of herd. I do see the utility in what `$dirname` proposes, however, user-facing changes come with a maintenance commitment to keep that interface stable, and I think that warrants its own PR and discussion.

maranget: The four files `linux-kernel.bell`, `linux-kernel.cat`, `linux-kernel.def`, and `lock.cat` are from the linux sources and I'd rather have them in the catalogue (with README) than in **herd7** standard libdir. Hence I'd rather have solution (3). However, `lock.cat` includes the files `cross.cat` and `cos-opt.cat`. Hence those files would have to be copied into `catalogue/linux/libdir`. Additionally a `stdlib.cat` would also have to be copied there. I do not find this that elegant.

maranget:
> Thanks for the additional context. As a data point, for the mere purpose of `make test` and/or CI, I believe it would be enough to write this in `ci.cfg`:
> 
> ```
> include catalogue/linux/cats
> local-conf linux-kernel.cfg
> ```
> 
> I realise this only works if `herd7` is executed in a working directory that corresponds to the repository root. Such assumption holds for CI, but not necessarily so if a user is running these tests independently from a non-root cwd. @relokin 's suggestions seem to offer more flexibility in that regard.

I follow your suggestion, which provides a simple and workable solution.

maranget:
Hi @relokin abd @fsestini. I have followed @fsestini's simple suggestion for configuring the CI testing of LKMM and removed the "$dirname" feature. I have also fixed a long-standing bug that the new test set unveiled, see commit 3867397.

fsestini (approved):
Thanks for the changes @maranget .

relokin (approved):
LGTM, thanks Luc!

maranget:
Merged thanks a lot @relokin and @fsestini for your suggestions.

PR #1944, opened 2026-07-31
[herd] Save dot/neato output with a shell redirection

maranget:
Small bug fix: for some reason using the option `-o filename` or `-ofilename` sometimes results into a segfault. Replacing the option with a redirection looks to be a workaround.

### DIFF
--- a/herd/show.ml
+++ b/herd/show.ml
@@ -72,7 +72,7 @@ module Make(O:PrettyConf.S) = struct
     with_temp_file ~keep:keep_tmp_pdf name_ps @@ fun () ->
     run_cmds
       [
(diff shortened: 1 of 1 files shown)


HadrienRenaud (approved):
Wow this is so strange. Thanks @maranget for the fix.

By curiosity, do you have a test that generates such a bug?
I'm also wondering if this is worth a bug report upstream to graphviz?

fsestini:
> By curiosity, do you have a test that generates such a bug? I'm also wondering if this is worth a bug report upstream to graphviz?

Hi @maranget , I'm also keen to hear your thoughts on these points. In addition, I'm wondering whether the same issue manifests itself on the macOS-specific `do_show_file_preview` path?

PR #1946, opened 2026-07-31
ASL compliance testing

Roman-Manevich:
This PR adds an ASL compliance test suite and a test runner --- `asltest` --- under asllib/compliance.

The runner:
- discovers all paired .asl / .yaml files under the `compliance` folder recursively.
- parses expected metadata from YAML
- runs `aslref` through the new new `Runner.RunResult` API
- compares expected and actual metadata
- supports `--promote` to rewrite expected YAML files
- reports all per-test failures

`compliance` also includes:
- The test schema under the `schema` sub-folder
- Test directories with a few sample tests.

Using the compliance test suite:
- run `dune runtest` from `asllib/compliance`
- To promote, run `dune exec -- asltest --promote`
- You can specify `--suite` to explicitly point to the test directory. By default, the test runner looks for the tests under `asllib/compliance` or just `compliance`.


asllib/Runner.ml
```diff
@@ -78,25 +78,35 @@ let default_args =
...
 
 (** Run ASLRef with the supplied arguments. This function never returns: it
-    raises an [Exit] exception containing ASLRef's exit code. *)
-let run_with (args : args) : unit =
+    raises an [Exit] exception containing ASLRef's exit code.
+
+    [on_error] is called with the structured ASLRef error before [Exit 1] is
+    raised. By default, it prints the error using [args.output_format].
+    [on_completed] is called before [Exit exit_code] is raised. *)
+let run_with ?on_error ?(on_completed = fun () -> ()) (args : args) : unit =
```
HadrienRenaud, line 86: I think a little refacto is needed here. Notably, I think we should not wrap individual subcalls with `or_exit`, and that we should wrap the whole content of run into various variants of `or_exit`, one for the command line that catch everything and raises `Exit exit_code`, one for the test runner that returns a `Result.t`, and potentially one for `jaslref` that does its own thing.

Roman-Manevich: We can do this in a future PR.

HadrienRenaud: No I am suggesting a refacto of the code implemented in this PR

See for example: 8a7e8ae4f47ef7538cb0a833f4038c5cc9103f40

asllib/dune-project
```diff
@@ -21,7 +21,8 @@
    (ocaml (>= 4.14))
    dune
    (menhir (>= 20231231))
-   (zarith (>= 1.13)))
+   (zarith (>= 1.13))
+   yaml)
```
HadrienRenaud: Does this work with any version of `yaml`?

Also please put it in `:with-test` because you don't need it to simply install the tool
```suggestion
   (yaml :with-test))
```

asllib/compliance/ASLSemanticsReference/SemanticsRule.SAssertNo.asl
hrutvik, line 1: Are these all duplicates of existing tests? They seem to be copied rather than moved

asllib/dune-project
Roman-Manevich: I pinned the `yaml` version to `3.2.0`.

HadrienRenaud: That looks good to me because it was release 3 years ago!

asllib/compliance/regressions/hello_world.asl
hrutvik, line 1: Why a `regressions` directory?

asllib/compliance/dune
```diff
@@ -0,0 +1,11 @@
...
+ (name asltest)
+ (public_name asltest)
+ (libraries yaml asllib unix))
+
+(rule
+ (alias runtest)
+ (deps
+  (source_tree .))
+ (action
+  (run ./asltest.exe --suite %{workspace_root}/asllib/compliance)))
```
hrutvik, line 10: Use a `test` stanza instead? Also, we should consider making this optional with `build_if` (see e.g. `menhir2bnfc` tests) so that users without `yaml` don't see errors

asllib/dune-project
```diff
@@ -21,7 +21,8 @@
    (ocaml (>= 4.14))
    (dune (>= 3.14))
    (menhir (>= 20231231))
-   (zarith (>= 1.13)))
+   (zarith (>= 1.13))
+   (yaml (= 3.2.0)))
```
hrutvik: Should this be `=` or `>=`?

herdtools7.opam
```diff
@@ -19,6 +19,7 @@ depends: [
   "dune"  {>= "3.14.0" }
   "menhir" {>= "20231231"}
   "zarith" {>= "1.13"}
+  "yaml" {= "3.2.0"}
```
hrutvik: Similarly should this be `>=`, but also does this dependency need to be surfaced for the whole project?

asllib/compliance/asltest.ml
```diff
@@ -0,0 +1,331 @@
...
+  *)
+  let directory_exists path = Sys.file_exists path && Sys.is_directory path
+
+  (** [default_suite_path ()] returns the default suite path for commands run
+      from either the repository root or [asllib]. *)
+  let default_suite_path () =
+    if directory_exists "asllib/compliance" then "asllib/compliance"
+    else if directory_exists "compliance" then "compliance"
+    else "asllib/compliance"
+
```
hrutvik, line 33: This all seems rather unnecessary and non-obvious behaviour, given `dune runtest` will always provide the directory. This can just error if no suite is provided.

asllib/compliance/asltest.ml
```diff
@@ -0,0 +1,331 @@
...
+  (** One discovered compliance test. *)
+
+  val discover_tests : string -> test list
+  (** [discover_tests suite_path] returns all sibling ASL/YAML test pairs.
+      Raises [Runner_error] when a source or metadata file is missing its
+      sibling. *)
+end = struct
+  type test = { asl_path : string; metadata_path : string }
+
+  (** [sorted_directory path] returns deterministic directory entries. *)
```
hrutvik, line 64: Does determinism matter here?

asllib/compliance/asltest.ml
```diff
@@ -0,0 +1,331 @@
...
+
+  val run_test : Discovery.test -> result
+  (** [run_test test] runs one compliance test. *)
+end = struct
+  type result =
+    | Completed of { expected : Metadata.t; actual : Metadata.t }
+    | Failed of string
+
+  (** [with_captured_stdout run] runs [run] while capturing stdout. *)
+  let with_captured_stdout run =
```
hrutvik, line 148: This seems a rather complex way to capture ASLRef output, given that we are calling ASLRef as an OCaml library. Can we instead parametrise ASLRef's error printer by output channel, so it can be stdout or something else as desired?

asllib/compliance/asltest.ml
```diff
@@ -0,0 +1,331 @@
...
+  let missing_metadata_error metadata_path =
+    Printf.sprintf "%s: expected metadata file is missing" metadata_path
+
+  (** [missing_asl_error asl_path] explains which ASL source file is missing. *)
+  let missing_asl_error asl_path =
+    Printf.sprintf "%s: expected ASL source file is missing" asl_path
+
+  (** [candidate_base path] returns the test base path represented by [path],
+      when [path] is an ASL source or YAML metadata file. *)
+  let candidate_base path =
```
hrutvik, line 79: The rest of this module seems unnecessarily verbose for what it's doing. I suggest borrowing from `parser_cmp.ml` in `menhir2bnfc`: a small function to list all files in the test suite, a single list traversal to build two sets (ASL names + YAML names, without file extensions), then compare the sets.

asllib/compliance/asltest.ml
```diff
@@ -0,0 +1,331 @@
...
+      default_args with
+      exec;
+      files = [ (NormalV1, asl_path) ];
+      (* The compliance suite is ASLv1-only. *)
+      no_stdlib0 = true;
+    }
+
+  (** [successful_metadata expected output] builds successful actual metadata.
+  *)
+  let successful_metadata expected output =
```
hrutvik, line 202: If this is used exactly once, I am not sure why it needs to be a separate function definition - this scatters the code unnecessarily

asllib/compliance/asltest.ml
```diff
@@ -0,0 +1,331 @@
...
+    try
+      let expected = Metadata.parse_file test.metadata_path in
+      let args = args_for_test expected test.asl_path in
+      let result, output =
+        with_captured_stdout (fun () ->
+            Asllib.Runner.RunResult.run_with_result args)
+      in
+      Completed { expected; actual = metadata_of_result expected result output }
+    with
+    | Metadata.Metadata_error message ->
```
hrutvik, line 250: Can this only happen when calling `Metadata.parse_file`? Why not isolate it to there?

asllib/compliance/asltest.ml
```diff
@@ -0,0 +1,331 @@
...
+end = struct
+  (** [location label path line] formats an editor-friendly labelled source
+      location. *)
+  let location label path line =
+    Printf.sprintf "%s %S, line %d, characters 0-0:" label path line
+
+  (** [actual_asl_location test actual] returns an editor-friendly ASL location,
+      using ASLRef's reported source line when available. *)
+  let actual_asl_location test actual =
+    let line = Option.value actual.Metadata.error_line ~default:1 in
```
hrutvik, line 270: If we don't have an actual line number, line 1 does not seem a sensible default

asllib/compliance/asltest.ml
```diff
@@ -0,0 +1,331 @@
...
+    location "Test" test.Discovery.asl_path line
+
+  (** [mismatch_message test expected actual] explains a metadata mismatch and
+      includes jump targets for both the expected YAML and actual ASL source
+      location. *)
+  let mismatch_message test expected actual =
```
hrutvik, line 276: These diff outputs are something we did not want to implement

asllib/compliance/asltest.ml
```diff
@@ -0,0 +1,331 @@
...
+      Arg.parse_argv argv specs
+        (fun arg ->
+          raise (Arg.Bad (Printf.sprintf "unexpected argument %S" arg)))
+        (usage program_name);
+      { suite_path = !suite_path; promote = !promote }
+    with Arg.Bad message -> raise_errorf "%s" message
+end
+
+module Discovery : sig
+  type test = { asl_path : string; metadata_path : string }
```
hrutvik, line 54: Why is this hidden within `Discovery`, when most parts of this file need it? A lot of the code ends up qualifying with `Discovery` everywhere

asllib/compliance/metadata.ml
```diff
@@ -0,0 +1,373 @@
...
+
+(** [string_of_mode mode] returns the YAML spelling of [mode]. *)
+let string_of_mode = function Exec -> exec_value | No_exec -> no_exec_value
+
+(** [string_of_outcome outcome] returns the YAML spelling of [outcome]. *)
+let string_of_outcome = function
+  | Success -> success_value
+  | Failure -> failure_value
+
+module Parser : sig
```
hrutvik, line 41: It feels like we are reinventing the wheel here on lots of parsing + validation, rather than relying on the schema that is checked in and an appropriate library.

asllib/compliance/metadata.ml
```diff
@@ -0,0 +1,373 @@
...
+    in
+    let yaml =
+      match Yaml.of_string text with
+      | Ok yaml -> yaml
+      | Error (`Msg message) -> raise_errorf "%s: %s" path message
+    in
+    parse_yaml path yaml
+end
+
+module Writer : sig
```
hrutvik, line 214: Similarly, this feels like we are not relying on an appropriate library

asllib/compliance/metadata.ml
```diff
@@ -0,0 +1,373 @@
...
+    | _ -> raise_errorf "%s: expected top-level object" diagnostic_path
+
+  (** [parse_file path] reads and validates one metadata YAML file. *)
+  let parse_file path =
+    let text =
+      try In_channel.with_open_text path In_channel.input_all
+      with Sys_error message -> raise_errorf "%s: %s" path message
+    in
+    let yaml =
+      match Yaml.of_string text with
```
hrutvik, line 207: This seems to be the only usage of the `yaml` library - is this right?

.github/workflows/make-test.yml
```diff
@@ -79,7 +79,7 @@ jobs:
       - name: Set-up OCaml
         uses: ocaml/setup-ocaml@v3
         with:
-          ocaml-compiler: ocamlformat.0.28.1
+          ocaml-compiler: ocamlformat.0.28.1,yaml.3.2.0
```
hrutvik, line 82: Why is `yaml` needed for a code formatting check?

asllib/compliance/asltest.ml
HadrienRenaud: I would do this in a separate PR merged before: it does not need the error codes.

asllib/compliance/dune
HadrienRenaud: I guess not doable as is because it takes the `--suite` argument. I think probably for now we are fine leaving this argument hard-coded in `asltest.exe`? The `DUNE_SOURCEROOT` variable is defined by dune.

herdtools7.opam
HadrienRenaud: Why not make the compliance test suite runner optional?

asllib/compliance/asltest.ml
HadrienRenaud: As mentioned above, I think this should be hard-coded.

asllib/compliance/asltest.ml
HadrienRenaud: With this needs to come a process to remove those cases where we don't have a location for the error

asllib/compliance/asltest.ml
```diff
@@ -0,0 +1,331 @@
...
+    {
+      mode = expected.mode;
+      outcome = Failure;
+      output;
+      error = error_code;
+      error_line;
+      info =
+        (match expected.info with
+        | Some _ -> expected.info
+        | None -> diagnostic);
```
HadrienRenaud, line 226:
```suggestion
        Option.value ~default:diagnostic expected.info;
```

asllib/compliance/dune
hrutvik: Is it not possible to replicate the approach from `menhir2bnfc/integration/dune`? That also passes an argument for the directory of tests

PR #1949, opened 2026-08-03
[asl] Fix a divergence between documentation/implementation

hrutvik:
Regarding for-loop limits:
  - Documentation: SemanticsRule.EvalFor ticks the limit before SemanticsRule.EvalForLoop executes the body.
  - Implementation: loop within eval_for seems to execute body then tick loop limit.

### DIFF
--- a/asllib/tests/loop-limits.t/run.t
+++ b/asllib/tests/loop-limits.t/run.t
@@ -108,9 +108,10 @@ Double loops
 For loops
   $ aslref for-correct.asl
   $ aslref for-incorrect.asl
-  File for-incorrect.asl, line 5, characters 4 to 26:
+  File for-incorrect.asl, line 4, character 2 to line 6, character 6:
+    for i = 1 to n looplimit 10 do
       counter = counter + 1;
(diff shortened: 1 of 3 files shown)


asllib/tests/loop-limits.t/run.t
```diff
@@ -165,3 +165,10 @@ Recursion limits:
                     ^^^^^^^^^^^^^
   ASL Dynamic error: recursion limit reached.
   [1]
+
+  $ aslref for-loop-zero.asl
+  File for-loop-zero.asl, line 4, characters 6 to 21:
+        println "BODY";
```
HadrienRenaud, line 173: We might need to fix this location also, we should probably put for loc the loop limit

hrutvik: Added a new commit with this

HadrienRenaud: Thanks, looks good.

Roman-Manevich (approved):
Looks good to me!

HadrienRenaud (approved):

PR #1951, opened 2026-08-05
[aslref] Refacto useless indirections in Runner.ml

HadrienRenaud:
This PR fixes some strange and useless indirections in `Runner.ml`. The resulting code is far less complicated and written far more directly.

Other notable change is that the choice of intercepting errors in now made by `aslref.ml` and not in library code in `Runner.ml`.

I did not change the patterns in `jaslref.ml` because I do not know how to test it.

### DIFF
--- a/asllib/Runner.ml
+++ b/asllib/Runner.ml
@@ -204,22 +204,22 @@ let run (args : args) : int =
 
   exit_code
 
+(** Run ASLRef with the supplied arguments. If an ASLException happens, it is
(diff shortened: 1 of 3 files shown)


Roman-Manevich:
I'm not sure it's best to do this refactor on top of `asl-compliance-testing`, which is a changing target. Perhaps do the refactor over `master` and then `asl-compliance-testing` can build on top of that.

HadrienRenaud:
Opened #1952.

PR #1954, opened 2026-08-06
[herd] Fix SWP register port annotation

relokin:


HadrienRenaud (approved):
Good catch, thanks for the fix @relokin.

relokin:
Thanks @HadrienRenaud !

PR #1957, opened 2026-08-07
[readers-guide] Refresh generated litmus descriptions and testing logic

fsestini:
This PR refreshes the generated litmus descriptions shipped in the Readers Guide and makes those files the authoritative snapshots for testing `litmus2desc`.

More precisely, it:

- updates `make descriptions` to use the in-repository `litmus2desc` tool and cat files by default;
- regenerates all LaTeX descriptions with the current `litmus2desc` via `make descriptions`;
- normalizes some description filenames for compatibility with case-sensitive filesystems;
- changes the cram testing logic in `litmus2desc` to use files from `readers-guide/litmus-descriptions` as authoritative snapshots for expectation testing, rather than duplicating them.

Note that the descriptions in `readers-guide/litmus-descriptions`, although auto-generated from `litmus2desc`, are kept checked-into version control. My initial idea was to remove them from version control and have `make` generate them instead. However, I'm now leaning towards keeping them checked in, for a couple of reasons I list below. Any comments on this are welcome.

- Keeping the descriptions checked-in makes it possible for someone to clone the repository, run `make`, and build the Readers Guide with LaTeX alone, without requiring an OCaml/opam toolchain to be installed;
- I think we need to be able to review and test changes to the exact prose of the descriptions we ship with the Guide, so that changes to the cat files or `litmus2desc` don't inadvertently introduce unwanted changes. This means that, in any case, _some_ descriptions snapshots would need to be kept under version control in some form or another. Currently, this form of snapshot testing is done through expectation files under `tools/litmus2desc/test`, but those files are to a large extent duplicating the contents of `readers-guide/litmus-descriptions`, and in some cases even lagging behind them. This PR proposes to use `readers-guide/litmus-descriptions` as the single, authoritative, and up-to-date source for both building the Readers Guide and snapshot-testing it.

A separate, follow-up PR will address the current duplication between `readers-guide/litmus-tests/*` and `catalogue/aarch64-readers-guide`.

CC @psafont


.gitignore
relokin, line 1: Shouldn't these go into a local gitignore? I mean inside the folder readers_guide?

tools/litmus2desc/test/rg_litmus_tests/CoRR.litmus
relokin, line 1: Can't we symlink the whole folder instead?

relokin (commented):
Overall the change looks good to me. Two minor suggestions/questions. I think we might also want to move the readers_guide folder inside doc? But this is not in the scope of this PR.

tools/litmus2desc/test/rg_litmus_tests/CoRR.litmus
fsestini: What the PR proposes is to actually delete the `rg_litmus_tests` folder altogether (since those files can already be found elsewhere).

.gitignore
fsestini: Yes, good idea.

readers-guide/litmus-descriptions/2+2w.tex
```diff
@@ -5,6 +5,6 @@
   \item The second Store instruction in program order on P0, i.e. STR W2,[X3], generates an Explicit Memory Write Effect of Location y with value 1;
   \item The first Store instruction in program order on P1, i.e. STR W0,[X3], generates an Explicit Memory Write Effect of Location y with value 2;
   \item The second Store instruction in program order on P1, i.e. STR W2,[X1], generates an Explicit Memory Write Effect of Location x with value 1;
-  \item The value of x is 2 in the end, which means that the Explicit Memory Write Effect of Location x on P1 generated by STR W2,[X1] is Coherence-before the Explicit Memory Write Effect of Location x on P0 generated by STR W0,[X1];
```
psafont, line 8: The second commit's message states that these descriptions where regenerated, but it doesn't say how. I think mentioning `cd readers-guide && make descriptions` would consolidate the knowledge inside the git history

tools/litmus2desc/test/readers-guide.t
```diff
@@ -1,495 +1,11 @@
 Run litmus2desc on readers-guide tests in sorted order and compare output.
 
-  $ for f in $(LC_ALL=C ls -1 rg_litmus_tests | sort); do
-  >   echo "== $f =="; echo;
-  >   cat "rg_litmus_tests/$f"; echo;
-  >   litmus2desc -set-libdir ./libdir --describe-dep-path "rg_litmus_tests/$f"; echo;
+  $ descriptions=../../../readers-guide/litmus-descriptions
```
psafont, line 3: It would be nice to use something like `$DUNE_SOURCEROOT/readers-guide/litmus-descriptions` to make the path only relative to the project root. But then I think it skips the sandbox.

tools/litmus2desc/test/rg_litmus_tests/CoRR.litmus
fsestini: To elaborate more on this: the PR proposes to just use the files in `readers-guide/litmus-tests` to run the cram test suite, thus eliminating the need for a separate `rg_litmus_tests` directory.

To your point, however, what we could do is to make `readers-guide/litmus-tests` a symlink to `catalogue/aarch64-readers-guide/tests`.

relokin: sorry I got the wrong impression that these files had become individual symlinks. But I agree, if I understand correctly, we have three copies of each test, what you are proposing removes one copy but ideally we should be able to do everything with a single copy.

fsestini: > sorry I got the wrong impression that these files had become individual symlinks.

No problem, and indeed the Github UI for this change is very confusing.

readers-guide/litmus-descriptions/2+2w.tex
fsestini: Good idea, I've now updated the commit description.

.gitignore
fsestini: Done.

fsestini:
Note to ourself for the future: we should probably also tweak the CI workflows a bit. Right now we are running the full `make test` workflow for this PR but I don't see the point in that.

relokin (approved):

tools/litmus2desc/test/readers-guide.t
fsestini: I double-checked, and indeed it does seem that `DUNE_SOURCEROOT` bypasses the sandbox.

PR #1959, opened 2026-08-10
[herd] Expose GCS effect event sets to cat

relokin:
This PR adds a dedicated `GCS` event set for herd and exposes AArch64 GCS barrier effects to cat files.

  The change:
  - Adds `is_gcs` to the explicit annotation interface.
  - Classifies AArch64 `NExp GCS` annotations as GCS events.
  - Adds the `GCS` set to the generated action sets exposed to cat.
  - Switches AArch64 barrier set generation to include the full barrier fold, exposing `GCSB.DSYNC` via its barrier tag.


fsestini (approved):
Thank you, this LGTM

relokin:
Thanks for the review Filippo!

PR #1960, opened 2026-08-11
Normalize machine-size names for 64-bit and 128-bit values

relokin:
This PR updates the machine-size terminology so that 64-bit values use Double/double and 128-bit values use Quad/quad.

  Changes include:

  - Rename the old 64-bit Quad size to Double.
  - Rename the old 128-bit S128 size to Quad.
  - Update architecture, parser, generator, scalar, variant, and litmus references to use the new names.
  - Keep s128 accepted as a compatibility alias where machine-size/variant tags are parsed.
  - Update C instruction tests and configs to use the new Quad variant naming.


ShaleXIONG:
The `diy` test should be fixed in 3b0d2eb2d44a9a06566eb96d4b094a7cecc8aeea in #1962 , please feel free to cherry pick. It is due to annotation name change from old `q0` to new `d0`.

PR #1963, opened 2026-08-12
[herd,asl,vmsa] Better handle double pte reads for hardware updates

HadrienRenaud:
This PR aims at restoring the correct behaviour of `herd -variant asl` on the long catalogue AArch64+VMSA tests.

To test it (warning very long):
```sh
make test.herd-asl.cata.aarch64-VMSA
```

Second commit is a bit irrelevant so if needed I'll split the PR in 2.

## Adding previously ignored tests

_first commit_

Those tests were previously not run because the ASL semantics for AArch64 (with VMSA activated) produced different results than the handwritten AArch64 semantics. After #1889, this is not the case anymore, so this PR adds them to the tests run with `make test-all`.

To test, use (after building and installing, with asl-pseudocode):
```sh
herd7 -speedcheck true -conf catalogue/aarch64-VMSA/cfgs/asl.cfg catalogue/aarch64-VMSA/tests/WW+RR-DMBST+DMBLD-SWP4+WREG.litmus catalogue/aarch64-VMSA/tests/RW+RR-DMBLD+DMBLD-SWP4+WREG.litmus
herd7 -speedcheck true catalogue/aarch64-VMSA/tests/WW+RR-DMBST+DMBLD-SWP4+WREG.litmus catalogue/aarch64-VMSA/tests/RW+RR-DMBLD+DMBLD-SWP4+WREG.litmus
```

## Ignoring the second call to PhysMemRead for hardware updates

_third commit_

There is a divergence in the model for hardware updates in herd and in the ASL in the Arm ARM: the herd semantics only perform a single read when the ASL in the Arm ARM performs 2 reads: a first, non-atomic, implicit effect reads the PTE, and if the PTE needs writing (in case of an hardware update), then a second atomic effect re-reads the PTE. If the values read in the two effects are different, then the translation is re-started.

Before this PR, we were generating a memory effect per call to `PhysMemRead`, but in order to avoid the expensive control-loop that ensures that the two values are equal, we were forcing the two values to be equal.

With this PR, we completely remove the second memory effect, and we return the same value for both calls to `PhysMemRead`.

This commit also rework the annotations placed on implicit effects. Some previous code was using the `accdesc.write` boolean flag of the access descriptor associated with the implicit accesses to understand if the explicit write was writing. This was an error as this flag only represents if the implicit access is writing. This commits thus defaults its value to true and inlines it.

## Flexible behaviour for DMBs

_second commit_

With #1890, we updated the ASL pseudocode to the version M.c of the Arm ARM. This version removes the uses of `DMB ISH`, `DMB OSH` and `DMB NSH`, only leaving `DMB SY`, and so #1890 was removing their semantics from `AArch64ASLSem.ml`.
This caused some problems for a few tests in the VMSA catalogue which now do not pass.
We fixed this by allowing the ASL semantics for AArch64 to simply call the handwritten semantics for those deprecated instructions.

### DIFF
--- a/catalogue/aarch64-VMSA/cfgs/asl.cfg
+++ b/catalogue/aarch64-VMSA/cfgs/asl.cfg
@@ -1,5 +1,6 @@
 variant asl,vmsa,strict,fatal,d128
 timeout 300
+
 ## ADR inside
 nonames S-I2V+dmb.st+data_Imp_TTD_R
 nonames S-I2V+dmb.st+ctrl_Imp_TTD_R
@@ -14,8 +15,8 @@ nonames MP-via_fault_STLR
 nonames R-via_fault_STR
 nonames CASAL+FAIL+FH
 nonames CASAL+FAIL2+FH
-## Allow vs. Forbidden kind, beacuse of two PTE reads?
-nonames RW+RR-DMBLD+DMBLD-SWP4+WREG,WW+RR-DMBST+DMBLD-SWP4+WREG
+
+## Producting graphs
 graph cluster
 squished true
 showevents noregs
--- a/herd/AArch64ASLSem.ml
+++ b/herd/AArch64ASLSem.ml
@@ -1753,6 +1753,8 @@ module Make (TopConf : AArch64Sig.Config) (V : Value.AArch64ASL) :
       match ii.A.inst with
        (* Specific -> get TLBI key *)
       | I_OP3 (V64,LSR,_,_,OpExt.Imm (12,0))
+       (* Unsupported by the Arm ARM since M.c: only SY is supported. *)
+      | I_FENCE (DMB ((NSH | OSH | ISH), _))
       (* Register or zero comparison, 64 is for addresses and pteval,
          32 is for instructions *)
       | I_OP3 ((V64|V32),SUBS,ZR,_,
--- a/herd/ASLSem.ml
+++ b/herd/ASLSem.ml
@@ -645,12 +645,6 @@ module Make (Conf : Config) = struct
     let read_memory ii ~n ~addr =
       do_read_memory ii addr n aneutral aexp avir
 
-    let read_pte ii ~n ~addr =
-      (* We do all the operations with 64 bits, even if the argument passed is different. *)
-      let* _ = n in
-      do_read_memory ii addr (M.unitT (V.intToV 64))
-        aneutral (AArch64Explicit.(NExp Other)) apte
-
     let read_memory_gen ii ~n ~addr ~accdesc ~access =
       let* accdesc = accdesc and* access = access in
       do_read_memory ii addr n (accdesc_to_annot Read accdesc)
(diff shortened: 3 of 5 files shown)


HadrienRenaud:
Extended tests running in CI [here](https://github.com/HadrienRenaud/herdtools7/actions/runs/31689102029) ✅ passed

`make test-all-asl` passes on my machine (although I have to kill some processes because timeout does not work)

maranget:
> This commit also rework the annotations placed on implicit effects. Some previous code was using the `accdesc.write` boolean flag of the access descriptor associated with the implicit accesses to understand if the explicit write was writing. This was an error as this flag only represents if the implicit access is writing. This commits thus defaults its value to true and inlines it.

This change introduces a new difference between native and asl mode. Consider the following test:
```
AArch64 LDR-HD
TTHM=HD
Variant=vmsa
{
[PTE(x)]=(af:0);
int x=1;
0:X0=x;
}
  P0         ;
 LDR W1,[X0] ;
forall 0:X1=1
```
In native mode the write labeled `b:` to `PTE(x)` resulting from the hardware update holds the annotation `AF`, 
[NATIVE.pdf](https://github.com/user-attachments/files/31064775/NATIVE.pdf)
while the annotation is `AFDB` in asl mode.
[ASL.pdf](https://github.com/user-attachments/files/31064761/ASL.pdf).
The native annotations looks more intuitive: `AF` meaning "update the AF flag", `AFDB` "update both flags". Notice that the read of PTE(x) holds the annotation `AFDB` in both diagrams

As far as I could  check and remember, those annotations are used to build the cat sets `AF` and `DB` that occur in `aarch64utils.cat`. Although the usage applies to writes and does not seem too critical I do.not feel very comfortable with the change.

maranget:
So much for intuition. Consider the following test:
```
AArch64 STR-HD
TTHM=HD
Variant=vmsa
{
[PTE(x)]=(af:0);
int x=0;
0:X0=x;
0:X1=1;
}
  P0         ;
 STR W1,[X0] ;
forall 0:X1=1
```
Here is the diagram in native mode:  [NATIVE2.pdf](https://github.com/user-attachments/files/31065370/NATIVE2.pdf). The annotation of write labelled b is `AFDB` although the write sets `AF` only...

We should probably reconsider why we are using those `AF`, `DB` and `AFDB` annotations....

HadrienRenaud:
Hi @maranget,

Thanks for looking into this.

I believe this is not a new behaviour. My experiments show that this is already the case on herd's master. See 
[LDR-HD-ASL-master.pdf](https://github.com/user-attachments/files/31065674/LDR-AF-ASL-master.pdf).

Indeed, what I tried to explain in the PR description is that the argument `is_write` is always set: the `accdesc.write` flag passed to `PhysMemRead`/`PhysMemWrite` only refers to whether the Implicit Effect is writing, not if the associated Explicit Effect is writing.

> We should probably reconsider why we are using those AF, DB and AFDB annotations...

I think probably on the ASL side, we should assume that we cannot, as for today, produce the correct annotations. This would need adding a new field in `AccessDescriptor` for those cases, which we can't really do from herd.

From the AArch64 side, we can also notice that the Implicit Read Effect is always tagged with the maximalist annotations, independently of whether the flags will be updated: if HD is set, it will be `AFDB`, else if AF is set, then `AF`, otherwise no annotation.

maranget:
Ah I missed that accdesc.write was always true. Anyway, I think the distinction between AF, DB, and AFDB annotations is now irrelevant. It was relevant when the execution of an instruction could perform flag updates in several steps as suggested by the only effective usage of the AF and DB sets in the cat model: the definition of AFtoDB.
```
let AFtoDB = same-instance & ((AF\DB) * (DB\AF))
let co0=co0 | AFtoDB

include "cos.cat"
```

maranget (approved):
LGTM, but somem day we should probably merge the AF,DB and AFDB annotations into one (say AFDB ?).

HadrienRenaud:
Thanks Luc, I've just rebased and will merge when CI pass.

PR #1964, opened 2026-08-14
[litmus] Two fixes for `-mode presi`

maranget:
This PR features two fixes:
  1. Use a local definition of the full fence `mbar` in the C implementation of the reverse sense synchronisation barrier.
  2. Make the difference between observed and displayed locations. Namely, those may differ when a filter refers to some locations and the final condition does not, as in the following test.
 
```
RISCV ISA03+SIMPLE
 
(* ISA V2.3, Fig 1.3, spinlock, adapt, test mutual exclusion *)

{
    0:a0=lock; 0:a5=a;
    1:a0=lock; 1:a5=a;
}
   P0                      | P1                      ;
  li t0,1                 | li t0,1                 ;
  amoswap.w.aq t0,t0,(a0) | amoswap.w.aq t0,t0,(a0) ;
  lw t1,0(a5)             | lw t1,0(a5)             ;
  addi t1,t1,1            | addi t1,t1,1            ;
  sw t1,0(a5)             | sw t1,0(a5)             ;
  amoswap.w.rl x0,x0,(a0) | amoswap.w.rl x0,x0,(a0) ;

filter (0:t0=0 /\ 1:t0=0)
forall (a=2)
```

### DIFF
--- a/litmus/libdir/_riscv/barrier.c
+++ b/litmus/libdir/_riscv/barrier.c
@@ -30,18 +30,22 @@ static void barrier_init (sense_t *p,int n) {
 
 /* C coding, play it safe with mbar's */
 
+inline static void mbar_loc(void) {
+  asm __volatile__ ("fence rw,rw" ::: "memory");
+}
+
 __attribute__ ((noinline)) static void barrier_wait(sense_t *p) {
-  mbar() ;
+  mbar_loc() ;
   int sense = p->sense ;
-  mbar();
+  mbar_loc();
   int rem = __sync_add_and_fetch(&p->c,-1) ;
-  mbar() ;
+  mbar_loc() ;
   if (rem == 0) {
     p->c = p->n ;
-    mbar();
+    mbar_loc();
     p->sense = 1-sense ;
   } else {
     while (p->sense == sense) ;
   }
-  mbar() ;
+  mbar_loc() ;
 }
--- a/litmus/libdir/_hash.c
+++ b/litmus/libdir/_hash.c
@@ -17,7 +17,7 @@
 /* Notice: this file contains public domain code by Bob Jenkins */
 
 typedef struct {
-  log_t key ;
+  hashlog_t key ;
 #ifdef STATS
   param_t p ;
 #endif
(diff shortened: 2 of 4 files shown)


litmus/libdir/_riscv/barrier.c
```diff
@@ -30,18 +30,22 @@ static void barrier_init (sense_t *p,int n) {
 
 /* C coding, play it safe with mbar's */
 
+inline static void mbar_loc(void) {
+  asm __volatile__ ("fence rw,rw" ::: "memory");
+}
```
psafont, line 35: I see this function is already available at `mbar.c`. I guess the previous code here relied on that definition being dumped on the same file before this one is.

What made presi mode diverge from the normal mode? And what are the reason that block `mbar.c` from being dumped in presi mode before `barrier.c` instead of having a local definition in barrier.c? Usually I tend to avoid duplicating code to avoid the code from diverging.

I would appreciate a reference to what's the problem being solved in the commit message, there's no hint that this code is being introduced to fix presi mode by looking at the information made available in git alone, which is very nice to have while inspecting the code's history in a repository.

maranget:
For some reason I do not see your comment on GitHub.

There is a difference between *observed* and *displayed* locations: locations that are appear in the filter and not in the final condition are observed and yet not displayed (see example in second commit message, which by the way is wrong)

Observed locations must be shared amongst threads, this is the purpose of the `out` field of the `ctx_t` record.
Displayed locations are saved in the histogram data structure (a hastable in presi and kvm modes).

Locations in histogram must be strictly limited to displayed locations, otherwise we can have duplicates in listed final states, due tu hidden parameters.

Notice that when all observed locations are displayed there is not specific `hashlog_t` type and field.

—Luc

psafont (commented):
I've only one comment for the first commit.

The second one is interesting to see how presi mode needs to gather the same information as the normal mode, while different threads run at the same time, which introduces a more complicated implementation.

psafont:
> For some reason I do not see your comment on GitHub.

Sorry about that, I deleted the comment since I did not intend it to end up it github, only to the local application I'm using.

Thank you very much for taking the time to explain the purpose of the new type, and how it related to the modes.

maranget:
I guess I have been afraid of including the definition of `mbar` twice.

maranget:
There is another reason, “barrier.c” uses the mbar primitive only when coded in C (and not in inline assembly). This is the case of RISCV only. So I guess I wanted to avoid useless insertion in the case of other architectures.

litmus/libdir/_riscv/barrier.c
maranget: > I would appreciate a reference to what's the problem being solved in the commit message, there's no hint that this code is being introduced to fix presi mode by looking at the information made available in git alone, which is very nice to have while inspecting the code's history in a repository.

i have just done that, with an explanation of why we define mbar locally instead of inserting the contents of the file mbar.c.

psafont (approved):

maranget:
Merged, thanks @psafont.

PR #1966, opened 2026-08-18
[asl] Capture of ASL print output and error messages

hrutvik:
Permit OCaml-level capture of ASLRef output.

- Covers ASL printed output (via `print`/`println`) and error messages reported by ASLRef.
- Enabled via the `Runner` module: two buffers (`Buffer.t`) can be supplied to enable capture.
- Parametrises the interpreter and the error printer to achieve this.

### DIFF
--- a/asllib/Native.ml
+++ b/asllib/Native.ml
@@ -487,40 +487,42 @@ module DeterministicInterpreter (I : Instrumentation.SEMINSTR) =
 
 module DeterministicInterpreterNoInstr =
   DeterministicInterpreter (Instrumentation.SemanticsNoInstr)
-
-module DeterministicInterpreterSingleSetInstr =
-  DeterministicInterpreter (Instrumentation.SemanticsSingleSetInstr)
+(** Fast path for interpretation: no instrumentation and no custom output *)
 
 let exit_value = function
   | NV_Literal (L_Int i) -> i |> Z.to_int
   | v -> mismatch_type v [ integer' ]
 
-let interpret ?instrumentation ?out_buffer static_env main_name ast =
-  match instrumentation with
-  | Some true ->
+let interpret ?(instrumentation = false) ?out_buffer static_env main_name ast =
+  match (instrumentation, out_buffer) with
+  | false, None ->
+      let res =
+        DeterministicInterpreterNoInstr.run_typed static_env main_name ast
+      in
+      (exit_value res, [])
+  | false, Some buf ->
+      let module Interpret =
+        Interpreter.Make
+          (DeterministicBackend)
+          (NativeConfig
+             (Instrumentation.SemanticsNoInstr)
+             (struct
+               let out_buffer = Some buf
+             end))
+      in
+      let res = Interpret.run_typed static_env main_name ast in
+      (exit_value res, [])
+  | true, _ ->
       let module B = Instrumentation.SemanticsSingleSetBuffer in
       B.reset ();
-      let res =
-        DeterministicInterpreterSingleSetInstr.run_typed static_env main_name
-          ast
+      let module Interpret =
+        Interpreter.Make
+          (DeterministicBackend)
+          (NativeConfig
+             (Instrumentation.SemanticsSingleSetInstr)
+             (struct
+               let out_buffer = out_buffer
+             end))
       in
+      let res = Interpret.run_typed static_env main_name ast in
       (exit_value res, B.get ())
-  | Some false | None -> (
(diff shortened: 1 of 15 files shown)


asllib/Runner.ml
```diff
@@ -49,6 +49,8 @@ type args = {
   no_stdlib0 : bool;
   v0_use_split_chunks : bool;
   version_eac1 : bool;
+  capture_output : (Buffer.t * Buffer.t) option;
```
HadrienRenaud: Wondering if passing channels would be more flexible, but I don't think this is critical for here.

asllib/tests/capture_output.ml
```diff
@@ -0,0 +1,20 @@
...
+      default_args with
+      exec = true;
+      capture_output = Some (stdout_buf, stderr_buf);
+      files = [ (NormalV1, "capture_output.asl") ];
+    }
+
+let code = Runner.safe_run args
+
+let _ =
+  Printf.printf "%d\n\nstdout:\n%s\n\nstderr:\n%s\n" code
```
HadrienRenaud:
```suggestion
  Printf.printf "exit code: %d\n\nstdout:\n%s\n\nstderr:\n%s\n" code
```

asllib/tests/capture_output.ml
```diff
@@ -0,0 +1,20 @@
...
+    {
+      default_args with
+      exec = true;
+      capture_output = Some (stdout_buf, stderr_buf);
+      files = [ (NormalV1, "capture_output.asl") ];
+    }
+
+let code = Runner.safe_run args
+
+let _ =
```
HadrienRenaud: nit:

```suggestion
let () =
```

asllib/Native.ml
```diff
@@ -485,7 +495,7 @@ let exit_value = function
   | NV_Literal (L_Int i) -> i |> Z.to_int
   | v -> mismatch_type v [ integer' ]
 
-let interpret ?instrumentation static_env main_name ast =
+let interpret ?instrumentation ?out_buffer static_env main_name ast =
   match instrumentation with
```
HadrienRenaud: A few small refactos on this function:
1. do a single match on `(instrumentation, out_buffer)`, and maybe only provide the case `(false, None)` as statically built;
2. put a default value of `false` for `instrumentation`, this way we avoid the strange equivalence between `None` and `Some false`.

asllib/StaticInterpreter.ml
```diff
@@ -37,6 +37,7 @@ module InterpConf = struct
   let display_call_stack_on_error = false
   let track_symbolic_path = false
   let bit_clear_optimisation = false
+  let out_buffer = None
```
HadrienRenaud, line 43: Maybe we should have one and check that it's empty at the end? Or is it guaranteed by type-system that this cannot happen?

HadrienRenaud (approved):
This looks good to me, I only have minor refactos that I think should be addresses in this PR.

---

For reference I was wondering if we could have a more flexible/generic type than `Buffer.t` in signatures, but there is not a good abstraction in the stdlib so I guess we can juste wait until this is needed.
In more details, in the end it comes down to replacing the
```ocaml
  capture_output : (Buffer.t * Buffer.t) option
```
By 2 optional fields of the type:
```ocaml
type captured_output = Buffer of Buffer.t
```
With the understanding that when we need to extend it we should add more variants as needed:
```ocaml
type captured_output =
  | Buffer of Buffer.t (** output to buffer *)
  | Outchan of out_channel (** output to channel *)
  | Removed (** output is a noop *)
  | Forbidden (** output is an error *)
```

But at this point more questions arise and I think it is too complicated and we will cross that bridge when we get there.

(e.g. at this point we should pass down a module or an object and not a value and abstract away the operations)

asllib/StaticInterpreter.ml
hrutvik: It should be guaranteed by the type system, in that `print` is considered impure

HadrienRenaud: Maybe good to add a little comment about this?

asllib/Native.ml
hrutvik: I believe I've implemented this in the latest commit - please could I ask for a re-review of this part?

hrutvik: Discussed offline - ideally a branch to select the modules would reduce duplication, but this is not a valid module-expr unfortunately. We'll move ahead with this implementation.

PR #1968, opened 2026-08-18
[asl] Tidying in errors

hrutvik:
- Remove various unused error reporting bits
- Slightly reorganise `error.ml` - group into modules
- Refactor `Types.lowest_common_ancestor` to avoid an optional return type, and directly raise an exception in the `None` case

Factors out part of #1935.


HadrienRenaud (approved):
Looks good to me thanks

---

On the first commit, `Operations.value_as_int` was removed in a previous PR.

PR #1969, opened 2026-08-18
[gen] Reduce `diy7` cycle-generation overhead

ShaleXIONG:
This PR reduces repeated work in the `diy*` cycle search and streamlines I/O in the test-preparation pipeline.

The DumpAll path streamlines the I/O pipeline by splitting test preparation into clear stages: atom variation, edge resolution, normalisation, duplicate detection, test construction, and final dumping.

- In `DumpAll`, we normalise the cycle and then check duplication. This eliminates various of `rotation` which was necessary without normalisation. 
- Prepared tests are reused across scope and name variants rather than being rebuilt before each output.

The cycle search now precomputes compatibility between relaxation chunks and rejects candidates that exceed process or instruction budgets before invoking DumpAll. This reduces repeated computation in the `can_precede` hot spot. Each selected relaxation is also counted once against the cycle-size budget.

For `diy7 -conf gen/libdir/forbidden.conf -size 5 -stdout`, the before-and-after results are:

| Branch | Wall times (seconds) | Mean |
|---|---|---:|
| After | 242.51, 240.75, 240.72, 241.16, 238.99, 237.39, 239.53, 238.44, 239.43, 238.52 | 239.744s |
| Before | 445.85, 440.64, 449.44, 441.91, 456.27, 428.40, 461.59, 451.11, 434.01, 427.93 | 443.715s |

After this change, the benchmark averages 239.744s, compared with 443.715s before: a 1.851× speed-up and a 45.97% reduction in wall time. The benefit is expected to grow as `-size` and `-ins` increase, expanding both the cycle-search space and the number of I/O operations.


gen/common/edge.ml
```diff
@@ -989,24 +989,24 @@ let fold_tedges f r =
 
 
 
-  let varatom es f r =
-    let rec var_rec ves es r = match es with
-    | [] -> f (resolve_edges (List.rev ves)) r
+  let varatom es =
```
fsestini: What is the reason for making this function return a list, instead of a fold like before?

Judging from this PR, `varatom` is only called in `dumpAll.ml` like this:

```ocaml
T.E.varatom es
          |> List.fold_left ...
```

In other words, the PR's version of `varatom` is allocating memory to materialise a whole list of lists at once, only for that list of lists to immediately be folder over. Wouldn't it be more efficient, then, to have `varatom` return a fold, like before?

ShaleXIONG: Yes you are right, the callback version gives me another 1%~ performance. I will recover the callback version. I was thinking the function are mostly identity from input list `l` to `[l]`.  The change is in https://github.com/herd/herdtools7/pull/1969/changes/c82e810fdd888cbddfdfcb66f7b84eefe3fffc83

gen/common/edge.ml
```diff
@@ -990,22 +990,22 @@ let fold_tedges f r =
...
               begin match dir_src e with
               | Dir d ->
                   A.varatom_dir d
-                    (fun a r -> var_rec ({e with a1=a}::ves)  es r)
-                    r
-              | NoDir ->  var_rec (e::ves) es r
+                    (fun a k -> var_rec ({e with a1=a}::ves)  es k)
+                    k
+              | NoDir ->  var_rec (e::ves) es k
               | Irr ->  assert false (* resolved at this step *)
```
TiberiuBucur, line 1006: This assertion fails when running `diy7 -arch AArch64 -size 2`. Presumably because the edges are not resolved prior to calling this function anymore.

ShaleXIONG: Fix, with one extra `resolve_edge` before `varatom`. The performance remains the same after.

gen/dumpAll.ml
```diff
@@ -246,8 +213,8 @@ module Make(Config:Config)(T:Builder.S)
         let cys = List.rev_map T.E.parse_edges cys in
         List.fold_left
           (fun k es ->
-            let xxx,_,k = comp_sig k es in
-            { k with sig_set = StringSet.add xxx k.sig_set;} )
+            let _,k = have_seen k es in
```
TiberiuBucur: Cycles in es need normalisation prior to this call.

gen/dumpAll.ml
```diff
@@ -166,51 +166,15 @@ module Make(Config:Config)(T:Builder.S)
...
-        if StringSet.mem xxx sigs.sig_set then
-          true,es,sigs
-        else
-          false,es,{ sigs with sig_set = StringSet.add xxx sigs.sig_set; }
+          let t = Array.of_list norm in
+          let xxx,sigs = sig_of_shift sigs t 0 in
+          if StringSet.mem xxx sigs.sig_set then
+            true,sigs
+          else
+            false,{ sigs with sig_set = StringSet.add xxx sigs.sig_set; }
```
TiberiuBucur, line 162: If we're removing this part, we're assuming that all cycles that make it to this are already in their canonical form. This is fine for most cases because the `check_dump` function calls `normalise` before this code is executed. However, the "blacklisted" cycles passed through `Config.no` end up in `sigs_init`, which does not normalise them before calling `have_seen`, therefore we might have duplicate entries modulo canonicalisation.

gen/dumpAll.ml
TiberiuBucur, line 207: Cycles in es need normalisation prior to this call.

```suggestion
            let es = T.E.resolve_edges es |> normalise in
            let _,k = have_seen k es in
```

gen/dumpAll.ml
TiberiuBucur: The `k` parameter here is no longer needed now that cycles are always canonical. In fact, the whole `sig_of_shift` name might be inaccurate, since it's no longer about a shift of a cycle, but a signature of a (already canonical) cycle.

TiberiuBucur: This comment is now stale.

gen/alt.ml
```diff
@@ -592,9 +616,11 @@ module Make(C:Builder.S)
     let last_check_call rej aset f rs po_safe res k =
       if is_empty_list res then k else
           let lst = Misc.last res in
-          if can_precede aset po_safe lst res then
-            let es = List.map snd res in
-            let le = List.flatten es in
+          let le = List.map snd res |> List.flatten in
+          let proc_count = procedure_count 0 le in
+          if proc_count > 0 && proc_count <= O.nprocs &&
```
clemenslosbichler-cloud: Filters look like they drop cycles that used to come out:
proc_count > 0 rejects all-internal cycles in every mode. The mode gate below only applies all_int for Default|Sc|ppo|mix, so all-internal cycles look intended to survive under Uni|Thin|Free|Critical|Transitive|Total. 

Should the <= O.nprocs bound stay and > 0 stay?

gen/alt.ml
```diff
@@ -385,23 +402,25 @@ module Make(C:Builder.S)
...
+        check_cycle r_suff reject
       then
-        let suff = r::suff
-        and n = n-sz r in
-        if O.verbose > 2 then eprintf "CALL: %i %s\n%!" n (pp_ess suff) ;
+        let n = n-1 in
+        if O.verbose > 2 then eprintf "CALL: %i %s\n%!" n (pp_ess r_suff) ;
         let k =
           if
+            can_precede_relax (Misc.last r_suff) r_suff &&
```
clemenslosbichler-cloud: pre-check resolves against r_suff's head, but the candidate emitted at 423 is prefix@r_suff.
diy7 -prefix Rfe -safe 'PodRW Coe' -size 3 drops Rfe ; PodRW ; Coe. Possibly already handled by can_prefix at 421?

gen/alt.ml
ShaleXIONG: Yes you are right here. Will update.

gen/alt.ml
ShaleXIONG: Yes, I will delay the check if `prefix` is not de facto empty. Probably best way is to fold the check in `can_prefix`.

gen/dumpAll.ml
ShaleXIONG: I think the best way to fix to to call normalise in `sigs_init` as you also mentioned. It is very low cost and only need to call once. In the hot path of `diy7`, this `Config.no` field is de facto empty.

gen/dumpAll.ml
ShaleXIONG: Here each `es`  from `cys` will be all normalised.

gen/dumpAll.ml
ShaleXIONG: I will remove them

ShaleXIONG: Yes you are right, since we now use normalised form to compute.

TiberiuBucur (approved):
LGTM, approved

PR #1970, opened 2026-08-18
[tools] Enhance compatibility of tools that read logs

maranget:
The printing of faults has changed over time:
  1. Initially faults did not hold a type string
  2. Introduction of type string.
  3. MMU faults are prefixed with "D-" or "I-" to specify the memory accessed (data or instruction).

Tools that compare or compute on logs , such as **mcompare7** and **msum7** are impacted. It is always possible to use the `-faulttype false` option to ignore fault types for tools to produce correct results.

However those results are not as precise as they should. This PR is an attempt to fix the problem by:
1. Ignore the "D-" and "I-" prefixes while comparing with a non-prefixed type string (standard behaviour).
2. Consider that all MMU faults are from data (activated by option `-mmu-faults-as-data true`).

Notice that the default for all tools are `-faulttype true mmu-faults-as-data true `.


tools/HashedFault.ml
```diff
@@ -46,14 +46,24 @@ let as_hash h = h.Hashcons.hkey
...
+  | _,_ -> String.compare s1 s2
+
 let compare h1 h2 =
   let p1,lab1,x1,ft1 = as_tt h1
   and p2,lab2,x2,ft2 = as_tt h2 in
   match Misc.int_compare p1 p2 with
   | 0 -> begin  match HashedStringOpt.compare lab1 lab2 with
          | 0 -> begin match HashedStringOpt.compare x1 x2 with
                 | 0 -> begin match HashedStringOpt.as_t ft1, HashedStringOpt.as_t ft2 with
-                       | Some ft1, Some ft2 -> String.compare ft1 ft2
```
psafont: Maybe you're interested in removing these nested patterns. It can be done with a simple function:

```ocaml
let ( <?> ) a b = if a = 0 then b else a in
let compare_string_opts a b =
  HashedStringOpt.as_t a, HashedStringOpt.as_t b with
  | Some a, Some b -> compare_ftype a b
  | None, _ | _, None ->
      if !warn_once then begin
        Warn.warn_always "Comparing faults with and without fault type, \
                          assuming same type";
        warn_once := false;
      end;
      0
in
    Misc.int_compare p1 p2
<?> HashedStringOpt.compare lab1 lab2
<?> HashedStringOpt.compare x1 x2
<?> compare_string_opts ft1 ft2
```

lib/checkName.ml
```diff
@@ -55,7 +55,12 @@ let parse_int32 int32 =
 let parse_faulttype ft =
    ("-faulttype", Arg.Bool (fun b -> ft := b),
     Printf.sprintf
-      "<bool> consider fault types, default %b" !ft);
+      "<bool> consider fault types, default %b" !ft)
+
+let parse_datafault ft =
+   ("-datafault", Arg.Bool (fun b -> ft := b),
```
fsestini: I wonder if we could use a more informative name for this option.  Maybe something along the lines of `-mmu-faults-as-data`?

lib/checkName.ml
```diff
@@ -55,7 +55,12 @@ let parse_int32 int32 =
...
+    Printf.sprintf
+      "<bool> all MMU faults are from data, default %b" !ft)
```
fsestini: > all MMU faults are from data

If I understand correctly, this is not quite the case. For example, a fault that already has a `I-` prefix is _not_ treated as being from data, regardless of the status of this option. If so, can we make this documentation text more precise?

lib/faultType.ml
```diff
@@ -209,3 +209,20 @@ module No = struct
...
 let is = AArch64.is
+
+let has_diprefix s =
+  String.length s > 2
+  &&
+  begin
+    match s.[0],s.[1] with
+    | ('D'|'I'),'-' -> true
+    | _,_ -> false
+  end
```
fsestini: Personal readability preference, but `String.starts_with` could also work here:
```suggestion
let has_diprefix s =
  String.(length s > 2 && (starts_with ~prefix:"D-" || starts_with ~prefix:"I-"))
```

lib/faultType.ml
```diff
@@ -209,3 +209,20 @@ module No = struct
...
+
+let split_at_diprefix s =
+  let len = String.length s in
+  if len > 2 then
+    match String.sub s 0 2 with
+    | ("D-"|"I-") as pref -> pref,String.sub s 2 (len-2)
+    | _ -> "",s
+  else "",s
```
fsestini: `split_at_diprefix` could reuse `has_diprefix` for checking the presence of a DI prefix.

Moreover, this function uses `""` as a sentinel to signal "no prefix", which feels stringly typed, since callers never actually use the first pair component and only need the stripped suffix when a prefix exists. Therefore, returning an optional like this would be clearer IMO:

```ocaml
  let strip_diprefix s =
    if has_diprefix s then
      let len = String.length s in
      Some (String.sub s 2 (len - 2))
    else
      None
```

lib/faultType.mli
```diff
@@ -59,3 +59,7 @@ module No : S
 
 (* For parse disambiguation  *)
 val is : string -> bool
+
+(* For compatibility *)
```
fsestini: I’m not sure this comment gives future readers enough context to understand what "compatibility" means here without already knowing the history behind this PR. By itself I feel like it may raise more questions than answers.

Can I suggest to either:

- expand this comment to explain, in brief, the issue with `MMU:...` vs `D-MMU:.../I-MMU:...`, or
- remove it and document the individual functions instead

tools/HashedFault.ml
```diff
@@ -46,14 +46,24 @@ let as_hash h = h.Hashcons.hkey
 
 let warn_once = ref true
 
+let compare_ftype s1 s2 =
```
fsestini: This function returns `0` on a pair of inputs `"MMU:foo"` and `"I-MMU:foo"`, which seems at odds with the idea of treating un-prefixed `MMU` faults as data, since if the inputs were `"D-MMU:foo"` and `"I-MMU:foo"`, the result wouldn't be 0.

I understand that under `-datafault true`, the input `"MMU:foo"` would be rewritten into `"D-MMU:foo"` before reaching `compare_ftype`. However, this makes the function’s semantics depend on a pre-processing step elsewhere in the parsing pipeline, which feels brittle and hard to reason about.

Moreover, the function seems to not be transitive:
- `compare_ftype "I-MMU:foo" "MMU:foo" == 0`
- `compare_ftype "MMU:foo" "D-MMU:foo" == 0`
- `compare_ftype "I-MMU:foo" "D-MMU:foo" != 0`

This goes against the convention that `compare` functions are a total order.

In light of this, I would suggest to document this function's assumptions w.r.t. its inputs, remove it from `compare`, and change its name to something else that doesn't suggest it's a genuine comparison function. Alternatively, we could keep it in `compare`, but we should then make sure it induces a total order.

tools/lexLog_tools.mll
```diff
@@ -262,7 +268,11 @@ and pline bds fs abs = parse
      let loc = Misc.map_opt Constant.old2new loc in
      let ftype =
        if O.faulttype then
-         match ftype with Some "kvm" -> None | _ -> ftype
+         Option.bind
+           ftype
+           (function
+             | "kvm" -> None
+             | f -> Some (norm_fault f))
```
fsestini, line 275: Personal preference, but I'd find a pattern match more readable, here.

tools/logState.ml
```diff
@@ -1297,6 +1326,39 @@ let simple_same out1 out2 t1 t2 k =
       else do_rec r1 r2 k in
   do_rec t1.s_tests t2.s_tests k
 
+(* Check presence of fault with no fault type in state *)
+let has_no_fault_type st =
+  let rec exists_no_type fs =
+   let open HashedFaults in
+   let open HashedFault in
+    match fs.Hashcons.node with
```
fsestini: Can we have two-space indendation here?

tools/logState.ml
```diff
@@ -1323,13 +1385,25 @@ let rec diff_not_empty  xs ys = match xs,ys with
...
 
-
 let diff_simple_states xs ys =
-  not (List.equal ( == ) xs ys)
+  select_diff
+    (fun xs ys ->
+      let xs = to_parsed_sts xs
+      and ys = to_parsed_sts ys in
+      do_diff_states xs ys [] true |> Misc.consp
+      || do_diff_states ys xs [] true |> Misc.consp)
```
fsestini: This computes the whole diff lists only to then check if they are non-empty. Not a blocker for this PR, but if this has the potential to become a performance issue, I'd consider revisiting this logic at some point.

fsestini (commented):
Hi @maranget , I've left some comments.

Could you also please provide some example inputs that I can use to reproduce the issue, and to see for myself how this PR impacts the output?

On a similar note, I think this PR should also include some dune cram tests, as those are designed to check expected input-output behaviour of CLI programs, which seems appropriate here. In particular it would be useful to have tests that show `mcompare` and `mcmp` side by side on equal inputs.

tools/logState.ml
```diff
@@ -381,28 +381,57 @@ let select_absent st1 st2 =
...
  * the presence of fault type implies that explicit
  * absent faults are also here (if some fault is
  * absent, of course).
+ * Notice that MMU fault types have been made even
+ * more precise by prefixing then with "D-" or "I-".
  *)
 let select_newer st1 st2 =
-  if st1 == st2 || state_has_fault_type st1 then st1
-  else if state_has_fault_type st2 then st2
+  if st1 == st2 then st1
```
fsestini: I understand this point is slightly tangential to this PR, but I think it's quite risky and error-prone to use `==` "openly", like so.

`==` may be semantically correct here because it is applied to two values of type `parsed_st`, and `parsed_st` happens to be made of two components where the first is irrelevant and the second is hash-consed. However this is brittle, because any change to `parsed_st` that invalidates these assumptions will end up silently breaking all usages of `==`, with no help from the compiler.

`==` is completely fine _as an implementation detail_, but I think its usage should be narrowly-scoped and hidden behind equality predicates that don't require callers to make assumptions about implementation (or that make those assumptions clear). For instance, my preference would be something like:

```ocaml
(* hashcons.mli *)
module type S = sig
  ...
  val equal_node : node -> node -> bool
  (** O(1) equality comparison. *)
end

(* hashcons.ml *)
  ...
  let equal_node = (==)
  ...

(* logState.ml *)
let equal_parsed_st st1 st2 = HashedState.M.equal st1.p_st st2.p_st
```

tools/HashedFault.ml
maranget: I have been tempted to follow your suggestion. Finally I did not because your code always evaluates all comparisons. I seized the opportunity to  reduce indentation and to fix a small glitch: ` (None,None)` should not trigger the warning.

lib/checkName.ml
maranget: Fixed, thanks.

lib/checkName.ml
maranget: Yes we can! What about _all non-specific MMU faults are from data (i.e. are implicitly prefixed with "D-")_

lib/faultType.ml
maranget: I am fine with this change. We can even get rid of `String.length > 2`.

lib/faultType.ml
maranget: Of course, `strip_diprefix` should reuse `has_diprefix". Ok for the new return type.

lib/faultType.mli
maranget: I have expanded the comment, as you suggest.

tools/HashedFault.ml
maranget: > This function returns `0` on a pair of inputs `"MMU:foo"` and `"I-MMU:foo"`, which seems at odds with the idea of treating un-prefixed `MMU` faults as data, since if the inputs were `"D-MMU:foo"` and `"I-MMU:foo"`, the result wouldn't be 0.
> 
> I understand that under `-datafault true`, the input `"MMU:foo"` would be rewritten into `"D-MMU:foo"` before reaching `compare_ftype`. However, this makes the function’s semantics depend on a pre-processing step elsewhere in the parsing pipeline, which feels brittle and hard to reason about.

Having the **mcompare7** tools to behave the same regardless of the  setting of `-mmu-faults-as-data`  is not our objective. The `-mmu-fault-as-data true` setting is a simple to understand measure that handle the most frequent case of logs of ruens that would combine the ifetch and vmsa variants.  The setting  `-mmu-fault-as-data false` offers an alternatibe that works when instruction faults and non-specified faults are both present and that may work when data faults are additionaly present.
> 
> Moreover, the function seems to not be transitive:
> 
> * `compare_ftype "I-MMU:foo" "MMU:foo" == 0`
> * `compare_ftype "MMU:foo" "D-MMU:foo" == 0`
> * `compare_ftype "I-MMU:foo" "D-MMU:foo" != 0`
> 
> This goes against the convention that `compare` functions are a total order.
> 
> In light of this, I would suggest to document this function's assumptions w.r.t. its inputs, remove it from `compare`, and change its name to something else that doesn't suggest it's a genuine comparison function. Alternatively, we could keep it in `compare`, but we should then make sure it induces a total order.

Nice catch, obserev that is absence of that this absence of transitivity predates this PR. Consider
  *  `compare Fault(L0,x,"MMU:Permission") Fault(L0,x) == 0`
  *  `compare Fault(L0,x) Fault(L0,x,"MMU:Translation") == 0`
  *  `compare Fault(L0,x,"MMU:Permission") Fault(L0,x,"MMU:Translation") != 0`

 The compare function is used:
  1. To normalise list of final states. In case "old" and "new" faults (_i.e._ no fault type and with fault types, or unspecified MMU faults and "D-"/"I-" prefixed MMU faults) do not mix, the sorted list of final states makes sense, as `compare` is transitive on those restricted sets.
  2. For merging (adding) states. In that case it is important for the "new" fault to take precedence over "old" ones (see `select_newer`), to maintain the invariant that "old" and "new" faults do not mix.
  3. Difference need a special treatement because a non-typed faults may equal  to several faults.

tools/lexLog_tools.mll
maranget: I usually prefer explicit pattern match too, but I had attempted a change here...

tools/logState.ml
maranget: Fixed.

tools/logState.ml
maranget: The `do_diff` function is rather complex, I see if I can write it in a generic style to both compute the difference (for **mcompare7**) and check teh difference for being non-empty (for **mcmp7**). Notice that this code applies to comparing legacy logs and new logs, a rare occurrence in my experience.

tools/logState.ml
maranget: Well, as  using `==` for hashconsed structure is the only use of `==` that I tolerate, I'll follow your suggestion that goes in that very direction.

tools/HashedFault.ml
psafont: I guess that could be achieved with a small tweak to delay the evaluation of the trailing expression, but then it's not as nice to use as it makes the intent harder to see.

```ocaml
let ( <?> ) a b = if a = 0 then b () else a in

(***)

    Misc.int_compare p1 p2
<?> (fun () -> HashedStringOpt.compare lab1 lab2)
<?> (fun () -> HashedStringOpt.compare x1 x2)
<?> (fun () -> compare_string_opts ft1 ft2)
```

tools/logState.ml
maranget: Finally, I have written a specific `do_diff_states_empty` function that follows `do_diff_states` structure.

lib/faultType.mli
maranget: Added many comments in a dedicated commit.

tools/HashedFault.ml
maranget: Hi @psafont. I have not followed your suggestion as it implies to compute all elementary "compare" functions.

maranget:
> Could you also please provide some example inputs that I can use to reproduce the issue, and to see for myself how this PR impacts the output?
> 
> On a similar note, I think this PR should also include some dune cram tests, as those are designed to check expected input-output behaviour of CLI programs, which seems appropriate here. In particular it would be useful to have tests that show `mcompare` and `mcmp` side by side on equal inputs.

I'll know little about cram test, I'll have a go.

tools/HashedFault.ml
HadrienRenaud: I think it'd be clearer to abstract on tuples with 4 elements:

```ocaml
(* compare lexicographically tuples with 4 elements *)
let compare_tuple4 (cmp1, cmp2, cmp3, cmp4) (x1, x2, x3, x4) (y1, y2, y3, y4) =
  let c1 = cmp1 x1 y1 in
  if c1 != 0 then c1 else
    let c2 = cmp2 x2 y2 in
    if c2 != 0 then c2 else
      let c3 = cmp3 x3 y3 in
      if c3 != 0 then c3 else
        cmp4 x4 y4

let compare h1 h2 =
  compare_tuple4
    ( Misc.int_compare,
      HashedStringOpt.compare,
      HashedStringOpt.compare,
      compare_string_opts)
    (as_tt h1) (as_tt h2)
```

(`compare_tuple4` could go in `Misc`, there's already `pair_compare`)

maranget: Le us do that... Thanks.

maranget:
Significant rewriting of diff and compare functions under way -> draft

maranget:
Hi @fsestini, @psafont and @relokin. I have finally decided not to rely on state lines to be totally ordered (which they are not) for merging, diffing and intersecting them. Instead, I have adopted a two  level algorithm: group lines by identical bindings, those groups of lines are totally ordered, and then rely on, naive, quadratic algorithms for operating inside these groups. See commit 7715bee31fe3121527e23660e6bfa89233bac3cd  for details.

I have also added a small cram test. The test checks that **mcompare7** and **mcmp7** produce the same result on a small example.

tools/HashedFault.ml
maranget: My comment below now is no longer exact: 

> The compare function is used:
> 
> 1. To normalise list of final states. In case "old" and "new" faults (_i.e._ no fault type and with fault types, or unspecified MMU faults and "D-"/"I-" prefixed MMU faults) do not mix, the sorted list of final states makes sense, as `compare` is transitive on those restricted sets.

I have relaxed this hypothesis: order on the binding part of final outcomes is total. I no longer assume the "compare" function on faults to induce an order.

lib/checkName.ml
```diff
@@ -55,7 +55,14 @@ let parse_int32 int32 =
 let parse_faulttype ft =
    ("-faulttype", Arg.Bool (fun b -> ft := b),
     Printf.sprintf
-      "<bool> consider fault types, default %b" !ft);
+      "<bool> consider fault types in comparisons, default %b" !ft)
```
psafont, line 58: Please bear with me for a minute. I'm trying to understand the reason why is this PR changing the code is changing, and I'm confused regarding the relation between parsing a boolean argument and the module "checkName".

From what I understand this module was made to parse names of litmus tests coming from user-provided CLI arguments.

What's has this with parsing the faults?

lib/faultType.ml
```diff
@@ -209,3 +209,9 @@ module No = struct
 end
 
 let is = AArch64.is
+
+let has_diprefix s = String.(starts_with ~prefix:"D-" s || starts_with ~prefix:"I-" s)
```
psafont: Similarly, I'm confused about the inclusion of the prefix detection and stripping of faults here.

From what I understand, these two functions should only occur in code related to the logs, which happens in mdiff and msum, which are part of `tools/`, and from what I can see, the module faultType is not related to the log-processing of the faults.

Would you mind explaining why are the function being added to this module? It would help me to structure code in a similar way in my future PRs.

tools/logState.ml
```diff
@@ -1371,25 +1372,48 @@ let simple_same out1 out2 t1 t2 k =
...
-    List.exists has_no_fault_type xs
-    || List.exists has_no_fault_type ys
-  then safe xs ys
-  else opt xs ys
+    _dbg &&
+    (FaultKinds.cardinal fk1 > 1 || FaultKinds.cardinal fk2 > 1)
+  then
+    Printf.eprintf "Found heterogeneous test %s\n%!" name ;
+  if FaultKinds.equal fk1 fk2
+  then opt xs ys
```
psafont, line 1483: Is the opt diffing safe to use when both tests are not homogeneous with regards to the fault kinds, even when these are equal?

tools/tests/mcompare.t/run.t
```diff
@@ -0,0 +1,14 @@
+Check that mcompare7 and mcmp7 yield the same result
+File V.00 is an authentic litmus log, OLD.00 is a hand made alteration of it.
```
psafont: Doing a regular diff here in the cram would exemplify what exactly are the textual differences that semantically are being ignored on purpose:
```
  $ diff OLD.00 NEW.00
  77,78c77,79
  < 398281:>[x]=1; fault(P0,x); ~fault(P0:L0);
  < 12595916:>[x]=0; fault(P0:L0,x);
  ---
  > 5864063:>[x]=0; fault(P0:L0,x,D-MMU:Permission);
  > 398281:>[x]=1; fault(P0,x,D-MMU:Permission); ~fault(P0:L0);
  > 6731853:>[x]=0; fault(P0:L0,x,D-MMU:Translation);
```

psafont (commented):
I'm surprised by the amount of files and code changes needed to do this change, but I can't really say I would approach it in a different way, or why. I do have some questions to understand these changes better

lib/checkName.ml
maranget: The module "CheckName" functionality is to select the input tests. It includes the definition  of command line options such as `-names`, `-excl` etc. The place looked convenient to host other command line definitions that are common to several tools (`-int32` for instance). This is a mistake and it sould be fixed, perhaps not in this PR?

lib/faultType.ml
maranget: The "compare" function on faults, is defined in `tools/HashedFaults.ml` (because it operates on hashed faults) and some helpers on fault types in `lib/faultType`. It does not look totally inappropriate.  Do you think that `logState.ml` would be a better place?

lib/checkName.ml
psafont: > This is a mistake and it sould be fixed, perhaps not in this PR?

Absolutely, I've seen other modules for handling args, so it might make sense to consolidate the code into a single module or something like that.

tools/logState.ml
maranget: I guess not. I had assumed that outcomes with equal bindings have fault types  of the same kind. There is no reason for this to be true. I'll fix it. Nice catch.

tools/tests/mcompare.t/run.t
maranget: I follow your suggestion, Thanks.

maranget:
> I'm surprised by the amount of files and code changes needed to do this change, but I can't really say I would approach it in a different way, or why. I do have some questions to understand these changes better

I am surprised too. Maybe the `-mmu-faults-as-data` option could belong to a different PR. I have also changed the "hashconsed" module a bit, to avoid using `==` directly, resulting in changing  files beyond  what is strictly necessary for the objective of handling the new "D-" and "I-" prefixes.

lib/faultType.ml
psafont: I do tend to prefer to responsibility-centric modules, and locate the code that deals with a responsibility in a module / file, or directory.

So anything that regards processing logs I would personally place it in `tools/`, probably in hashedFault, because it's the only code that uses it. If other uses regarding logs are foind it could go into logState or other shared code.

maranget: This looks reasonable. Let us move the functions `has_diprefix` and `split_diprefix` to `tools/HashedFaults.ml`. BTW I was (unconsciously) following a different policy: place simple functionality close to data definition (here the strings that represent fault types).

psafont (approved):

tools/logState.ml
psafont: I think `union_faults` solves it nicely now

maranget: Well, some work still under way. I have decided to have a proper compare function on faults that will distinguish fauts syntactically, as correct hashconsing of states rely on it. That some syntactically different faults are the same will be handled in a more explicit way, by a `equivalent` function.

fsestini: > Well, some work still under way. I have decided to have a proper compare function on faults that will distinguish fauts syntactically, as correct hashconsing of states rely on it. That some syntactically different faults are the same will be handled in a more explicit way, by a `equivalent` function.

@maranget I don't see these changes reflected in the PR, so just wanted to check whether you are still planning to work on this, or if you think this PR is ready for re-review.

maranget:
Hi @psafont and @relokin . I have squashed and rebased this PR and would like to merge today. I'd value any comment.

tools/tests/mcompare.t/run.t
```diff
@@ -0,0 +1,25 @@
...
+  < 12595916:>[x]=0; fault(P0:L0,x); 
+  ---
+  > 5864063:>[x]=0; fault(P0:L0,x,D-MMU:Permission); 
+  > 398281:>[x]=1; fault(P0,x,D-MMU:Permission); ~fault(P0:L0);
+  > 6731853:>[x]=0; fault(P0:L0,x,D-MMU:Translation); 
+  [1]
+
+  $ mcompare7 OLD.00 NEW.00 -pos P -neg N >/dev/null 2>/dev/null
+  $ cat P N | wc -l
+  0
```
fsestini: This fails on my macOS machine due to silly whitespace padding:

```
$ dune runtest tools/tests/mcompare.t
File "tools/tests/mcompare.t/run.t", line 1, characters 0-0:
diff --git a/_build/default/tools/tests/mcompare.t/run.t b/_build/default/tools/tests/mcompare.t/run.t.corrected
index b772f3a81..9d4faf07a 100644
--- a/_build/default/tools/tests/mcompare.t/run.t
+++ b/_build/default/tools/tests/mcompare.t/run.t.corrected
@@ -13,7 +13,7 @@ standing for what a very old version of litmus could have produced.
 
   $ mcompare7 OLD.00 NEW.00 -pos P -neg N >/dev/null 2>/dev/null
   $ cat P N | wc -l
-  0
+         0
   $ mcmp7 OLD.00 NEW.00 2>/dev/null
   $ msum7 OLD.00 NEW.00 > ALL.00 2>/dev/null
   $ herd7 -set-libdir ../libdir src/@all > Herd
```

maranget: Funny.... Pipe into `tr -d ' '`?

maranget: We can also use the test command with the -s option. I'll do that,

tools/tests/mcompare.t/run.t
fsestini, line 1: I think we should add a few more cases to this test. In particular, the current cram test does not exercise the new `-mmu-faults-as-data` flag in the `false` case. In would appear that `mcompare7` and `mcmp7` disagree under `-mmu-faults-as-data false`, which I don't think is intended:

```sh
$ cat old.log
Test T Allowed
States 1
~Fault(P0:L0,x,MMU:Translation);
Ok
Hash=x
$ cat new.log
Test T Allowed
States 1
~Fault(P0:L0,x,D-MMU:Translation);
Ok
Hash=x
$ dune exec mcompare7 -- -mmu-faults-as-data false old.log new.log
*Diffs*
 |Kind | old.log                           new.log
--------------------------------------------------
--------------------------------------------------
T|Allow| [~fault(P0:L0,x,MMU:Translation)] ==
 |Ok   |

$ dune exec mcmp7 -- -mmu-faults-as-data false old.log new.log
T
```

maranget: Interesting. If I am not wrong **mcompare7** ignores negative fault occurrences. The fast path of **mcmp7** does not ignore them, as it relies on hashedconsed identity nodes to perform comparisons.

maranget: Well, if **mcompare7** ignore negative fault occurrences, **mcmp7** can also ignore them, by simply not inserting them in its internal data structure. I'll implement this technique.

tools/logState.ml
maranget: I had overwritten those changes by an untimely "git push --force". They are now present again. See [here](https://github.com/herd/herdtools7/pull/1970/changes#diff-6d431d0a3d8e6157c74ffee49c7cc040b4449e0d0d02a23260d890dfa0727491) for the definitions of the  `equivalent` and `compare` functions on hashconsed faults and [here](https://github.com/herd/herdtools7/blob/a703b5085b67fda9f4b1b34c2ac1dcd1718e9446/tools/logState.ml#L377) for the call to `HashedFault.equivalent` in `tool/logState.ml`.

maranget:
Here is a copy of  a comment that might otherwise get lost.

I had overwritten significant changes  changes by an untimely "git push --force". Namely I have separated the ordering of state llnes (or *outcomes*) and the equivalence of faults. They are now present again. See [here](https://github.com/herd/herdtools7/pull/1970/changes#diff-6d431d0a3d8e6157c74ffee49c7cc040b4449e0d0d02a23260d890dfa0727491) for the definitions of the  `equivalent` and `compare` functions on hashconsed faults and [here](https://github.com/herd/herdtools7/blob/a703b5085b67fda9f4b1b34c2ac1dcd1718e9446/tools/logState.ml#L377) for the call to `HashedFault.equivalent` in `tool/logState.ml`.

lib/checkName.ml
maranget: TODO in a future PR.

maranget:
Hi @psafont and @relokin (and maybe @artkhyzha, I guess you have been involved in adding the "D-" and "I-" prefixes) . This PR experienced many adventures, resulting in partial and total rewrites. Things  went far beyond to where I initially expected: fixing the  compare tools so that I can compare new herd vmsa logs to old logs from before the introduction of the "D-" and "I-" prefixes...

Thanks to all the remarks and suggestions that have been made, the situation of log comparisions is now much cleaner than before: from series of fixes to the clear understanding that we cannot totally order the lines of a test log, especially when those logs have been recorded at different times. But the compare tools must still evolve now. I'd really like to have your informed opinion once more before merging.

lib/checkName.ml
```diff
@@ -55,7 +55,14 @@ let parse_int32 int32 =
...
+
+let datafault_key = "-mmu-faults-as-data"
+
+let parse_datafault ft =
+   (datafault_key, Arg.Bool (fun b -> ft := b),
+    Printf.sprintf
```
relokin, line 64: I think this is fine but also not necessary, all old logs that said `MMU:<type>` should match with `D-MMU:<type>`. I don't see a situation where we would set the parameter to false.

tools/HashedFault.ml
```diff
@@ -44,33 +44,76 @@ let as_t h =
...
+  if has_diprefix s then Some String.(sub s 2 (length s-2))
+  else None
+
 let warn_once = ref true
 
+let equivalent_ftype_names s1 s2 =
+  match strip_diprefix s1,strip_diprefix s2 with
+  | None,Some s2 -> String.equal s1 s2
+  | Some s1,None -> String.equal s1 s2
+  | _,_ -> String.equal s1 s2
```
relokin, line 75: Maybe I have misunderstood this, but wouldn't we want to just change D-MMU to MMU for backwards compatibility but maintain I-MMU?

lib/checkName.ml
maranget: Hi @relokin. I agree with you that MMU: in old logs most certainly refer to D-MMU:. Setting the switch "-mmu-faults-as-data" to false may help debugging by having the internal functions to operate on stuctures as close as possible to the input.

tools/HashedFault.ml
maranget: This would contradict  the intuitive meaning of faults in final conditions. For instance `exists Fault(..,..,MMU:Permission)` is intuitively understood as: "there exists a permission fault, would it be from data or instruction".  This "intuitive" semantics already governs the behaviour of **herd7**, while PR #1988 adjusts **litmus7**.

I mean I saw no reason for tools and **herd7** to interpret prefixless MMU fault types differently. Given our hypothesis on old logs tests generating only "D-MMU:..." faults, there should be no practical implications for log comparisons and sums.

maranget:
PR #1970 merged, thanks @psafont, @fsestini and @relokin for your suggestions and reviews.

PR #1971, opened 2026-08-20
[asl] Add `error_handling_time` to various interpreter errors

hrutvik:
Some errors were reported as dynamic, even though static. Add some tests to show this.

Assists https://github.com/herd/herdtools7/pull/1935.

### DIFF
--- a/asllib/tests/regressions.t/run.t
+++ b/asllib/tests/regressions.t/run.t
@@ -774,3 +774,36 @@ If test environment reversion bug
   ASL Grammar error: Obsolete syntax:
     Deprecated slice syntax, use "3*2 +: 2" instead.
   [1]
+
+Static errors:
+  $ aslref --no-exec static-unreachable.asl
+  File static-unreachable.asl, line 3, characters 2 to 14:
+    unreachable;
+    ^^^^^^^^^^^^
+  ASL Static error: unreachable reached.
+  [1]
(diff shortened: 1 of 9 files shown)


asllib/tests/regressions.t/run.t
```diff
@@ -774,3 +774,36 @@ If test environment reversion bug
...
+  File static-unreachable.asl, line 3, characters 2 to 14:
+    unreachable;
+    ^^^^^^^^^^^^
+  ASL Static error: unreachable reached.
+  [1]
+  $ aslref --no-exec static-assertion.asl
+  File static-assertion.asl, line 3, characters 9 to 14:
+    assert FALSE;
+           ^^^^^
+  ASL Static error: Assertion failed: FALSE.
```
HadrienRenaud, line 789: In general, I'm a bit worried that people are reading `Assertion failed` as an assertion in the interpreter that failed, and not an ASL assertion that failed, but I don't really know what we can do against this, and I don't even know if people are going to think this.

HadrienRenaud (approved):
Looks good to me, thanks

asllib/tests/regressions.t/run.t
hrutvik: Ah interesting - I suppose the location information in the ASL file should help, but I agree we can also think of better wordings

PR #1972, opened 2026-08-21
[herd,riscv] Signed and unsigned extensions of size 64 avoided.

maranget:
As native size is 64 bit those two operations are the identity. However, the generic implementation cannot detect the situation. As a result, loading and storing symbolic addresses yielded fatal errors. Fix consists in not even applying those extension operations for quad (or 64bit) size

Typical test:
```
RISCV T
{
  int *p; int x;
  0:x1=p; 0:x2=x;
}
  P0          ;
  sd x2,0(x1) ;
```

### DIFF
--- a/herd/RISCVSem.ml
+++ b/herd/RISCVSem.ml
@@ -50,8 +50,21 @@ module
       let (>>!) = M.(>>!)
       let (>>::) = M.(>>::)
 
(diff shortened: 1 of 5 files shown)


maranget:
Simple bug fix, merging.

PR #1974, opened 2026-08-27
[litmus] Fix cram test

maranget:
In the test, **litmus7** options were `... -a 4 -s 1k -r 100 -mach x86_64 ...` The configuration file [`x86_64`](https://github.com/herd/herdtools7/blob/master/litmus/libdir/x86.cfg) defines all explicitly set options, which, as options are processed left-to-right, amounts to redefine those options.

In particular _avail_ (option `-a`) is re-defined to zero, and affinity is enabled (`affinity = incr0` in configuration file `litmus/libdir/x86_64`). As a result, the number of available cores corresponds to the number of logical cores of the tested machine! Hence, the test fails on my machine that features 24 logical cores. Fix consists in having the option `-mach x86_64` first. Additionally we change the `-a` setting to `-a 2` for safety. Then, we have 200k test runs on any machine.

Folowup to PR #1926 that introduced the cram test for litmus.

### DIFF
--- a/litmus/tests/X86_64/A009.t
+++ b/litmus/tests/X86_64/A009.t
@@ -4,23 +4,23 @@ Use litmus7 to generate code from a litmus test
   $ mkdir "$TEST"
   $ litmus7 -set-libdir ../../libdir -o "$TEST" \
   > "../../../herd/tests/instructions/X86_64/$TEST.litmus"  \
-  > -mode std -a 4 -s 1k -r 100 \
-  > -mach x86_64
+  > -mach x86_64 -mode std -a 2 -s 1k -r 100
+
 
 Compile and run the litmus test natively, avoid printing the timing, it's not
 stable
 
   $ cd $TEST
   $ make > /dev/null
-  $ "./$TEST.exe" | sed '$d'
+  $ "./$TEST.exe" | grep -v -e ^Time
   Test A009 Required
   Histogram (1 states)
-  4000000:>0:rcx=-1;
+  200000:>0:rcx=-1;
   Ok
   
   Witnesses
-  Positive: 4000000, Negative: 0
+  Positive: 200000, Negative: 0
   Condition forall (0:rcx=-1) is validated
   Hash=7ca3c35015d75a877ccf509d75062e79
-  Observation A009 Always 4000000 0
+  Observation A009 Always 200000 0
 


psafont (approved):

psafont:
I think it's worth documenting these interaction between arguments in the `-help` output.

maranget:
I have a few remarks  about the AArch64 litmus7 cram tests:
  + **litmus7** options are `-mode std -a 4 -s 1k -r 100 -mach aarch64` it would be safer to write `-mach aarch64 -mode std -a 4 -s 1k -r 100 ` so that a number of test runs would not depend on the settings of the configuration file `aarch64.cfg`. This remark applies to all tests `-mach `_config_ better precedes other options. 
   + The AArch64.kvm tests are not run in kvm mode, which looks difficult as running this tests depends on kvm-unit-tests being installed.
   + The reference for test `AArch64.kvm/A005.t` includes compiler diagnostics that not all compilers produce.

Moreover,  the "cross" tests `litmus/tests/cross/AArch64` (apparently) do not run on my system. I do not understand all the subtelties of dune cram tests. This can be normal behaviour.

maranget:
> I think it's worth documenting these interaction between arguments in the `-help` output.

The interaction of the left-to-right processing order  of options and of configuraton files is documented [in the manual](https://diy.inria.fr/doc/litmus.html#sec31). I am not sure it belongs to the `-help` output.

psafont:
> I have a few remarks about the AArch64 litmus7 cram tests:
> 
>     * **litmus7** options are `-mode std -a 4 -s 1k -r 100 -mach aarch64` it would be safer to write `-mach aarch64 -mode std -a 4 -s 1k -r 100 ` so that a number of test runs would not depend on the settings of the configuration file `aarch64.cfg`. This remark applies to all tests `-mach `_config_ better precedes other options.

I'll take a note to modify them

>     * The AArch64.kvm tests are not run in kvm mode, which looks difficult as running this tests depends on kvm-unit-tests being installed.

I'll remove these, seeing that they are problematic
>     * The reference for test `AArch64.kvm/A005.t` includes compiler diagnostics that not all compilers produce.

The should be easy to suppress in the cram test

> Moreover, the "cross" tests `litmus/tests/cross/AArch64` (apparently) do not run on my system. I do not understand all the subtelties of dune cram tests. This can be normal behaviour.

What OS is it running? These tests are only set up to run under linux, with a crosscompiler and qemu in user mode installed. There's a readme with more information on how to run these: https://github.com/herd/herdtools7/blob/master/litmus/tests/cross/readme.md

Please let me know how can I complete the information to be able to install dependencies and run the tests.

To make it easier to install qemu, I can contribute to the official opam repository a package to handle the installation of the qemu packages. Since there are no official packages that would depend on it, I have to see whether it's acceptable to contribute it or not.



> > I think it's worth documenting these interaction between arguments in the `-help` output.
> 
> The interaction of the left-to-right processing order of options and of configuraton files is documented [in the manual](https://diy.inria.fr/doc/litmus.html#sec31). I am not sure it belongs to the `-help` output.

I'm used to tools like dune and opam, that display manpages when `-help` is provided. This is provided with the Cmdliner library. I think using the library in the library would help to provide interesting features for herdtools, like manpages, or autocompletion for the shell. But it needs care to not break backwards compatibility. I haven;t looked into it yet.

maranget:
Merged, thanks @psafont.

maranget:
> > Moreover, the "cross" tests `litmus/tests/cross/AArch64` (apparently) do not run on my system. I do not understand all the subtelties of dune cram tests. This can be normal behaviour.
> 
> What OS is it running? These tests are only set up to run under linux, with a crosscompiler and qemu in user mode installed. There's a readme with more information on how to run these: https://github.com/herd/herdtools7/blob/master/litmus/tests/cross/readme.md

My bad, I can run the test by using  with the proper alias, as described in the readme.md file. I was attempting th emore standard `dune runtest` that does not work purposely because of the `(runtest_alias false)` declaration in the `dune` file.

PR #1979, opened 2026-09-02
[herd] return non-zero exit codes whenever there's an error

### LINKED ISSUE #1977: [herd] herd7 exits successfully when the default cat model fails to parse
When the default cat model selected through `-set-libdir` contains a syntax error, `herd7` aborts without running the test but exits with status 0. The error message it prints out is also slightly confusing.

Run from the herdtools7 repository root:

```sh
mkdir -p /tmp/herd7-cat-error-repro

cat >/tmp/herd7-cat-error-repro/aarch64.cat <<'EOF'
(* unterminated comment
EOF

cat >/tmp/herd7-cat-error-repro/minimal.litmus <<'EOF'
AArch64 Minimal
{ 0:X1=x; }
P0;
LDR W0,[X1];
exists (0:X0=0)
EOF

dune exec herd7 -- -set-libdir /tmp/herd7-cat-error-repro /tmp/herd7-cat-error-repro/minimal.litmus
echo $?
```

This will print out the following:

```
Warning: File "/tmp/herd7-cat-error-repro/aarch64.cat", line 2, character 0: Lex error eof in skip_comment (in model) (User error)
0
```

Supplying the same malformed model explicitly behaves differently:

```sh
dune exec herd7 -- -model /tmp/herd7-cat-error-repro/aarch64.cat /tmp/herd7-cat-error-repro/minimal.litmus
echo $?
```

This reports a fatal error and exits with status 2:

```
Fatal error: exception Misc.UserError("File \"/tmp/herd7-cat-error-repro/aarch64.cat\", line 2, character 0: Lex error eof in skip_comment (in model)")
2
```

## Problems

- A model parsing failure is described as a 'warning' even though execution is aborted and the error message itself mentions 'User error' at the end.
- The process exits successfully with exit code 0 despite reporting an error and not running the test.
- Different selection modes of the same cat file handle the same error inconsistently.
- For the specific case of syntax errors arising from unterminated comments:
    - `eof in skip_comment` exposes lexer implementation terminology rather than explaining the syntax problem.
    - The diagnostic message reports the location of EOF instead of the location where the unterminated comment began.

psafont:
Now herd only returns 0 when there are no errors during the running of
all tests. The error code corresponds to the last exception that was
raised while running all the tests.

Note that -exit still governs whether herd exits immediately after
encountering an error while executing more than one litmus test.

Fixes #1977

### DIFF
--- a/herd/tests/other/cat-parse-error.t/run.t
+++ b/herd/tests/other/cat-parse-error.t/run.t
@@ -1,0 +1,9 @@
+Parsing a malformed model must result in exit status 2.
+
+  $ herd7 -model aarch64.cat minimal.litmus
+  herd7: File "./aarch64.cat", line 2, character 0: Lex error eof in skip_comment (in model)
+  [2]
+
+  $ herd7 -set-libdir . minimal.litmus
+  Warning: File "./aarch64.cat", line 2, character 0: Lex error eof in skip_comment (in model) (User error)
+  [2]
--- a/herd/herd.ml
+++ b/herd/herd.ml
@@ -22,6 +22,13 @@ open Archs
 open Opts
 open OptNames
 
+
+(* exit codes *)
+
+let exit_code_of_exn = function
+  | Misc.Exit | Misc.UserError _ | Misc.Fatal _ -> 2
+  | _ -> 1
+
 (* Command line arguments *)
 let args = ref []
 let get_cmd_arg s = args := s :: !args
@@ -547,7 +554,9 @@ let () =
         Printf.eprintf "Error: %s\n %s \n" first_line usg_default;
         exit 2
       end
-  | Misc.Fatal msg -> eprintf "%s: %s\n" prog msg ; exit 2
+  | Misc.Fatal msg as e ->
+      eprintf "%s: %s\n" prog msg ;
+      exit (exit_code_of_exn e)
 
 (* Read generic model, if any *)
 
@@ -565,10 +574,12 @@ let model,model_opts = match !model with
       let (fname,((b,_,_) as r)) = P.find_parse ~opt:true fname in
       Some (Model.Generic (fname,r)),b
     with
-    | Misc.Fatal msg -> eprintf "%s: %s\n" prog msg ; exit 2
-    | Misc.Exit ->
+    | Misc.Fatal msg as e ->
+        eprintf "%s: %s\n" prog msg ;
+        exit (exit_code_of_exn e)
+    | Misc.Exit as e ->
         eprintf "Failure of generic model parsing\n" ;
-        exit 2 end
+        exit (exit_code_of_exn e) end
 | Some _ as m -> m,ModelOption.compat
 | None -> None,ModelOption.default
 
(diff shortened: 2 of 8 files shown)


lib/misc.ml
```diff
@@ -22,6 +22,10 @@ exception UserError of string
 exception Fatal of string
 exception NoIsync
 
+let exit_code_of_exn = function
+  | Exit | UserError _ | Fatal _ -> 2
```
fsestini: I understand the PR is mostly just adopting the existing convention, but it would be nice to know (and document) the rationale behind this choice of exit codes here. Maybe @maranget can provide more context.

[Some people claim](https://stackoverflow.com/a/40484670) that the general Unix convention would be to use 1 for program errors and 2 for user errors. This is not a POSIX mandate and we don't need to follow it, however I do think it could be useful if program errors and user errors generated different exit codes. I don't have a strong opinion on this though.

herd/tests/other/cat-parse-error.t/run.t
fsestini, line 1: Would you mind also adding a cram test to demonstrate the `-exit` case?

lib/misc.ml
psafont: > I understand the PR is mostly just adopting the existing convention, but it would be nice to know (and document) the rationale behind this choice of exit codes here.

I agree, I also would like to add exit codes to the docs as well. I'll see how to do this.

> [Some people claim](https://stackoverflow.com/a/40484670) that the general Unix convention would be to use 1 for program errors and 2 for user errors. This is not a POSIX mandate and we don't need to follow it, however I do think it could be useful if program errors and user errors generated different exit codes. I don't have a strong opinion on this though.

Exit codes seem to be platform specific because there's no POSIX standard. As you say, using the CLI wrong (wrong number of arguments, on invalid arguments can use exit code 2, but using `man sysexits` in macOS shows the exit code should be 64 (EX_USAGE), and if the input data is incorrect 65 (EX_DATAERR) should be used). On linux we can see these match: https://man7.org/linux/man-pages/man3/sysexits.h.3head.html But in general I haven't seen application using it

I think for the time being we should encode existing behaviour and maybe try to separate user errors from internal errors. We can think of reworking the error handling, and hence exit codes if we want to have better reporting

herd/tests/other/cat-parse-error.t/run.t
psafont: Sure, let me know what you think about it

lib/misc.ml
fsestini: I was surprised by `man sysexits` as well.

I think it's fine to keep the existing project convention for this PR.

I only have one minor comment on the code side: I'm not sure I would define `exit_code_of_exn` as a "global" API function exposed from the shared `lib`. My understanding is that stuff in `lib` should be generally useful repo-wide. However, it's not hard to imagine that different tools may want to return different exit codes for different exceptions in a way that does not follow `exit_code_of_exn`. One example is custom exceptions that are locally defined within the tool's code: it's not clear why `exit_code_of_exn My_tools_custom_exception` should return 1. Similarly, in my own tools I tend to use 1 for `Misc.Fatal`-like errors, so I personally wouldn't be using  `exit_code_of_exn` in my tools, as it is written.

Maybe `exit_code_of_exn` should be made local to `herd.ml` for now, since it seems to mainly reflect herd7 conventions. We can always move it to `lib` later, if and when we manage to hammer down a coherent repo-wide convention.

herd/tests/other/cat-parse-error.t/run.t
fsestini: Amazing, thank you

fsestini (approved):
LGTM, thanks @psafont . If you give it a rebase on master I can merge it for you.

PR #1980, opened 2026-09-02
[tools/mprog7] Add ability to read input litmus test from stdin

fsestini:
This PR adds the conventional `-` operand to `mprog7` for reading one litmus test from standard input. For example:

```sh
$ cat ./tools/tests/mprog/stdin.t/A.litmus
AArch64 A
{}
P0;
MOV X0,#1;
exists (0:X0=1)
$ cat ./tools/tests/mprog/stdin.t/A.litmus | mprog7 -mode text -
AArch64 A
{
}
 P0        ;
 MOV X0,#1 ;

exists (0:X0=1)
```

This functionality can be quite handy when using `mprog7` as a litmus test formatter, for example within an editor like vim.

- As before this PR, absent `-` or positional arguments, stdin is interpreted as a list of filenames.
- `-` may be used alongside file paths.
- `-` can only be specified at most once.
- A `-` read from a filename list (be it from stdin or from a `@list` file) is interpreted as a literal filename.
- When `-o DIR` is specified, the litmus test read from stdin input is saved as `DIR/stdin.litmus`.

This PR updates the input mechanism of `mprog7` to support the new stdin operand and brings its batch error handling in line with conventional CLI behaviour. Previously, errors associated with an individual input were reported but not reflected in the exit code, so a batch could finish with status zero even when one or more inputs had failed. With this PR, `mprog7` now retains its useful behaviour of continuing batch processing of multiple inputs, while recording whether any input failed and returning status `1` after the batch if so. Command-line errors and failures encountered while opening an `@list` continue to return status `2`.


tools/mprog.ml
```diff
@@ -324,15 +352,41 @@ module X =
...
+        eprintf "User error: %s\n%!" msg
+      else
+        eprintf "User error: %a %s\n%!" Pos.pp_pos0 name msg ;
+      false
+
 let () =
-  Misc.iter_argv_or_stdin
-    (fun fname ->
-      try X.zyva fname with
-      | Misc.Exit -> ()
```
fsestini, line 331: I've removed this line as I did not understand why it was there: no code path downstream of `mprog7` seems to be raising `Misc.Exit`, and even if it does, I'm not sure it's a good idea to silently swallow the exception anyway.

tools/mprog.ml
```diff
@@ -324,15 +352,41 @@ module X =
...
-      | Misc.Fatal msg|Misc.UserError msg ->
-          Warn.warn_always "%a %s" Pos.pp_pos0 fname msg ;
-          ()
-      | e ->
-          Printf.eprintf "\nFatal: %a Adios\n" Pos.pp_pos0 fname ;
-          raise e)
```
fsestini, line 337: I've removed this as generally I'm not a fan of catch-all branches, and this branch doesn't seem to really add much to the error it catches, in terms of error message. In fact, catching and rethrowing exceptions risks garbling the stack trace of the original exception, which makes debugging more difficult.

tools/mprog.ml
maranget: Well, I do not know if the handler is appropriate, probably not. The usual reaction to Misc.Exit is not doing anything, assuming that the error has been flagged before, or that there is no error to flag, See `tools/mnames.ml` for an example of the second possibility.

tools/mprog.ml
maranget: I agree with deleting this handler. Notice that OCaml stack traces should remain understandable here: signaling that the exception is "re-raised".

fsestini: Yes, I was being overly pedantic here. While stack trace preservation is somewhat heuristic and not guaranteed in all scenarios, the compiler should be smart enough in this particular case, as you correctly point out.

maranget:
LGTM.

 I am wondering if we can seize the oportunity to fix a slight inconsitency: command line arguments are processed right to left (_i.e._ in reverse order), while lists of files from standard input or index files are processed in order.
Consider two litmus tests `A1.litmus` and `A2.litmus` with respective names `A1` and `A2. We have
```
 % ( echo A1.litmus && echo A2.litmus ) | mprog7 -mode text | grep AArch
AArch64 A1
AArch64 A2
```
While we have:
```
% mprog7 -mode text A1.litmus A2.litmus | grep AArch64
AArch64 A2
AArch64 A1
```
To that aim it would suffice to change the function add_input:
```OCaml
let add_input input = args := !args @ [input]
```
I do not see the quadratic behaviour as a problem here. If so, it would suffice to keep `add_input` as it is and to reverse the list of arguments before using it:
```OCaml
let () =
  let no_errors = List.rev !args |> iter_inputs |> iter_check_all do_input in
  if not no_errors then exit 1
```

fsestini:
> I am wondering if we can seize the oportunity to fix a slight inconsitency: command line arguments are processed right to left (_i.e._ in reverse order), while lists of files from standard input or index files are processed in order.

Thanks for the comment. We sure can fix this. Should be addressed by [4e02409](https://github.com/herd/herdtools7/pull/1980/commits/4e0240940b4e10cc7104a4e20f8fe746f0bb5a7a) as per your suggestion.

fsestini:
The PR checks seem to be failing on some mcompare7 tests, for some reason 🤔

maranget:
> The PR checks seem to be failing on some mcompare7 tests, for some reason 🤔

The tests of mcompare work again when one erases the line `(applies_to miaou)` from the file `tools/tests/dune`.

tools/tests/dune
```diff
@@ -1,4 +1,5 @@
 (cram
+ (applies_to miaou)
```
maranget, line 2: For some reason this line invalidates the `source_tree` declarations in `mcompare.t/dune`. Perhaps can we move the tests of **miaou7** into a `miaou.t` directory, with its own `dune` file.

maranget: The following commit does precisely that.

[move.txt](https://github.com/user-attachments/files/32280362/move.txt)

fsestini: Thanks @maranget for looking into this. I had another look as well.

> For some reason this line invalidates the `source_tree` declarations in `mcompare.t/dune`.

I think the reason is that dune is simply ignoring the `mcompare.t/dune` file. The cram test convention is that a file named `foo.t` denotes a cram test. This could be either a single file, or a directory containing some assets plus a run.t file. When the cram test is a directory, everything inside it is treated as test data. In particular, any `dune` files put in it will have no effect. See the [dune docs](https://dune.readthedocs.io/en/stable/tests.html#directory-tests):

> The name of the test directory must end with .t and must include a run.t as the test script. Everything else in that directory is treated as raw data for the test. It’s not possible to define rules using dune files in such a directory.

Thus when `dune` scans the file system and sees `mcompare.t`, it interprets it as a cram test and runs it according to the rule specified by the last dune file it found in the hierarchy, which is `tools/tests/dune`.
Adding a `dune` file to `miaou.t` won't work for the same reason.

I think this should be fixable by moving the cram stanza for mcompare into the `tools/tests/dune` file. Seems to work on my machine at least. [e2a6f22](https://github.com/herd/herdtools7/pull/1980/commits/e2a6f227d9cfc120443e831bf0ad45fa9d9aa1ad) does this.

maranget: Thanks, I did not notice that the `dune` file is ignored in `mcompare.t`...

fsestini: No problem. Planning to merge this later today unless there are other concerns.

PR #1985, opened 2026-09-03
Aligning implementation/specification error codes - phase 1/5

Roman-Manevich:
### Non-variable collection field access (`TE_UT`)
Collection storage is accessed through a variable. For expression whose base has collection type but is not a variable, the specification previously had no error case and the OCaml implementation raised the generic,
uncoded `UnsupportedExpr` diagnostic.

This change:
- adds the new error kind `CollectionBaseNotVariable`, mapped to the `TE_UT`;
- reports the actual base expression in the diagnostic;
- uses it for single-field reads, multi-field reads, and the corresponding multi-field assignment path;
- adds matching `TypeError(TE_UT)` cases to `asl.spec`;
- adds a new regression test using a conditional expression as a collection base.

### Mismatched bitvector widths (`TE_UT`)
The specification reports unequal bitvector widths as `TE_UT`. The implementation raised the uncoded `UnreconcilableTypes` diagnostic, which incorrectly described the failure as a missing common ancestor.

This change:
- adds the new error kind `MismatchedBitvectorWidths`;
- maps it to the existing `TE_UT` code;
- uses it in `check_bits_equal_width`;
- removes the redundant `TE_BO` check following `check_bits_equal_width` in the single-pattern specification rule.

### Overlapping implementations (`TE_OE`)
The OCaml implementation reports overlapping `implementation` subprograms as `TE_OE`, while the corresponding specification check previously produced a generic type error without a code.

This change:
- adds an explicit `TE_OE` check to `check_implementations_unique`;
- fixes `signatures_match` to return its calculated result rather than always returning `True`.

Updated the console outputs in `ErrorCodes.tex`.

### DIFF
--- a/asllib/doc/ErrorCodes.tex
+++ b/asllib/doc/ErrorCodes.tex
@@ -37,8 +37,8 @@ File ../tests/ASLDefinition.t/TypingErrorReporting.asl, line 3,
   characters 11 to 22:
     return 5 + "hello";
            ^^^^^^^^^^^
-ASL Type error (TE_BO): Illegal application of operator + on types
-  integer {5} and string.
+ASL Type error: Illegal application of operator + on types integer {5}
+  and string.
 \end{Verbatim}
 % CONSOLE_END
 
@@ -57,8 +57,7 @@ followed by an example of a report from \aslrefterm{} in a \linuxbashshellterm.
 % CONSOLE_BEGIN CONSOLE_STDERR CONSOLE_CMD aslref \definitiontests/DynamicErrorReporting.asl
 \begin{Verbatim}[fontsize=\footnotesize, frame=single]
 > aslref ../tests/ASLDefinition.t/DynamicErrorReporting.asl
-ASL Dynamic error (DE_BO): Illegal application of operator DIV for values 128
-  and 7.
+ASL Dynamic error: Illegal application of operator DIV for values 128 and 7.
 \end{Verbatim}
 % CONSOLE_END
 
@@ -82,7 +81,7 @@ File ../tests/ASLDefinition.t/AssertionStatement.asl, line 5,
   characters 11 to 22:
     assert a + b < 256;
            ^^^^^^^^^^^
-ASL Dynamic error (DE_DAF): Assertion failed: ((a + b) < 256).
+ASL Dynamic error: Assertion failed: ((a + b) < 256).
 \end{Verbatim}
 % CONSOLE_END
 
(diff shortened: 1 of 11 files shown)


asllib/error.ml
```diff
@@ -306,8 +306,7 @@ module ErrorCode = struct
         None
     | CannotParse _ (* used in lexing too *) -> None
     | UnreconcilableTypes _ (* LCA failures *) -> None
-    | EmptyConstraints (* does this need to be reflected in reference? *) ->
-        None
+    | EmptyConstraints (* An internal invariant *) -> None
```
HadrienRenaud: Maybe that needs to be moved to the next section, after line 317 for example

asllib/doc/ErrorCodes.tex
HadrienRenaud, line 1: Maybe would be nice to rebase and run it after each commit

HadrienRenaud (approved):
LGTM

asllib/error.ml
```diff
@@ -304,7 +305,7 @@ module ErrorCode = struct
     (* dynamic ATC but also mismatched integers for loop limits *) ->
         None
     | CannotParse _ (* used in lexing too *) -> None
-    | UnreconcilableTypes _ (* both LCA and check_bit_widths_equal *) -> None
+    | UnreconcilableTypes _ (* LCA failures *) -> None
```
hrutvik: If this is only LCA failures now, should it now:
- generate the appropriate error code
- be moved to the right section

?

asllib/doc/ErrorCodes.tex
```diff
@@ -37,8 +37,8 @@ \section{How Implementations Should Handle Errors\label{sec:How Implementations
   characters 11 to 22:
     return 5 + "hello";
            ^^^^^^^^^^^
-ASL Type error (TE_BO): Illegal application of operator + on types
-  integer {5} and string.
+ASL Type error: Illegal application of operator + on types integer {5}
```
hrutvik: I'm not sure I understand this commit - it seems to remove error codes, which seems like a mistake?

asllib/error.ml
Roman-Manevich: moved to line 317

asllib/doc/ErrorCodes.tex
Roman-Manevich: will do.

asllib/doc/ErrorCodes.tex
Roman-Manevich: Right. I ran `python3 doclint.py -cm`, which ran an outdated `aslref`. Fixed now.

asllib/error.ml
Roman-Manevich: Renamed to `UnreconcilableTypes ` to `NoCommonAncestor` and mapped to `Typing LCA`.
Moved to earlier section.

PR #1987, opened 2026-09-03
Aligning implementation/specification error codes - phase 2/5

Roman-Manevich:
### Binary operator priority (`BE_BOP`)
Unparenthesised binary operators with incompatible associativity or equal priority were reported as uncoded `CannotParse`. This change adds `BadBinopPriority`, mapped to `BE_BOP`, for both parser checks.

### Invalid all-discard local declaration (`BE_BD`)
All-discard local tuple declarations were reported as uncoded `CannotParse`, while the `BE_BD` documentation described only top-level declarations. This change adds:
1. `BadDeclarationSyntax`, mapped to `BE_BD`, and 
2. **updates the documentation to cover both local and top-level declarations**.

### Type satisfaction failures (`TE_TSF`)
Failures from `check_type_satisfies` and `check_can_be_initialized_with` were reported as `TE_UT`. This change adds `TypeSatisfactionFailure`, mapped to `TE_TSF`, for both helpers.

### Bad tuple index (`TE_BTI`)
Out-of-range tuple item selectors were reported as `TE_BF`. This change adds a dedicated `BadTupleIndex` diagnostic mapped to `TE_BTI`, as specified.

### DIFF
--- a/asllib/Parser.mly
+++ b/asllib/Parser.mly
@@ -88,24 +88,22 @@ let check_is_associative ~loc (op : AST.binop) =
   | _ ->
       Error.(
         fatal_from loc
-          (CannotParse
-             (Some
-                (Format.sprintf
-                   "Binary operator `%s` is not associative - parenthesise to \
-                    disambiguate."
-                   (PP.binop_to_string op)))))
+          (BadBinopPriority
+             (Format.sprintf
+                "Binary operator `%s` is not associative - parenthesise to \
+                 disambiguate."
+                (PP.binop_to_string op))))
 
 
 let check_not_same_prec loc op op' =
   if prec op = prec op' then
     Error.(
       fatal_from loc
-        (CannotParse
-           (Some
-              (Format.sprintf
-                 "Operators `%s` and `%s` have the same priority - parenthesise \
-                  to disambiguate."
-                 (PP.binop_to_string op) (PP.binop_to_string op')))))
+        (BadBinopPriority
+           (Format.sprintf
+              "Operators `%s` and `%s` have the same priority - parenthesise \
+               to disambiguate."
+              (PP.binop_to_string op) (PP.binop_to_string op'))))
 
 let check_not_binop_same_prec op e =
   match e.desc with
--- a/asllib/Parser.mly
+++ b/asllib/Parser.mly
@@ -477,8 +477,7 @@ let decl_item :=
   | vs=plist2(discard_or_identifier) ; {
       if List.for_all is_local_ignored vs then
         Error.fatal_here $startpos $endpos @@
-          Error.BadDeclarationSyntax
-            "A local declaration must declare at least one name."
+          Error.AllDiscardLocalDeclaration
       else LDI_Tuple vs
     }
 
--- a/asllib/Parser.mly
+++ b/asllib/Parser.mly
@@ -477,7 +477,8 @@ let decl_item :=
   | vs=plist2(discard_or_identifier) ; {
       if List.for_all is_local_ignored vs then
         Error.fatal_here $startpos $endpos @@
-          Error.CannotParse (Some "A local declaration must declare at least one name.")
+          Error.BadDeclarationSyntax
+            "A local declaration must declare at least one name."
       else LDI_Tuple vs
     }
 
--- a/asllib/Typing.ml
+++ b/asllib/Typing.ml
@@ -2188,7 +2188,8 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
                     E_GetItem (e2, index) |> add_pos_from ~loc:e,
                     ses1 )
                 else
-                  fatal_from ~loc (Error.BadField (field_name, t_e2))
+                  fatal_from ~loc
+                    (Error.BadTupleIndex { index; length = List.length tys })
                   |: TypingRule.EGetTupleItem
             (* End *)
             (* Begin EGetBadField *)
--- a/asllib/Typing.ml
+++ b/asllib/Typing.ml
@@ -39,9 +39,6 @@ let add_pos_from ~loc = add_pos_from loc
 let conflict ~loc expected provided =
   fatal_from ~loc (Error.ConflictingTypes (expected, provided))
 
-let type_satisfaction_failure ~loc expected provided =
-  fatal_from ~loc (Error.TypeSatisfactionFailure (expected, provided))
-
 let plus e1 e2 = binop `ADD e1 e2
 let minus e1 e2 = binop `SUB e1 e2
 let t_bits_bitwidth e = T_Bits (e, [])
@@ -581,7 +578,9 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
         Format.eprintf "@[<hv 2>Checking %a@ <: %a@]@." PP.pp_ty t1 PP.pp_ty t2
     in
     if Types.type_satisfies env t1 t2 then ()
-    else type_satisfaction_failure ~loc [ t2.desc ] t1
+    else
+      fatal_from ~loc
+        (Error.TypeSatisfactionFailure { expected = t2; provided = t1 })
 
   (* CheckStructureBoolean *)
 
@@ -2751,8 +2750,11 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
     | _ -> Types.type_satisfies env t s
 
   let check_can_be_initialized_with ~loc env s t () =
-    if can_be_initialized_with env s t then ()
-    else type_satisfaction_failure ~loc [ s.desc ] t
+    if can_be_initialized_with env s t then
+      () |: TypingRule.CheckCanBeInitialisedWith
+    else
+      fatal_from ~loc
+        (Error.TypeSatisfactionFailure { expected = s; provided = t })
   (* End *)
 
   (* Begin ShouldRememberImmutableExpression *)
--- a/asllib/Typing.ml
+++ b/asllib/Typing.ml
@@ -39,6 +39,9 @@ let add_pos_from ~loc = add_pos_from loc
 let conflict ~loc expected provided =
   fatal_from ~loc (Error.ConflictingTypes (expected, provided))
 
+let type_satisfaction_failure ~loc expected provided =
+  fatal_from ~loc (Error.TypeSatisfactionFailure (expected, provided))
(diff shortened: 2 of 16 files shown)


asllib/Parser.mly
```diff
@@ -88,24 +88,22 @@ let check_is_associative ~loc (op : AST.binop) =
...
         fatal_from loc
-          (CannotParse
-             (Some
-                (Format.sprintf
-                   "Binary operator `%s` is not associative - parenthesise to \
-                    disambiguate."
-                   (PP.binop_to_string op)))))
+          (BadBinopPriority
+             (Format.sprintf
+                "Binary operator `%s` is not associative - parenthesise to \
```
HadrienRenaud, line 93: Does the verb "parenthesise" exist? I'd suggest "Add parentheses" instead

asllib/Parser.mly
```diff
@@ -88,24 +88,22 @@ let check_is_associative ~loc (op : AST.binop) =
...
       fatal_from loc
-        (CannotParse
-           (Some
-              (Format.sprintf
-                 "Operators `%s` and `%s` have the same priority - parenthesise \
-                  to disambiguate."
-                 (PP.binop_to_string op) (PP.binop_to_string op')))))
+        (BadBinopPriority
+           (Format.sprintf
+              "Operators `%s` and `%s` have the same priority - parenthesise \
```
HadrienRenaud, line 104: Same here

asllib/error.ml
```diff
@@ -31,6 +31,7 @@ type error_desc =
   | BadField of string * ty
   | MissingField of string list * ty
   | BadSlices of error_handling_time * slice list * int
+  | BadTupleIndex of int * int
```
HadrienRenaud: nit: maybe a record would be nice?

```ocaml
| BadTupleIndex of { index: int; length: int }
```

asllib/Parser.mly
Roman-Manevich: It's a proper verb (https://www.merriam-webster.com/dictionary/parenthesize).

psafont: It's more common with a z, but the verb does exist :)

hrutvik: Sometimes "-ise" is considered British English, and "-ize" is considered American English

asllib/Typing.ml
```diff
@@ -39,6 +39,9 @@ let add_pos_from ~loc = add_pos_from loc
 let conflict ~loc expected provided =
   fatal_from ~loc (Error.ConflictingTypes (expected, provided))
 
+let type_satisfaction_failure ~loc expected provided =
```
hrutvik: I would prefer not to introduce more of these shorthands that are rarely used and don't really save much verbosity

asllib/doc/ErrorCodes.tex
hrutvik, line 1: Does the type satisfaction error description also need updating to mention `check_can_be_initialized_with`?

asllib/Parser.mly
HadrienRenaud: Oh wow, the british spelling does not show up on the first page of results

All good then :)

asllib/doc/ErrorCodes.tex
Roman-Manevich: I think it is clear enough.

asllib/error.ml
```diff
@@ -49,8 +50,11 @@ type error_desc =
   | InvalidExpr of expr
   | MismatchType of string * type_desc list
   | ConflictingTypes of type_desc list * ty
+  | TypeSatisfactionFailure of type_desc list * ty
```
HadrienRenaud: A few questions on this:
1. I think only the case for a singleton is used here, can we remove the list?

```suggestion
  | TypeSatisfactionFailure of type_desc * ty
```

2. Can we use a record?

```suggestion
  | TypeSatisfactionFailure of { expected: type_desc; provided: ty }
````

3. Can we leave the location annotations on the first member?

```suggestion
  | TypeSatisfactionFailure of { expected: ty; provided: ty }
```

Feel free to argue against any of those suggestions and only take the combination of what seems reasonable.

asllib/Typing.ml
Roman-Manevich: It is symmetric to `conflict`. If we have a shorthand for one, why not for the other?

asllib/error.ml
```diff
@@ -49,8 +50,11 @@ type error_desc =
...
   | AssertionFailed of error_handling_time * expr
   | CannotParse of string option
+  | BadBinopPriority of string
+  | BadDeclarationSyntax of string
```
HadrienRenaud: Why leave a message on this one if it's only used once? Are there plans to add more cases to it?

asllib/error.ml
```diff
@@ -49,8 +50,11 @@ type error_desc =
   | InvalidExpr of expr
   | MismatchType of string * type_desc list
   | ConflictingTypes of type_desc list * ty
+  | TypeSatisfactionFailure of type_desc list * ty
   | AssertionFailed of error_handling_time * expr
   | CannotParse of string option
+  | BadBinopPriority of string
```
HadrienRenaud, line 56: Now would be the time to refactor this one if you wanted:

```ocaml
type bpp_cases =
  | BPP_NonAssociativeOp of binop
  | BPP_SamePriorityOps of binop * binop
```

```suggestion
  | BadBinopPriority of bpp_cases
```

And move the error messages to `error.ml`.

But IMHO totally optional

asllib/doc/ErrorCodes.tex
hrutvik: It seems inconsistent with the other descriptions - these distinguish between "examples" of rules that produce the error, and "the" rule that produces the error. Please either:
- _(preferred)_ add TypingRule.CheckCanBeInitializedWith
- clarify that TypingRule.TypeSatisfaction is only an example of a rule that produces TE_TSF

asllib/Typing.ml
hrutvik: `conflict` wasn't introduced in this PR, and in any case has far more usages.

I also notice that `type_satisfaction_failure` is only ever called with a singleton list argument - is this right?

HadrienRenaud: `conflict` is used 22 times in when `type_satisfaction_failure` is used only twice.

IMHO this is a good reason for having one and not the other

asllib/error.ml
hrutvik: > I think only the case for a singleton is used here, can we remove the list?

Agreed on all, but this in particular

asllib/error.ml
hrutvik: The other usage should be in ASTRule.SetBuiltin, which seems to be `Builder.make_builtin` in the implementation - this currently just prints to `stderr` and exits. I think we should align it as part of this PR if possible

asllib/Typing.ml
Roman-Manevich: okay, I inlined the function.

asllib/error.ml
Roman-Manevich: I changed the constructor to accept a single type. I'm keeping the constructor as a tuple, since it allows to compactly match:
```
| ConflictingTypes ([ expected ], provided)
| TypeSatisfactionFailure (expected, provided) ->
        pp_err Typing "a subtype of@ %a@ was expected,@ provided %a."
          pp_type_desc expected pp_ty provided
```

HadrienRenaud: Wouldn't you be able to do:
```ocaml
| ConflictingTypes ([ expected ], provided)
| TypeSatisfactionFailure { expected; provided } ->
        pp_err Typing "a subtype of@ %a@ was expected,@ provided %a."
          pp_type_desc expected pp_ty provided
```

And what about point 3?

Roman-Manevich: done.

asllib/error.ml
Roman-Manevich: Replaced `BadDeclarationSyntax of string` with `AllDiscardLocalDeclaration`.

asllib/doc/ErrorCodes.tex
Roman-Manevich: I'm not sure I understand. I added `|: TypingRule.CheckCanBeInitialisedWith` to `check_can_be_initialized_with` - is that what you mean by "add TypingRule.CheckCanBeInitializedWith"?

Where were you think the clarification would go? In a new diagnostic message for perhaps a new error kind like `CannotBeInitialisedWith`?

asllib/error.ml
HadrienRenaud: @Roman-Manevich what about point 3?

asllib/error.ml
HadrienRenaud: What about Hrutvik's comment on the case in `Builder.make_builtin`?

asllib/error.ml
Roman-Manevich: done. this required separating the pattern branches.

asllib/error.ml
Roman-Manevich: `make_builtin` now uses `NonFunctionBuiltinDeclaration`.

asllib/doc/ErrorCodes.tex
hrutvik: This comment is on `ErrorCodes.tex`, and the suggested change is for that file too. The suggestion is as you have done with BadDeclaration - add a back-reference in the prose description of the TE_TSF error that mentions the other rule that can cause this error, i.e. TypingRule.CheckCanBeInitialisedWith.

Roman-Manevich: I added text to the description of `TE_SEF` to that effect.

HadrienRenaud (approved):
I haven't had a look at the reference, but otherwise LGTM

asllib/doc/ErrorCodes.tex
```diff
@@ -246,6 +247,8 @@ \section{Type Errors\label{sec:TypingErrors}}
 \item[$\TypeSatisfactionFailure$]
   \textit{Type satisfaction failure.}
   See \TypingRuleRef{TypeSatisfaction}.
+  For example, a value of \booleantypeterm{} is assigned to a variable of \integertypeterm{}
+  (See \TypingRuleRef{CheckCanBeInitializedWith}).
```
hrutvik, line 251: This remains inconsistent with the rest of this section.

PR #1989, opened 2026-09-04
Aligning implementation/specification error codes - phase 3/5

Roman-Manevich:
## Distinguish lexical and parse failures
Lexer-detected invalid symbols now use `UnknownSymbol` and report `BE_LE`, preserving suggested replacements, while the remaining `CannotParse` errors report `BE_PE`.

## Classify global-initialization exceptions
Exceptions escaping global initialization now report `DE_UE`, and `build_genv` in `asl.spec` converts the corresponding throwing result to `DynamicError(DE_UE)`.

## Classify asserted-type-conversion failures
Failed asserted type conversions now use `ATCFailure`, reporting `DE_TAF` during dynamic evaluation and `TE_SEF` during static evaluation as specified by `asl.spec`.

## Classify repeated tuple-field writes
Repeated fields in tuple assignments now report `TE_IAD`, matching the specification's use of `check_no_duplicates` rather than the rework PR's disputed `BE_PE` mapping.

## Distinguish call and tuple arity failures
Static call arity failures now report `TE_BC` through `BadCallArity`, while tuple arity failures report `TE_UT` through `BadTupleArity`; unchecked runtime arity failures remain uncoded.

### DIFF
--- a/asllib/Interpreter.ml
+++ b/asllib/Interpreter.ml
@@ -530,7 +530,8 @@ module Make (B : Backend.S) (C : Config) = struct
         let* b = is_val_of_type e1 env v t in
         (if b then return_normal (v, new_env)
          else
-           fatal_from e1 env (Error.MismatchType (B.debug_value v, [ t.desc ])))
+           fatal_from e1 env
+             (Error.ATCFailure (C.error_handling_time, B.debug_value v, t.desc)))
         |: SemanticsRule.ATC
     (* End *)
     (* Begin EvalEVar *)
--- a/asllib/Interpreter.ml
+++ b/asllib/Interpreter.ml
@@ -531,7 +531,8 @@ module Make (B : Backend.S) (C : Config) = struct
         (if b then return_normal (v, new_env)
          else
            fatal_from e1 env
-             (Error.ATCFailure (C.error_handling_time, B.debug_value v, t.desc)))
+             (Error.ATCExecutionFailure
+                (C.error_handling_time, B.debug_value v, t)))
         |: SemanticsRule.ATC
     (* End *)
     (* Begin EvalEVar *)
--- a/asllib/Lexer.mll
+++ b/asllib/Lexer.mll
@@ -453,9 +453,9 @@ and token = parse
     | '<'                      { LT                               }
     | ">>"                     { SHR                              }
     | "&&"                     { BAND                             }
-    | "-->"                    {
+    | "-->" as symbol          {
         fatal lexbuf
-          (UnknownSymbol { symbol = "-->"; alternative = Some "==>" })
+          (UnknownSymbol { symbol; alternative = Some "==>" })
       }
     | "==>"                    { IMPL                             }
     | "<<"                     { SHL                              }
@@ -467,9 +467,9 @@ and token = parse
     | '{'                      { LBRACE                           }
     | "!="                     { NE                              }
     | '-'                      { MINUS                            }
-    | "<->"                    {
+    | "<->" as symbol          {
         fatal lexbuf
-          (UnknownSymbol { symbol = "<->"; alternative = Some "<=>" })
+          (UnknownSymbol { symbol; alternative = Some "<=>" })
       }
     | "<=>"                    { BEQ                              }
     | '['                      { LBRACKET                         }
--- a/asllib/Lexer.mll
+++ b/asllib/Lexer.mll
@@ -453,7 +453,10 @@ and token = parse
     | '<'                      { LT                               }
     | ">>"                     { SHR                              }
     | "&&"                     { BAND                             }
-    | "-->"                    { fatal lexbuf (CannotParse (Some "Did you mean `==>`?")) }
+    | "-->"                    {
+        fatal lexbuf
+          (UnknownSymbol { symbol = "-->"; alternative = Some "==>" })
+      }
     | "==>"                    { IMPL                             }
     | "<<"                     { SHL                              }
     | ']'                      { RBRACKET                         }
@@ -464,7 +467,10 @@ and token = parse
     | '{'                      { LBRACE                           }
     | "!="                     { NE                              }
     | '-'                      { MINUS                            }
-    | "<->"                    { fatal lexbuf (CannotParse (Some "Did you mean `<=>`?")) }
+    | "<->"                    {
+        fatal lexbuf
+          (UnknownSymbol { symbol = "<->"; alternative = Some "<=>" })
+      }
     | "<=>"                    { BEQ                              }
     | '['                      { LBRACKET                         }
     | "[["                     { LLBRACKET                        }
--- a/asllib/Typing.ml
+++ b/asllib/Typing.ml
@@ -1609,8 +1609,12 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
                List.length params )
       else if List.compare_lengths func_sig.args args != 0 then
         fatal_from ~loc
-        @@ Error.BadArity
-             (Static, name, List.length func_sig.args, List.length args)
+        @@ Error.BadCallArity
+             {
+               name;
+               expected = List.length func_sig.args;
+               provided = List.length args;
+             }
     in
     (* Check that call parameters are statically evaluable and type-satisfy the
        declaration parameters *)
@@ -1726,8 +1730,12 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
     let () =
       if List.compare_lengths callee.args args1 != 0 then
         fatal_from ~loc
-        @@ Error.BadArity
-             (Static, name, List.length callee.args, List.length args1)
+        @@ Error.BadCallArity
+             {
+               name;
+               expected = List.length callee.args;
+               provided = List.length args1;
+             }
     in
     let eqs2 =
       let folder acc (_x, ty) (t_e, _e, _ses) =
@@ -2516,11 +2524,8 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
           | T_Tuple tys ->
               if List.compare_lengths tys les != 0 then
                 Error.fatal_from le
-                  (Error.BadArity
-                     ( Static,
-                       "LEDestructuring",
-                       List.length tys,
-                       List.length les ))
+                  (Error.BadTupleArity
+                     { expected = List.length les; actual = List.length tys })
               else
                 let lhs_tys, les', sess =
                   List.map2 (annotate_lexpr_ty env) les tys |> list_split3
@@ -2799,11 +2804,8 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
     | T_Tuple lhs_tys, T_Tuple rhs_tys ->
         if List.compare_lengths lhs_tys rhs_tys != 0 then
           fatal_from ~loc
-            (Error.BadArity
-               ( Static,
-                 "tuple initialization",
-                 List.length rhs_tys,
-                 List.length lhs_tys ))
+            (Error.BadTupleArity
+               { expected = List.length lhs_tys; actual = List.length rhs_tys })
         else
           let lhs_tys' =
             List.map2
@@ -2837,11 +2839,8 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
           | T_Tuple tys when List.compare_lengths tys names = 0 -> tys
           | T_Tuple tys ->
               fatal_from ~loc
-                (Error.BadArity
-                   ( Static,
-                     "tuple initialization",
-                     List.length tys,
-                     List.length names ))
+                (Error.BadTupleArity
+                   { expected = List.length names; actual = List.length tys })
           | _ -> conflict ~loc [ T_Tuple [] ] ty
         in
         let new_env =
--- a/asllib/Typing.ml
+++ b/asllib/Typing.ml
@@ -580,7 +580,7 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
     if Types.type_satisfies env t1 t2 then ()
     else
       fatal_from ~loc
-        (Error.TypeSatisfactionFailure { expected = t2; provided = t1 })
+        (Error.TypeSatisfactionFailure { expected = t2; actual = t1 })
 
   (* CheckStructureBoolean *)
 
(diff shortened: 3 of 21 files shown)


asllib/Lexer.mll
```diff
@@ -453,7 +453,10 @@ and token = parse
     | '<'                      { LT                               }
     | ">>"                     { SHR                              }
     | "&&"                     { BAND                             }
-    | "-->"                    { fatal lexbuf (CannotParse (Some "Did you mean `==>`?")) }
+    | "-->"                    {
+        fatal lexbuf
+          (UnknownSymbol { symbol = "-->"; alternative = Some "==>" })
+      }
```
hrutvik:
```suggestion
    | "-->" as symbol                    {
        fatal lexbuf
          (UnknownSymbol { symbol; alternative = Some "==>" })
      }
```
(similarly below)

asllib/error.ml
```diff
@@ -747,7 +762,7 @@ module PPrint = struct
           (ErrorKind.of_error_handling_time t)
           "array@ length@ expression@ %a@ has@ negative@ length:@ %i." pp_expr
           e_length length
-    | MultipleWrites id -> pp_err Parse "multiple@ writes@ to@ %S." id
+    | MultipleWrites id -> pp_err Typing "multiple@ writes@ to@ %S." id
```
hrutvik, line 764: This error is reported during AST building (desugaring specifically) - why is it a typing error?

asllib/error.ml
```diff
@@ -42,21 +42,24 @@ type error_desc =
       found_call_type : subprogram_type;
     }
   | BadArity of error_handling_time * identifier * int * int
+  | BadCallArity of { name : identifier; expected : int; provided : int }
+  | BadTupleArity of { expected : int; actual : int }
   | BadParameterArity of error_handling_time * version * identifier * int * int
```
hrutvik, line 47: What are the status of `BadArity` and `BadParameterArity` after the new `Bad{Call,Tuple}Arity`? For example, is `BadArity` an V0-only error, and only used with `Dynamic`? If so, we should remove `error_handling_time` from `BadArity`.

asllib/error.ml
```diff
@@ -270,24 +274,26 @@ module ErrorCode = struct
     | ExpectedNamedType _ | UnexpectedCollection | MismatchedBitvectorWidths _
     | CollectionBaseNotVariable _ ->
         Some (Typing UT)
-    | MismatchedCallType _
+    | MismatchedCallType _ | BadCallArity _
     | BadParameterArity (Static, _, _, _, _)
```
hrutvik, line 278: nit: perhaps we should also match on the version here, i.e. `BadParameterArity (Static, V1, ...)`?

asllib/error.ml
```diff
@@ -42,21 +42,24 @@ type error_desc =
...
+  | BadCallArity of { name : identifier; expected : int; provided : int }
+  | BadTupleArity of { expected : int; actual : int }
   | BadParameterArity of error_handling_time * version * identifier * int * int
   | UnsupportedBinop of error_handling_time * binop * literal * literal
   | UnsupportedUnop of error_handling_time * unop * literal
   | UnsupportedExpr of error_handling_time * expr
   | UnsupportedTy of error_handling_time * ty
   | InvalidExpr of expr
   | MismatchType of string * type_desc list
+  | ATCFailure of error_handling_time * string * type_desc
```
hrutvik: This is a bit confusing with `BadATC` - could we rename to clarify? E.g.
- `ATCExecutionFailure`
- `ATCTypecheckFailure`

or similar?

asllib/error.ml
```diff
@@ -588,14 +598,19 @@ module PPrint = struct
...
+        let codes = List.map Char.code (List.of_seq (String.to_seq symbol)) in
         let not_printable code = code < 33 || code > 126 in
         if List.exists not_printable codes then
           pp_err Lexical "Unknown symbol (ASCII code point(s): %a)."
             (pp_comma_list pp_print_int)
             codes
-        else pp_err Lexical "Unknown symbol."
+        else
+          match alternative with
+          | None -> pp_err Lexical "Unknown symbol."
```
hrutvik:
```suggestion
          | None -> pp_err Lexical "Unknown symbol %S." symbol
```

asllib/error.ml
```diff
@@ -588,14 +598,19 @@ module PPrint = struct
...
+          | Some alternative ->
+              pp_err Lexical "Unknown symbol %S.@ Did you mean %S?" symbol
+                alternative)
```
hrutvik: Or better:

```suggestion
          let pp_alternative fmt alt = fprintf fmt "@ Did you mean %S?" alt in
          pp_err Lexical "Unknown symbol %S.%a" (pp_option pp_alternative) alternative
```

asllib/Typing.ml
```diff
@@ -1726,8 +1730,12 @@ module Annotate (C : ANNOTATE_CONFIG) : S = struct
     let () =
       if List.compare_lengths callee.args args1 != 0 then
         fatal_from ~loc
-        @@ Error.BadArity
-             (Static, name, List.length callee.args, List.length args1)
+        @@ Error.BadCallArity
```
hrutvik, line 1733: This is an ASL0-only error site - does it need to change?

Roman-Manevich: As long as we support ASL0, this seems like an improvement in terms of the error code and diagnostic message.

asllib/Lexer.mll
Roman-Manevich: done

asllib/error.ml
Roman-Manevich: This is a bit subtle. `AssignableExpressions.tex` defines the (AST) desugaring rule by calling `check_no_duplicates`, which returns a type error. We can roll back to `Parse` but then there's a discrepancy between the implementation and documentation.
Alternatively, we could change the spec (say, by creating a dedicated duplicates check) so that raises a build error.

asllib/error.ml
Roman-Manevich: `BadArity` is used in both ASLv0 and ASLv1 contexts and its `error_handling_time` depends on the configuration passed to `Make` in `Interpreter.ml` so both static/dynamic are possible. The situation is similar with `BadParameterArity`.

asllib/error.ml
Roman-Manevich: `MismatchedCallType` is used both for ASLv0 and ASLv1 in `eval_subprogram`.

asllib/error.ml
Roman-Manevich: Renamed to `ATCExecutionFailure`.

asllib/error.ml
Roman-Manevich: done. By the way, the error location already highlights the symbol.

asllib/error.ml
Roman-Manevich: ok

asllib/error.ml
```diff
@@ -49,6 +49,7 @@ type error_desc =
   | UnsupportedTy of error_handling_time * ty
   | InvalidExpr of expr
   | MismatchType of string * type_desc list
+  | ATCFailure of error_handling_time * string * type_desc
```
HadrienRenaud: Why not?
```suggestion
  | ATCFailure of error_handling_time * string * ty
```

asllib/error.ml
```diff
@@ -498,6 +500,10 @@ module PPrint = struct
           "Mismatch type:@ value %s@ does not subtype any of those types:@ %a" v
           (pp_comma_list pp_type_desc)
           li
+    | ATCFailure (t, v, ty) ->
+        pp_err
+          (ErrorKind.of_error_handling_time t)
+          "Value %s does not satisfy the asserted type %a." v pp_type_desc ty
```
HadrienRenaud: This is a bit ugly on some types, please add break spaces *and* the box

```suggestion
          "Value %s@ does@ not@ satisfy@ the@ asserted@ type@ @[%a@]." v pp_type_desc ty
```

asllib/Interpreter.ml
```diff
@@ -530,7 +530,8 @@ module Make (B : Backend.S) (C : Config) = struct
         let* b = is_val_of_type e1 env v t in
         (if b then return_normal (v, new_env)
          else
-           fatal_from e1 env (Error.MismatchType (B.debug_value v, [ t.desc ])))
+           fatal_from e1 env
+             (Error.ATCFailure (C.error_handling_time, B.debug_value v, t.desc)))
```
HadrienRenaud: Maybe we should have something a bit more flexible than `debug_value`, ideally a formatter

Probably not for this PR.

asllib/error.ml
```diff
@@ -42,6 +42,8 @@ type error_desc =
       found_call_type : subprogram_type;
     }
   | BadArity of error_handling_time * identifier * int * int
```
HadrienRenaud, line 44: What is `BadArity` used for then? Maybe add a little comment to explain this?

asllib/tests/lexer.t
```diff
@@ -47,7 +47,7 @@
   File println6.asl, line 3, character 0:
   
   
-  ASL Lexical error (BE_LE): Unknown symbol.
+  ASL Lexical error (BE_LE): Unknown symbol "".
```
HadrienRenaud, line 50: Oh that's ugly, should we create an issue about this?

asllib/error.ml
hrutvik: The spec needs changing.

asllib/error.ml
hrutvik: Uses of `BadArity` seem to be:

- In `Interpreter.ml` for:
    - incorrect argument/parameter arity for a subprogram call
    - `protected_multi_assign`
    - `run_typed`
- in `Native.ml` for primitives

From a brief look, these all seem to be for untyped specifications - I believe ASL1 type-checking would catch all of these before interpretation. Am I missing something?

asllib/error.ml
hrutvik: I am not sure what this comment means. ASL0 errors do not need to produce error codes, as error codes are only defined for ASL1. This is why matching on version would make sense here. As this is only a nit, this can be left as-is.

asllib/tests/lexer.t
hrutvik: Why has this regressed?

asllib/Typing.ml
hrutvik: ASL0 does not have error codes

asllib/error.ml

asllib/error.ml
Roman-Manevich: okay

asllib/Typing.ml
Roman-Manevich: Still, this is consistent with the same check for ASLv1. I don't see how this hurts.

asllib/error.ml
Roman-Manevich: answered offline.

asllib/error.ml
Roman-Manevich: Reverted `Typing` back to `Parse`.
Changed the spec to compare the cardinality of the set of fields with the length of the list and have it emit `BE_PE` if they differ.

asllib/error.ml
Roman-Manevich: How about
```
An arity mismatch detected during evaluation: a subprogram or
          primitive receives the wrong number of arguments, a tuple assignment
          receives the wrong number of values from a subprogram call, or an
          entry point returns the wrong number of values.
 ```
?

Comment (HadrienRenaud) [reply asllib/error.ml:44]: I like it, what about adding quickly at the start that this should not happen if type-checking worked correctly?

File: asllib/error.ml [hunk 14]
```diff
@@ -42,21 +42,28 @@ type error_desc =
...
+      (** An arity mismatch detected during evaluation: a subprogram or
+          primitive receives the wrong number of arguments, a tuple assignment
+          receives the wrong number of values from a subprogram call, or an
+          entry point returns the wrong number of values. *)
+  | BadCallArity of { name : identifier; expected : int; provided : int }
+  | BadTupleArity of { expected : int; actual : int }
```
HadrienRenaud: There's a naming mismatch between those 2 declaration: should we use `provided` or `actual`?

asllib/error.ml
```diff
@@ -42,21 +42,28 @@ type error_desc =
       found_call_type : subprogram_type;
     }
   | BadArity of error_handling_time * identifier * int * int
+      (** An arity mismatch detected during evaluation: a subprogram or
+          primitive receives the wrong number of arguments, a tuple assignment
+          receives the wrong number of values from a subprogram call, or an
+          entry point returns the wrong number of values. *)
```
HadrienRenaud: This is nice, just a tiny edit:

```suggestion
      (** [BadArity (time, name, expected, provided)] is raised when there is 
          an arity mismatch detected during evaluation when it should have been
          detected at type-checking: a subprogram or
          primitive receives the wrong number of arguments, a tuple assignment
          receives the wrong number of values from a subprogram call, or an
          entry point returns the wrong number of values. *)
```

HadrienRenaud (approved):
This looks ok to me, but please double check with Hrutvik that he's happy with it

asllib/Typing.ml
hrutvik: I'm not sure why it needed to be in the scope of this PR, but fine

asllib/error.ml
hrutvik: I vote `actual`

hrutvik (approved):
I think just a few minor bits to clear up before merging:

- Hadrien's comment on `provided` vs. `actual`
- Hadrien's comment on `BadArity` docstring

And then we can continue offline discussion on `BadArity`

asllib/error.ml
Roman-Manevich: answered offline.

asllib/tests/lexer.t
Roman-Manevich: done

asllib/error.ml
Roman-Manevich: I replaced `provided` with `actual`.

asllib/error.ml
Roman-Manevich: okay

PR #1991, opened 2026-09-04
[litmus] Warn when timebase counter is not available in standard mode.

maranget:
By contrast with presi/kvm mode, if no timebase counter  is available in standard mode, there is no fatal error.

Instead, synchronisation is performed by a simple reverse sense barrier, without the extra poll loop for time based synchronisation. Warn users about this.


psafont (approved):

maranget:
Merged, thanks @psafont.

PR #1992, opened 2026-09-04
Slight output format changes for **herd7** and **litmus7**

maranget:
Those changes are:
  1. **herd7**: get rid of one spurious space.
  2.  **litmus7**: end each test output with a blank line, to match other tools behaviour.
Those mundane modification entail some modifification of commited reference logs.

Modification 2. originates in the remark from PR #1973 that finding the end of  test outputs in a **litmus7** log is more involved than in the logs of other tools. Modification 1 results in nicer **herd7** logs.


fsestini (approved):
Thanks @maranget . Having a more unified log format is a desirable goal, so I welcome this change.

PR #1993, opened 2026-09-05
Aligning implementation/specification error codes - phase 4/5

Roman-Manevich:
## Classify unsupported parameter expressions
Expressions that cannot define subprogram parameters now report `TE_BSPD` through `BadParameterExpr`, matching `params_of_expr` in `asl.spec`.

## Classify invalid collection field types
Non-bitvector collection fields now report `TE_UT` through `check_structure_bits`, matching the per-field `check_structure_label` checks in `asl.spec`.

## Classify unsupported parameter types
Types not permitted as subprogram argument or return types now report `TE_BSPD` through `BadParameterType`; `asl.spec` and a documented regression now cover pending-constrained integers, and the obsolete `UnsupportedTy` diagnostic is removed.

## Classify out-of-bounds indices
Out-of-bounds native vector accesses now report `DE_BI`, or `TE_SEF` during static evaluation, through `BadIndex` instead of the uncoded `MismatchType` diagnostic.

## Classify arbitrary negative-length arrays
Parameterized arrays whose length evaluates negatively now report `DE_AET` through `ArbitraryEmptyType`, matching `EArbitrary` in `asl.spec`.

## Classify static evaluation failures
Static evaluation that does not produce a literal now reports `TE_SEF` through `StaticEvaluationFailure`, matching `static_eval` in `asl.spec`.

## Classify implementation integer overflow
Integers exceeding OCaml `int` limits now report a structured internal `ImplementationIntegerOverflow` diagnostic instead of the uncoded `UnsupportedExpr`.

### DIFF
--- a/asllib/error.ml
+++ b/asllib/error.ml
@@ -54,8 +54,8 @@ type error_desc =
   | BadParameterArity of error_handling_time * version * identifier * int * int
   | UnsupportedBinop of error_handling_time * binop * literal * literal
   | UnsupportedUnop of error_handling_time * unop * literal
-  | UnsupportedExpr of error_handling_time * expr
   | StaticEvaluationFailure of expr
+  | ImplementationIntegerOverflow of Z.t
   | InvalidExpr of expr
   | MismatchType of string * type_desc list
   | ATCExecutionFailure of error_handling_time * string * ty
@@ -325,8 +325,9 @@ module ErrorCode = struct
         Some (Typing SEF)
     | BadIndex (Dynamic, _, _) -> Some (Dynamic BI)
     | NoCommonAncestor _ (* LCA failures *) -> Some (Typing LCA)
-    (********** TODO tidy up - does not cleanly correspond to a code **********)
-    | UnsupportedExpr _ -> None
+    (********** Errors without specification codes **********)
+    (* Implementation limitations are not ASL errors. *)
+    | ImplementationIntegerOverflow _ -> None
     (********** Should not happen **********)
     (* e.g. skipped type-checking, ASL0, internal option or invariant *)
     | MismatchType _ (* Skipped type-checking or violated typing invariant *) ->
@@ -485,12 +486,11 @@ module PPrint = struct
           (ErrorKind.of_error_handling_time t)
           "Illegal application of operator %s for value@ %a."
           (unop_to_string op) pp_literal v
-    | UnsupportedExpr (t, e) ->
-        pp_err
-          (ErrorKind.of_error_handling_time t)
-          "Unsupported expression %a." pp_expr e
     | StaticEvaluationFailure e ->
         pp_err Typing "Static evaluation of expression %a failed." pp_expr e
+    | ImplementationIntegerOverflow z ->
+        pp_err Internal "Integer %a exceeds aslref implementation limits."
+          Z.pp_print z
     | InvalidExpr e -> pp_err Typing "invalid expression %a." pp_expr e
     | MismatchType (v, [ ty ]) ->
         pp_err Dynamic "Mismatch type:@ value %s does not belong to type %a." v
@@ -879,8 +879,8 @@ module CSV = struct
     | BadParameterArity _ -> "BadParameterArity"
     | UnsupportedBinop _ -> "UnsupportedBinop"
     | UnsupportedUnop _ -> "UnsupportedUnop"
-    | UnsupportedExpr _ -> "UnsupportedExpr"
     | StaticEvaluationFailure _ -> "StaticEvaluationFailure"
+    | ImplementationIntegerOverflow _ -> "ImplementationIntegerOverflow"
     | InvalidExpr _ -> "InvalidExpr"
     | MismatchType _ -> "MismatchType"
     | ATCExecutionFailure _ -> "ATCExecutionFailure"
--- a/asllib/error.ml
+++ b/asllib/error.ml
@@ -90,6 +90,7 @@ type error_desc =
   | ParameterWithoutDecl of identifier
   | BadParameterDecl of identifier * identifier list * identifier list
       (** name, expected, actual *)
+  | BadParameterExpr of expr
   | BaseValueEmptyType of ty
   | ArbitraryEmptyType of ty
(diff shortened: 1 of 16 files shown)


asllib/error.ml
```diff
@@ -31,6 +31,7 @@ type error_desc =
   | BadField of string * ty
   | MissingField of string list * ty
   | BadSlices of error_handling_time * slice list * int
+  | BadIndex of error_handling_time * int * int
```
HadrienRenaud: worth a record I think

asllib/tests/regressions.t/run.t
```diff
@@ -813,6 +813,12 @@ Static errors:
   ASL Static error (TE_SEF):
     FloorLog2 (primitive) expected an argument greater than 0
   [1]
+  $ aslref --no-exec static-evaluation-non-literal.asl
+  File static-evaluation-non-literal.asl, line 5, characters 13 to 24:
+  constant C = R { x = 1 };
+               ^^^^^^^^^^^
+  ASL Type error (TE_SEF): Static evaluation of expression R { x = 1 } failed.
```
HadrienRenaud: Oh I had forgotten we had this limitation. Could we put it on a ticket please?

asllib/error.ml
```diff
@@ -487,6 +484,8 @@ module PPrint = struct
         pp_err
           (ErrorKind.of_error_handling_time t)
           "Unsupported expression %a." pp_expr e
+    | StaticEvaluationFailure e ->
+        pp_err Typing "Static evaluation of expression %a failed." pp_expr e
```
hrutvik: Would it be more descriptive to say "static evaluation ... did not successfully produce a literal"?

hrutvik (approved):

asllib/error.ml
Roman-Manevich: I prefer records, but why this specific one and not all other constructors with multiple fields?

asllib/tests/regressions.t/run.t
Roman-Manevich: I'm not sure what you want the ticket to say.

asllib/error.ml
Roman-Manevich: okay

asllib/error.ml
hrutvik: I can see at least two reasons:

- A record clearly differentiates between the two arguments with the same `int` type.
- This constructor was added in this PR - we can make it better quality than other constructors without being inconsistent.

asllib/tests/regressions.t/run.t
```diff
@@ -822,10 +813,22 @@ Static errors:
...
   $ aslref tuple-arity-mismatch.asl
   File tuple-arity-mismatch.asl, line 3, characters 2 to 25:
     let (x, y) = (1, 2, 3);
     ^^^^^^^^^^^^^^^^^^^^^^^
   ASL Type error (TE_UT): Tuple arity mismatch:
     expected 2 element(s); provided 3.
   [1]
+
+Implementation errors:
+  $ aslref --gnu-errors implementation-integer-overflow.asl
```
hrutvik: nit: why `--gnu-errors` here?

hrutvik (approved):
Please ensure @HadrienRenaud has approved before merging.

asllib/tests/regressions.t/run.t
HadrienRenaud: Well that ASLRef does not support initialising constants with compound expressions when it is not forbidden by type-checking?

asllib/error.ml
HadrienRenaud: In general, I want to prevent this type of error: you have switched your arguments around and the type system does not catch it.
For example, it is not obvious that this code is wrong, and it will be accepted by the ocaml type-checker:
```ocaml
  fatal ~loc (Error.BadIndex (Static, actual, expected))
```
By naming things we can make this class of errors far more difficult to happen:
```ocaml
  fatal ~loc (Error.BadIndex { error_handling_time = Static; actual; expected })
```

This is the same reasoning for naming arguments to functions. The only exception to this is when it things are left-to-right ordered, e.g. `plus x y` makes as much sense as `plus ~lhs:x ~rhs:y`.

Roman-Manevich: No argument there - as I said, I prefer records for these very reasons. I was just questioning the ad-hoc application of this refactoring to this one constructor, as opposed to uniformly switching all similar constructors to records.
In any case, I'll switch this one.

asllib/tests/regressions.t/run.t
Roman-Manevich: But the type system does forbid it, albeit indirectly, by invoking static evaluation, which returns an error for non-literal results.

asllib/tests/regressions.t/run.t
Roman-Manevich: I thought it would be a good idea to exercise this option and using it for implementation-specific issues seems like the right test. That said, I don't mind removing it.

asllib/error.ml
hrutvik: As above, improvement on new additions is a good step even in the absence of a complete refactor.

asllib/tests/regressions.t/run.t
hrutvik: This seems like a conflation of concerns, and we already have tests exercising this. I suggest removing because I can imagine us asking this question each time we see this test.

asllib/tests/regressions.t/run.t
HadrienRenaud: Moving offline.

HadrienRenaud (approved):
LGTM, thanks @Roman-Manevich

asllib/tests/regressions.t/run.t
Roman-Manevich: okay

hrutvik (approved):
Please fix CI before merging

Roman-Manevich:
> Please fix CI before merging

I'm re-runningn the failed but, but it isn't related to this PR. We seem to be having a rise in intermittent issues related to difference of hashes to packages required for `apt-get`.

PR #1994, opened 2026-09-06
Aligning implementation/specification error codes - phase 5/5

Roman-Manevich:
## Clarify empty integer constraints
Removed a stale TODO: ASL syntax and `asl.spec` require well-constrained integers to contain at least one constraint, so the implementation's empty case is an uncoded internal invariant.

## Align reserved-identifier errors
Reserved identifiers now produce `BE_RI` in the lexical formalism, matching the implementation.

## Classify invalid runtime slices
Runtime slices with a negative start or length now produce `DE_BI`, matching `asl.spec`; static slice failures remain `TE_BS` and use reason-specific `bad_slices` cases.

### DIFF
--- a/asllib/error.ml
+++ b/asllib/error.ml
@@ -428,8 +428,7 @@ end
       in reference?
     - Various errors are overused in several places - need to clearly
       distinguish between ASL1 errors and e.g. ASL0 non-typechecked errors,
-      assertion failures, cases we don't expect to hit etc.
-    - TypingRule.TInt mismatch on empty case *)
+      assertion failures, cases we don't expect to hit etc. *)
 (* TODO: BE_RI unused in reference *)
 
 module PPrint = struct
--- a/asllib/error.ml
+++ b/asllib/error.ml
@@ -26,18 +26,26 @@ open AST
 
 type error_handling_time = Static | Dynamic
 
+type bad_slices =
+  | NonPositiveLength of { slice : slice; length : int }
+      (** Converting a slice to positions during typing requires a positive
+          length. *)
+  | OutOfBitvectorBounds of slice list * int
+  | NegativeStartOrLength of error_handling_time * slice list
+      (** Native slicing permits a zero length, but requires non-negative starts
+          and lengths. *)
+
 type error_desc =
   | ReservedIdentifier of string
   | BadField of string * ty
   | MissingField of string list * ty
-  | BadSlices of error_handling_time * slice list * int
+  | BadSlices of bad_slices
   | BadIndex of {
       handling_time : error_handling_time;
       start : int;
       length : int;
     }
   | BadTupleIndex of { index : int; length : int }
-  | BadSlice of slice
   | EmptySlice
   | TypeInferenceNeeded
   | UndefinedIdentifier of error_handling_time * identifier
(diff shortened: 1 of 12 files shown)


asllib/error.ml
```diff
@@ -26,14 +26,22 @@ open AST
...
 type error_desc =
   | ReservedIdentifier of string
   | BadField of string * ty
   | MissingField of string list * ty
-  | BadSlices of error_handling_time * slice list * int
+  | BadSlices of bad_slices
   | BadIndex of error_handling_time * int * int
   | BadTupleIndex of { index : int; length : int }
-  | BadSlice of slice
   | EmptySlice
```
HadrienRenaud: Should empty slice be part of `BadSlices`?

asllib/error.ml
```diff
@@ -510,19 +517,27 @@ module PPrint = struct
...
+          "Slice selection %a includes a position outside the bounds of a \
+           bitvector of length %d."
+          pp_slice_list slices length
+    | BadSlices (NegativeStartOrLength (t, slices)) ->
         pp_err
           (ErrorKind.of_error_handling_time t)
-          "Cannot extract from bitvector of length %d slice %a." length
+          "Slice %a is invalid: its start and length must be non-negative."
           pp_slice_list slices
+    | BadSlices (NonPositiveLength { slice; length }) ->
```
hrutvik, line 540: Do we have a test showing this?

asllib/error.ml
Roman-Manevich: okay, done.

asllib/error.ml
Roman-Manevich: Added a test.

asllib/tests/ASLTypingReference.t/run.t
```diff
@@ -559,6 +559,15 @@ ASL Typing Tests / annotating types:
...
   ASL Warning: pragma implementation_hidden will be ignored.
   $ aslref --no-exec TypingRule.BitfieldSliceToPositions.asl
+  $ aslref --no-exec TypingRule.BitfieldSliceToPositions.bad.asl
+  File TypingRule.BitfieldSliceToPositions.bad.asl, line 1, character 0 to
+    line 3, character 2:
+  var myData: bits(16) {
+      [5+:0] data
+  };
+  ASL Static error (TE_BS):
+    Slice 5+:0 has length 0; slice lengths must be at least 1.
```
hrutvik: I realise this is a bit misleading - it's not the case that all slices must have length of at least one (we have zero-width slices). Please could we clarify this error message to explicitly restrict the cases in which a slice length must be at least 1? I believe this is in bitfields and on left-hand sides, but please confirm.

HadrienRenaud (approved):
Looks good to me, please wait resolution on Hrutvik's side before merging

asllib/tests/ASLTypingReference.t/run.t
Roman-Manevich: Right, this is indeed for both bitfields and left hand slide slices. Perhaps 
`"Slice %a has length %d; but the length of this slice must be at least 1."` is clearer?

hrutvik: Sounds good to me

hrutvik (approved):

PR #1995, opened 2026-09-09
ASL compliance testing tool

Roman-Manevich:
Add ASL compliance test runner prototype

This PR adds an ASL compliance testing tool (`asltest`) and test files under `asllib/compliance`. 
Tests are specified by pairs of `test.asl` and `test.yaml` (a.k.a metadata) files.

The tool can be run from `asllib/compliance` as `dune runtest`.

Details:
- Add a YAML schema for compliance test metadata.
- Add initial sample compliance tests under `asllib/compliance/tests`.
- Generate dune test rules dynamically from discovered .asl/.yaml test pairs.
- Add `asltest`, a test executable that runs `aslref`, and emits actual metadata.
- Compare expected and actual metadata using `diff`.
- Use structured `aslref` error codes and source positions from the `Runner`/`Error` interfaces.
- Keep YAML/schema dependencies optional so the compliance tests are skipped when the libraries are unavailable.


asllib/compliance/asltest.ml
```diff
@@ -0,0 +1,152 @@
+let read_file path = In_channel.with_open_text path In_channel.input_all
+
+let parse_args () =
```
HadrienRenaud, line 11: I still don't understand why we're doing this `--base BASENAME` when it is actually a mandatory argument. This could be a single positional argument which is a one liner:
```ocaml
let parse_args () =
  if Array.length Sys.argv != 2 then print_usage_and_exit ()
  else Sys.argv.(1)
```

asllib/compliance/asltest.ml
```diff
@@ -0,0 +1,152 @@
...
+    mode : mode;
+    outcome : outcome;
+    output : string option;
+    error_code : string option;
+    error_line : int option;
+    info : string option;
+  }
+end
+
+open TestCase
```
HadrienRenaud, line 53: @psafont was considering that it was useless to have a module then open it. He was suggesting just using `TestCase.foo` or abbreviate it to `TC.foo` if needed

asllib/compliance/dune
```diff
@@ -0,0 +1,21 @@
...
+
+(subdir
+ run
+ (dynamic_include ../generate/dune.inc))
+
+(subdir
+ generate
+ (rule
+  (deps
+   (glob_files_rec ../tests/*.asl)
```
HadrienRenaud, line 19: I'm not certain that this is the right specification for deps. I'm confused about the semantics of `glob_files_rec` and the dune specification does not say anything about it. Could you reassure me by testing it? For example with a nested directory in tests?

asllib/compliance/gen.ml
```diff
@@ -0,0 +1,60 @@
...
+
+let generate_rules actual_stems base =
+  (* Dune doesn't support targets in other directories (including directories
+     nested in the current one) until 3.24. Therefore, though the test directory
+     may have nested structure, we have to flatten it into a single directory
+     when generating the temporary "actual" tests outputs that we will diff with
+     the "expected" ones. The use of a map below ensures that we do not
+     accidentally clash names of "actual" files. *)
+  let actual, actual_stems =
+    let name = Filename.basename base in
```
HadrienRenaud, line 34: Should we add a test checking that tests base names are not of the form `foo.32`? Or do we feel confident with writing this assumption somewhere?

hrutvik (changes requested):
- Please stack the PRs to allow for review.
- Please remove `--compare` and associated code:
  - This duplicates the `dune` functionality for diffing and promotion unnecessarily.
  - We do not want to ignore the `info` field.
- Please remove the rename of `base` to `context` - this does not add anything to the PR, but creates a lot of noise for review.
- Please remove the schema validation (but keep the YAML schema). We are not yet decided on per-run schema validation or CI schema validation.
- As per the comment on `yaml_value_to_testcase`, and in the commit message for "[PoC] Round-tripping YAML files", please refactor to only consider the `mode` field on input.
- Please ensure you have addressed the following comments visible per-commit [here](https://github.com/herd/herdtools7/compare/master...hrutvik:herdtools7:compliance-testing):
  - Hadrien ([link](https://github.com/herd/herdtools7/commit/081627293b9700978c477f9242e08649d2474404#r197016784)), Hrutvik ([link](https://github.com/herd/herdtools7/commit/081627293b9700978c477f9242e08649d2474404#r198580832)), Pau ([link](https://github.com/herd/herdtools7/commit/081627293b9700978c477f9242e08649d2474404#r199052426))
  - I think the remaining comments we can handle another time (e.g. some are validation-related, or can be further discussed)

asllib/compliance/dune
hrutvik: The tests on my PoC branch should assure this (see [this commit](https://github.com/herd/herdtools7/commit/3d399f3a37c5a000f62ede58d69ef1eae7a5b0c2)) - we can add similar tests over here. I think I took the description of `glob_files_rec` from [here](https://dune.readthedocs.io/en/stable/concepts/dependency-spec.html#glob) in the docs, but I appreciate this is not overly detailed.

asllib/compliance/gen.ml
hrutvik: I think the invariants are perhaps:
- file discovery ensures that the "expected" files end in `.yaml`
- the "actual" files end in `.yaml` or `.yaml.<number>`

So I think if we had `foo.32` as a base name, then `foo.32.yaml.1` could be a possible "actual" file. Does this help your concern, or am I missing something?

asllib/compliance/asltest.ml
Roman-Manevich: okay.

hrutvik (changes requested):
- Please stack the PRs as previously requested.
- The commits in this PR no longer seem coherent for review or merge - please could we:
  - Fix-up commits appropriately: commits titled "Add initial compliance tests" and "Run aslref on generated rules" seem duplicated (?) and should be fixed-up
  - Rather than keeping my previous commits that added schema validation and then deleting it later, please drop the commits that add validation (a separate commit can introduce the YAML schema). The commits to delete are titled "Validate test cases for schema compliance" and "Validate wrt YAML schema instead of JSON schema"

asllib/compliance/asltest.ml
HadrienRenaud: Thank you!

asllib/compliance/asltest.ml
```diff
@@ -0,0 +1,195 @@
...
+        {
+          testcase with
+          outcome = Failure;
+          output = output_of_buffer stdout_buffer;
+          error_code = Some (Asllib.Error.ErrorCode.to_string code);
+          error_line =
+            (if Asllib.ASTUtils.is_dummy_pos error then None
+             else
+               let position = error.Asllib.AST.pos_start in
+               Some position.Lexing.pos_lnum);
```
HadrienRenaud, line 139: nit: I think this could be extracted into a function `error_line_of_annotated`

asllib/compliance/asltest.ml
```diff
@@ -0,0 +1,195 @@
...
+  let mapping map =
+    `O
+      {
+        m_anchor = None;
+        m_tag = None;
+        m_implicit = true;
+        m_members =
+          List.filter_map
+            (fun (k, v, style) ->
+              Option.map (fun v -> (key k, simple_scalar v style)) v)
```
HadrienRenaud, line 166: nit: I find this a bit easier to read:

```suggestion
            (function
             | (k, Some v, style) -> Some (key k, simple_scalar v style)
             | (_, None, _) -> None)
```

asllib/compliance/asltest.ml
```diff
@@ -0,0 +1,195 @@
...
+      [
+        ("mode", Some (string_of_mode testcase.mode), `Plain);
+        ("outcome", Some (string_of_outcome testcase.outcome), `Plain);
+        ("output", testcase.output, `Literal);
+        ("error", testcase.error_code, `Plain);
+        ("error_line", Option.map Int.to_string testcase.error_line, `Plain);
+        ("info", testcase.info, `Literal);
+      ]
+  in
+  to_output
```
HadrienRenaud, line 181: Why declare `to_output` just to use it after? simplify to just the call to `mapping`

asllib/compliance/asltest.ml
```diff
@@ -0,0 +1,195 @@
...
+    error_code : string option;
+    error_line : int option;
+    info : string option;
+  }
+end
+
+open TestCase
+
+(** Read a YAML file to a testcase. *)
+let yaml_value_to_testcase ~base (yaml : Yaml.value) =
```
hrutvik, line 56: I see that this docstring comment has been edited from my PoC, but it has not actually been addressed. The comment read:

> This is only for proof-of-concept: for the actual testsuite, we only need the mode to know how to run the file, and diffing will take care of the rest.

This was noted in a previous review:

> As per the comment on yaml_value_to_testcase, and in the commit message for "[PoC] Round-tripping YAML files", please refactor to only consider the mode field on input.

asllib/compliance/asltest.ml
```diff
@@ -0,0 +1,195 @@
...
+
+let read_testcase_from_yaml_file path =
+  let yaml_value =
+    match read_file path |> Yaml.of_string with
+    | Ok y -> y
+    | Error (`Msg s) -> fatal ~base:path s
+  in
+  yaml_value_to_testcase ~base:path yaml_value
+
+let read_testcase_from_file ~test_stem =
```
hrutvik, line 92: nit: separating this into two functions is unnecessary - please revert

asllib/compliance/gen.ml
```diff
@@ -0,0 +1,65 @@
+let find_compliance_files dir =
+  let dir_contents =
+    let rec loop result = function
+      | f :: fs when Sys.is_directory (Filename.concat dir f) ->
+          Sys.readdir (Filename.concat dir f)
+          |> Array.to_list
+          |> List.map (Filename.concat f)
+          |> List.append fs |> loop result
```
HadrienRenaud, line 8: Can be done in one array traversal, and it might even be more explicit

```ocaml
    Sys.readdir (Filename.concat dir f)
    |> Array.fold_left (fun fs f' -> (Filename.concat f f') :: fs) fs
    |> loop result
```

asllib/compliance/gen.ml
```diff
@@ -0,0 +1,65 @@
...
+      | f :: fs -> loop (f :: result) fs
+      | [] -> result
+    in
+    let start = Sys.readdir dir |> Array.to_list in
+    loop [] start
```
HadrienRenaud, line 13: Isn't `start` just `[ "" ]`?

```suggestion
    loop [] [ "" ]
```

HadrienRenaud, line 13: I think you don't really need to keep dir and can simplify it to just handle the complete paths:

```ocaml
    let rec loop result = function
      | f :: fs when Sys.is_directory f ->
          Sys.readdir f
          |> Array.to_list
          |> List.map (Filename.concat f)
          |> List.append fs |> loop result 
      | f :: fs -> loop (f :: result) fs
      | [] -> result
    in
    loop [] [ dir ]
```

asllib/compliance/gen.ml
```diff
@@ -0,0 +1,65 @@
...
+  in
+  let asl_suffix = ".asl" and yaml_suffix = ".yaml" in
+  let chop_suffix fname =
+    if Filename.check_suffix fname asl_suffix then
+      Filename.chop_suffix_opt ~suffix:asl_suffix fname
+    else Filename.chop_suffix_opt ~suffix:yaml_suffix fname
```
HadrienRenaud, line 19: nit: no need to call `check_suffix`

```suggestion
    match Filename.chop_suffix_opt ~suffix:asl_suffix fname with
    | Some _ as o -> o
    | None -> Filename.chop_suffix_opt ~suffix:yaml_suffix fname
```

hrutvik (commented):
A few things remaining from previous reviews:
 
- using only the `mode` field of a YAML file on input
- using `TestCase.foo` or `TC.foo`
- following up on Hadrien's questions about `glob_files_rec` and clashes between base names

I also realise that there is no change to CI in this PR. Please can we add corresponding CI updates to run the compliance suite, and ensure that CI is picking up discrepancies?

asllib/compliance/gen.ml
HadrienRenaud: Ahh I see, so no possible collision because of using `.yaml` as a separator. That makes sense, thanks.

HadrienRenaud (commented):
This now looks like a very nice PR. Just some general simplification comments, nothing too blocking

asllib/compliance/gen.ml
HadrienRenaud: Actually un-resolving it because the actual names are `foo.32.1.yaml` so I don't think the algorithm you described Hrutvik is the one in the file. For example I think there would be a collision with those 3 files:
- `a/foo.yaml`
- `b/foo.yaml`
- `c/foo.1.yaml` (can also be in `a` or `b`, it does not really matter)

Can we test this scenario please?

hrutvik: Ah good point - I think this means:

- we should refactor to use the algorithm I described above
- we should add the tests you describe

asllib/compliance/asltest.ml
Roman-Manevich: folded `read_testcase_from_yaml_file` into `read_testcase_from_file`.

asllib/compliance/asltest.ml
Roman-Manevich: removed `to_output`.

asllib/compliance/asltest.ml
Roman-Manevich: okay

asllib/compliance/asltest.ml
Roman-Manevich: - Renamed `yaml_value_to_testcase` to `yaml_value_to_mode` 
- Removed unused `outcome_of_string`, `extract_string_opt`, `extract_int_opt`, and other unused code.
- Renamed `read_testcase_from_file` to `read_mode_from_file`,
- `run_aslref` now only takes the mode.

asllib/compliance/asltest.ml
Roman-Manevich: I find grouping related definitions in a module helpful on its own. I'd prefer to to add `TC.` all across the file.

asllib/compliance/gen.ml
Roman-Manevich: Discussed offline and postpone to a future PR.
Also, I think it makes sense to require that all test cases have different base names - it makes it easier to communicate which test is referred to unambiguously.

asllib/compliance/dune
Roman-Manevich: The `tests` folder now contains several sub-folders and dune seems to be running the relevant tests in them. For example, changing `error_line` to `7` in `tests/build/ASTRule.EBinop.bad1.yaml` yields
```
File "asllib/compliance/tests/build/ASTRule.EBinop.bad1.yaml", line 1, characters 0-0:
diff --git a/_build/default/asllib/compliance/tests/build/ASTRule.EBinop.bad1.yaml b/_build/default/asllib/compliance/run/ASTRule.EBinop.bad1.yaml.actual
index 1cc00bf98..95aa73671 100644
--- a/_build/default/asllib/compliance/tests/build/ASTRule.EBinop.bad1.yaml
+++ b/_build/default/asllib/compliance/run/ASTRule.EBinop.bad1.yaml.actual
@@ -1,7 +1,7 @@
 mode: no-exec
 outcome: FAILURE
 error: BE_BOP
-error_line: 7
+error_line: 6
 info: |-
   File ../tests/build/ASTRule.EBinop.bad1.asl, line 6, characters 20 to 25:
           let p_a_s = a + b - c;
```

asllib/compliance/asltest.ml
Roman-Manevich: okay

asllib/compliance/gen.ml
Roman-Manevich: okay

asllib/compliance/gen.ml
Roman-Manevich: done

Roman-Manevich: That won't work becayse the dune file uses relative paths:
```
   (glob_files_rec ../tests/*.asl)
   (glob_files_rec ../tests/*.yaml))
```

asllib/compliance/gen.ml
Roman-Manevich: okay

Roman-Manevich:
> I also realise that there is no change to CI in this PR. Please can we add corresponding CI updates to run the compliance suite, and ensure that CI is picking up discrepancies?

Added `(yaml (and (>= 3.2.0) :with-test)))` to `asllib/dune-project` to install `yaml` only in a test run and a dedicated CI workflow step to install the correct dependency when tests are run.

asllib/compliance/asltest.ml
hrutvik: > I'd prefer to to add TC. all across the file. 

That is one of the suggestions above - feel free to go ahead

.github/workflows/make-test.yml
```diff
@@ -55,6 +55,11 @@ jobs:
       - name: Install opam dependencies
         run: opam install . --deps-only --yes --with-test
 
+      - name: Install ASLRef test dependencies
+        run: |
+          opam exec -- dune build asllib/aslref.opam
+          opam install ./_build/default/asllib/aslref.opam --deps-only --yes --with-test
```
hrutvik: I don't think we should need to hardcode build directories, as I believe `aslref.opam` gets promoted into the source tree
```suggestion
          opam install ./asllib/aslref.opam --deps-only --yes --with-test
```

asllib/dune-project
```diff
@@ -21,7 +21,8 @@
    (ocaml (>= 4.14))
    (dune (>= 3.14))
    (menhir (>= 20231231))
-   (zarith (>= 1.13)))
+   (zarith (>= 1.13))
+   (yaml (and (>= 3.2.0) :with-test)))
```
hrutvik, line 25: Is this what we want? I'm concerned that:

- it might require propagating the `with-test` dependency to `herdtools7`
- it's not quite accurate, as `yaml` isn't strictly a test dependency because those tests are optional from `dune`'s perspective

I would have thought that adding `yaml` to the `ocaml-compiler` config in the GitHub workflow files would do the job (but we need to confirm that the CI picks up any failing tests).

@HadrienRenaud what do you think?

hrutvik (commented):
- As before, the commits no longer seem coherent for review/merge. Please could you fix-up the commit "[asltest] addressed review comments" appropriately into earlier commits?
- I can't see that the `TC.foo` suggestion has been implemented
- I'm a bit unsure on the CI side - I've left a couple of comments

asllib/compliance/gen.ml
HadrienRenaud: Nice thanks

asllib/compliance/gen.ml
HadrienRenaud: I don't understand why that would be a problem. I've made a PR here: https://github.com/herd/herdtools7/pull/2008

HadrienRenaud: Thanks

asllib/compliance/gen.ml
HadrienRenaud: Thanks

asllib/compliance/asltest.ml
HadrienRenaud: Thanks

.github/workflows/make-test.yml
HadrienRenaud:
```suggestion
          - ocaml-compiler: "dune,zarith,menhir,logs,yaml"
```

asllib/dune-project
HadrienRenaud: I think this is fine, yaml is stable enough that it would not be a problem to let this one as a dependency for the tests even if it's not strictly necessary.

asllib/compliance/gen.ml
Roman-Manevich: To be resolved in https://github.com/herd/herdtools7/pull/2008

.github/workflows/make-test.yml
Roman-Manevich: okay

.github/workflows/make-test.yml
Roman-Manevich: okay, but shouldn't we also add `yaml` to the minimal version job?

HadrienRenaud: Ah I had forgotten about this. Yes I think this it should be added to the minimal version job too.

asllib/compliance/gen.ml
HadrienRenaud: Can't we merge #2008 in #1995 before we merge #1995 please? I think it would be cleaner to be squashed into the correct commit

asllib/dune-project
hrutvik: Sounds reasonable - do we need to propagate to `herdtools7` at all, or is this sufficient?

.github/workflows/make-test.yml
Roman-Manevich: added yaml to both versions.

asllib/compliance/asltest.ml
Roman-Manevich: I removed the module opening and aliased `TC` for `TestCase` and.

.github/workflows/make-test.yml
```diff
@@ -24,13 +24,13 @@ jobs:
           - ubuntu-latest
         cfg:
           # Minimal versions: we want to test both ends of the supported space
-          - ocaml-compiler: "ocaml-base-compiler.4.14.0,dune.3.15.3,zarith.1.13,menhir.20231231,logs.0.7.0"
+          - ocaml-compiler: "ocaml-base-compiler.4.14.0,dune.3.15.3,zarith.1.13,menhir.20231231,logs.0.7.0,yaml"
```
hrutvik: `yaml` needs a minimal version number matching the `dune-project` one

asllib/compliance/tests/build/ASTRule.EBinop.bad1.asl
hrutvik, line 1: As commented in #2006, please move rather than add (and later delete) these

asllib/compliance/gen.ml
Roman-Manevich: integrated into this PR. Sorry, this was a response to a different comment, please ignore.

asllib/compliance/gen.ml
Roman-Manevich: done

.github/workflows/make-test.yml
Roman-Manevich: changed to `yaml.3.2.0` to match `(yaml (and (>= 3.2.0) :with-test))`

asllib/compliance/tests/build/ASTRule.EBinop.bad1.asl
Roman-Manevich: That would break the references to those files from the ASL Reference .tex files. It would also require making the changes in https://github.com/herd/herdtools7/pull/2006 to `doclint.py` and `extended_macros.py`.  `asllib/menhir2bnfc/tests/integration/parser_cmp.expected` would also need to be updated.
I think it's best to maintain the sample copy tests here and make sure they are moved with their history in https://github.com/herd/herdtools7/pull/2006. What do you think?

hrutvik: Understood - I would suggest some one or two brand new example tests for this PR (could be a simple "hello world"), and all existing tests can be moved for for the next PR.

asllib/compliance/gen.ml
hrutvik: I'm not sure I see this - please could you clarify?

asllib/compliance/gen.ml
HadrienRenaud: Thank you!

asllib/dune-project
HadrienRenaud: I don't think so: the tests are just for `asllib` afterall, and would not break herdtools tests because they are optional.

HadrienRenaud (approved):
From my side of things, this looks good. Please wait for @hrutvik's approval before merging.

hrutvik (approved):

PR #1996, opened 2026-09-10
[gen] `diy7` performance improvement

ShaleXIONG:
Following previous #1969, this PR further improvements the performance in `diy7`:

- Use stable identifiers for adjacency lookups. Previously, adjacency lookups recovered hash-table searches Each relaxation chunk now carries a numeric identifier used directly to index the adjacency array. 4f71e2f99
- Cache process and instruction counts. Previously, search algorithm repeatedly traversed the edges to calculate these counts. We now store the counts when they are created, so subsequent checks use cached values. 651a2065f and 1af2cbece
- Reject invalid candidates earlier. Previously, incompatible concrete edge directions, annotations, and RMW combinations could enter `cycle.ml` before being rejected. They are now rejected earlier during parsing and searching. b6f17396b, 7f57480e7, a3396590b and 3e3bd662e
- Avoid redundant validation. Previously, the completed cycle repeated an adjacency check already enforced during search. The duplicate final check is removed. ced1652a6
- Assign read and write values in one cycle pass. Previously, `cycle.ml` traversed each location once to assign writes and again to assign reads. A single stateful traversal now assigns both. 7e733b341

We measured wall-clock time on `-size 5` and `-size 6`, specifically `diy7 -conf gen/libdir/forbidden.conf -size 6 -stdout >/dev/null`. the `-stdout` removed the I/O noises so we focused on time spending on searching algorithm. We run 5 times for each configuration and took the average. The result are the following:

| Size | new | `master` baseline | Improvement |
|---:|---:|---:|---:|
| 5 | 51.910s | 248.545s | 79.11% |
| 6 | 524.548s | 4,309.706s | 87.83% |

As we expected, when size goes up, the performance gains become more significant.


gen/alt.ml
```diff
@@ -291,44 +291,64 @@ module Make(C:Builder.S)
...
-    let can_precede safes po_safe (_,xs) k = match k with
-    | [] -> true
-    | (_,ys)::_ -> FilterImpl.can_precede safes po_safe xs ys
+    module Chunk : sig
+      type chunk
+      type t
+
+      val to_relax : chunk -> C.R.relax
+      val pp_list : chunk list -> string
+      val make :
```
psafont, line 300: Maybe it's because I'm not familiar enough, but I'm having trouble understanding what are the properties that `make` is supposed to maintain, or what does each parameter mean. Could a doc-comment be added here explaining some of these properties?

gen/alt.ml
```diff
@@ -309,6 +311,13 @@ module Make(C:Builder.S)
           id : int ;
           relax : C.R.relax ;
           process_count : int ;
+          (* Internal edges before the first external edge. *)
```
psafont, line 314: These can be turned into doccoments by using `(** ... *)` and moving them right after the field they document.

gen/cycle.ml
```diff
@@ -920,9 +923,26 @@ let check_cycle c =
     n.evt <- e ;
     st
 
-  (* `do_set_write_val` returns true when variable next_x has been used
+  (* TODO: this is wrong for Store CR's: consider Rfi Store PosRR *)
```
psafont, line 926: Is this TODO still significant?

psafont (commented):
I only have some minor comments.

Do you know if the caching being done has a measurable impact on memory usage? I'm curious to know its magnitude, or even if it's relevant.

gen/cycle.ml
ShaleXIONG: I actually do not this TODO. Let me check and decide if I will keep this `TODO`

gen/alt.ml
ShaleXIONG: Done.

ShaleXIONG:
> I only have some minor comments.
> 
> Do you know if the caching being done has a measurable impact on memory usage? I'm curious to know its magnitude, or even if it's relevant.

I do not have detail profile but let's me check in practice. However, in concept level, we use extra `O(n^2)` memory, where `n` is the number of input `relax`+`safe`, because we introduce a new look-up table, 2-dimension boolean array, to check adjacency. 

[update] Following my previous comment, the `type Chunk` is 8 bytes extra, so itself is a small data type.
My profile by checking the resident set size via `ps -o rss=` shows we use extra 2% more memory. So I think it is small cost here.

gen/alt.ml
ShaleXIONG: Done.

gen/cycle.ml
ShaleXIONG: [update] it is no longer valid comment. I removed it.

PR #1997, opened 2026-09-10
Fix GCS instruction semantics

relokin:


herd/tests/other/gcs-dot.t/GCSPOPM.litmus
```diff
@@ -0,0 +1,11 @@
+AArch64 GCSPOPM-dot
```
psafont, line 1: The commit message for `GCSPUSHM` explains how the semantics of the instructions is changed, but this is not done for the one corresponding to`GCSPOPM`, could an explanation be added for it?

herd/tests/other/gcs-dot.t/BL.litmus
```diff
@@ -0,0 +1,13 @@
+AArch64 BL
```
psafont, line 1: I'm having a hard time understanding why these changes produce a more desirable behaviour, maybe a little explanation in the commit message would help.

Image

Image

psafont (commented):
I think it would help to understand why some of these changes are desirable if there was an explanation on the commit messages

herd/tests/other/gcs-dot.t/run.t
```diff
@@ -0,0 +1,433 @@
...
+  }
+  
+  /* the intra_causality_data edges */
+  
+  eiid0 -> eiid1 [label="iico_data", color="black", fontcolor="black"];
+  eiid0 -> eiid5 [label="iico_data", color="black", fontcolor="black"];
+  eiid1 -> eiid7 [label="iico_data", color="black", fontcolor="black"];
+  eiid4 -> eiid0 [label="iico_data", color="black", fontcolor="black"];
+  eiid4 -> eiid8 [label="iico_data", color="black", fontcolor="black"];
+  eiid6 -> eiid5 [label="iico_data", color="black", fontcolor="black"];
```
fsestini: Image

This PR introduces an additional `iico_data` edge from the register read "g" to the branch "f". I believe this is incorrect?

herd/tests/other/gcs-dot.t/run.t
```diff
@@ -0,0 +1,433 @@
...
+  Test GCSPOPM-dot Required
+  States 1
+  
+  Ok
+  Witnesses
+  Positive: 1 Negative: 0
+  Flag Guarded-Control-Stack-is-work-in-progress
+  Condition forall (true)
+  Observation GCSPOPM-dot Always 1 0
+  Hash=b42c2d302c23e4a44044dd980148c33a
```
fsestini: This part of herd's output doesn't seem to really contribute to the test. Could we streamline this by printing out only the dot graphs? Perhaps something like:

```sh
$ herd7 ... -o . GCSPOPM.litmus > /dev/null
$ cat GCSPOPM.dot
```

We could go even further and just print out the nodes and edges:

```sh
$ herd7 ... -o . GCSPOPM.litmus > /dev/null
$ grep eiid GCSPOPM.dot
```

Ditto for the other cram tests.

PR #2000, opened 2026-09-14
[tools] Fix bug in state comparison

maranget:
Very unfortunate error in implementing the difference of two sorted lists, introduced by PR #1970.


tools/logState.ml
```diff
@@ -409,20 +412,20 @@ let diff_faults sts1 sts2 =
...
   let rec diff stss1 stss2 =
     match stss1,stss2 with
     | (_,[])|([],_) -> List.concat stss1
-    | sts1::stss1,sts2::stss2 ->
+    | sts1::r1,sts2::r2 ->
        let r = cmp_by_env sts1 sts2 in
-       if r < 0 then sts1@diff stss1 (sts2::stss2)
-       else if r > 0 then diff stss1 (sts2::stss2)
+       if r < 0 then sts1@diff r1 stss2
+       else if r > 0 then diff stss1 r2
```
psafont, line 426: This is the only change in behaviour I could find, the list passed as second parameter for diff has one fewer element now.

psafont (approved):
Separating the refactorings from the behavioral change, as well as explaining that the exact error was in the commit message would help in reviewing the code.

From what I understood, the issue was that that the diffing was not reducing the problem on the recursive call when the second argument is more than the first one.

tools/logState.ml
maranget: Hi @psafont, you are correct. Sorry for having hidden the fix.

maranget, line 426: I have updated the commit message.

psafont: thanks!

maranget:
Merged, thanks @psafont for the quick review.

PR #2002, opened 2026-09-15
[litmus] silence warnings emitted by C compilers in tests

psafont:
The output can change easily and hence the test fails.


psafont:
#2003 does this better

PR #2003, opened 2026-09-15
[herd] Create a snapshot of aarch64.cat corresponding to Arm ARM issue M.d

relokin:


relokin:
@psafont this is a fairly straightforward PR but can you please have a look?

Makefile
```diff
@@ -327,7 +328,7 @@ test.herd-asl.cata.%: asl-pseudocode
 		-variant strict \
 		-herd-path $(HERD) \
 		-herd-timeout $(TIMEOUT) \
-		-libdir-path ./herd/libdir \
+		-libdir-path $(HERD_LIB_DIR) \
```
psafont, line 331: I see another 2 instances of -libdir-path: line 435 and 449 for the diymicro tests, shouldn't these be changed as well?

relokin: Thanks Pau. I wasn't sure about the diy ones but let's go ahead and change them.

ShaleXIONG: It is for `diymicro` which Augustin upstreams. Thanks for taking care that part too.

relokin:
I am planning to merge this by the end of the day because we need to make the corresponding release. Please let me know if you have any objections.

fsestini:
Hi @relokin , sorry, I only started looking at this earlier today.

Running tests with different libdirs doesn't seem to work for me. For example `$ make test HERD_LIB_DIR=herd/libdir/aarch64/ArmARM-M.c` raises some errors.

Codex gave an explanation which I think is very plausible upon inspection: the new `HERD_LIB_DIR` variable only instructs the test runner to set the specified libdir, however the runner still passes an additional `-cat` parameter to herd7, regardless of which libdir we specify. E.g. for the tests in `catalogue/aarch64`, the test runner runs herd7 with `-cat catalogue/aarch64/cats/aarch64.cat`, which links back to `herd/libdir/aarch64.cat`.

herd/libdir/aarch64/ArmARM-M.d/aarch64.cat
```diff
@@ -0,0 +1,201 @@
...
+irreflexive [Exp & R]; rmw; [Exp & W]; rf as coRW1-Exp-rmw
+irreflexive [Imp & Tag & R]; (po & same-loc); [Exp & Tag & W]; rf as coRW1-MTE
+irreflexive [Exp & W]; (po & same-loc); [Exp & W]; ca as coWW-Exp
+irreflexive [Exp & W]; (po & same-loc); [Exp & R]; ca as coWR-Exp
+irreflexive [Exp & Tag & W]; (po & same-loc); [Imp & Tag & R]; ca as coWR-MTE
+irreflexive 
+  ( [Exp & R]; po & same-loc; [Exp & W] 
+  | [Exp & R]; rmw; [Exp & W] 
+  | [Imp & Tag & R]; (po & same-loc); [Exp & Tag & W]
+  ); rf as coRW1
```
fsestini: Some of these axioms seem duplicated, for example `coRW1-Exp-rmw` appears at line 161 and 168.

relokin: Good point I will revert these. I need to add a check in my script to check that the newly created snapshot is the same as the current model.

relokin:
> Hi @relokin , sorry, I only started looking at this earlier today.
> 
> Running tests with different libdirs doesn't seem to work for me. For example `$ make test HERD_LIB_DIR=herd/libdir/aarch64/ArmARM-M.c` raises some errors.
> 
> Codex gave an explanation which I think is very plausible upon inspection: the new `HERD_LIB_DIR` variable only instructs the test runner to set the specified libdir, however the runner still passes an additional `-cat` parameter to herd7, regardless of which libdir we specify. E.g. for the tests in `catalogue/aarch64`, the test runner runs herd7 with `-cat catalogue/aarch64/cats/aarch64.cat`, which links back to `herd/libdir/aarch64.cat`.

That’s a good point. We might want to remove these, but I think this might have consequences for the web interface. I’ll leave this for @TiberiuBucur.

The regressions are also failing for another reason. Sometimes, between releases, we make promote the references files (for example, .expected files). For instance, a small typo fix such as https://github.com/herd/herdtools7/commit/75a26f5c582950121ef8a45f6a38e464f9590737, together with the corresponding promotion, means that this test will be flagged as failing against older releases.

fsestini:
> That’s a good point. We might want to remove these, but I think this might have consequences for the web interface. I’ll leave this for @TiberiuBucur.

Should we then leave the commit introducing `HERD_LIB_DIR` out of this PR until we figure this out? As things are today, setting `HERD_LIB_DIR` doesn't fully achieve what the commit description claims, instead having the effect of running the test suite on a mix of cat files (the provenance of which is non-obvious unless one is familiar with the test runner logic). I'm fine leaving this in if it anyway helps your workflow, though.

relokin:
> > That’s a good point. We might want to remove these, but I think this might have consequences for the web interface. I’ll leave this for @TiberiuBucur.
> 
> Should we then leave the commit introducing `HERD_LIB_DIR` out of this PR until we figure this out? As things are today, setting `HERD_LIB_DIR` doesn't fully achieve what the commit description claims, instead having the effect of running the test suite on a mix of cat files (the provenance of which is non-obvious unless one is familiar with the test runner logic). I'm fine leaving this in if it anyway helps your workflow, though.

I could do that, but I don’t think it hurts to have a Makefile variable for libdir either. I can rewrite the commit message so that it doesn’t promise anything about regressions, but having a variable for a commonly used path instead of hard-coding it seems like a good idea regardless of the regressions.

fsestini:
> I can rewrite the commit message so that it doesn’t promise anything about regressions, but having a variable for a commonly used path instead of hard-coding it seems like a good idea regardless of the regressions.

Sounds good.

fsestini (approved):

TiberiuBucur:
Hi @relokin @fsestini . I think removing the `/cats` folder from a catalogue is not necessarily the best way to proceed from the point of view of the website. The files under that folder are the models that the website loads for each respective catalogue, and the ones the user can select out of.
Perhaps a better idea would be to slightly rework the logic employed by the runner to prioritise a cat file passed by hand over the one it finds in the `ci-shelf.py` file.
Either way, I think catalogue tests are not expected to pass when overriding the libdir path.
What do you guys think?

fsestini:
> Hi @relokin @fsestini . I think removing the `/cats` folder from a catalogue is not necessarily the best way to proceed from the point of view of the website. The files under that folder are the models that the website loads for each respective catalogue, and the ones the user can select out of. 

Looking at the catalogues used by the website, it seems all the models in  `cats/` are symlinks to files in `herd/libdir` anyway, so _as far as the website is concerned_ I'm not sure we really need these directories. As long as each catalogue has a way to specify a list of "preferred models" in the form of metadata, it seems to me that the actual cat files could just be read directly from `herd/libdir`. Do you agree? That said, whether other catalogues/test suites rely on `cats/` in a fundamental way is another question.

> Perhaps a better idea would be to slightly rework the logic employed by the runner to prioritise a cat file passed by hand over the one it finds in the `ci-shelf.py` file.

That should help, I agree. Even if we end up going down this route though, I think it would still be valuable to make the catalogues' structure more streamlined if possible.

> Either way, I think catalogue tests are not expected to pass when overriding the libdir path.

I think it depends on the catalogue and libdir. For example, I would expect catalogues shipped with herd-www to largely pass with most, if not all, libdirs that the user can select from the web UI, like [previous Arm ARM releases](https://github.com/herd/herdtools7/pull/1686) up to the last N releases or so. We don't have test recipes in place to check this at the moment, but it would be good to add this kind of checks at some point, IMO.

TiberiuBucur:
> Looking at the catalogues used by the website, it seems all the models in cats/ are symlinks to files in herd/libdir anyway, so as far as the website is concerned I'm not sure we really need these directories. As long as each catalogue has a way to specify a list of "preferred models" in the form of metadata, it seems to me that the actual cat files could just be read directly from herd/libdir.

I think that makes sense. Catalogues work by default on the latest version of `aarch64.cat`, which is why the files inside `/cats` are symlinked. This may not be the case for Oban update catalogues, but those have their own self-contained libdirs, so your point still stands.

> I think it depends on the catalogue and libdir. For example, I would expect catalogues shipped with herd-www to largely pass with most, if not all, libdirs that the user can select from the web UI, like https://github.com/herd/herdtools7/pull/1686 up to the last N releases or so.

My understanding is that this does not hold in the general sense. A new Arm ARM release may change some behaviours or some aspects of the memory model, which lead to differences in the `.expected` files. In fact, I think this is what you observed when you selected the M.c release libdir and ran `make test`. This is one of the necessities of providing versions of the model corresponding to releases of the Arm ARM for each catalogue in the website, so that the user may observe the changes in behaviour.

In terms of the testing infrastructure, we do not keep old `.expected` files after a promotion, but one can still run the test suite/catalogues with a different libdir and observe the outputs. Overriding the libdir variable in the Makefile and running `make test` may just not be the most efficient way to do this.

fsestini:
> My understanding is that this does not hold in the general sense. A new Arm ARM release may change some behaviours or some aspects of the memory model, which lead to differences in the `.expected` files. In fact, I think this is what you observed when you selected the M.c release libdir and ran `make test`. This is one of the necessities of providing versions of the model corresponding to releases of the Arm ARM for each catalogue in the website, so that the user may observe the changes in behaviour.
> 
> In terms of the testing infrastructure, we do not keep old `.expected` files after a promotion, but one can still run the test suite/catalogues with a different libdir and observe the outputs. Overriding the libdir variable in the Makefile and running `make test` may just not be the most efficient way to do this.

Agreed, we probably can't expect the full `make test` to keep working wholesale as we modify the model. But I don't think a hypothetical cross-libdir test recipe necessarily needs to run the full `make test` worth of catalogues. It could focus on a narrow selection of them, using stability w.r.t. model changes as one of the selection criteria. For example, currently the only aarch64 catalogue that uses `.expected` files is `aarch64-BBM`. Perhaps that one could stay out of the cross-libdir recipe. On the other hand, I think it would be reasonable to check that `make test.herd.cata.aarch64` can run without failure on the last couple of Arm ARM releases. I think that was the distinction I was trying to capture by saying “it depends on the catalogue”.

TiberiuBucur:
> On the other hand, I think it would be reasonable to check that make test.herd.cata.aarch64 can run without failure on the last couple of Arm ARM releases

In that case I would proceed to make the change I described above to the test runner. I can open an issue and assign it to me if you think that makes sense.

fsestini:
> In that case I would proceed to make the change I described above to the test runner. I can open an issue and assign it to me if you think that makes sense.

Thanks, I'd say let's open an issue and continue the discussion there. I have a couple of questions about the proposed change to the test runner (e.g. how do we tell the test runner to look for the main cat file elsewhere? With an explicit `-cat` option like herd? Something else?) and a dedicated issue would be a good place to spell out and comment on those details.

PR #2004, opened 2026-09-16
[AARCH-24967] Consolidate the requirements on Coherence-before and Reads-from-memory

relokin:


tools/miaou.ml
```diff
@@ -180,6 +182,12 @@ and cons_seqs (fs:exp list) (es:exp list) =
       let name = tr_id s in
       check_nodefs loc name
 
+    let find_inline_def id =
+      if StringSet.mem id O.inline then
+        try Some (StringMap.find id !inline_defs)
+        with Not_found -> None
```
fsestini: This is equivalently a single `find_opt`:
```suggestion
        StringMap.find_opt id !inline_defs
```

tools/miaou.ml
```diff
@@ -662,6 +673,8 @@ and cons_seqs (fs:exp list) (es:exp list) =
     let rec pp_def pref s = function
       | Item txt ->
          printf "%s %s%s\n" pref txt s
+      | List { items = [i]; _ } ->
+          pp_def pref s i
```
fsestini: Have you had the chance to investigate what is causing these spurious singleton lists to be emitted?
The simplification makes sense for the particular coherence axioms this PR is dealing with, but I wonder whether there are any legitimate uses of a singleton list like `List { items = [i]; ... }` that we should not be simplifying away.

tools/miaou.ml
```diff
@@ -745,6 +758,14 @@ and cons_seqs (fs:exp list) (es:exp list) =
          pp_txt indent s1 txt ;
          pp_txts indent s1 s2 s3 txts
 
+    let pp_check_def loc id postcond precond =
+      match postcond,precond with
+      | Item txt1, _ ->
+        let pref = "It is architecturally forbidden that " ^ txt1 ^ " if" in
+        pp_def pref "." precond
+      | _,_ ->
+        eprintf "%a: %S: Unsupported test shape\n" TxtLoc.pp loc id
```
fsestini: `precond` doesn't seem to be contributing to this pattern match.

tools/miaou.ml
```diff
@@ -873,6 +893,28 @@ and cons_seqs (fs:exp list) (es:exp list) =
...
+         end
+      | Pvar _|Ptuple _ -> defs
+
+    let rec collect_inline_defs fname defs =
+      let collect_bds defs bds =
+        List.fold_left add_inline_def defs bds in
+      let collect_ins defs = function
+        | Let (_,bds) -> collect_bds defs bds
+        | Rec _ -> defs
+        | Include (_,fname) -> collect_inline_defs fname defs
```
fsestini: Should this respect `O.expand`, like the corresponding match clause of `tst_ins`?

tools/miaou.ml
```diff
@@ -873,6 +893,28 @@ and cons_seqs (fs:exp list) (es:exp list) =
...
+        | _ -> defs in
+      let (_,_,ast) = P.parse fname in
+      List.fold_left collect_ins defs ast
```
fsestini: The overall logic of this function `collect_inline_defs` overlaps quite a bit with `tst_ast` and co., in that they both traverse the cat file, collect bindings, recurse into included files, all while processing the bindings somehow. Not a blocker for this PR, but I think it would be good if this common bindings-traversal process could be extracted into a single shared function.

tools/miaou.ml
```diff
@@ -941,6 +983,7 @@ and cons_seqs (fs:exp list) (es:exp list) =
         bds
 
     let tr_ast name =
+      inline_defs := collect_inline_defs name StringMap.empty;
```
fsestini: One thing to note is that `collect_inline_defs` fully traverses the cat file to the end, before any translation takes place. This can be an issue if bindings are shadowed/redefined:
```
let foo = ...
let bar = ... foo ...
let foo = ...
```
If we ask miaou7 to inline `foo`, it will register the second definition of `foo` as the one to inline, and then proceed to translate `bar` using the wrong inlining of `foo`. Instead, it should inline the first `foo`.

Perhaps this is not a pattern that is very used in `aarch64.cat` and other cat files we'd normally use miaou7 on, however the snipped above is valid cat, and I think miaou7 should either handle it correctly or raise an error if an inlinable binding gets redefined. This could be a very simple check in `add_inline_def` that rejects `ids` already present in `defs` with an exception.

PR #2005, opened 2026-09-17
[ci] remove sporadic failures due to apparmor profiles and latex container

psafont:
In ubuntu 24.04, the apparmor profile does not allow opam to do certain
operations neded for sanboxing:
ocaml/opam#5968

This issue should be fixed in 26.04, but ubuntu-latest still uses 24.04

Change ubuntu-latest to ubuntu-24.04 to controlb etter the os used.

Disable the apparmor profile completely, with comment to remember to
remove the workaround in the future.

ON top of this, build-asl-reference used a container for setting up the latex environment. This caused intermittent issues. This has been replaced by using a different tex distribution, tectonic.

The generated PDF looks like it's rendering fine, and @hrutvik has told me that the document is not used for anything official so it should be quite safe to replace.


psafont:
The trick for controlling whether the test dependencies are installed or not is working fine and that can be seen in the logs for the job runs

fsestini:
Would disabling opam sandboxing in CI jobs also be an option?

psafont:
> Would disabling opam sandboxing in CI jobs also be an option?

I would guess so.  I don't think it makes much of a difference the exact workaround once that a shared action contains it.

asllib/doc/ASLmacros.tex
```diff
@@ -1,4 +1,27 @@
...
+  \catcode`\–=\active
+  \catcode`\—=\active
+  \protected\def“{``}
+  \protected\def”{''}
+  \protected\def’{'}
+  \protected\def®{\textregistered{}}
+  \protected\def™{\texttrademark{}}
+  \protected\def©{\textcopyright{}}
+  \protected\def–{--}
+  \protected\def—{---}
```
Roman-Manevich, line 23: How about something like the following to make the macro names visually distinct:
```
\DeclareUnicodeCharacter{201C}{``}                 % U+201C “ LEFT DOUBLE QUOTATION MARK
\DeclareUnicodeCharacter{201D}{''}                 % U+201D ” RIGHT DOUBLE QUOTATION MARK
\DeclareUnicodeCharacter{2019}{'}                  % U+2019 ’ RIGHT SINGLE QUOTATION MARK
\DeclareUnicodeCharacter{00AE}{\textregistered{}}  % U+00AE ® REGISTERED SIGN
\DeclareUnicodeCharacter{2122}{\texttrademark{}}   % U+2122 ™ TRADE MARK SIGN
\DeclareUnicodeCharacter{00A9}{\textcopyright{}}   % U+00A9 © COPYRIGHT SIGN
\DeclareUnicodeCharacter{2013}{--}                 % U+2013 – EN DASH
\DeclareUnicodeCharacter{2014}{---}                % U+2014 — EM DASH
```

maranget:
> Disable the apparmor profile completely, with comment to remember to remove the workaround in the future.

Thanks for this. Hi @psafont. I know little about .yml files and I may be wrong, but I could not see the comments. Anyway getting rid of spurious CI failures is very useful and I'd like to have this PR merged soon.

.github/workflows/build-asl-reference.yml
```diff
@@ -20,23 +20,17 @@ jobs:
...
         with:
           ocaml-compiler: dune,zarith,menhir,logs,conf-which
-          opam-disable-sandboxing: true
-          dune-cache: true
+          with-test-dependencies: "false"
 
-      - name: Install opam dependencies
-        run: opam install . --deps-only --yes
+      - name: Install dependencies for doclint.py
+        run: sudo apt-get install python3
```
HadrienRenaud: Maybe good to add a comment saying why `apt-get update` is not needed?

.github/workflows/build-asl-reference.yml
```diff
@@ -20,23 +20,17 @@ jobs:
...
-
       - name: Checkout tree
         uses: actions/checkout@v7
 
-      - name: Set-up OCaml
-        uses: ocaml/setup-ocaml@v3
+      - name: Set up OCaml and dependencies
+        uses: ./.github/workflows/setup-opam
         with:
           ocaml-compiler: dune,zarith,menhir,logs,conf-which
```
HadrienRenaud, line 29: Optionally we could also use opam to install python3:

```suggestion
          ocaml-compiler: dune,zarith,menhir,logs,conf-which,conf-python3
```

asllib/doc/ASLmacros.tex
psafont: I had to declare a macro because xetex doesn't have `\DeclareUnicodeCharacter`

.github/workflows/build-asl-reference.yml
psafont: I've moved the mention of doclint.py to the ocaml-setup step

psafont:
> > Disable the apparmor profile completely, with comment to remember to remove the workaround in the future.
> 
> Thanks for this. Hi @psafont. I know little about .yml files and I may be wrong, but I could not see the comments. Anyway getting rid of spurious CI failures is very useful and I'd like to have this PR merged soon.

I can see why my wording would be confusing, I didn't meant a comment in the technical sense, but in the textual sense: I added the reason the workaround is needed, and when it should be removed in the name of the workflow step.

```
    - name: Work around faulty apparmor profile for opam in ubuntu 24.04
```

Roman-Manevich:
Optional: squash `[[ci] remove sporadic failure due to apparmor profiles](https://github.com/herd/herdtools7/pull/2005/changes/4cd083eaf72d8a9f26e2dccf59cbb976d7960461)` into `[[ci] use an action to share the opam setup among workflows](https://github.com/herd/herdtools7/pull/2005/changes/28015929638d6dd85e458660091fc4215edb6a0f)`.

.github/actions/setup-opam/action.yml
```diff
@@ -0,0 +1,41 @@
...
+    description: "Install test dependencies"
+    required: false
+    default: "true"
+
+runs:
+  using: "composite"
+  steps:
+    - name: Update Ubuntu repositories
+      shell: bash
+      run: sudo apt-get update
```
Roman-Manevich, line 22: Perhaps in a subsequent PR we can look into running `apt-get update` only when it is needed.

By the way, `make-asldoc-html` already runs `sudo apt-get update`

asllib/doc/Makefile
```diff
@@ -28,11 +28,15 @@ ASL_HTML_FAIL_ON_STDERR_WARNING = messages="$$($1 2>&1)"; \
...
 LATEXMK=latexmk
+TECTONIC=tectonic
 
 ASLReference.pdf: generated_spec *.tex *.bib
 	if command -v $(LATEXMK) >/dev/null 2>&1; \
 	then \
 		$(LATEXMK) -pdf ASLReference.tex; \
+	elif command -v $(TECTONIC) >/dev/null 2>&1; \
+	then \
+		$(TECTONIC) --outfmt=pdf --keep-logs ASLReference.tex; \
```
Roman-Manevich, line 39: I understand the intent is to make it possible to use Tectonic, but this is somewhat indirect. If `latexmk` becomes available in the workflow later on, this will revert to the existing behavior. Perhaps there's a way to explicitly select the desired latex installation?

.github/workflows/build-asl-reference.yml
psafont: This didn't work as expected, I've opted into the previous solution, with the comment you suggested

Roman-Manevich (approved):
Thanks Pau!
A few suggestions that can also be postponed to a later PR.

.github/actions/setup-opam/action.yml
psafont: > Perhaps in a subsequent PR we can look into running apt-get update only when it is needed.

From experience, it's always needed when using the setup-ocaml action. This is because apt can become out of date in any run, which causes setup-ocaml to error out

Roman-Manevich: okay

PR #2009, opened 2026-09-18
 [gen] Check both exclusive accesses for faults

ShaleXIONG:
This PR add checks on both the load and store accesses of an AArch64 exclusive pair, `LxSx` for faults.

Previously, only the `LDXR` instruction was labelled and included in the fault condition. This could miss a fault raised by the corresponding `STXR`. For example `diyone7 -arch AArch64 -variant memtag -metadata false -oneloc T Rfe LxSx Coe`
```
AArch64 CoRW2+rmw+T
Variant=memtag
{
 0:X0=x:red;
 1:X0=x:red; 1:X1=x:green;
}
 P0                   | P1          ;
 MOV W2,#1            | STG X0,[X1] ;
 Loop00:              |             ;
 L01: LDXR W1,[X0]    |             ;
 L00: STXR W3,W2,[X0] |             ;
 CBNZ W3,Loop00       |             ;

exists (0:X1=1 /\ not (fault(P0:L00,x)) /\ not (fault(P0:L01,x))) (* new check on `L00` *)
```
The `herd7` observation confirms that both `check` can fail:
```
0:X1=0;  ~Fault(P0:L01,x); ~Fault(P0:L00,x);
0:X1=0; Fault(P0:L00,x:red,TagCheck); ~Fault(P0:L01,x);
0:X1=0; Fault(P0:{L01,Loop00},x:red,TagCheck); ~Fault(P0:L00,x);
```

A recently experience on `diy7` test generation provide another example that it is necessary for the following forbidden test:
```
AArch64 MP+popa-amo.swpal-polt+rmwpl-pola
Variant=memtag
Prefetch=0:x=F,0:z=W,1:z=F,1:x=T
Com=Rf Fr
Orig=PodWR A Amo.Swp L PodWW T Rfe LxSx L PodWR A Fre
"PodWR A Amo.Swp L PodWW T Rfe LxSx L PodWR A Fre"
{
 0:X1=x:green; 0:X2=y:green; 0:X5=z:red; 0:X6=z:green;
 1:X1=x:green; 1:X5=z:red;
}
 P0               | P1                    ;
 MOV W0,#1        | MOV W3,#1             ;
 STR W0,[X1]      | Loop00:               ;
 MOV W4,#1        | L00: LDXR W0,[X5]     ;
 SWPAL W4,W3,[X2] | L01: STLXR W4,W3,[X5] ;
 STG X5,[X6]      | CBNZ W4,Loop00        ;
                  | LDAR W7,[X1]          ;

exists (0:X3=0 /\ 1:X0=0 /\ 1:X7=0 /\ not (fault(P1:L00,z)) /\ not (fault(P1:L01,z))) (* Fault check on L01 is needed *)
```


psafont (approved):
It makes sense to have this pair of instruction handled in a special case, although it's strange that for other instructions the choice between write and read events must be made.

As a curiosity, is it intended that the label are generated in reverse order (L01 before L00)?

ShaleXIONG:
> It makes sense to have this pair of instruction handled in a special case, although it's strange that for other instructions the choice between write and read events must be made.
> 
> As a curiosity, is it intended that the label are generated in reverse order (L01 before L00)?

The label does not matter as long as it match with the expected check in post-condition. The label assignment is based on where do we start to generate the cycle. I will look into it the precise reason.

psafont:
> I will look into it the precise reason.

No need to look into it, I was asking in case you already knew

PR #2013, opened 2026-09-22
[asl] Avoid compliance testing name clashes

hrutvik:
This PR avoids name clashes in "actual" files generated by compliance testing.

- Previously: clashing file-stems `foo` would have actual files `foo.yaml.actual` and `foo.1.yaml.actual`. This causes an issue if there is also a file-stem `foo.1`.
- Now: use `foo.yaml.actual` and `foo.yaml.actual.1` instead, by modifying actual generation in `gen.ml`. This should avoid clashes, as:
  - the numbering is added after the file extension
  - only `*.yaml` files are detected for use in generating actual files, so there can be no trailing numbers

This PR tests the changes by:
- introducing the `framework` directory, for tests that aim to exercise the test framework itself (rather than exercising ASL)
- introducing name-clashing tests within this directory

It also moves the `hello-world*` tests to this directory.


asllib/compliance/gen.ml
```diff
@@ -29,7 +29,7 @@ let generate_rules actual_stems base =
      the "expected" ones. The use of a map below ensures that we do not
      accidentally clash names of "actual" files. *)
   let actual, actual_stems =
-    let name = Filename.basename base in
+    let name = Filename.basename base ^ ".yaml.actual" in
     match StringMap.find_opt name actual_stems with
     | None -> (name, StringMap.add name 1 actual_stems)
```
Roman-Manevich: Three `foo.yaml` clash by generating `foo.yaml.actual`, `foo.yaml.actual.1`, `foo.yaml.actual.1`.
Appending a running counter to the names should fix this.

Roman-Manevich (changes requested):

asllib/compliance/gen.ml
hrutvik: Thanks - fixed the incorrect shadowing of name by distinguishing `name` (what gets stored in the `StringMap`) and `actual_name` (what gets returned by the function). Added another `foo.yaml` in `first-dir`, so that we have three `foo.yaml`s and tested that we get `foo.yaml.actual.1` and `foo.yaml.actual.2`.

Roman-Manevich (approved):
Looks good to me!

HadrienRenaud (approved):
LGTM

PR #2014, opened 2026-09-22
[asl][doc] Fix some error-code wording

hrutvik:
This was:
- inconsistent with the surroundings (in usage of "see" and "for example" to clarify how many rules may generate the error)
- referred to the wrong rule ("TypeSatisfaction" instead of "CheckTypeSatisfaction")
- small typo (capitalised "See" in middle of sentence)


Roman-Manevich (approved):
Thanks

HadrienRenaud (approved):

